PR #6, opened 2018-06-20
Cleaned import and used argparse for options

vuillaut:
- removed all unused import (matplotlib, some astropy.__ , some ctapipe.__ , sys, copy)
- introduced argparse to deal with options
   - filename
   - particle_type
   - outdir
   - storeimg
- Removed all absolute paths
- Introduced default settings for options
   - filename = gamma_test from ctapipe
   - particle_type is guessed from filename or set to 'unknown'
   - outdir = './results'
   - storeimg = False

The code is cleaner and can be tested by anyone that has ctapipe installed with a simple 
``` python LST1_Hillas.py```
(before needed to manually change absolute paths)

### DIFF
--- a/reco/LST1_Hillas.py
+++ b/reco/LST1_Hillas.py
@@ -9,35 +9,95 @@ USAGE: python LST1_Hillas.py 'Particle' 'Simtelarray file' 'Store Img(True or Fa
 
 """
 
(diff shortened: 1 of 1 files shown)


rlopezcoto:
Looks good, thanks!

PR #7, opened 2018-06-20
Add functions for alt,az transformation to Disp.py

misabelber:
I added the functions alt_to_theta and az_to_phi to Disp.py, since they were deprecated from ctapipe.
I guess this is temporary until we find out how to use the coordinate transformations from ctapipe.


rlopezcoto:
Ok, thanks, let's keep them here for the moment, but I would suggest to open an issue in the ctapipe repo to report on this and try to figure out how these transformations are now performed

PR #8, opened 2018-06-21
Updated Hillas parameters to be compliant with new version of ctapipe.

vuillaut:
The HillasContainer replaces 'cen_x' with 'x', 'cen_y' with 'y' and 'size' with 'intensity'


vuillaut:
Done.

misabelber:
Ok! I'll check it out, no problem. Thanks!

PR #9, opened 2018-06-27
HDF5 format files and pandas DataFrames

misabelber:
I added the possibility to store events as HDF5 file, the final format (fits or hdf5) can be selected as an argument when running LST1_Hillas.py. The type must be specified in the "RF" scripts.
Also changed the way data is read in the "RF" files, now pandas DataFrames are used, which makes it easier to do cuts and select different features for training.


rlopezcoto:
It looks like there are conflicts with the master branch, could you solve them and I'll merge the PR? Thanks!

vuillaut:
Ha!
I think this is a really good PR, I was actually working on doing exactly that as I work mainly with hdf5 files myself and I think it will the prefered format for ctapipe intermediate dumps.
I will go through the code asap.

vuillaut:
Hi @misabelber 

Thank you for the PR. I have several comments about it:
- line 210: why are there changes concerning `cen_x`? This is not consistent with the current version of master in the repo. Did you pull the last version before doing the PR? (it is not so important here but I would be afraid there are other changes elsewhere not consistent with current master branch)

- file reading should be transparent to the user. If we want to use different file formats (and I am in favour of supporting hdf5), I think a little separate script file to handle writing and reading would come handy. That could be done in another PR but if you have time, doing it in this one would be great.

- I see you use pandas. which is great. However, if you do, why exoprting and importing files with another library just to convert your data in a panda dataframe right after? You could use directly `dataframe.to_hdf()` and `dataframe.read_hdf()` no?

misabelber:
Hi!
I think I've resolved the conflicts, it seems that maybe my version was not fully updated to the master branch or something, so now the variables: x, y, intensity, are correctly updated.
I'm very new to using pandas, and even to reading and managing data with python so probably my way of reading/writing data is not the most efficient. I will look after what you suggested, probably things can be done much more simple than what I did.
Thanks for the comments!

rlopezcoto:
Hi, thanks for the updates, for me the code looks already good. Point 2 can be opened as a feature and regarding the last point from @vuillaut, I think this can be improved (maybe opening another issue to keep track of it? @vuillaut could you maybe open them?) but should not be a stopper for merging the code, do you agree?

vuillaut:
Hi @rlopezcoto.
I agree this should not be a stopper.
I will open more issues.

kosack:
If you're using ctapipe already as a dependency, why not use it's HDF5TableWriter, as in the example at the end of the notebook here:

https://github.com/cta-observatory/ctapipe/blob/master/examples/notebooks/ctapipe_handson.ipynb

That automatically takes the output from the `hillas_parameters()` routine (or any other ctapipe Container) and writes it to an output table. You can give the call to write() a list of containers liek so, `writer.write([hillas, mc])`, and you'll get an output table with all the combined columns.  We're going to add FITS support in the future.

kosack:
As an example:

```py3
with HDF5TableWriter(filename='hillas.h5', group_name='dl1', overwrite=True) as writer:
    
    for event in source:
        calib.calibrate(event)
    
        for tel_id, tel_data in event.dl1.tel.items():
            tel = event.inst.subarray.tel[tel_id]
            mask = tailcuts_clean(tel.camera, tel_data.image[0])
            params = hillas_parameters(tel.camera[mask], tel_data.image[0][mask])
            writer.write("hillas", params)
```

The output is *pandas* complient, so you can read the output file like:

```py3
import pandas as pd
hillas = pd.read_hdf("hillas.h5")

hillas.hist()
```

download

vuillaut:
@kosack How can we add additional information on the event please?
For example, here we would like to add new column `mcEnergy`, `mcAlt`, etc...

kosack:
You can just add more Containers to the all to write:

```py3

writer.write("hillas", [params, event.mc])
```

Right now the behavior is that it just adds those as new columns, but there could be name conflicts.  In the near future we will allow one to set an attribute of the Container class to use as a "tag", so that for instance the mc  fields become "mc_energy, mc_core_x", etc. rather than just "energy", and "core_x".

The other advantage of using the `ctapipe.io.TableWriter` classes is that units and other metadata are automatically stored in the output file (including any information stored in the Container's meta dictionary, e.g. `params.meta`)

PR #13, opened 2018-06-29
Code cleaning: file opening and calibration

vuillaut:
Doing some code cleaning 
- file opening
- calibration now using ctapipe classes
- some minor changes here and there

Note that I kept the `LocalPeakIntegrator` that was used in the code but I think we should use the `NeighbourPeakIntegrator` if there is no specific reason for the Local one

### DIFF
--- a/reco/LST1_Hillas.py
+++ b/reco/LST1_Hillas.py
@@ -12,18 +12,15 @@ USAGE: python LST1_Hillas.py 'Particle' 'Simtelarray file' 'Store Img(True or Fa
 import numpy as np
 import os
 from ctapipe.image import hillas_parameters, hillas_parameters_2, tailcuts_clean
-from ctapipe.io.eventsourcefactory import EventSourceFactory
-from ctapipe.image.charge_extractors import LocalPeakIntegrator
-from astropy.table import vstack,Table
+from ctapipe.io import event_source
+from ctapipe.calib import CameraCalibrator
+from astropy.table import vstack, Table
 from astropy.io import fits
 import argparse
 import h5py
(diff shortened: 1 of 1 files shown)


rlopezcoto:
Thanks, it looks good to me. Did you check that results are the same as before for the Hillas parameters calculations? (We should also start adding test modules...)

vuillaut:
Hi.
No the results are not identical to the previous version.
I am trying to figure out why (it must be due to calibration of course).
However, as I am using the standard and up-to-date calibration chain from ctapipe, I would trust this one over the step-by-step one done by hand.

I also wanted to add test modules but will be very busy in the coming days/weeks.

vuillaut:
Here I compare the Hillas parameters from:
- 'manual calibration' is what is now implemented in master
- 'ctapipe calibration' is using the ctapipe method (this PR)
- 'true pe image' uses the image in pe that one can find in the simulation 

compare

As you see, the Hillas parameters are quite different.
However, the intensity (in log) - which is really what changes with the calibration - is closer to ground truth. It might indicate that the cleaning (which I didn't touch) is not adapted anymore.
I will try to find a better cleaning threshold for LST.

rlopezcoto:
The cleaning levels from the manual calibration are not optimized either, but it would be nice if you could check if this is the only source for this difference. I think it is probably better wait for your tests and see if we find the reason of these differences before merging, is that ok?

rlopezcoto:
Hi Thomas, any progress on these checks? Thanks

rlopezcoto:
These changes should be applied to the code after refactoring now anyways, so I'll close it since it does not apply anymore.

PR #17, opened 2018-07-18
correct coordinate rotation

yiwamura:
Changed one line in Disp.py to correct multiplication of rotation matrix.


rlopezcoto:
Thanks for the PR. I would like @misabelber to review it to see if everything is in order and then I'll merge it

PR #18, opened 2018-07-18
Magic event display using ctapipe

labsaha:
This PR shows how to display a MAGIC event through ctapipe. It includes the MAGIC camera geometry file and 10 MAGIC events stored in an hdf5 file. At present, the MAGIC camera geometry file (fits file) is not a part of ctapipe. Hence I need to provide it here. I'll soon commit this file to ctapipe repo. Once that is done, we don't need to have this fits file here.


rlopezcoto:
Hi Lab, thanks! Looking forward to test it!

PR #26, opened 2018-11-08
Update timepars

misabelber:
It seems that the function timing_parameters() has been updated in ctapipe, so I had to update is usage in the function get_events() from "calib_dl0_to_dl1.py".


rlopezcoto:
Changes look good to me, I would say that it would be nice if you make the code pep8 compliance because this repo will in the end be merged with ctapipe, but ok by now.
Regarding the notebook, thanks for preparing it, I would accept it now and have a look together with Abelardo to suggest changes for the workshop.

Sorry I got a bit confused with the two PRs, shall I merge this one and rule #25 out?

misabelber:
Well, actually I thought I already did try the adaptation to pep8, but since I'm not very familiarized with it, I guess there are a lot of things that are still not in proper format. I'll take a look at it again.

I tried to separate the "subjects" of the commits in different pull requests using different branches but I'm not sure if it works like that... I see that in this pull request are included also the notebook and the other changes so yes, I think you should just merge this one and forget about the #25.

PR #28, opened 2018-11-15
changing variables names to be pep8 and the same as in ctapipe

vuillaut:
Hi

I propose to go step by step with the refactoring as it might quickly grow big.
The first one I'd like to make concerns variable names as it impacts every step of the pipeline.


vuillaut:
As a refactored the whole repo, the variables names should have been changed in all files.
But yes indeed there seem to be redundant scripts in the repo and some might be depreciated.
I asked @misabelber on the ctapipe's slack (if you have time during the day I have a bunch of questions for you Mab 😄 )

misabelber:
Yes, I will erase the files that are no longer needed. 
@vuillaut Of course! I'm not sure if today I will have the time, but if you want to ask anything, tomorrow morning I can take a look.

PR #29, opened 2018-11-15
Deleted old files that are no more necessary 

misabelber:


vuillaut:
That's an important clean-up, thank you Mab.

What is `rate_tests.py`?
The name is not clear to me.
Is it intended as a kind of unit test?

Also, could  you move `guess_type` into `utils.py` please?

misabelber:
Ahhh sorry, this commit wasn't supposed to be here yet. It is a quick script I was working on to include spectral weighting, but I'm still doing test on it... that file will be deleted at some point!

vuillaut:
You can just remove it from the PR then ;-)

rlopezcoto:
Good, a very important cleanup, thanks @misabelber. I'll wait for you to delete the temporary script for the weighting and will merge the PR afterwards.

misabelber:
I think I managed to delete it (sorry, sometimes I feel like a duck managing github)

Issue #11, opened 2018-06-28
Transparent file reading/writing


vuillaut:
Make the intermediate data (DL1) writing/reading transparent to the user. This is related to issue #10.

vuillaut:
hi @rlopezcoto 
My idea was to wait for the last PR from @misabelber and to work from there.
Next week I'll focus on:
- making all the code more pep8 compliant
- getting the code more in-line with ctapipe functionalities and in particular, the files writing/reading (this issue)
- merging my work with the one from @misabelber 

How does that sound for both of you?

rlopezcoto:
it sounds verty good and helpful, thanks!

misabelber:
Yes! Thank you very much. Any help with this is very welcomed

vuillaut:
Solved by #30

Issue #10, opened 2018-06-28
Simplifying table data writing and reading 


vuillaut:
I feel like this could be simplified, for example, using directly pandas writing and reading functions.
I shall have a look into it quickly but I open the issue to keep track of the idea.

vuillaut:
Solved by #30.

PR #30, opened 2018-11-15
adding new io container for DL1 mono reconstruction and example notebook on how to use it

vuillaut:
Changes:
- add a new io container for DL1 mono reconstruction
- example notebook to show how to use it
- add `utils.get_event_pos_in_camera` function to get the position of the source in camera using ctapipe coordinates procedures
- add `utils.disp` function to compute disp (in meters) using `get_event_pos_in_camera` and straight out of ctapipe containers

This PR is to show how I propose to solve issues #10 and #11.
The advantages of using a custom ctapipe container are:
- fix the names of variables for the rest of the chain
- be able to add custom variables such as disp or mc_core_distance
- handle units correctly
- use ctapipe containers writing and reading procedures

If that looks ok with you, let me know but don't merge it right away, I'll use the same PR to make the actual changes in the scripts.


rlopezcoto:
ok, thanks @vuillaut, this will affect the way data are read in for lstpipe but I see that you already plan to include this in the PR. If @misabelber es ok with the changes for the reading, you can go ahead with the changes to the rest of the scripts and I'll merge them afterwards

rlopezcoto:
ok, so could you please write some test modules so we can be checking that the code is not being broken? 
I would also like to see that the change of coordinate system is not screwing things up with the reconstruction. Whenever @vuillaut has implemented those in the new lstpipe scripts, could one of you @misabelber or @vuillaut check that things are still properly calculated (Hillas parameters...)? And that you get similar results as before?

vuillaut:
@misabelber The containers dump pandas dataframe in hdf5 (see last notebook cell) so it won't change the rest of the analysis.

misabelber:
@vuillaut Ahh no problem then!
@rlopezcoto Of course, I can test everything when the changes are implemented.

vuillaut:
@rlopezcoto a question relative to the coordinates system and change from sky to camera frame.

In the last notebook I pushed, you can see that the results with custom code (from Mab) and the coordinates change with ctapipe coordinates system (based on astropy) give the same results.

However, the change of frame with astropy coordinates is quite slow. In this case the custome code appears to be about 10 times faster.
But it might be more error-prone if camera rotation changes or ctapipe coordinates are handled differently in the future.

What approach do you think we should keep by default (in any case I'd be in favor of keeping both codes in the repo for now and explain this in the docstring) ?
Thanks

vuillaut:
Hi.
I finished the refactoring to use ctapipe container and data dumping.
- new custom function to make the calibration
- new function to make dl1 reduction step
- new function to make r0_to_dl1
- new function to get event type number from a particle name
- notebook to compare previous chain and new one --> get the same results

I flag the previous `get_events()` chain as deprecated but did not remove it for now.
You can see in the notebook that we get the exact same results for both chains.

rlopezcoto:
Ok, thanks Thomas, could you please @misabelber have a look to the final code and see if you agree with all the changes before merging it?

misabelber:
I just reviewed all the changes and they seem perfect to me. Thank you @vuillaut  for taking all this tasks. Also the installation package looks very good and useful.

PR #33, opened 2018-11-19
Installation package

vuillaut:
Hi.
Here is another refactoring.
I have reorganised the repository for clarity and to make a package out of it:
- scripts and functions are now separated
- functions are now part of a package 'lstchain'
- refactoring of files into submodules: calib, reco, visualization, io, magic-lst
- installation procedure with python setup.py install
- unit test for imports and the main pipeline steps: `d0_to_dl1`,`buildModels`, `ApplyModels`
- add imports in notebooks

After installation, one can now do
`from lstchain.reco import ...  `
It will prevent a lot of issues with relatives paths import.

The unit tests I set-up are really simple and not meant to check results but to ensure the pipeline is running. This way we can have a continuous integration and be sure future modification won't impact running the pipeline.

ps: this PR should probably come after #30

PR #34, opened 2018-11-21
Revised new factorisation. Corrected a couple bugs and adapted/splitted lstchain notebook.

misabelber:
Hi!
I've checked the refactorization of the code that @vuillaut did,everything runs smoothly now, but had to modify a couple of things:
- The new r0_to_dl1() function from calib_dl0_to_dl1.py module was storing "nan" width values, which was problematic later for training the Random Forests. I included an if statement similar to the one in get_events() functions, to avoid nan "w" values. 
- Corrected some small bugs due to a substitution of the word "Disp" by "disp_".

About the notebook:
- lstchain.ipynb : Replaced old io function by the new one. Moved the training RF part to other notebook, this one reads stored RF.
- RF-performance.ipynb: This is a new notebook to check the performance of the RF, exploring the possible cuts in energy or other variables. The trained RF created here are stored and can be used by lstchain.ipynb


vuillaut:
Hi @misabelber 

> The new r0_to_dl1() function from calib_dl0_to_dl1.py module was storing "nan" width values, which was problematic later for training the Random Forests. I included an if statement similar to the one in get_events() functions, to avoid nan "w" values.

I would tend to disagree with that choice.

The fact that one parameter (why `width` rather than another one ?) fails and ends-up with a `nan` value should not be evaluated a this point.
The reconstructed parameters can still be stored and the later stage of the pipeline decide what parameters they use (`width` might not be used by RF after all) and decide what to do with `nan` values (they can be replaced by zeroes, the whole set of parameters might be deleted, etc...).

ps: when you load a dataframe with pandas, you can use `df.dropna()` to remove these bad boys.

misabelber:
Well, width is one of the most important parameters for the RF, (it is even evaluated twice, with w and w/l) so I think is better to just drop the events with bad width values. I will try your suggestion and update the commit, thanks!

misabelber:
I removed the if statement in r0_to_dl1(), instead, I included the "df.dropna()" line after reading df from files when doing the reconstruction.

rlopezcoto:
Hi, I would also go for the old solution for now, we can open an issue to keep this in mind to settle it down in future PRs

rlopezcoto:
Whenever this is solved, I can merge #33 and #34 together if there is no opposition. Thanks a lot for all the hard work settling this now before the workshop

misabelber:
@rlopezcoto what do you mean the old solution? The if statement in the r0_to_dl1() function, or the dropna() method implemented now?

rlopezcoto:
Sorry for the confusion. I meant that if there is any danger that all data gets dropped due to the dropna() method, I would rather keep the if statement. @vuillaut do you think it is ok if we leave the dropna() as it is now? I found it a bit dangerous in its currant implementation as you explained it, but you may comment further.

rlopezcoto:
It seems that after the merging of #33 there are some conflicts with this merging, @misabelber, could you please correct them? Thanks

misabelber:
Yes, I see, with the new refactorization everything I changed in this PR is in conflict. I will fix it and do a commit.

misabelber:
Still don't know what to do with the nan values of width: Do the keep the if statement or the dropna() until we discuss a definitive silution?

vuillaut:
You may follow @rlopezcoto suggestion for a quicker fix and we can discuss a more elegant fix later.

misabelber:
I did a commit resolving the conflictive files, are they ok already?

rlopezcoto:
I still see the same conflicts as before...

vuillaut:
Hi @misabelber.
Seeing the number of files in conflict, you are probably not in synch with the master branch.
Here are my suggestions:
- fetch and pull last changes into your master branch after #39 has been accepted
- create new branch from there to apply corrections about width and disp_ (maybe these ones have been corrected already).
- create another branch from master for the new notebook. As it's only adding a new independent file it cannot be in conflict.

misabelber:
I just closed this PR because I'm having a lot of trouble trying to sync with the upstream repository. I will pull the last changes with #39 and see if I then I can upload the notebooks.

vuillaut:
Let me know if you need help.
I’m available on slack.

Thomas

PR #39, opened 2018-11-22
Correcting some bugs:

vuillaut:
- some due to last PR and related to disp refactored in disp_
- scripts could not work if the result dir did not exist before
- models could not be loaded if the path didnt end with /
- bug do to refactoring of ApplyModels in number of arguments
- in lstpipe.py path_models arguments was not passed
- also removed `impact` features that was a duplication of `mc_core_distance`
- corrected bug due to renaming "w/l" as "wl"

I retested all scripts that are working now.


vuillaut:
ps: tests are now passing on my travis.

PR #40, opened 2018-11-23
Gammalearn

mikael10j:
This PR adds a notebook for my tutorial and the necessary functions in a subpackage named gammalearn.


rlopezcoto:
good @mikael10j, and thanks for preparing a minimal gammalearn solution for the workshop.

rlopezcoto:
I do not have time this weekend to test the notebooks, I expect to get it done Monday morning but I hope you already did, so I'll merge it for people to get it for the bootcamp

rlopezcoto:
If possible, I would rather get the `gammalearn` folder outside `lstchain` by now, probably in a folder for the temporary material for the workshop (if we would like to get gammalearn later, we can get the repo).

mikael10j:
These are not really part of gammalearn, only utils functions to load networks without gammalearner framework. How do you want to proceed ?

rlopezcoto:
what about a `utils` folder?

vuillaut:
Maybe not `utils` as there is already one in the lstchain.

rlopezcoto:
there is a `utils.py` script but not a `utils` folder

vuillaut:
@rlopezcoto you were proposing a bootcamp directory for specific notebooks and/or presentations.
(As an example, in gammapy there is another repository on the side for coding sprints, bootcamp etc... but it might be overkill for us)

vuillaut:
> there is a `utils.py` script but not a `utils` folder
Yes, I was just afraid it could be confusing.

rlopezcoto:
yes, it might be overkilling for us, probably a folder should be enough. In any case it may be a bit late and we may rearrange everything during or after the Bootcamp

PR #41, opened 2018-11-23
Avoid NaNs in width, and small corrections

misabelber:
I redid the old PR with the changes I needed to make all scripts run. 

- I included the if statement for the NaNs, it is necessary for running on larger simtelarray files other than the example one from ctapipe., until we decide how to manage it better.
- Just for aesthetics and coherence with cta-lstchain-extra: RFreg_disp -> RFreg_Disp
- Usage of "r0_to_dl1()" in lst-recopipe instead of get_events()
- Added a new argument to lst-recopipe, to limit the maximum number of events to analyze (for quick tests on larger simtelarray files)
- In visualization: plot_disp_ -> plot_disp, plot_E() -->  plot_e()


rlopezcoto:
Ok, if it works I will approve it and we will try to rearrange a bit the repo after the bootcamp

PR #42, opened 2018-11-23
Notebooks for LST1 bootcamp

misabelber:
These are the two notebooks I prepared to run during the bootcamp:

- RF-performance: Shows how to train and store RF for reconstruction and separaion and so some performance plots. Also shows how to do cuts in energy or any other variable to improve performance.
- cta-lstchain: Takes a simtelarray file as imput and performs the full analysis chain to dl2 format, using already trained RF.
I also deleted a notebook that is no more necessary.


rlopezcoto:
Thanks, Mab. Are you also planning to show these in the workshop? Keep in mind that hands-on demonstration take time and that you may need more time in this case

misabelber:
I thought that was the idea, right? To show the notebooks and run them with the people. In any case the notebooks are very quick, and if they use the files in cta-lstchain-extra, which are smaller than the ones I have, the time will be even shorter.

rlopezcoto:
Yes, this is the idea, I just wanted to know to adjust time for contributions.

PR #43, opened 2018-11-23
Add script to test the performance of RF

misabelber:
This is a script I had for some time, but which was not already merged. 
It is for testing the performance of the RF, with a larger "test" dataset than when using lst-recopipe. 
It's useful to tune the parameters of RF, testing new features os applying cuts.

### DIFF
--- a/scripts/lst-rfperformance.py
+++ b/scripts/lst-rfperformance.py
@@ -1,0 +1,114 @@
+"""Pipeline for test the performance of Random Forests.
+
+Usage:
+
+$>python lst-rfperformance.py
+
+"""
+import sys                                                   
+sys.path.insert(0, '../')
+import numpy as np
+import pandas as pd
+import argparse
+import matplotlib.pyplot as plt 
+from sklearn.externals import joblib
+import sys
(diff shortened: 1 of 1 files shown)


scripts/lst-rfperformance.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+import sys                                                   
+sys.path.insert(0, '../')
+import numpy as np
+import pandas as pd
+import argparse
+import matplotlib.pyplot as plt 
+from sklearn.externals import joblib
+import sys
+sys.path.insert(0, '../')
```
vuillaut: This line can be removed since you import lstchain

misabelber: I get an error "No module named 'lstchain'" if I don't include the line :s

vuillaut: Did you install it ?

misabelber: Yep, now it's working, sorry! We can do a cleaning of the code because I'm sure there are other parts where that line remains. Thank you

PR #44, opened 2018-11-23
Gammalearn

mikael10j:
I moved the gammalearn utils functions from lstchain to a new utils folder.


rlopezcoto:
Good, we can discuss alternate solutions next week.

Issue #50, opened 2018-11-25
cta-dev3 as environment name? 


vuillaut:
I think there is a typo in the name of the environment in the environment.yml file.

rlopezcoto:
Sorry I was making tests and changing notebooks and pushed the environment.yml by accident. It is fixed now

PR #51, opened 2018-11-25
Simplifies install

vuillaut:
- ctapipe is now the only dependency
- changed the README accordingly

I think this simplifies the install:
if people don't have ctapipe already, we suggest them to install it a cta conda environment.
In all cases the install of lstchain is just a `python setup.py install`.

This is to avoid the issue you had on Friday @rlopezcoto 
And solves issue #50


rlopezcoto:
I think instructions were clear now, right? Is this just to avoid to create a `cta-dev` environment and create a `cta` one instead?

vuillaut:
- I thought we could avoid having any other dependency than ctapipe.  But it seems that even though it is used in ctapipe, h5py is not in their setup so we need to keep in ours
- yes at the end not much change, I just thought it was clearer this way

rlopezcoto:
ok, I also think it is clearer this way, I'll merge it in case today it is still needed

PR #54, opened 2018-12-03
Coordinates notebook corrected.

thomasgas:
Coordinate usage corrected after discussion with @MaxNoe on the implementation of the coordinates with astropy. One should first create a frame and then create points in this frame with SkyCoord from astropy.
Related to https://github.com/cta-observatory/ctapipe/pull/842 
@MaxNoe could you check this notebook? I'm doing the transformation from camera to AltAz explicitely for each frame.


MaxNoe:
I proposed some minor updates here:
https://github.com/thomasgas/cta-lstchain/pull/1

thomasgas:
Thanks max for the review ;) 
I'll just delete one file (the plot_on_sky.py) since it's now fully working.
I'll add it later on to ctapipe directly.

MaxNoe:
I think this belongs to ctapipe, you should also put it there

thomasgas:
> I think this belongs to ctapipe, you should also put it there

Ok I'll do it.

rlopezcoto:
I would try to avoid duplicities, so if this is going to go to ctapipe maybe you can completely move it there?

thomasgas:
> I would try to avoid duplicities, so if this is going to go to ctapipe maybe you can completely move it there?

I agree...I can remove it from here as long as the PR in ctapipe is merged.

rlopezcoto:
Sorry, I could not follow all the changes from last week... any progress on this point?

thomasgas:
I guess that i can remove the notebook since last week there has been a major coordinates refactoring. We can really remove it from here and there will just be the updated one in the ctapipe folder. let me do this and then you can merge the PR.

thomasgas:
ok you can merge the PR @rlopezcoto.

PR #56, opened 2018-12-07
Added notebook with examples to read LST1 data

FrancaCassol:
This note book will work as soon as the LST reader will be accepted in ctapipe


rlopezcoto:
This looks good and very necessary now that we have the first data

PR #57, opened 2018-12-18
Add a simple notebook to plot LST images

rlopezcoto:
Since there is some excitement regarding the first LST data, I share here a simple notebook to plot telescope real data in a simple way. @FrancaCassol since you were the one implementing the routines for LST data reading, could you please have a look? 
This could be added in the future to #56, that is more complete, but I wanted to have something simple now to have a quick look to the data.


FrancaCassol:
Hi Ruben, 
concerning the reading of data, two things: 

1- in case of parallel zfit writers you call read all the ordered events using the wildchard ' * ' to open several files at the time. This is easy starting form run 76 where the file names take into accounts the streams number (1,3,5,7), so you will used:

LST-1.*.Run00076.0000.fits.fz 

in order to read the parallel written files:

LST-1.1.Run00076.0000.fits.fz
LST-1.3.Run00076.0000.fits.fz
LST-1.5.Run00076.0000.fits.fz
LST-1.7.Run00076.0000.fits.fz

2- the LSTCam-002 camera geometry is directly connected to the events by the reader in the "inst" container, so you can use 

  disp0 = CameraDisplay(ev.inst.subarray.tels[0].camera, ax=ax)

instead of 

  geom = CameraGeometry.from_name("LSTCam-002") 
  disp0 = CameraDisplay(geom, ax=ax)

But as you do is also correct, just one has to remember the geometry camera name

rlopezcoto:
Thanks for the comments Franca, I added them to the notebook. I'll merge them in case somebody wants to use it

PR #58, opened 2019-01-11
Going step by step in the selection/reduction/calibration of DL0 to DL1 data

misabelber:
Hi!
In order to try to be able to detect the possible problems that are leading to the bad performance of the energy reconstruction with lstchain, @rlopezcoto asked me to write a document to go throughout all steps we are taking to treat the data from dl0 to the calculation of hillas and timing parameters. To do so I thought that it would be much more graphic to write a notebook explaining each step, in which is possible to plot the results from pedestal substraction, charge extraction, claning, etc.
If you want to take a look and check if everything makes sense, if anything is being done wrong, or if I should include any other explanation, plot or something, just tell me. 

Thank you very much!


scripts/lst-rfperformance.py
```diff
@@ -21,13 +21,19 @@
...
+'''
 parser.add_argument('--gammafile', '-fg', type=str,
                     dest='gammafile',
                     help='path to the dl1 file of gamma events for training')
 
 parser.add_argument('--protonfile', '-fp', type=str,
                     dest='protonfile',
                     help='path to the dl1 file of proton events for training')
+'''
+parser.add_argument('--inpath', '-i', action='store', type=str,
```
vuillaut: why this change of API?

vuillaut:
Hi Mab.
Very good initiative.
I will start the review of the notebook right away.

scripts/lst-rfperformance.py
misabelber: Hi Thomas! Thanks for the review of the notebook.
Ah, the other commits seems to be things I did  after the bootcamp due to some comments I received. 
The change in the RFperformance.py was in order to always use the same train and test events, instead of building them randomly each time the script is ran.

vuillaut: I see. But this just moves the issue from pointing to training files to a directory where train/test splitting has already been applied, no?
I think a more versatile (and still easy and consistent) way would be to do the train/validation splitting in the script with a default random seed so that we know the training dataset is always the same when running for the same file.

In any case, could you limit the PR to its intended purpose and have only the notebook? It's better to have multiple PR for different purposes so we can discuss changes separately.
Thank you.

vuillaut: Thank you.
PR can be merged I think.

misabelber:
Thanks! Yes of course, it's better to have sepparate pull requests (it's just me still struggling with github), I will review the other topic and do a proper PR when it's ready.

rlopezcoto:
thanks @vuillaut and @misabelber for the PR and the review! I had a quick look and it looks fine and helpful for the purpose we discussed. I'll merge it now so people can start using it and we can have further discussion about the things to check in the analysis/reconstruction

PR #59, opened 2019-01-15
Updating step by step notebook with disp reconstruction

vuillaut:
Hi.

Small PR to update the step by step notebook with the `disp` computation and reconstruction.

I also set `tel_id=1` from start so we can more easily try the notebook with other telescopes


rlopezcoto:
It looks good to me, if @misabelber agrees I'll merge it

PR #60, opened 2019-01-15
remove hillas_parameters_2

vuillaut:
`hillas_parameters_2` has been removed from ctapipe


rlopezcoto:
Thanks, simple change

PR #65, opened 2019-01-22
Updates on Step_by_Step_dl0_to_dl1 notebook

misabelber:
Hi everyone,
I've been working on adding some new checks to this notebooks, suggested by @moralejo .

- We wanted to compare the hillas intensity (or size) that we calculate after cleaning, with the number of MC photoelectrons in the camera. I extract this info using pyhessio and then I plot intensiy vs mc phe  to check that the relation is linear. 

- Also, I've added a part on the calculation of the spectral weight for each event, to convert their spectra to the Crab (if events are gammas), and obtain a rate in the histograms, instead of the number of events. I'm not sure if this part is correct, so if you can take a look and see if everything make sense, it will be very appreciated. 

Something I noticed is that reading  files using pyhessio is quite different from using ctapipe event readers, and it is redundant to go over the same simtel file twice to access to all the information. Maybe I'm missing something and there is a  better way to access to all the MC information + calculate hillas/timing parameters than going first with ctapipe(and using our custom containers and everything) and then using pyhessio, but I don't know. Suggestions are highly welcomed.


misabelber:
Hi!

After reviewing the results with @rlopezcoto  and @moralejo  I've introduced some new features in the notebook and corrected things that were worng:

- I've added two functions aimed to visualize the results on energy resolution. They include a representation of the mean and rms of the quantity: log(Erec/Etrue) per energy bin (function "plot_e_resolution()) and a way to present the resolution as: The integral of the region around 0 containing 68% of the events in the histogram of log(Erec/Etrue), instead of just the rms of the distribution (function "calc_resolution()"). If you are happy with these representantions, I can add them to the visualization module of lstchain.

- I've included some lines of code to detect pixels that saturate the waveform(sample > 4094) and thus, low gain data should be used for them. The resulting image will be a combination of high gain/low gain pixels. If you are ok with this, I will include this code to the calib module.

- There's a test on the comparison of high gain and low gain number of photoelectrons.

- I've corrected the spectral weighting calculation, which was wrong.

If you think there's anything missing or that needs a change, just comment! Thank you!

rlopezcoto:
Thank you very much @misabelber! As you said, we've been discussing about it, so if nobody opposes I'll merge it.

PR #66, opened 2019-01-28
Small corrections to make the notebook work

rlopezcoto:
The notebook was not working due to some packages were not defined at the beginning. 
Also it would be nice to clarify: 

- the variable for unit conversion, is it MeVtoTeV? -> then the value has to be changed
- is it MeVtoGeV? -> then the name of the variable should be changed

I'm asking now because I see that the energies emax, emin... are given in TeV, if the energy of the events is also given in TeV, then it would be better to work on TeV. 
I did not have time yet to check if the absolute results of the weighting are ok, the spectral index, as @moralejo commented out, seems to be ok.

### DIFF
--- a/notebooks/Step_by_Step_dl0_to_dl1.ipynb
+++ b/notebooks/Step_by_Step_dl0_to_dl1.ipynb
@@ -33,7 +33,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 523,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


misabelber:
Hi Ruben
Thanks for the fixing. The units of the energies given to the RF are in GeV, so that's why I wanted to work in GeV, so the name of the variable should be MeVtoGeV. I'm preparing a commit with the new features of the notebooks and was going to include this correction, but if you want to do it right now it's ok.

rlopezcoto:
No, I guess I'll merge this one not to make a mess with yours, I only changed a couple of lines so it should be safe, is it ok?
Please keep in mind that you are mixing TeV (emin,emax) with GeV for the weighting. The easier would be to work with astropy units

misabelber:
Yes, I have that in mind. I'll try to find a proper way to use the units when I include the final function into the lstchain code.

PR #69, opened 2019-01-31
Reconstruction of diffuse events in mono

vuillaut:
This PR introduces some changes regarding the `disp` parameter in order to improve the mono reconstruction.

Here are the major changes:
- disp is now considered a vector
   - the old disp being its norm
   - a class `DispContainer()` is holding the different parameters of the disp vector: `(dx, dy, norm, angle, sign, miss)`
- the reconstruction of the disp vector can now be made in different ways:
   - only reconstruct the `disp_norm` (as before)
   - reconstruct the `disp.norm` and the `disp.sign` allowing for the reconstruction of diffuse events.
   - reconstruct the `disp_norm`, the `disp_sign` and the `disp_angle`. In many cases, the `disp_angle` actually differs from `hillas.psi`. The example in the notebook will show that in a dramatic way.
   - reconstruct the disp vector: `disp_dx` and `disp_dy`. This should be equivalent to the previous one but I actually get better physics results this way (and there is one less model to train).
- some visualisation functions to overlay the source position, disp and hillas major axis to `CameraDisplay`
- a notebook to demonstrate these changes
- some unit tests

The current reconstruction in `reco_dl1_to_dl2` is unchanged and reconstruct only the `disp_norm` and still works only for point source gammas


misabelber:
HI @vuillaut , that's a lot of work, thank you! I've done a quick test running the notebook and everything seems to run smoothly. 
I did notice the issue of the disp.angle and hillas.psi to be very different in certain cases, when there are pixels with a considerable amount of light far from the actual shower. I explored it changing the cleaning parameters and calculating the number of islands. Taking only the main island to calculate the hillas parameters can solve the issue, but I'm still not sure on how to deal with this.
I'll take a deeper look at all the new features. 
Once this PR is merged, I'll start working on the pyeventio thing.

misabelber:
Hi @vuillaut , can this method of reconstructing the disp vector be used with point gammas? I tried to run it over a file of pointlike gammas and the reconstruction is surprisingly perfect, so I guess that we must be introducing some kind of bias,  but I'm not sure why, probably I'm missing something. Or or maybe I'm just doing something wrong (I make sure that I splitted the gamma events in train and test, but still) .

rlopezcoto:
Hi @misabelber, are you using point-like gammas for train and test (I guess you are using the RF from lstchain, right?)? If you use point-like gammas coming from the center of the camera the RF may get too specialized in reconstructing that particular direction. Can you share with us these performance figures and also try training with diffuse gammas?

vuillaut:
Hi @misabelber and @rlopezcoto.
Yes, the vector method can be used for point-like gammas but as @rlopezcoto suggests, you must train on diffuse gammas.
In my humble opinion, that should be the case for any method actually, otherwise, the model will be biased to reconstruct events at a specific position in the camera.
The fact that training only for the `disp_norm` does not give a perfect result comes from the fact that when doing the reconstruction `hillas.psi` differs from `disp.angle`. But there is still an important bias I think. If you were to train to reconstruct `disp.angle` on point-like gammas as well, the reconstruction would also be perfect. 

The model can only reconstruct events exactly at the position it was trained on. Feed it with whatever events at it will always predict an event exactly at this position (in your case giving a perfect reconstruction for point-like gammas from the same source position).

vuillaut:
Thank you for taking the time to review the code and actually testing in on data :)

misabelber:
Ok, now I understand, so the best thing is to always train with diffuse gammas to avoid any bias. Perfect then!

rlopezcoto:
Hi, @misabelber you've been having a deeper look to this PR, can you review it in more detail before accepting it? Thanks

misabelber:
Everything looked perfect to me, the only question was about the reconstruction of point gammas (and was answered), but the modifications on the code itself work fine and will be very useful.

PR #70, opened 2019-02-01
Training refactoring

vuillaut:
This is work in progress (mainly waiting for PR #69) to refactor the training functions.

- make changes to be more in line with pep8 
- introducing external configuration for models
   - either with standard configuration (for now defined and loaded in `dl1_to_dl2.py` - this should change in the future)
   - or by using a configuration file
   - only supporting RandomForest for now (
- solves some bug in `plot_dl2` introduced by PR #69


rlopezcoto:
thanks @vuillaut. What did you exactly have in mind for the configuration files introduced?

vuillaut:
Hi @rlopezcoto 

Well, today, a lot of parameters are hardcoded, which prevents proper tests, versatility, sharability, etc...
First I was aiming at making the whole pipeline parameterizable with a config file.
But then I realized that was a bit too much to do at once and that it was probably better to go small and incrementally. So I used the fact that I was doing some code cleaning in the machine learning part to introduce a bit of parameterisation through a [JSON file](https://github.com/cta-observatory/cta-lstchain/blob/274e814b4009c009bea0d195c79ade05d4ea34f3/utils/config_training.json).

The idea is later to build on it and add more and more parameterization to the pipeline, at all stages (calibration, cleaning, etc...).

I merged master so I think this PR is ready to be reviewed.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -0,0 +1,517 @@
...
+                                'random_state': 42,
+                                'verbose': 0,
+                                'warm_start': False,
+                                }
+
+
+random_forest_classifier_args = {'max_depth': 2,
+                                 'min_samples_leaf': 10,
+                                 'n_jobs': 4,
+                                 'n_estimators': 'warn',
```
misabelber: What does the 'warn' option mean for n_estimators?
I'm getting an error here because n_estimators is not supposed to be a string: ValueError: n_estimators must be an integer, got <class 'str'>.

vuillaut: Absolutely right @misabelber, my mistake.
This is fixed now.

rlopezcoto:
Hi @vuillaut. I like the idea of making the pipeline parametrizable, the only question would be to make it the same way as it is planned for ctapipe, any info on the format decided for these config files?

misabelber:
When running test_change_frame_camera_sky() from the test_lstchain.py script, I'm getting an error about the Nominal Frame: TypeError: Coordinate frame got unexpected keywords: ['array_direction', 'pointing_direction', 'x', 'y']. 
I've checked and NominalFrame takes only the arguments "origin", "obstime" and "location". Maybe this has changed in the lastest versions of ctapipe. I guess we should use it following the example in: https://github.com/cta-observatory/ctapipe/blob/master/examples/coordinate_transformations.py function "cam_to_nom" to transform from camera coordinates to sky coordinates.

vuillaut:
Hi Mab.

The coordinates transform have changed in ctapipe lately.
I didn’t want to include these changes before they were included in a new ctapipe release.

However, since yesterday, there is a new ctapipe release !
So I think you are right and this should be updated.

I can do that next week I think.

Shall we merge this PR knowing that it works with ctapipe v0.6.1 or shall we wait that I have time to update it?

Thomas

misabelber:
Ahh, ok! No problem. On my side it's ok if @rlopezcoto  wants to merge it this way because I don't think the coordinates will affect to the things I'm checking right now. Thank you!

vuillaut:
Side comment: we should probably fix a ctapipe release in the requirements… 

Thomas

rlopezcoto:
Hi,

I also noticed some functionalities (and notebooks!) are not working anymore, but was waiting for ctapipe core developers to finish with the refactoring. I would go for a merging now, I would just like to know if somebody is aware of what was agreed on ctapipe for the config files.

vuillaut:
Hi @rlopezcoto 
There is nothing decided (or discussed) among the ctapipe dev. team concerning config files.
I would like to move forward with this anyway, even if it means throwing it away later and replacing it with the solution chosen for ctapipe.

rlopezcoto:
ok, thanks, I'm happy with it. I'll merge it.

PR #72, opened 2019-02-04
adding scripts for the r0 to dl1 reco part

vuillaut:
Adding two simple scripts for the first part of the reconstruction:

- r0 to dl1
- merging hfd5 files (especially after r0 to dl1)


rlopezcoto:
ok this one can be approved right away

PR #71, opened 2019-02-01
WIP: Getting mc info using `pyeventio` to compute IRF

vuillaut:
Hi @misabelber @rlopezcoto 

I am opening this PR to work together on building functions to get mc info and later compute IRF.
You may contribute by pulling and pushing on the PR.
Depending how you called your remote, you may use:
```
git fetch upstream pull/71/head:mc_info
git checkout mc_info
```
Update code and push again:
```
git push upstream pull/71/head
```

Adding a function to recompute the energy distribution from mc info and a notebook as an example.

PR #73, opened 2019-02-05
Notebook to calculate sensitivity with real data

misabelber:
Hi!

I've modified the notebook created by @rlopezcoto (Calculate_sensitivity.ipynb) to be used with simulated data. You can find the simtelfiles (for gammas and protons) I've used in the notebook here:

- https://www.mpi-hd.mpg.de/personalhomes/bernlohr/cta-raw/Prod-3/LaPalma/flashcam-prod3h/gamma_20deg_0deg_run8___cta-prod3-lapalma-2147m-LaPalma-FlashCam.simtel.gz

- https://www.mpi-hd.mpg.de/personalhomes/bernlohr/cta-raw/Prod-3/LaPalma/flashcam-prod3h/proton_20deg_0deg_run194___cta-prod3-lapalma-2147m-LaPalma-FlashCam.simtel.gz

And the reconstructed dl2 LST events can be found in cta-lstchain-extra/reco/sample_data/dl2.

I think units are correct along the notebook, but  pay attetion if you review it to any mistake. 
I've faced the problem that the number of LST triggered events compared to the simulated number of events is pretty low (specially for protons), so I couldn't manage many bins in energy/hadroness/theta_2 because some of them remained empty. 

Cell 25 looks bad when changing the binning and I wasn't able to properly fix this yet.

Do you think these results make any sense?


rlopezcoto:
Thanks @misabelber and @vuillaut for the PR!
@misabelber: Thanks for the sensitivity notebook using real data, I think we can substitute the old one by the one you are pushing here. Could you change the paths to the files from '/home/queenmab/DATA/LST1/Gamma/.....' to '../../cta-lstchain-extra/.....' so it is compatible to anybody using it (people may not know these files are in cta-lstchain-extra).

@vuillaut: thanks for the implementation of the weighting function and the notebook, do you think it would be better to work on the minimization functions in a different PR?

If this is ok, I could merge this PR. To avoid conflicts, it may be better to merge this one that already include the changes from #71 and (I think) there would not be any authorship conflicts, would that be ok?

vuillaut:
Hi.
I think you can merge this already and then we can move forward and build on top of it.
At least the basic tools will be this notebook for the time being.

PR #75, opened 2019-02-10
Included function for gain selection and spectral weighting

misabelber:
Hello!
I have done some implementations based on the latests notebooks we've been working on:

- I've included a function in the calib module to do the gain selection after calibration. It's the same I showed in the step_by_step_dl0_to_dl1 notebook, it looks for waveform samples that go over a thresholds and for those pixels, it changes to low gain charge values. 
I've included this selection in the r0_to_dl1() function from dl0_to_dl1 moduel, so the selection it's done right before calculating the hillas and timing parameters. 
About this: I've searched in ctapipe for a function that does the gain selection, and in fact it exist  (ctapipe/ctapipe/calib/camera/gainselection.py, there's a class named "GainSelector") and basically does the same thing but it reads the thresholds depending on each camera type from a file which is inside ctapipe-extra. The problem is I think the number for the LST cam is wrong (it says 100, but it should be using the number of counts (4095), not photoelectrons) and I couldn't manage to make it work, even editing the file myself.

- I've included an option in dl0_to_dl1 module to chose between the standard calibration (CameraCalibrator from ctapipe) and our custom calibration. Probably using the standard is the best option. 

- Included functions in dl0_to_dl1 module to calculate the spectral weight of an event.


misabelber:
I've managed to make work the GainSelector class of ctapipe. Still, it is necessary to manually set the threshold for the LSTCam to 4094, because the one in the default file is wrong (100). I will open an issue in ctapipe to comment this.

rlopezcoto:
For this PR, I would prefer to wait a little bit to see what is the answer about this Issue:
https://github.com/cta-observatory/ctapipe/issues/962
is that ok, @misabelber?

misabelber:
Actually I think it can be merged so the new functions are already in the code, because I think any answer to the issue would barely affect one line of the code (it has more to do with the tabulated thresholds stored in ctapipe-extra than with the usage of GainSelector) but it's up to you :)

rlopezcoto:
sorry, I just realized I forgot to send these comments that I wrote last week...
Now there seem to be some conflicts, could you please work on the comments, solve the conflicts and I'll merge it (no need to wait for ctapipe issue to be solved...)

misabelber:
Hi! I just answered the comments and solved conflicts (it was just because of the change of name from containers to lstcontainers module). Thank you!!

PR #78, opened 2019-02-12
PR related only to containers and lsteventsource module 

FrancaCassol:
Moved old containers module to lstcontainers module
Added to lstchain the last version of the lsteventsource module
(eventually to be PR to ctapipe)
Added a container module with the present version of the monitoring data
container:

lstchain.io.containers.MonitoringDataContainer:
                   flatfield.*: Relative flat field data
                    pedestal.*: Pedestal data
                pixel_status.*: Container of masks with pixel status

Hopefully I changed everywhere it was necessary the "containers" name with "lstcontainers. 

Is there a reason why there are no test directories? (even if sometime heavy they are useful to easily test the code)


rlopezcoto:
I've been testing the PR and the changes look good to me. It would be good, however, that somebody from the main core developers team have a look and comment on the ctapipe compliance of the PR. @vuillaut, could you also have a look?
There are problems running notebooks due to the coordinates refactoring in ctapipe. We should solve this in a separate PR...

vuillaut:
Hi.
Tests are passing and the scripts are running, so the refactoring is ok.
I now need to review the new code.

lstchain/io/containers.py
```diff
@@ -1,154 +1,268 @@
...
+]
+
+
+class R0CameraContainer(Container):
+    """
+    Storage of raw data from a single telescope
+    """
+    trigger_time = Field(None, "Telescope trigger time, start of waveform "
+                               "readout, None for MCs")
+    trigger_type = Field(0o0, "camera's event trigger type if applicable")
```
vuillaut: Out of curiosity, why `0o0` ?

lstchain/io/lsteventsource.py
```diff
@@ -0,0 +1,379 @@
...
+from ctapipe.instrument import TelescopeDescription, SubarrayDescription, \
+    CameraGeometry, OpticsDescription
+from ctapipe.io.eventsource import EventSource
+from .containers import LSTDataContainer
+
+
+__all__ = ['LSTEventSource']
+
+
+class LSTEventSource(EventSource):
```
vuillaut: It would be nice to have a small test file to unit test `LSTEventSource`.

lstchain/io/lsteventsource.py
```diff
@@ -0,0 +1,379 @@
...
+            event_container.event_counter.append(unpacked_counter[2])
+            event_container.trigger_counter.append(unpacked_counter[3])
+            event_container.local_clock_counter.append(unpacked_counter[4])
+
+        event_container.chips_flags = event.lstcam.chips_flags
+        event_container.first_capacitor_id = event.lstcam.first_capacitor_id
+        event_container.drs_tag_status = event.lstcam.drs_tag_status
+        event_container.drs_tag = event.lstcam.drs_tag
+
+    def fill_r0_camera_container_from_zfile(self, container, event):
```
vuillaut: `container` is not self-explanatory.
A short docstring would help understanding on what objects these functions are working.

lstchain/io/lsteventsource.py
```diff
@@ -0,0 +1,379 @@
...
+
```
vuillaut: What do you think about asserting the type of container passed here?

lstchain/io/lsteventsource.py
```diff
@@ -0,0 +1,379 @@
...
+            n_gains = 1
+        elif event.waveform.shape[0] == (self.camera_config.num_pixels *
+                                         container.num_samples * 2):
+            n_gains = 2
+        else:
+            raise ValueError("Waveform matrix dimension not supported: "
+                             "N_chan x N_pix x N_samples= '{}'"
+                             .format(event.waveform.shape[0]))
+
+        container.pixel_status = np.zeros([self.n_camera_pixels])
```
vuillaut: This line fails.
Currently `type(self.data) == lstchain.io.containers.LSTDataContainer` (line 280)
However `LSTDataContainer` inherits from `ctapipe.io.containers.DataContainer`.
Thus `type(self.data.r0) == ctapipe.io.containers.R0Container` which does not have `pixel_status`.

lstchain/io/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .containers import DL1ParametersContainer
+from .lstcontainers import DL1ParametersContainer
 from .config import read_configuration_file
```
vuillaut: You probably want the new modules imported as well, no?
(especially LSTEventSource for users)

lstchain/io/containers.py
```diff
@@ -1,154 +1,268 @@
...
-    """
-    dx = Field(nan, 'x coordinate of the disp_norm vector')
-    dy = Field(nan, 'y coordinate of the disp_norm vector')
-
-    angle = Field(nan, 'Angle between the X axis and the disp_norm vector')
-    norm = Field(nan, 'Norm of the disp_norm vector')
-    sign = Field(nan, 'Sign of the disp_norm')
-    miss = Field(nan, 'miss parameter norm')
+from astropy import units as u
+#from astropy.time import Time
```
vuillaut: Can be deleted.

lstchain/io/containers.py
FrancaCassol: :-) no idea, I wondered also the meaning of this.

Anyway, I just changed the code in order not to have the R0CameraContainer in lstchain

lstchain/io/lsteventsource.py
FrancaCassol: This test is already in ctapipe, no problem to add it in lstchain, where should I put it?
Should we create a test subdirectory?

lstchain/io/lsteventsource.py
FrancaCassol: I agree, I will add it

lstchain/io/lsteventsource.py
vuillaut: Are you therefor adding this piece of code here for future developments?
Otherwise better just call the ctapipe version which is tested no?

If the test subdirectory is missing, you can create it yes.

FrancaCassol: Yes, the problem of ctapipe is that PR acceptance takes too much time also they want to take specific telescope readers out of ctapipe and have them as plugin. I am actually waiting for Dominik to indicate how to proceed. In the meanwhile, I would like to share in LST this new version of the reader

vuillaut: Ok great.
In this case, a unit test would be welcome.
Thanks.

lstchain/io/lsteventsource.py
FrancaCassol: You are write, I cleaned this part of code.

FrancaCassol: "right"

vuillaut:
Hi @FrancaCassol 

Thank you for taking my review into account.
I have made a small PR to your branch to add `LSTEventSource` in the importable functions.

The rest of the code looks good and is functional.

rlopezcoto:
Hi @vuillaut, thanks for reviewing it. I could not see your changes in the PR, am I missing them or is there something missing. After this I'll merge

vuillaut:
The PR is on @FrancaCassol repo:
https://github.com/FrancaCassol/cta-lstchain/pull/1/files

I am not a maintainer so I can't propose changes to this repo directly.

FrancaCassol:
I got it. Done. 
Ruben, I think you can now accept the PR.

PR #82, opened 2019-02-13
Fix bug with merge hdf5 files

vuillaut:


rlopezcoto:
Hi @vuillaut, thanks! Could you report what is exactly the problem with the keys access of the hdf5 files before I merge it?

vuillaut:
- For an unknown reason (maybe related to second point), for some files, `pandas.HDFStore.keys()` was not returning any key.
- Moreover, I realized while testing on other files that the present code wasn't able to crawl to find deeper datasets (e.g. it could find `/alpha` but not `beta/alpha`)

This PR fixes both issues.
However, I had to use `h5py`. It's the only place where we use it, should I add it to the install requirements?

rlopezcoto:
h5py is a dependency from ctapipe, so it should be installed whenever we install ctapipe and cta-lstchain needs to use ctapipe, so I would say that it is not needed.
Ok, thanks for the report, it makes it easier to keep track of the problems and solutions. If there are no additional changes, I'll merge

vuillaut:
Ho I thought `h5py` was removed from ctapipe but I see it is still in the environment.
So there is no problem there, sorry for that.
Thanks.

Thomas

vuillaut:
ps: there are no additionnal changes btw so this can be merged

PR #83, opened 2019-02-19
Reader

FrancaCassol:
Corrected bug:  dragon counters where not  initialised at each event


rlopezcoto:
Hi, good looks to me, thanks. Maybe we can think on adding some test modules to catch these bugs

rlopezcoto:
shall I wait for some test modules to be thought/added? Otherwise we could open and issue and merge

FrancaCassol:
Hi Ruben,
please merge this version (there is a working test).

PR #86, opened 2019-02-28
Low level calib drs4

pawel21:
I have created code to low level calibration:

- pedestal subtraction

- time lapse correction (delta_t)

- interpolate spikes.

I added notebook `low_level_corrections.ipynb`. 
Pedestal file needed to pedestal subtraction, I added in cta-lstchain-extra.


misabelber:
Thank you Pawel! I've been waiting for this : ). I'll check the notebook as soon as possible.

FrancaCassol:
Hi Pawel,
in order to be consistent with ctapipe I would better place your code in a "camera" subdirectory of the "calib" directory:

lstchain/calib/camera

where I will put also the flatf-field calibration code.

pawel21:
@FrancaCassol 
Done.
I'm sorry, it was my mistake.

lstchain/calib/camera/r0.py
```diff
@@ -0,0 +1,288 @@
...
+                        time_diff = time_now - last_time_array[nr_clus, gain, pix, posads]
+                        val = waveform[gain, pixel, k] - ped_time(time_diff / (133.e3))
+                        waveform[gain, pixel, k] = val
+                    if (k < 39):
+                        last_time_array[nr_clus, gain, pix, posads] = time_now
+
+@jit
+def ped_time(timediff):
+    return 29.3 * np.power(timediff, -0.2262) - 12.4
+
```
rlopezcoto: where are these numbers (29.3, 0.2262, 12.4) coming from? I would add a comment about those

lstchain/calib/camera/r0.py
```diff
@@ -0,0 +1,288 @@
...
+                position = int((fc_cap[nr, gain, pix]) % size4drs)
+                waveform[gain, pixel, :] = \
+                    (event_waveform[gain, pixel, :] -
+                    pedestal_value_array[nr, gain, pix, position:position + 40])
+    return waveform
+
+@jit(parallel=True)
+def do_time_lapse_corr(waveform, expected_pixel_id, local_clock_list, fc, last_time_array, number_of_modules):
+    size4drs = 4096
+    for nr_clus in prange(0, number_of_modules):
```
rlopezcoto: All these for loops are likely making the code slower, would there be a more pythonic way of doing it using numpy functionalities instead of going though all the loops?

lstchain/calib/camera/r0.py
```diff
@@ -0,0 +1,288 @@
...
+        Get first capacitor values from event for nr module.
+        Parameters
+        ----------
+        event : `ctapipe` event-container
+        nr_module : number of module
+        """
+        fc = np.zeros((2, 7))
+        first_cap = event.lst.tel[0].evt.first_capacitor_id[nr_module * 8:
+                                                            (nr_module + 1) * 8]
+        for i, j in zip([0, 1, 2, 3, 4, 5, 6], [0, 0, 1, 1, 2, 2, 3]):
```
rlopezcoto: could you add a comment on where are these arrays coming from?

lstchain/calib/camera/r0.py
```diff
@@ -0,0 +1,288 @@
...
+            stored in a numpy array of shape
+            (n_clus, n_gain, n_pix).
+        n_modules : int
+            Number of modules
+        """
+        roisize = 40
+        size4drs = 4096
+        n_gain = 2
+        n_pix = 7
+        for nr_clus in prange(0, n_modules):
```
rlopezcoto: Same as comment below about the loops, although here it seems to be more difficult to simplify...

misabelber:
Hi Pawel,
I'm very interested in understanding how to extract the information of the first capacitor id, 
why is that curious structure in the data? I see from your function _get_first_capacitor() that from event.lst.tel[0].evt.first_capacitor_id is possible to get the first capacitor id for every pixel and two gains, but this array has a shape of 2120 (instead of (1855,2) which is what I would have expected). Can you know why is that and why that "[0, 1, 2, 3, 4, 5, 6], [0, 0, 1, 1, 2, 2, 3]" structure is needed? Thank you very much!

pawel21:
Hi Mabel,
This is according to the Dragon data format. Information you can find here: https://portal.cta-observatory.org/WG/lst/DAQ/SitePages/Home.aspx (Dragon-v5_manual.pdf page 28)
In data format of Dragon board:
We have 7 pixels (for each we have first capacitor value) and DRS_TAG (whatever it is, not important for us) value. In LST we have 265 boards, so 8*265 = 2120.
We have the same value of first capacitor for two pixels (HG-high gain, LG-low gain): 
HG0,HG1 - first_capacitor_id[0]
HG2,HG3 - first_capacitor_id[1]
HG4,HG5 - first_capacitor_id[2]
HG6,TAG - first_capacitor_id[3]
LG0, LG1 - first_capacitor_id[4]
LG2, LG3 - first_capacitor_id[5]
LG4, LG5 - first_capacitor_id[6]
LG6, TAG - first_capacitor_id[7]
In my code I want have store value of first capacitor for each module in array with shape (2, 7) (2 gain: low and high, 7 pixel)
I'm sending you email with discussion about it.
If you have any further questions, do not hesitate to write.

lstchain/calib/camera/r0.py
pawel21: @rlopezcoto Thank you,
I will improve code according comments.

misabelber:
Hi Pawel!
I've tried the notebook and seems to work fine. Are you planning to include somewhere the script to create pedestal fits files? I remember from the LST Bootcamp that you provided a script to do that, but I guess probably there have been changes since then.

pawel21:
Hi Mabel,
In next step, I will add script to create pedestal file in cta--lstchain. Now, you can use script 'create_pedestal_file.py' from my github: https://github.com/pawel21/low_level_calib_drs4/tree/master/Pedestal_drs4. Let me know how it works for you

rlopezcoto:
Hi Pawel,
it would be nice if you could also add this script to cta-lstchain, so we do not depend on personal scripts tu run some functionalities. 
In any case, if you need to clean up or make changes, do not worry, I would merge this PR as soon as it is a bit cleaner/commented and the comments are taken into account and you can work on the script a bit later

pawel21:
Hi Ruben,

Of course, I will add script to create pedestal file.

misabelber:
Hi Pawel! I've tried the script for creating the pedestal file and it works fine, also in the notebook, thank you!

lstchain/calib/camera/r0.py
pawel21: Done

lstchain/calib/camera/r0.py
pawel21: Numbers come from curve fitting. I added comment. Now, this is fixed numbers for all pixels, but I will investigate how those coefficients depends of pixel.

lstchain/calib/camera/r0.py
pawel21: I improved code a bit. But now, I would like leave code as simple as possible, because in next step I will modify code to take account another small effects.

pawel21:
Dear All,

I've added script to create pedestal file.
I've changed a lot in code (in function to subtract pedestal), because We found problem with pixel mapping.
Now works fine, and we can use pedestal file from different runs.
I will swap pedestal file in cta-lstchain-extra.
@rlopezcoto could you check code? Is OK or should I fix something?

Thank you

rlopezcoto:
Hi @pawel21, thanks for all the work. It looks goo to me now. @misabelber has also been having a look, so I'll wait for her to give the ok as well and then I'll merge

misabelber:
Everything works fine for me so let's go merge it :)

PR #87, opened 2019-03-01
Fix coordinate transformations to new ctapipe frames

misabelber:
Hi!
I've done the changes necessary to make the coordinates transformations work  again with the lastest changes in ctapipe (plus the substitution of HorizonFrame to AltAz).
I've checked and everything seems to work fine, but if you can, take a look and confirm that it makes sense.

(Also did a small change on dl0_to_dl1 (which actually only affects a deprecated function) due to the elimination of EventSourceFactories)


lstchain/reco/utils.py
```diff
@@ -201,17 +201,18 @@ def get_event_pos_in_camera(event, tel):
     -------
     (x, y) (float, float): position in the camera
     """
-    array_pointing = HorizonFrame(alt=event.mcheader.run_array_direction[1],
+    array_pointing = AltAz(alt=event.mcheader.run_array_direction[1],
```
thomasgas: The `array_pointing' should be defined in this way
```python
from astropy.coordinates import SkyCoord, EarthLocation, AltAz
from astropy.time import Time

location = EarthLocation.of_site('Roque de los Muchachos')
obstime = Time('2018-11-01T02:00')
horizon_frame = AltAz(location=location, obstime=obstime) 
array_pointing = SkyCoord(az=event.mcheader.run_array_direction[0],
                          alt=event.mcheader.run_array_direction[1], 
                          frame=horizon_frame)
```
so that at first you define the frame and then specify the coordinates in that frame.

lstchain/reco/utils.py
```diff
@@ -201,17 +201,18 @@ def get_event_pos_in_camera(event, tel):
...
                                   az=event.mcheader.run_array_direction[0])
-    event_direction = HorizonFrame(alt=event.mc.alt.to(u.rad),
+    event_direction = AltAz(alt=event.mc.alt.to(u.rad),
```
thomasgas: this should become:
```python
event_direction = SkyCoord(alt=event.mc.alt, az=event.mc.az, frame=horizon_frame)
```
And I don't think you need to do any change to u.rad

lstchain/reco/utils.py
```diff
@@ -277,16 +278,17 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
     y = np.array([1,1]) * u.m
 
     """
-    pointing_direction = HorizonFrame(alt=pointing_alt, az=pointing_az)
+    pointing_direction = AltAz(alt=pointing_alt, az=pointing_az)
```
thomasgas: same change as above here

lstchain/reco/utils.py
thomasgas: so you define the `horizon_frame` just once and then you re-use it where you need.

lstchain/reco/utils.py
```diff
@@ -303,20 +305,18 @@ def sky_to_camera(alt, az, focal, pointing_alt, pointing_az):
     -------
 
     """
-    pointing_direction = HorizonFrame(alt=pointing_alt, az=pointing_az)
-
-    event_direction = HorizonFrame(alt=alt, az=az)
-    nom_frame = NominalFrame(array_direction=pointing_direction,
-                             pointing_direction=pointing_direction)
+    pointing_direction = AltAz(alt=pointing_alt, az=pointing_az)
```
thomasgas: same change here to be done here

lstchain/reco/utils.py
```diff
@@ -277,16 +278,17 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
...
-                                        y=pos_y/focal * u.rad,
-                                        )
+    camera_frame = CameraFrame(focal_length=focal,
+                               telescope_pointing=pointing_direction
+    )
 
-    return source_pos_in_camera.transform_to(pointing_direction)
+    camera_coord = SkyCoord(pos_x, pos_y, frame=camera_frame)
 
+    horizon = camera_coord.transform_to(AltAz())
```
thomasgas: here you can also use the `horizon_frame` defined above without  using a new AltAz()

lstchain/reco/utils.py
```diff
@@ -303,20 +305,18 @@ def sky_to_camera(alt, az, focal, pointing_alt, pointing_az):
...
-                              x=event_dir_nom.x.to(u.rad).value * focal,
-                              y=event_dir_nom.y.to(u.rad).value * focal,
-                              )
-
-    # return focal * (event_dir_nom.x.to(u.rad).value, event_dir_nom.y.to(u.rad).value)
+    camera_frame = CameraFrame(focal_length=focal,
+                               telescope_pointing=pointing_direction
+    )
+    
+    event_direction = AltAz(alt=alt, az=az)
```
thomasgas: same change here

lstchain/reco/utils.py
```diff
@@ -201,17 +201,18 @@ def get_event_pos_in_camera(event, tel):
...
                                    az=event.mc.az.to(u.rad))
 
-    nom_frame = NominalFrame(array_direction=array_pointing,
-                         pointing_direction=array_pointing)
-
-    event_dir_nom = event_direction.transform_to(nom_frame)
     focal = tel.optics.equivalent_focal_length
-    return focal * (event_dir_nom.x.to(u.rad).value, event_dir_nom.y.to(u.rad).value)
+
+    camera_pos = sky_to_camera(event_direction.alt, event_direction.az,
```
thomasgas: you can transform directly from AltAz to CameraFrame using the `transform_to()` method.
So you have your coordinates in the `camera_frame` and you can just do (provided `array_pointing` is in AltAz):
`camera_pos = array_pointing.transform_to(camera_frame)`

rlopezcoto:
Hi Mab,

Thank you very much for taking care of this. I did not have time to check
the code yet, but do you or anybody else think it would be a good idea to
include some test modules to check that the coordinates transformation is
1) working 2) working properly?

thomasgas:
> Hi Mab, Thank you very much for taking care of this. I did not have time to check the code yet, but do you or anybody else think it would be a good idea to include some test modules to check that the coordinates transformation is 1) working 2) working properly? 

Hi ruben! All the coordinates tests are already in ctapipe, so i don't think we should have those tests here too. Were you thinking about other tests?

thomasgas:
@misabelber In order to show a little bit better how the coordinates transformation works, I'm working on a notebook to show this. It's work in progress in a PR in ctapipe and I'll soon finish it :sweat_smile:

vuillaut:
Thank you @thomasgas for the review.

rlopezcoto:
Hi Thomas,

Yes, I was thinking about that, although if coordinates are massively
changed again, we will lose compatibility as we already did. I was thinking
on a test to check that the modules are at least callable.
As for the second point, I guess it is properly handled on ctapipe already
(although did not have the chance to follow all the massive refactoring)

vuillaut:
Hi.
There is a test already `test_change_frame_camera_sky()` to check that going from camera to sky and back works. That should do I think.

misabelber:
Thank you very much @thomasgas ! I will now take care of the comments you did in the review.
About the tests, yes, I used the tests functions in lstchain/tests/test_lstchain.py

misabelber:
Hi! I updated the functions to the changes suggested by @thomasgas . They work fine when doing the tests on tests/test_lstchain.py

rlopezcoto:
Ok, thanks, changes look good, would it be possible to also get an update of all the calls in the different notebooks involving the coordinate frames?

PR #88, opened 2019-03-05
Fix import

misabelber:
There was a mistake in the imports of dl0_to_dl1 module (event_source was giving error due to refactoring of ctapipe.)


vuillaut:
Hi @misabelber .
You just removed the import here, this won't fix the issue: all the calls to `event_source` in the file will now fail.

misabelber:
It is imported below! It was being imported twice

rlopezcoto:
There are currently two event_source calls in the file:

from ctapipe.io.eventsource import event_source
from ctapipe.io import event_source

@misabelber: is the one you removed the outdated one? have you tested the functioning of the script?

misabelber:
Yes, I tested it and now it won't complain for not being able to import event_source from ctapipe.io.eventsource

vuillaut:
Ha, my bad that wasn't clear in the PR.
Sorry for that, looks good then.

rlopezcoto:
Our messages just crossed... if the code is tested, I think it can be merged without any further changes

Issue #91, opened 2019-03-06
Undefined name 'sign' in line 427 of utils.py


dneise:
https://github.com/cta-observatory/cta-lstchain/blob/aec11315151d75a64e4d2601a9a221129835ccf4/lstchain/reco/utils.py#L425-L431

So for `x == 0` this function will raise an exception.

dneise:
Fixed by #95

PR #94, opened 2019-03-06
Solve definition of sign before calling it

rlopezcoto:
Solves #91


vuillaut:
Hi @rlopezcoto .
Thanks to #91  I actually realised there was a big mistake in this function (my bad I guess I was the author).
This is solved in #95

PR #95, opened 2019-03-06
Fix wrong calculations in `cartesian_to_polar`

vuillaut:
The calculation was actually wrong.
Also solves #91.

I added a unit test to check that this is now working as expected.


rlopezcoto:
ok, thanks @vuillaut, I'll merge this one and cancel #94

PR #98, opened 2019-03-08
Solve reweighting error

rlopezcoto:
When reweighting the MC to a Crab-like/proton-like spectrum, there was a mistake because of dividing by E0 twice. There was also another mistake on the printing of the rate.
@misabelber could you please have a look? I'll slowly start to get all this into functions as we agreed some time ago...


rlopezcoto:
@vuillaut I think I found a bug in `lstchain/mc/mc.py`, I wrote a test unit to check it. Also, shouldn't we use astropy units here?

rlopezcoto:
Anybody against merging?

lstchain/mc/tests/test_mc.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
+from numpy.testing import assert_allclose
```
vuillaut: if you already import `numpy`, I'd suggest using directly `np.testing.assert_allclose` in the test.

lstchain/mc/tests/test_mc.py
```diff
@@ -0,0 +1,16 @@
...
+from lstchain.mc import power_law_integrated_distribution
+
+def test_integrated_distribution():
+
+    emin = 50.     # u.GeV
+    emax = 500.e3  # u.GeV
+    Nevents = 1e6
+    spectral_index = -2
```
vuillaut: Note that this test would actually pass with the previous version of the code, because `spectral_index = -2` is a special case.
I would suggest testing with another value (not integer).

rlopezcoto: Thanks for the review @vuillaut. I took into account your comments in the PR

vuillaut: Well, thank you for the fix.
Good to merge.

Issue #93, opened 2019-03-06
do not use bare except


dneise:
There are some bare excepts in `dl1_to_dl2.py` and in other places, like 10 or so ...

misabelber:
Hi, I will take a look at this as soon as possible.

rlopezcoto:
Solved in #101

PR #100, opened 2019-03-12
Fix bare excepts

misabelber:
Hi!
I tried to solve issue #93 and set a certain error in the bare excepts. I don't have a lot of experience in error handling in Python so I'm not completely sure if I used the proper error type. If anyone can take a look and see if seems correct, I will be thankful.


misabelber:
I don't know why for some reason commits from PR #98 are appearing here :s

rlopezcoto:
Thanks @misabelber. @dneise could you please have a look at the correction?
Regarding the appearing of #98 commits... maybe you picked the upstream branch and merge it with your master before creating fix-excepts?

misabelber:
Probably yes, and now I don't know how to fix it...

rlopezcoto:
Since there is only one real change, probably the easiest would be to close this PR, stash all the changes and merge it with the upstream/master. Not sure if the order I'm putting here is correct but should be something like:

`git stash drop`
`git fetch upstream`
`git reset --hard upstream/master`
`git push --force`

and then you can add your changes. If you manage to have your branch even with cta-lstchain, it should be solved and you can commit the new changes and cancel this PR.

misabelber:
Ok! I'll try it

PR #101, opened 2019-03-12
Fix bare excepts

misabelber:
Reopen of this pull request (sorry!)

I tried to solve issue #93 and set a certain error in the bare excepts. I don't have a lot of experience in error handling in Python so I'm not completely sure if I used the proper error type. If anyone can take a look and see if seems correct, I will be thankful.

### DIFF
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -72,7 +72,7 @@ class DL1ParametersContainer(Container):
         for k in hillas.keys():
             try:
                 self[k] = hillas[k]
-            except:
+            except StopIteration:
                 print("{} cannot be copied in container".format(k))
                 pass
 
@@ -91,7 +91,7 @@ class DL1ParametersContainer(Container):
             self.mc_x_max = event.mc.x_max
             self.mc_alt_tel = event.mcheader.run_array_direction[1]
             self.mc_az_tel = event.mcheader.run_array_direction[0]
-        except:
+        except ValueError:
             print("mc information not filled")
 
     def fill_event_info(self, event):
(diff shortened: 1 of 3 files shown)


rlopezcoto:
Thanks @misabelber! @dneise could you please have a look to this fix?

lstchain/io/lstcontainers.py
```diff
@@ -72,7 +72,7 @@ def fill_hillas(self, hillas):
         for k in hillas.keys():
             try:
                 self[k] = hillas[k]
-            except:
+            except StopIteration:
```
dneise: I think StopItereation is wrong here.

lstchain/reco/dl0_to_dl1.py
```diff
@@ -93,7 +93,7 @@ def get_dl1(calibrated_event, telescope_id):
             dl1_container.set_disp([dl1_container.src_x, dl1_container.src_y], hillas)
             return dl1_container
 
-        except:
+        except RuntimeError:
```
dneise: Which of the functions in this try...except block raises a `RuntimeError` ?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -106,7 +106,7 @@ def train_energy(train,
             config = read_configuration_file(config_file)
             regression_args = config['random_forest_regressor_args']
             print("Configuration loaded from {}".format(config_file))
-        except:
+        except OSError:
```
dneise: Why do you expect an OSError here?

misabelber:
Haha, I tried and I failed! I guess I'm not understanding the error types...
I think maybe @vuillaut can help here. Thomas, I think you where who introduced most of these bare excepts during one of the refactorings of the code, do you know what kind of exceptions are going to be raised in each case? Thank you very much!

rlopezcoto:
Solved in #104

PR #104, opened 2019-03-15
Bare excepts

dneise:
I looked a bit at these bare excepts and think most of them hurt more than they help. 
Therefore here I propose to remove many of them and actually raise exceptions in case a function is called with wrong input.

### DIFF
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -96,12 +96,19 @@ class DL1ParametersContainer(Container):
         self.event_id = event.r0.event_id
 
     def get_features(self, features_names):
-        return np.array([self[k].value if isinstance(self[k], Quantity) else self[k]
-                         for k in features_names])
+        return np.array([
+            self[k].value
+            if isinstance(self[k], Quantity)
+            else self[k]
+            for k in features_names
+        ])
 
     def set_mc_core_distance(self, event, telescope_id):
         tel_pos = event.inst.subarray.positions[telescope_id]
-        distance = np.sqrt((event.mc.core_x - tel_pos[0]) ** 2 + (event.mc.core_y - tel_pos[1]) ** 2)
+        distance = np.sqrt(
+            (event.mc.core_x - tel_pos[0]) ** 2 +
+            (event.mc.core_y - tel_pos[1]) ** 2
+        )
         self.mc_core_distance = distance
 
     def set_disp(self, source_pos, hillas):
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -91,12 +91,9 @@ class DL1ParametersContainer(Container):
             print("mc information not filled")
 
     def fill_event_info(self, event):
-        try:
-            self.gps_time = event.trig.gps_time
-            self.obs_id = event.r0.obs_id
-            self.event_id = event.r0.event_id
-        except:
-            print("event information not filled")
+        self.gps_time = event.trig.gps_time
+        self.obs_id = event.r0.obs_id
+        self.event_id = event.r0.event_id
 
     def get_features(self, features_names):
         return np.array([self[k].value if isinstance(self[k], Quantity) else self[k]
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -70,11 +70,7 @@ class DL1ParametersContainer(Container):
         (e.g. conserving unit definition)
         """
         for key in hillas.keys():
-            try:
-                self[key] = hillas[key]
-            except AttributeError:
-                print("{} cannot be copied in container".format(key))
-                pass
+            self[key] = hillas[key]
 
     def fill_mc(self, event):
         """
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -65,15 +65,15 @@ class DL1ParametersContainer(Container):
         """
         fill Hillas parameters
 
-        hillas: HillasContainer
+        hillas: HillasParametersContainer
         # TODO : parameters should not be simply copied but inherited
         (e.g. conserving unit definition)
         """
-        for k in hillas.keys():
+        for key in hillas.keys():
             try:
-                self[k] = hillas[k]
-            except:
-                print("{} cannot be copied in container".format(k))
+                self[key] = hillas[key]
+            except AttributeError:
+                print("{} cannot be copied in container".format(key))
                 pass
 
     def fill_mc(self, event):
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -87,7 +87,7 @@ class DL1ParametersContainer(Container):
             self.mc_x_max = event.mc.x_max
             self.mc_alt_tel = event.mcheader.run_array_direction[1]
             self.mc_az_tel = event.mcheader.run_array_direction[0]
-        except:
+        except (AttributeError, IndexError):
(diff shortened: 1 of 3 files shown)


lstchain/reco/dl0_to_dl1.py
```diff
@@ -68,107 +75,120 @@ def get_dl1(calibrated_event, telescope_id):
...
-                    l = np.rad2deg(np.arctan2(dl1_container.length, foclen))
-                    dl1_container.width = w.value
-                    dl1_container.length = l.value
-                    
-                    if w>=0:
-                        writer.write(camera.cam_id, [dl1_container])
-
-
-    
+                try:
```
vuillaut: Maybe I am wrong but isn't possible that `get_dl1` fails or return `None` for another reason than `HillasParameterizationError`?
In this case, the code would break no?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -102,12 +102,8 @@ def train_energy(train,
     The trained model
     """
```
vuillaut: What if a `config_file` is passed but the loading of the config fails?
What is the interest of removing this try/except?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -144,12 +140,8 @@ def train_disp_vector(train, features,
...
     if config_file is not None:
-        try:
-            config = read_configuration_file(config_file)
-            regression_args = config['random_forest_regressor_args']
-            print("Configuration loaded from {}".format(config_file))
-        except:
-            print("Configuration could not be loaded from file. Standard configuration applies")
+        config = read_configuration_file(config_file)
```
vuillaut: Same comment for the next ones

lstchain/io/lstcontainers.py
```diff
@@ -121,13 +121,10 @@ def set_disp(self, source_pos, hillas):
...
 
     def set_timing_features(self, geom, image, peakpos, hillas):
-        try:
-            peak_time = Quantity(peakpos) * u.Unit("ns")
-            timepars = time.timing_parameters(geom, image, peak_time, hillas)
-            self.time_gradient = timepars['slope'].value
-            self.intercept = timepars['intercept']
-        except:
-            pass
+        peak_time = Quantity(peakpos) * u.Unit("ns")
```
vuillaut: I agree that `except` should be handled differently than `pass`.
However, I might want to continue and get the dl1 container even without the reconstructed timing.

lstchain/reco/dl1_to_dl2.py
vuillaut: Sorry I didn't see the comment in your commit.
I see your point now.

lstchain/reco/dl0_to_dl1.py
dneise: If something goes wrong, an exception will be raised, yes. And that is a good thing, because this way bugs can be found and fixed. 

If there is a place where it is "normal" or common that something does not work, then I think it would be good to catch that common-case and ignore it or log it or print it or whatever. But bugs should not be caught in a too narrow net and handled just like the common case "size == 0".

It would be good to have these cases covered by tests. So to refer to the example of the function `hillas_parameters()` it would be good to have (at least) one event with "size == 0" and (at least) one event with "size != 0"  in the test set.

Accordingly, if you are aware of more cases, which are commonly not working, then it would be good to have at least a single example of such a case in the tests. So one can point clearly with a finger to that event and say: Look ... this case ... it raised an exception ... but it is a common case and should not raise ... but it should only log or print.

lstchain/reco/dl1_to_dl2.py
dneise: So I think we can resolve this conversation?

lstchain/reco/dl1_to_dl2.py
dneise: If it is the same comment, then could we discuss it only in one place and accodingly resolve this convversation?

lstchain/reco/dl1_to_dl2.py
vuillaut: Yes 👍

lstchain/reco/dl1_to_dl2.py
vuillaut: yes

rlopezcoto:
It seems that there was a problem with conda http access and that is why checks are not passing, could you please change something in the PR so travis analysis it again? Thanks

rlopezcoto:
Also, would this PR supersede #101 completely?

dneise:
>  would this PR supersede #101 completely?

Yes

dneise:
> could you please change something in the PR so travis analysis it again? 

One can just press this button  in order to restart a build. (I just did it)

Screenshot from 2019-03-18 10-02-00

rlopezcoto:
Thanks @dneise, I think there is still one opened point from @vuillaut, afterwards I can merge this.

lstchain/io/lstcontainers.py
dneise: I think it would be good to have an example event at hand, where the timing cannot be reconstructed. This way, the feature you want can be implemented and it can be ensured that this feature works forever. For me (and possibly for everybody) having such an example event would make it easier to understand what is failing exactly, so we can catch the correct exception. 

At the moment, I can only assume the failing part would be somewhere in
[`ctapipe.image.timing_parameters.timing_parameters()`](https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/image/timing_parameters.py#L15), but looking at that code, I think this cannot fail unless the hillas parameters were not calculated, but we already catch that case. 

Anyway I might be guessing in the entirely wrong direction ... I assume you have an example event somewhere, for which the timing reconstruction fails?

dneise:
Thank you for reviewing this @vuillaut  and @rlopezcoto. I know bare excepts look like just a minor cosmetical issue, and I opened the issue #104 only because my linter freaked out about them, when I had a look at lst-chain some weeks ago. I did not want to cause work for anybody, so I thought I'd quickly fix them here.

lstchain/io/lstcontainers.py
vuillaut: Thank you for replying.
I don't actually have an example in mind.
I think I am dealing with bare expects in a too broad sense.

vuillaut:
Hi.
Sorry, the review was more a way for me to understand the thought process behind the bare expects.
Thank you very much for taking the time to reply to my questions, it's really appreciated.
The code looks good to me and can be merged.

dneise:
> I think I am dealing with bare expects in a too broad sense.

@vuillaut 
I personally believe, too many `try...except` or too many `if isinstance()` can sometimes be seen by C++ developers. Due to the missing strict types in Python one can get the general impression: "OMG basically everything can go wrong here, so I have to check for that". 

I think, taking care of too many cases bloats the code base and makes the "normal" case harder to see, thus making the code more difficult to understand and consequently leads to programmers making mistakes. 

In order to deal with this, I personally try to use the test suite. By having tests, that show what works, and also some tests that show what does not work (including what kind of exceptions are thrown, in such cases) I have the feeling I can a) document the behavior of the product much nicer to my fellow devs, than (often outdated) documentation ever could and b) I am quite sure what works and what not. 

Especially features, which I personally want to work, I have to test. So that a fellow developer who wants to refactor code, which was originally written by me, can do so without fear to break anything. As long as the tests are untouched, everybody can refactor to their own liking and will only make it better ... never worse. 

----

But that all is my personal take on things, I am sure there are other ways to deal with this matter.

PR #111, opened 2019-03-26
allow to load model not strictly indentical with the network

mikael10j:
Modification of the way to load pretrained deep learning models. The only file impacted is utils/gammalearn/gl_utils.py.


rlopezcoto:
Sorry, this was not passing travis builds, that is why I did not accept it, but it was just an http error, I re-run it and now it looks fine

PR #115, opened 2019-04-10
A bit of refactoring for gain selection

vuillaut:
- the code is twice faster this way (removed some loop)
- added a combine image function to update the dl1 container with combined image and peakpos
- added a test for gain selection


vuillaut:
This PR should come after #117 as been merged.

PR #122, opened 2019-04-29
fixing new ctapipe calibration API

vuillaut:
Fixing new ctapipe calibration API.

Also adding the registry download for astropy in tests while investigating further the issue.

### DIFF
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -31,6 +31,7 @@ import math
 from . import utils
 from ..calib.camera import lst_calibration, gain_selection
 from ..io.lstcontainers import DL1ParametersContainer
+from ctapipe.image.extractor import NeighborPeakWindowSum
 
 ### PARAMETERS - TODO: use a yaml config file
 
@@ -43,7 +44,7 @@ threshold = 4094
 # Add option to use custom calibration
 
 custom = False
-cal = CameraCalibrator(extractor_name="NeighborPeakWindowSum")
+cal = CameraCalibrator(image_extractor=NeighborPeakWindowSum())
 
 
 cleaning_method = tailcuts_clean
@@ -144,7 +145,7 @@ def r0_to_dl1(
             if i % 100 == 0:
                 print(i)
             if not custom:
-                cal.calibrate(event)
+                cal(event)
                 # for telescope_id, dl1 in event.dl1.tel.items():
             for ii, telescope_id in enumerate(event.r0.tels_with_data):
                 if custom:
--- a/tests/test_lstchain.py
+++ b/tests/test_lstchain.py
@@ -1,7 +1,5 @@
 from ctapipe.utils import get_dataset_path
 import numpy as np
-from astropy.coordinates import EarthLocation
-EarthLocation._get_site_registry(force_download=True)
 
 def test_import_calib():
     from lstchain import calib
--- a/tests/test_lstchain.py
+++ b/tests/test_lstchain.py
@@ -1,5 +1,7 @@
 from ctapipe.utils import get_dataset_path
 import numpy as np
+from astropy.coordinates import EarthLocation
+EarthLocation._get_site_registry(force_download=True)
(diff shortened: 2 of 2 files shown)


tests/test_lstchain.py
```diff
@@ -1,5 +1,7 @@
 from ctapipe.utils import get_dataset_path
 import numpy as np
+from astropy.coordinates import EarthLocation
+EarthLocation._get_site_registry(force_download=True)
```
rlopezcoto: why do you want to add the EarthLocation to the lstchain test? This seems like a general functionality that could be better suited, if necessary, to ctapipe

vuillaut: There seems to be an issue with astropy recently where the registry is not _always_ downloaded.
I have had a lot of trouble tracking the issue because it does not seem to be fully reproducible (or at least I could not).
See the following [Travis build](https://travis-ci.org/cta-observatory/cta-lstchain/builds/525712712) in a recent PR that failed because
```
astropy.coordinates.errors.UnknownSiteException: "Site 'Roque de los Muchachos' not in database. Use the 'names' attribute to see available sites."
```

I needed to force the download to fix it. I am investigating this but in the meantime added it to have a fix.

vuillaut:
The problem appeared in ctapipe as well.
Discussing with @MaxNoe, he thinks there was an issue with the astropy server (but this seems to happen a lot recently...). If that's the case, the force_download is indeed unnecessary and would not solve the issue anyway, I removed it and now the tests are passing...

tests/test_lstchain.py
MaxNoe: You should probably just define a constant `ORM` with fixed values instead of relying on astropy's database.

MaxNoe:
You could consider adding a constant `ORMLocation` or similar with fixed values, instead of relying on the name resolution.

vuillaut:
> You could consider adding a constant `ORMLocation` or similar with fixed values, instead of relying on the name resolution.

Right. Thank you.
Just for bookkeeping, that would be replacing
```location = EarthLocation.of_site('Roque de los Muchachos')```
by
```location = EarthLocation.from_geocentric(5327448.9957829, -1718665.73869569, 3051566.90295403, u.m)```

**Comment by MaxNoe** [comment]
Or something more human readable using lat, lon and height.

**Comment by vuillaut** [comment]
> Or something more human readable using lat, lon and height.

```location = EarthLocation.from_geodetic(-17.88*u.deg, 28.75833333 *u.deg, 2327*u.m)```

**Comment by MaxNoe** [comment]
The LST position and the whole array will be lower, checking on Google Earth:
The LST is at

 28.761518° North
-17.891495° West
and 2187 m altitude.

What you have seems to be the Willam Herrschel Telescope.

**Comment by vuillaut** [comment]
> The LST position and the whole array will be lower, checking on Google Earth:
> The LST is at
> 
> 28.761518° North
> -17.891495° West
> and 2187 m altitude.
> 
> What you have seems to be the center of the upper observatory, with the Willam Herrschel Telescope and the Isaac Newton Telescope.

I got the position from the astropy 'Roque de los muchachos' to be consistent with what we had.
But sure we could use the true LST1 position :)

File: lstchain/reco/dl0_to_dl1.py [hunk 2]
```diff
@@ -43,7 +44,7 @@
 # Add option to use custom calibration
 
 custom = False
-cal = CameraCalibrator(extractor_name="NeighborPeakWindowSum")
+cal = CameraCalibrator(image_extractor=NeighborPeakWindowSum())
```
misabelber: Ah! Thanks for this, I had my ctapipe version outdated and this was driving me crazy because for some reason it worked but gave wrong hillas parameters... Nice!

vuillaut:
I think this PR can be merged and I'll open a new one to update the position with the LST1 one.

MaxNoe:
orm

rlopezcoto:
I'll merge this one, we continue with the discussion of the LST1 position in #123

PR #123, opened 2019-04-30
changing location to the exact LST1 position

vuillaut:
Following the suggestion from @MaxNoe (thanks) in PR #122 .

This will also solve the issue we have sometimes with getting the position of Roque de Los Muchachos from astropy servers.

The position is taken from Google Earth.
image

### DIFF
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -20,7 +20,7 @@ from astropy.coordinates import AltAz, SkyCoord, EarthLocation
 from astropy.time import Time
 
 
-location = EarthLocation.of_site('Roque de los Muchachos')
+location = EarthLocation.from_geodetic(-17.89139 * u.deg, 28.76139 * u.deg, 2184 * u.m) # position of the LST1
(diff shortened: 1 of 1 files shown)


rlopezcoto:
Are these the same as the LST1 position in simtelarray? I remember that those positions are taken from elsewhere, but cannot remember the name of the survey. For sure @moralejo knows, could you please comment on that?

rlopezcoto:
ok I was talking to @moralejo and he does not remember either, but it should not be very relevant for this anyways and it is better to solve this for tests to pass

moralejo:
> Are these the same as the LST1 position in simtelarray? I remember that those positions are taken from elsewhere, but cannot remember the name of the survey. For sure @moralejo knows, could you please comment on that?

Here is a file with the coordinates of the 4 LSTs of CTA-N (source: J. Cortina, Dec 20, 2016):
[LST1-4_coordinates.txt](https://github.com/cta-observatory/cta-lstchain/files/3136150/LST1-4_coordinates.txt)

Issue #31, opened 2018-11-16
particle type convention


vuillaut:
Hi

It seems the convention for particle type in simtel is:
0: gamma
1: electron
2: muon
100+N+Z: nuclei

Shall we follow the same convention for the pipeline?
Adopting the same convention would probably help making the work more general, however it makes thing more tricky for classification algorithms that usually work with classficiation between 0 and 1. Plus hadroness or gammaness is easy to define and understand in the case of gamma=0 and proton=1.

rlopezcoto:
Hi @vuillaut, I think that this issue is more general than and probably the solution adopted should be taken the same as in ctapipe

PR #124, opened 2019-05-03
Improvements in DL1 reconstruction

vuillaut:
- create the dl1 container outside the event loop and then fill it -> improve speed.
- fill dl1_container with mc_type from `event.mc` info (from simtel info)
- do not compute parameters if image size = 0 after cleaning
- use `mc_type` and removes `hadroness`

### DIFF
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -163,6 +163,9 @@ class DL1ParametersContainer(Container):
         self.src_x = source_pos[0]
         self.src_y = source_pos[1]
 
+    def set_mc_type(self, event):
+        self.mc_type = event.mc.shower_primary_id
+
 
 class DispContainer(Container):
     """
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -38,7 +38,6 @@ class DL1ParametersContainer(Container):
     disp_angle = Field(None, 'disp_angle [rad]', unit=u.rad)
     disp_sign = Field(None, 'disp_sign')
     disp_miss = Field(None, 'disp_miss [m]', unit=u.m)
-    hadroness = Field(None, 'hadroness')
     src_x = Field(None, 'source x coordinate in camera frame', unit=u.m)
     src_y = Field(None, 'source y coordinate in camera frame', unit=u.m)
     time_gradient = Field(None, 'Time gradient in the camera')
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -55,20 +55,23 @@ cleaning_parameters = {'boundary_thresh': 3,
                        }
 
 
-def get_dl1(calibrated_event, telescope_id):
+def get_dl1(calibrated_event, telescope_id, dl1_container=None):
     """
-    Return a DL1ParametersContainer of extracted features from a calibrated event
+    Return a DL1ParametersContainer of extracted features from a calibrated event.
+    The DL1ParametersContainer can be passed to be filled if created outside the function
+    (faster for multiple event processing)
 
     Parameters
     ----------
     event: ctapipe event container
-    telescope_id:
+    telescope_id: int
+    dl1_container: DL1ParametersContainer
 
     Returns
     -------
     DL1ParametersContainer
     """
-    dl1_container = DL1ParametersContainer()
+    dl1_container = DL1ParametersContainer() if dl1_container is None else dl1_container
 
     tel = calibrated_event.inst.subarray.tels[telescope_id]
     dl1 = calibrated_event.dl1.tel[telescope_id]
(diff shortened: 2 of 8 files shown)


vuillaut:
Hi @rlopezcoto 
We had the conversation in issue #31 .
The function `particle_number` is not used anywhere else.
However, since you asked, I realise that the particle numbering will be different if we take the one from Simtel (and we should, but one has to be aware of that if dealing with old DL1 files where 1 = proton and 0 = gamma).

vuillaut:
Actually let me check that everything works with classification of protons as well...

vuillaut:
This wasn't the initial intention of the PR but I also removed hadroness and use `mc_type` instead to train.
The reconstructed value `hadro_rec` has been removed as well, to be replaced with `reco_type`.

vuillaut:
Training goes fine with the particle numbering > 1

lstchain/io/lstcontainers.py
```diff
@@ -38,7 +38,6 @@ class DL1ParametersContainer(Container):
     disp_angle = Field(None, 'disp_angle [rad]', unit=u.rad)
     disp_sign = Field(None, 'disp_sign')
     disp_miss = Field(None, 'disp_miss [m]', unit=u.m)
-    hadroness = Field(None, 'hadroness')
```
rlopezcoto: not that it matters that much, but up to know, according to the code of @misabelber, we were using `gammaness` instead of `hadronness`, did you decide to go for hadronness in the last DL1 data format call?

lstchain/reco/dl0_to_dl1.py
```diff
@@ -55,20 +55,23 @@
                        }
 
 
-def get_dl1(calibrated_event, telescope_id):
+def get_dl1(calibrated_event, telescope_id, dl1_container=None):
```
rlopezcoto: do we need to initialize `dl1_container` to a None value?

scripts/lst-recopipe.py
```diff
@@ -63,10 +63,8 @@
     #dl1_file=args.datafile
     #data = pd.read_hdf(args.datafile, key='events/LSTCam')
     intensity_min = np.log10(50)
-    r_max = 10#0.8
 
-    data = dl1_to_dl2.filter_events(pd.read_hdf(dl1_file, key='events/LSTCam'), 
-                             r_max=r_max, intensity_min=intensity_min)
+    data = dl1_to_dl2.filter_events(pd.read_hdf(dl1_file, key='events/LSTCam'), intensity_min=intensity_min)
```
rlopezcoto: `r_max` was the maximum impact parameter or the Hillas parameter `r`? If it is the former, it may be useful to apply cuts on it to perform tests, but from its previous value it does not look like

scripts/lst-rfperformance.py
```diff
@@ -164,6 +161,6 @@
     plot_dl2.plot_importances(reg_disp_vector, features)
     plt.show()
 
-    plt.hist(dl2[dl2['hadroness']==1]['gammaness'], bins=100)
-    plt.hist(dl2[dl2['hadroness']==0]['gammaness'], bins=100)
+    plt.hist(dl2[dl2['mc_type']==101]['gammaness'], bins=100)
```
rlopezcoto: 101?

scripts/lst-recopipe.py
misabelber: Hi @rlopezcoto , r is the hillas parameter (distance between center of camera and c.o.g. of the image). I was using cuts in r_max before we had the leakage parameter, to avoid showers in the border of the camera. Now it makes no sense to do cuts in r_max anymore.

lstchain/io/lstcontainers.py
vuillaut: If @misabelber could confirm but it seemed to me that we decided that:
- we would use the particle type from simtel to indentify them
- we would use gammaness that is more broadly used today

misabelber: Yes, that's correct!

lstchain/reco/dl0_to_dl1.py
vuillaut: It allows using the function without initializing a DL1ParametersContainer before

scripts/lst-recopipe.py
vuillaut: `r_max` is not an option of the dl1_to_dl2 reconstruction anymore. I think it @misabelber removed it in some previous PR but it still appeared in some functions. I just fix the code for the functions to work here.

scripts/lst-rfperformance.py
vuillaut: 101 is the particle number for a hadron

rlopezcoto: in simtelarray?

vuillaut: Yes. The numbering goes as follow (actually it might even be from CORSIKA 🤔):
MC shower primary ID:
- 0 (gamma)
- 1(e-)
- 2(mu-)
- 100*A+Z for nucleons and nuclei
- negative for antimatter

rlopezcoto: ok, maybe it is right, but according to the corsika user guide:
https://web.ikp.kit.edu/corsika/usersguide/usersguide.pdf

image

This only works for A > 2. Maybe this works, I had always used 14 for protons (and the numbers you are quoting there are different from those from CORSIKA user guide).

Anyways, this looks good, thanks @vuillaut!

vuillaut: Ho ok so it does come from CORSIKA but from simtel (I just checked with a proton file and pyeventio and the shower primary id is 101).
Thank you for the info.

vuillaut:
Thank you for the review!

PR #105, opened 2019-03-18
WIP: Sensitivity calculation

rlopezcoto:
PR including all the functions needed to minimize and calculate the sensitivity of the telescope using two sets of gamma-ray and proton MC (from [Calculate_sensitivity.ipynb](https://github.com/cta-observatory/cta-lstchain/blob/master/notebooks/Calculate_sensitivity.ipynb) and [Calculate_sensitivity_eventio.ipynb](https://github.com/cta-observatory/cta-lstchain/blob/master/notebooks/Calculate_sensitivity_eventio.ipynb)). Still some things missing:
- Add tests, clean up and add comments to some of the functions.
- Test everything to check its consistency
- Add electrons to the background calculation
- Separate the minimization of gammaness/theta2 cuts from their application (set validation and test datasets)

PR #128, opened 2019-05-15
fix imports and references

vuillaut:
My bad, we merged too soon.
Here is a quick fix.

### DIFF
--- a/lstchain/mc/plot_utils.py
+++ b/lstchain/mc/plot_utils.py
@@ -130,10 +130,12 @@ def sens_minimization_plot(eb, gb, tb, e, sens):
     Returns
     --------
     """
+
+    #TODO : To be changed!!!
+
     # if (eb == 12):
     #     figarr, axarr = plt.subplots(4,3, sharex=True, sharey=True, figsize=(13.2,18))
 
-    # To be changed!!!
     figarr, axarr = plt.subplots(4,3, sharex=True, sharey=True, figsize=(13.2,18))
 
     # The minimum sensitivity per energy bin
--- a/lstchain/mc/plot_utils.py
+++ b/lstchain/mc/plot_utils.py
@@ -1,4 +1,5 @@
 import matplotlib.pyplot as plt
+from matplotlib.colors import LogNorm
 from lstchain.spectra.crab import crab_hegra
 import numpy as np
 import astropy.units as u
@@ -19,7 +20,7 @@ def fill_bin_content(ax, sens, energy_bin, gb, tb):
     return ax
 
 
-def format_axes(ax,pl):
+def format_axes(ax, pl):
     """
 
     Parameters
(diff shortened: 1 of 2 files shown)


lstchain/mc/plot_utils.py
```diff
@@ -125,8 +130,8 @@ def sens_minimization_plot(eb, gb, tb, e, sens):
     Returns
     --------
     """
-    if (eb == 12):
-        figarr, axarr = plt.subplots(4,3, sharex=True, sharey=True, figsize=(13.2,18))
+    # if (eb == 12):
+    #     figarr, axarr = plt.subplots(4,3, sharex=True, sharey=True, figsize=(13.2,18))
 
     # To be changed!!!
```
rlopezcoto: could you move the comment up, before the commented lines? So I do not forget what is it that needs to be changed?

lstchain/mc/sensitivity.py
```diff
@@ -62,7 +62,7 @@ def process_mc(simtelfile, dl2_file):
     events = pd.read_hdf(dl2_file)
 
     e_reco = 10**events.mc_energy * u.GeV
-    n_reco = e_reco.shape[0]
+    # n_reco = e_reco.shape[0]
     gammaness = events.gammaness
     theta2 = (events.src_x - events.src_x_rec)**2 + \
         (events.src_y - events.src_y_rec)**2 * u.deg**2
```
rlopezcoto: If you prefer to change the theta2 definition, please do so

lstchain/mc/sensitivity.py
```diff
@@ -166,13 +166,13 @@ def sensitivity(emin_sens, emax_sens, eb, gb, tb, noff, obstime = 50 * 3600 * u.
...
 
-                ep_w_sum = np.sum(ep_reco_pw[(e_reco_p < E[i+1].to_value()) & (e_reco_p > E[i].to_value()) \
+                ep_w_sum = np.sum(e_reco_pw[(e_reco_p < E[i+1].to_value()) & (e_reco_p > E[i].to_value()) \
                                          & (gammaness_p > g[j]) & (theta2_p < t[k])])
             
                 final_gamma[i][j][k] = eg_w_sum * obstime
                 final_hadrons[i][j][k] = ep_w_sum * obstime
                     
-    sens = calculate_sensititity(final_gamma.to_value(), final_hadrons.to_value(), 1/noff)
+    sens = calculate_sensitivity(final_gamma.to_value(), final_hadrons.to_value(), 1/noff)
```
rlopezcoto: That one was very bad, thanks! ;-)

lstchain/mc/plot_utils.py
vuillaut: done

lstchain/mc/sensitivity.py
vuillaut: Will do in a future PR.
I'd like to introduce the use of other libraries for common stuff like this.

vuillaut: I just wanted to make a quick fix here.

rlopezcoto: which other libraries? 
well we can discuss about this later, but I would try not to make it too complex if the computing time overhead is too large.

vuillaut: We are here trying to re-implement many things that are already implemented in `astropy`, `gammapy` or `ctaplot` which means maintenance that we should avoid (and potentially also means wrong implementations).
But we can discuss that later.

vuillaut:
@rlopezcoto can we merge this small patch to have at least working functions?
As discussed in pr #105, improvements on all that shall be coming later.

PR #134, opened 2019-05-29
Update calibration code to master of ctapipe

FrancaCassol:
This PR update the calibration code to be compatible to the master of ctapipe which now contains the calibration containers and Components


lstchain/calib/camera/r0.py
```diff
@@ -30,6 +30,10 @@ class CameraR0Calibrator(Component):
         Set to None if no Tool to pass.
     kwargs
     """
+    offset = Int(
+        300,
+        help='Define the offset of the baseline'
+    ).tag(config=True)
```
rlopezcoto: this is the same offset as the one intended in `ctapipe_io_lst`, right?

lstchain/tools/camera_calibration_param.json
```diff
@@ -34,11 +34,11 @@
        "r1_sample_start": 2,
       "r1_sample_end": 38
    },
-    "LocalPeakIntegrator":{
+    "LocalPeakWindowSum":{
      "window_shift": 5,
      "window_width":12
     },
```
rlopezcoto: this values we may need to discuss/optimize. A window width of 12 corresponds to 12 ns, right? This is certainly too much for normal pulses, not sure if there is any requirement that needs to be fulfilled here, @moralejo, do you know?

lstchain/calib/camera/r0.py
FrancaCassol: In principle yes, but this time the offset is set by Pawel's code and he use 300 as baseline.
Till now this is the active offset, for that the offset subtraction is done in this code for the moment

lstchain/tools/camera_calibration_param.json
FrancaCassol: @ruben, this committed code is not what we are going to use, it is just an example. 
This are configuration files and the values can be change as one want. In the data of last days I am using 9 ns, but as you say one should do a study before fixing it

FrancaCassol:
@vuillaut, @ruben, I don't understand the error, could you perhaps help?

I would like to have the code committed so that we can install it in La Palma so that people can play with it. By the way, where should I install it?

rlopezcoto:
Found! 
The problem is this [PR 1079](https://github.com/cta-observatory/ctapipe/pull/1079) in ctapipe. So I would wait until tomorrow to see if it is fixed there and then we merge, or we merge either way since it is not a problem of our repo.

rlopezcoto:
I'll merge this since it is urgent. Let's hope the ctapipe bug will be solved soon.

PR #136, opened 2019-06-04
WIP: Debugging sensitivity calculation

misabelber:
I leave here the latests things I've been playing in the sensitivity code to try to achieve a correct sensitivity calculation (which is being really hard...) so maybe if you want to take a look and spot any source of error that could be very helpful.


vuillaut:
Hi @misabelber.
Merge master again for the tests to pass.

lstchain/mc/sensitivity.py
```diff
@@ -148,9 +160,18 @@ def calculate_sensitivity_lima(nex, nbg, alpha):
...
         n_off=nbg,alpha=alpha,significance=5,method='lima')
+
+    for i in range(0, eb):
+        for j in range(0, gb):
+            for k in range(0, tb):
+                if nex_5sigma[i][j][k] < 10:
+                    nex_5sigma[i][j][k] = 10
+                if nex_5sigma[i,j,k] < 0.05 * nbg[i][j][k]/5:
+                    nex_5sigma[i,j,k] = 0.05 * nbg[i][j][k]/5
+
```
rlopezcoto: I do not like this loop inside `calculate_sensitivity_lima`, but I understand why you are putting it here.

The purpose of this function should have been to calculate `nex_5sigma` for a given bg, if this excess does not fulfill certain conditions, it can be changed inside `sens`. The problem is that we would need to recompute the sensitivity there, so it may be ok to leave it here so everything is re-computed and we do not mess so much the main function `sens`

lstchain/mc/sensitivity.py
```diff
@@ -289,10 +328,10 @@ def sens(simtelfile_gammas, simtelfile_protons,
 
     e_reco_gw = ((e_reco_g / crab_par['e0'])**(crab_par['alpha'] - mc_par_g['sp_idx'])) \
                 * w_g
-    e_reco_pw = ((e_reco_p / proton_par['e0'])**(proton_par['alpha'] - mc_par_g['sp_idx'])) \
+    e_reco_pw = ((e_reco_p / proton_par['e0'])**(proton_par['alpha'] - mc_par_p['sp_idx'])) \
```
rlopezcoto: grrr... sorry!

lstchain/mc/sensitivity.py
```diff
@@ -319,33 +362,106 @@ def sens(simtelfile_gammas, simtelfile_protons,
...
-                             or (final_gamma[i,j,k] < min_num_events) \
                              or (final_hadrons[i,j,k] < min_num_events) \
-                             or (not final_gamma[i,j,k] > final_hadrons[i,j,k] * 0.05)
+                             or (pre_gamma[i,j,k] < min_pre_events) \
+                             or (pre_hadrons[i,j,k] < min_pre_events)
                 if conditions:
                     sens[i][j][k] = np.inf
 
-    # Calculate the minimum sensitivity per energy bin
+    #Quantities to show in the results
```
rlopezcoto: This part of the code could use some cleanup before it is merged

scripts/lst-sensitivity.py
```diff
@@ -21,20 +22,28 @@
 
 args = parser.parse_args()
 
-nfiles_gammas = 100*0.2 #Pointlike gammas
-nfiles_protons = 5000*0.2
+nfiles_gammas = 100 #Pointlike gammas
+nfiles_protons = 5000*0.8
```
rlopezcoto: These numbers (0.2 and 0.8) are coming from the division between train and test, right? what about `nfiles_gammas`?

scripts/lst-sensitivity.py
```diff
@@ -21,20 +22,28 @@
...
-gb = 10 #Number of gammaness bins
+eb = 20 # Number of energy bins
+gb = 11 #Number of gammaness bins
 tb = 10 #Number of theta2 bins
-obstime = 50 * 3600 *u.s
+obstime = 50 * 3600 * u.s
 noff = 5
 
-E, best_sens = sensitivity.sens(args.simtelfile_gammas, args.simtelfile_protons,
+E, best_sens, result, units, rate_g, rate_p, w_g, w_p, nex_5sigma, final_hadrons = sensitivity.sens(args.simtelfile_gammas, args.simtelfile_protons,
```
rlopezcoto: this function has now maaaaany outputs, it would be nice to break it down

lstchain/mc/sensitivity.py
```diff
@@ -289,10 +328,10 @@ def sens(simtelfile_gammas, simtelfile_protons,
 
     e_reco_gw = ((e_reco_g / crab_par['e0'])**(crab_par['alpha'] - mc_par_g['sp_idx'])) \
                 * w_g
-    e_reco_pw = ((e_reco_p / proton_par['e0'])**(proton_par['alpha'] - mc_par_g['sp_idx'])) \
+    e_reco_pw = ((e_reco_p / proton_par['e0'])**(proton_par['alpha'] - mc_par_p['sp_idx'])) \
                 * w_p
 
-    p_contained, ang_area_p = ring_containment(angdist2_p, 0.4 * u.deg, 0.1 * u.deg)
+    p_contained, ang_area_p = ring_containment(angdist2_p, 0.4 * u.deg, 0.2 * u.deg)
```
rlopezcoto: we should think on making this as an input instead of having it hardcoded

lstchain/mc/plot_utils.py
```diff
@@ -15,33 +17,34 @@ def fill_bin_content(ax, sens, energy_bin, gb, tb):
     """
     for i in range(0,gb):
         for j in range(0,tb):
-            text = ax.text((j+0.5)*0.05, (i+0.5)*0.1, "%.2f %%" % sens[energy_bin][i][j],
-                       ha="center", va="center", color="w")
+            theta2 = 0.005+0.005/2+((0.05-0.005)/tb)*j
```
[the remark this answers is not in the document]
misabelber: It seems, but this is only for the plotting, probably 0.0075 is the position of the center of the bin. The actual definition of the cuts is in sensitivity.py, funcion bin_definition() and the first cut is 0.005.

scripts/lst-sensitivity.py
misabelber: Yes, they come from there. Gammas are not divided because I trained with diffuse gammas and these are point like gammas, so I'm using all of them for test. 
Still I have to include another separation of files, because we need to use different events for optimizing the cuts and for calculating the sensitivity.

scripts/lst-sensitivity.py
misabelber: Ah yes, sorry, I was doing tests and needed all of those and forgot to remove them before merging. The only ones needed are the first 4 actually.

rlopezcoto:
Hi @misabelber, everything here was reviewed, so no reason not to merge it, right? Could you please resolve the conflict here so it can be merged? It would be great to have this done before next week

misabelber:
I think is fixed now so it can be merged. I will open a new PR before next week with new features (like separating the data set for calculating best cuts and evaluating sensitivity)

rlopezcoto:
Travis is complaining about some of the tests I wrote (I guess some backward incompatibility with the changes of this PR), could you please solve that? Thanks

misabelber:
Sure, I'll check it asap!

rlopezcoto:
still some conflicts... I'll merge it whenever they are solved!

misabelber:
Tests passed!

rlopezcoto:
thanks! it already looks good, I'll merge

PR #137, opened 2019-06-05
fixing bug with unit concerning pulse_time

vuillaut:
Fixing bug appearing after ctapipe PR https://github.com/cta-observatory/ctapipe/pull/1079


rlopezcoto:
this looks good and fixes the broken tests, thanks @vuillaut!

PR #139, opened 2019-06-06
Add notebook for the analysis of muon real data

rlopezcoto:
Very quick and dirty notebook to analyze muon real data. The intention of this notebook is not to stay as a notebook example (at least not the way it is now), but in case somebody wants to analyze the muons present on the real data that were recently taken


vuillaut:
Hi @rlopezcoto 
Could you just add a markdown cell on top of the notebook adding the date and explaining that this is not an example notebook and not intended to stay.
It will help keeping track of it in the future.
Thanks.

rlopezcoto:
Thanks @vuillaut, not sure if we want to leave this notebook in the repo at all. I just uploaded it here in case some urgent analysis was needed while I was away, I should include some of the analysis in the repo or directly in ctapipe... 
Until that point, I think that we could leave the PR open and if at some point we want to merge, I'll clean it up.

rlopezcoto:
I reduced the number of the images inside the notebook and added a markdown cell at the beginning. Since this will be needed in the near future, we probably want it to be merged for people to use it in case it is needed and I can re-arrange it later whenever the relevant code is integrated into the repo

PR #133, opened 2019-05-22
Add tests for the mc module

rlopezcoto:
Add tests for the sensitivity.py and mc.py

PR #141, opened 2019-06-17
Adding simulation info to DL1 files

### LINKED ISSUE #119: Metadata header for dl1 and dl2 data
- [ ] simtel file name and version
- [ ] MC info
- [ ] histograms

vuillaut:
This PR only adds simulation infos to DL1 files following the structure in v5 discussed in https://github.com/cta-observatory/ctapipe/issues/1059
Somehow fixes #119 .

As the DL1 structure is still being discussed, I think we can move forward with the simu info part to at least have this info available to create IRFs.

The events table structure stays untouched so the rest of the code will still work without any change.

I also modify the merge script to append the simu info from each merged file.

### DIFF
--- a/lstchain/io/__init__.py
+++ b/lstchain/io/__init__.py
@@ -1,2 +1,3 @@
 from .config import *
 from .lstcontainers import *
+from .io import *
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -1,10 +1,12 @@
 from ctapipe.io import HDF5TableReader
 from ctapipe.io.containers import MCHeaderContainer
 from tables import open_file
+import h5py
 import numpy as np
 
 __all__ = ['read_simu_info_hdf5',
            'read_simu_info_merged_hdf5',
+           'get_dataset_keys',
            ]
 
 def read_simu_info_hdf5(filename):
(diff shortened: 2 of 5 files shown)


vuillaut:
This will allow to go further and e.g. avoid requiring simtel files for the sensitivity calculation.
This can be done in another PR after changes from #133 have been merged.

vuillaut:
Hi 

> Hi Thomas! I'm taking a look at the format and dl1 seems to be fine, I could read the mc data without problem. However, I ran lst-recopipe to check how the format is inherited in the dl2 file and I'm finding a weird structure. When I walk through the groups of the dl2 file like this:
> `f1 = tables.open_file(dl2_file)`
> `for group in f1.walk_groups():`
> `print(group)`
> 
> I obtain this list:
> / (RootGroup) ''
> /events (Group) ''
> /simulation (Group) ''
> /events/LSTCam (Group) ''
> 
> Shouldn't the last one have been erased before writing it to the dl2 file?

Hi @misabelber,

What do you mean the last one should be erased?
`events` is just a node with here only the `LSTCam` node inside as usual.
`events/LSTCam` contains the DL2 table (same as the DL1 table + the reconstructed fields from the reco dl1 to dl2).

misabelber:
Ahhh, ok, nevermind then, I was misunderstanding the structure. In that case, no problem!

PR #142, opened 2019-06-18
Configurable pipeline

vuillaut:
This PR introduces a scheme for lstchain configuration with
- a clear standard configuration (which was until now hidden in many functions)
- a way to override the standard configuration with custom configuration
- custom configuration can now be passed as a file to all scripts 

This, I believe, will both improve the reproducibility of results and the ability to play with configurations.

It is a first step. A more perfect solution would be using ctapipe Tools everywhere to load and keep track of configs. Baby steps...


vuillaut:
Is someone reviewing this PR?

I could keep improving it but it's already large and is now self-consistent (no more hidden hard-coded configuration settings).

I have used this PR to reduce the last production so it's working but I'd like to have feedback.

Thank you.

misabelber:
I'm on it! Tomorrow morning I'll give you my feedback :)

vuillaut:
Super, thank you so much for the review @misabelber !
I have added some more unit tests for future changes.

Issue #144, opened 2019-06-21
Gain Selector issue introduced by ctapipe/pull #1093


vuillaut:
See ctapipe merged PR: https://github.com/cta-observatory/ctapipe/pull/1093
lstchain build is now failing.

PR #145, opened 2019-06-21
Use new gain selector from ctapipe

vuillaut:
Fixes issue #144


vuillaut:
If someone can have a quicklook at this, that would be great since it's breaking lstchain.

vuillaut:
Thank you for the review @misabelber

PR #148, opened 2019-07-02
Improve create pedestal script

pawel21:
Hi,

I've modified the code to low level correction:
1) I've speeded up significantly code (using numpa prange) to create pedestal file. 
2) I've changed coefficients for dt_corr, from study run 97 I think coefficients for temperature 30 fit better.
3) I've changed way to give path **LST-1.*.Run00097.0000.fits.fz -> LST-1.1.Run00097.0000.fits.fz** according to reader.


scripts/create_pedestal_file.py
```diff
@@ -1,21 +1,22 @@
 import sys
 sys.path.insert(0,'/home/pawel1/Pulpit/Astrophysics/CTA/soft/ctapipe_io_lst')
+sys.path.insert(0, '/home/pawel1/Pulpit/Astrophysics/CTA/cta-lstchain')
```
vuillaut: Can we avoid this kind of import that depends on a specific path and prevent other users to use the code?
Why would we need such an import anyway?

lstchain/calib/camera/r0.py
```diff
@@ -369,9 +369,9 @@ def ped_time(timediff):
     """
     Power law function for time lapse baseline correction.
     Coefficients from curve fitting to dragon test data
-    at temperature 40 degC
+    at temperature 30 degC
     """
-    return 23.03 * np.power(timediff, -0.25) - 9.73
+    return 27.33 * np.power(timediff, -0.24) - 10.4
```
vuillaut: Hardcoded values are always error-prone, especially if not documented.
Are these values subject to change in the future for different conditions?

lstchain/calib/camera/drs4.py
```diff
@@ -35,27 +36,29 @@ def fill_pedestal_event(self, event, nr):
     def _fill_pedestal_event_jit(nr, waveform, expected_pixel_id, first_cap, meanped, numped):
         size4drs = 4096
         roisize = 40
-        for i in range(0, 2):
-            for j in range(0, 7):
+        for i in prange(0, 2):
```
vuillaut: Out of curiosity: by how much does this improve the efficiency?
Have you tried without loops using numpy only (in a similar fashion to what I proposed by email some time ago)?

scripts/create_pedestal_file.py
pawel21: Yes, sorry my mistake. We don't need it.  I removed.

vuillaut: Thank you.
If you can have a look to my other comments we could merge this.

lstchain/calib/camera/drs4.py
vuillaut: Nice.

vuillaut:
> After discussed with Seiya Nozaki yesterday, I will change code a bit, so please give me 1 or 2 days.

No problem, thanks for the update.

pawel21:
I changed and cleaned up code to create pedestal file. I added option to deactivate delta T correction.

pawel21:
Dear @vuillaut,

Could you look on code? 
Now, it is version, which I would like pull into master.

lstchain/calib/camera/drs4.py
```diff
@@ -1,68 +1,71 @@
...
-        first_cap = event.lst.tel[self.telid].evt.first_capacitor_id[nr * 8:(nr + 1) * 8]
+        fc = np.zeros((2, 7))
+        first_cap = event.lst.tel[self.tel_id].evt.first_capacitor_id[nr * 8:(nr + 1) * 8]
         # First capacitor order according Dragon v5 board data format
         for i, j in zip([0, 1, 2, 3, 4, 5, 6], [0, 0, 1, 1, 2, 2, 3]):
-            fc[self.high_gain, i] = first_cap[j]
+            fc[0, i] = first_cap[j] # high gain
         for i, j in zip([0, 1, 2, 3, 4, 5, 6], [4, 4, 5, 5, 6, 6, 7]):
-            fc[self.low_gain, i] = first_cap[j]
+            fc[1, i] = first_cap[j] # low gain
```
rlopezcoto: why changing the clear general definition of high/low gain by these two numbers (and corresponding comments)?

scripts/create_pedestal_file.py
```diff
@@ -30,6 +29,9 @@
 parser.add_argument("--max_events", help="Maximum numbers of events to read",
                     type=int, default=5000)
 
+parser.add_argument("--deltaT", help="Use deltaT correction",
+                    type=str, default='yes')
```
rlopezcoto: wouldn't it be better a True/False boolean rather than a string?

lstchain/calib/camera/drs4.py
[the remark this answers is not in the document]
pawel21: I wanted reduce number default values, but I think this was bad idea. So I will back to previous version.

lstchain/calib/camera/drs4.py
```diff
@@ -1,68 +1,71 @@
...
-                        numped[i, j, posads] += 1
+        for nr_module in prange(0, 265):
+            first_cap = first_cap_array[nr_module, :, :]
+            for gain in prange(0, 2):
+                for pix in prange(0, 7):
+                    fc = first_cap[gain, pix]
+                    pixel = expected_pixel_id[nr_module * 7 + pix]
+
+                    posads0 = int((2 + fc) % size4drs)
+                    if posads0 + 40 < 4096:
```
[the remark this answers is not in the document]
pawel21: Done

scripts/create_pedestal_file.py
```diff
@@ -30,6 +29,9 @@
...
+
```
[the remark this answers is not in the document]
pawel21: You right, I added boolean argument.

rlopezcoto:
This looks already good @pawel21, let's hope the script works fine. For the near future, it would be nice to have more tests for all this code, specially for the part that will remain even after EvB3. 
If no further functionalities are going to be included in the PR, I'll merge it

pawel21:
I am not going include new functionalities now, so we can merge it.

lstchain/calib/camera/drs4.py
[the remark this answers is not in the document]
vuillaut: what is `size4drs`?
what's its meaning? why 4096?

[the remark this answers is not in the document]
pawel21: This is number of capacitors in each pixel which using DRS4.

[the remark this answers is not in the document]
vuillaut: Ok thank you.
This is exactly the kind of comment and reference that should go directly into the code for future references.

PR #149, opened 2019-07-09
Improve speed by moving change of frames out of event loop

vuillaut:
Hi

This is the first part of some lstchain speed-up, starting by a part that was taking most of the part: coordinates change of frames to compute the disp parameter.

This leads to a significant speed-up, of about a factor 2 on the analysis of a 300MB file:

before, `get_event_pos_in_camera` was taking 100s alone:

before

after, the total analysis is ~90s:
after


lstchain/reco/utils.py
```diff
@@ -271,13 +224,9 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
     y = np.array([1,1]) * u.m
 
     """
-    pointing_direction = SkyCoord(alt=pointing_alt, az=pointing_az,
-                               frame=horizon_frame
-    )
+    pointing_direction = SkyCoord(alt=pointing_alt, az=pointing_az, frame=horizon_frame)
```
rlopezcoto: tab here to make it pep8 compliant?

lstchain/reco/utils.py
```diff
@@ -271,13 +224,9 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
...
 
-    camera_frame = CameraFrame(focal_length=focal,
-                               telescope_pointing=pointing_direction
-    )
+    camera_frame = CameraFrame(focal_length=focal, telescope_pointing=pointing_direction)
```
rlopezcoto: same as above

rlopezcoto: (also a couple below)

vuillaut:
> Before moving to a deeper review, I would like to understand the benefit of moving the disp calculation from utils to a new file: if there is no benefit and it is just a matter of taste, I would leave it as it was because this will result in many notebooks, etc... incompatibilities

Yes I believe having everything concerning disp calculation in one place is clearer.

Good thinking about the notebook incompatibility, thanks! However, it will be the case even if I left the functions in utils.py as half of them have been refactored.

lstchain/reco/utils.py
vuillaut: It wasn't correctly aligned.

lstchain/reco/utils.py
vuillaut: Yes, the parenthesis shouldn't hang there.
Small things, I just spotted while refactoring and my IDE complained so I fixed it...

PR #150, opened 2019-07-22
adding comment on size4drs for future references

vuillaut:


lstchain/calib/camera/drs4.py
```diff
@@ -4,9 +4,14 @@
 
 __all__ = ['DragonPedestal']
 
+# `size4drs` is number of capacitors in each pixel which using DRS4.
+# "The sampler has 1024 capacitors per channel (pixel).
```
rlopezcoto: I would not say here `(pixel)`, since each pixel has four channels of a chip connected, for a total of 4096 capacitors per pixel

vuillaut: Sorry I mindlessly copied the comment from @pawel21 comment in PR #148 
Just removing the "(pixel)" part would be more clear?

rlopezcoto: yeah, ok from my side, but let's wait for @pawel21 to see if it is also ok for him

pawel21: Ok from my side.

vuillaut: Ok thanks.

PR #152, opened 2019-07-25
Drivereport container

labsaha:
This PR address the issue of adding a drive report container following the structure of the drive report. This will be followed by a script which will use the container class to add a container (or group) to the DL1 data. I'll put another script where I show how the drive report appears in the DL1 h5 file

DriveReport


rlopezcoto:
@labsaha could you please correct the errors so all test pass before starting the review?
@francacassol can you have a look to the PR to try to homogenize it to what it was discussed by email? 
Thanks

labsaha:
@rlopezcoto Yes, I am correcting now that error.

labsaha:
@rlopezcoto @FrancaCassol  I'll close this. The correct one is now at https://github.com/cta-observatory/cta-lstchain/pull/153#issue-301115175

FrancaCassol:
Hi,
actually I think we should better plan a short meeting to understand this question of post inserting informations in the DL1 hdf5 file. I understand that in MAGIC did like that, but in the case of lstchain we better try to keep things the most as possible compatible with ctapipe. And a table like that is definitely not. The pointing information is instead stored event per event in the TelescopePointingContainer.

vuillaut:
I am closing this PR, let's keep discussing in #153

PR #153, opened 2019-07-25
Drive Report Container & adding it to DL1 files

labsaha:


lstchain/io/lstcontainers.py
```diff
@@ -178,3 +179,21 @@ class DispContainer(Container):
     norm = Field(nan, 'Norm of the disp_norm vector')
     sign = Field(nan, 'Sign of the disp_norm')
     miss = Field(nan, 'miss parameter norm')
+
+class DriveReport(Container):
```
vuillaut: A class is created and not used anywhere?

vuillaut: Ok I see this is related to PR #153 .

labsaha: No, it will be used. I'll upload another script soon which uses this class.

vuillaut: Why not directly including this in PR #153 ?
This PR makes little sense on its own.

vuillaut: > No, it will be used. I'll upload another script soon which uses this class.

Then please do it all at once to help the review process.
One cannot judge the relevance of this code with anything using it.

labsaha: Actually, this can be part of ctapipe as well.

vuillaut: Even better.

labsaha:
Now I have added one script which does the job of adding a drive report to the DL1 file https://github.com/cta-observatory/cta-lstchain/pull/153/commits/f8ee8c2b3a5185a0cfd0f2595327fc087d68b1c7. Note that at present, it just adds the drive report to a DL1 file. However, this is not doing an event-by-event basis. The idea is to take the timing information of the first event and last event of a run (or subrun) and then take the portion of the drive report relevant to that time slot of the run. However, there is no timing information in the DL1 file so far. Hence, the drive report cannot be attached based on timing information. I'll try to add some dummy values of timing information to the DL1 file and based on that. I'll do that in the next version. Adding the pointing information event-wise should be part of a separate code which includes also a correction to the pointing information.

As an example, I give here the drive report container how it appears.  
DriveReport

vuillaut:
Hi

I think we should discuss a bit the approach here.
If I understand correctly, this information is basically the pointing information that we need for the whole chain.
If so, this information should not be manually added to yet another file (why to DL1?), this is perpendicular to the "file agnostic" approach of ctapipe.
The information should be read by `ctapipe_io_lst` and provided in ctapipe containers in a transparent way. So that we don't have to develop custom code for real data vs MC data that will require double maintenance.
And then the information will be automatically dumped at all data level without custom code.

labsaha:
Yes, we need definitely some discussion on this. Since there is no effort towards this yet, it is good to start with this. We thought of adding it after the DL1 because this is the way it is done in MAGIC. It also has some advantages because we will have to add some more information, whose data flow is still not clear. This way of doing makes us enable to add them all together and should be much more flexible. We also think the content of the drive report is quite simple and suggest just adding it as a table of an additional h5 group. The table would only contain the drive records relevant for the observation. At first, we think of adding the drive report. Then the drive report (which comes every 2 seconds) can be used to interpolate pointing information for all the events. Here drive report is just an example. Other information such as star guider, atmospheric transmission, etc can be added in a similar fashion. We see this procedure as an effective one.

vuillaut:
Copied from @FrancaCassol https://github.com/cta-observatory/cta-lstchain/pull/152#issuecomment-515389661
> Hi,
actually I think we should better plan a short meeting to understand this question of post inserting informations in the DL1 hdf5 file. I understand that in MAGIC did like that, but in the case of lstchain we better try to keep things the most as possible compatible with ctapipe. And a table like that is definitely not. The pointing information is instead stored event per event in the TelescopePointingContainer.

Couldn't agree more.
There is already a container to add the pointing information.
Once it's been added there, it will automatically be added to the DL1 file when we create it, no need for additional code.
If it's not imported in some way in the chain (and thus in the pointing container), it cannot be used anyway... 

We might want to have a `DriveContainer` that will provide more information than just pointing.
But in this case, we need:
- [ ] the container
- [ ] a way to **read** the drive file and fill the container

Then this DriveContainer can be dumped in all files as all other containers. no need for `scripts/add_drivereports.py`.

labsaha:
@vuillaut Yes, one can do this in several ways. One can just consider one `DriveContainer` which does all the job. Eventually, that container will be part of DL1. So it means if I understand correctly, there won't be any drive report container in DL1 file, it will be a Container which will provide pointing information for each of the events.  Then I think the following should be part of the `DriveCoantainer` (to be a bit specific)

1. method to read the drive report
2. method to Interpolate for each of the events
3. method to correct interpolated values using star guider information
4. methods to add star guider corrections
5. something elses

Is there anything already available in ctapipe which does all these or partially?

@FrancaCassol  Anyway, I have asked @moralejo to keep a time slot in the next LST-reco meeting on Monday to discuss this.

FrancaCassol:
Hi @labsaha 

>One can just consider one `DriveContainer` which does all the job. 

Actually, in ctapipe the containers do non have methods, they are just containers. 
I guess what you want to create is a Component class, let's call it DriveCalculator, that eventually fill a DriveContainer (do we need it?)
 
> 
> 1. method to read the drive report
> 2. method to Interpolate for each of the events
> 3. method to correct interpolated values using star guider information
> 4. methods to add star guider corrections
> 5. something else
> 
> Is there anything already available in ctapipe which does all these or partially?
No. ctapipe supposes you have already done all this and filled the correct pointing position in the event Pointing container. 
However, we can think to place some of the methods in lstchain  (which treats also R0 data), only let try to coordinate the development so to keep a coherent code structure (ctapipe compatible). 
It seems to me that for what concerns point 1 and 2, they can be easily done at the data reader level. 
For what concerns point 3-5, they could be performed in Component class, which has the real to better "calibrate" the pointing position. If we do it properly we could then even think to move the it to ctapipe for other telescopes.
I wonder, which type of data are those of the star guider?  How do you think to access them? 

> 
> @FrancaCassol Anyway, I have asked @moralejo to keep a time slot in the next LST-reco meeting on Monday to discuss this.

Perfect :-)

labsaha:
@vuillaut @FrancaCassol 
For the time being and for the LST1, to consider the pointing calculation at the DL0/DL1 level, do you consider pointing position calculator class for event-wise at `io/lstcontainers.py` or `lstchain/tools/`? or somewhere else? Let me know which place you think would be suitable for that. I can implement that accordingly.

vuillaut:
Hi @labsaha 
After this morning call, I realise that there has been some confusion.
Here is what I understand and how I see things (subject to discussion of course):
- AUX data should be read in parallel with RAW data and added to specific containers (such as `MON` container or `lst` container...). You may create new containers but consider the ones already existing and you may update them (see https://github.com/cta-observatory/ctapipe_io_lst/blob/master/ctapipe_io_lst/containers.py).
- then the information is processed from the containers
- then the containers are passed to the HDF5 writer and automatically written to DL>1 levels

My concern, as expressed previously is mainly to keep a chain in line with ctapipe code and organisation and agnostic of the data source (be it MC from simtel, HDF5 or else or real-data from ZFITS, etc... ).
Concerning the pointing information, we are dealing with two different things here (maybe another source of confusion) :
a. the drive AUX data (the one addressed in this PR). For this one, as other AUX data, a specific container is probably the way to go.
b. the event wise pointing. This one comes from the processing of the first one (by interpolation) and is added to the container `event.pointing`. 

In my opinion, a. and b. should be read/created when reading the raw data.
To answer to your specific question:
- You may keep the current DriveContainer.
- Add a function to interpolate from the info in DriveContainer and update the info in event.pointing

FrancaCassol:
Hi @labsaha,

if you want to keep all the driver table, you should define your container and add it to the monitoring container, I order to be able to do it I will create a LSTMonitoring container (daughter of the Monitoring container ) so that you can add your container there in. Is it ok @vuillaut ?

For what concerns the class, in order to do things properly, I would create a subdirectory 
calib/pointing and put the the class (which should be a Component class) in a new module named DriverPosition.py or similar. Though, if this class makes only the interpolation of the position between two times it is probably overdone. Is this class expected to do also something else?

labsaha:
@vuillaut 

> AUX data should be read in parallel with RAW data and added to specific containers (such as MON container or lst container...). You may create new containers but consider the ones already existing and you may update them (see https://github.com/cta-observatory/ctapipe_io_lst/blob/master/ctapipe_io_lst/containers.py).
> then the information is processed from the containers
> then the containers are passed to the HDF5 writer and automatically written to DL>1 levels

Sorry, I didn't know this container. Anyway, we can update the container. The container class I have written was also following ctapipe similar to the container clasees present in `ctapipe_io_lst/containers.py`. So it is in line with ctapipe, only the place is different. An then I filled the container with the specific values and passed the container to HDF5 writer which is again following some example file present in ctapipe. 

> a. the drive AUX data (the one addressed in this PR). For this one, as other AUX data, a specific container is probably the way to go.
> b. the event wise pointing. This one comes from the processing of the first one (by interpolation) and is added to the container event.pointing.
> 
Yes, we think, in the long run, we may have to go with this. Of course, this needs to be discussed within a big forum, maybe in ctapipe. For the time being, we can just test it for LST and then bring the issues in ctapipe.

labsaha:
@FrancaCassol  

> if you want to keep all the driver table, you should define your container and add it to the monitoring container, I order to be able to do it I will create a LSTMonitoring container (daughter of the Monitoring container ) so that you can add your container there in. Is it ok @vuillaut ?

Yes, the container class which is present in this PR can be part of Monitoring container. For other reports, we can just go adding things.

> For what concerns the class, in order to do things properly, I would create a subdirectory
calib/pointing and put the the class (which should be a Component class) in a new module named DriverPosition.py or similar. Though, if this class makes only the interpolation of the position between two times it is probably overdone. Is this class expected to do also something else?

When we talk about pointings we consider the correction as well. At present, we may not need to think about that. But this is to plan how things should go. So the class should do the interpolation as well as correction. The class can be named as `PointingPosition` to make it clear. I can take care of writing this class if you don't mind.

FrancaCassol:
> @FrancaCassol
> 

> Yes, the container class which is present in this PR can be part of Monitoring container. For other reports, we can just go adding things.

Ok. 

> When we talk about pointings we consider the correction as well. At present, we may not need to think about that. But this is to plan how things should go. So the class should do the interpolation as well as correction. The class can be named as `PointingPosition` to make it clear. I can take care of writing this class if you don't mind.

Please go ahead, just, for curiosity, on which type of data and/or methods are based the corrections?

labsaha:
@FrancaCassol  Thanks!  I think the data is similar to what we see in the drive report. I am trying to find the reference for the algorithm used for MAGIC for star guider correction. I didn't get it yet. I'll let you know if I get it.

rlopezcoto:
ok, so after agreeing on the way that the drive information should be added and accepting [the monitoring container](https://github.com/cta-observatory/ctapipe_io_lst/pull/9) , shall we open a new PR to keep things cleaner?

labsaha:
Hi @rlopezcoto,
Let's wait for a few more times on this here. I have almost the working version for adding pointing information ready. I am just now making it compatible for `ctapipe/lstchain` framework.  I may add some more comments here just for discussion, then I can make a new PR with a clean version.

labsaha:
Hi @vuillaut and  @FrancaCassol,
I am just looking for the place where I can put the classes required for handling drive reports, pointing position. You might provide some information on the right place to include them. 
1. In this PR I have the `DriveReport` container which I renamed as `DriveContanier` now. This can be part of `lstconatnier` in this lstchain repo. I don't see any place suitable other than this. I can include this `ctapipe_io_lst` repo. I am a bit confused with the need for additional repo like `ctapipe_io_lst`, when many of the classes are defined are at `lstchain/io`

2. I have now another class `PointingPosition`. This has two methods, one reads the drive reports (`read_drive_report`) & and other calculates pointing positions `cal_pointingposition`. I tested the code with available report files and some dummy event times and it provides pointing positions event-wise.  The most suitable place for this class seems to me at `lstchain/pointing` which we can create. @FrancaCassol mentioned earlier to have a directory like `lstchain/calib/pointing`. But I think it shouldn't be part of the calibration directory. It should be a separate one. 

Let me know what you think.

FrancaCassol:
Hi @labsaha 
> 1. In this PR I have the `DriveReport` container which I renamed as `DriveContanier` now. This can be part of `lstconatnier` in this lstchain repo. I don't see any place suitable other than this. I can include this `ctapipe_io_lst` repo. I am a bit confused with the need for additional repo like `ctapipe_io_lst`, when many of the classes are defined are at `lstchain/io`

you have to put the DriveContainer container as a field of the class LSTMonitoringContainer in the containers.py module in ctapipe_io_lst. I can do it for you, but, please pass me version where the content of the fields is better described together with the units of each field (if any)

> 2.  @FrancaCassol mentioned earlier to have a directory like `lstchain/calib/pointing`. But I think it shouldn't be part of the calibration directory. It should be a separate one.

Yes, as far as I understood you want to provide a correction of the pointing values, isn't it? (because otherwise we don't need a class for this, we can do it directly at the reader level as foreseen before) So, please put it in a directory calib/pointing. Also it would be better that your class is a daughter of the ctapipe Component class [https://cta-observatory.github.io/ctapipe/api/ctapipe.core.Component.html#ctapipe.core.Component](https://cta-observatory.github.io/ctapipe/api/ctapipe.core.Component.html#ctapipe.core.Component)
so that it has Trailets  and Logging

labsaha:
@FrancaCassol 
> you have to put the DriveContainer container as a field of the class LSTMonitoringContainer in the containers.py module in ctapipe_io_lst. I can do it for you, but, please pass me version where the content of the fields is better described together with the units of each field (if any)

Are we importing any classes from `ctapipe_io_lst` in lstchain?  

>Yes, as far as I understood you want to provide a correction of the pointing values, isn't it? (because otherwise we don't need a class for this, we can do it directly at the reader level as foreseen before) So, please put it in a directory calib/pointing. Also it would be better that your class is a daughter of the ctapipe Component class https://cta-observatory.github.io/ctapipe/api/ctapipe.core.Component.html#ctapipe.core.Component
so that it has Trailets and Logging

We want to make the star guider correction but that has nothing to do with calibration (as far as I can see). Hence I don't understand why the pointing should be part of `calib`. Am I missing something? Can you explain a bit more?

labsaha:
Hi @FrancaCassol, I just make a PR of the container in `ctapipe-io-lst`.  Please have a look.

labsaha:
I am closing this PR. I'll create a new one following the discussion we had here.

PR #155, opened 2019-07-26
Fix lstchain notebook 

garciagenrique:


rlopezcoto:
Thanks for fixing this @garciagenrique

PR #156, opened 2019-07-26
clean a bit the test_lstchain with a specific dir and testfile names

vuillaut:
- Move the general tests in lstchain folder (where they belong)
- Define general file names for test files at DL1 and DL2 level so they can be used globally


vuillaut:
Note: I'd like this PR considered before PR #120 as it will ease the tests update (and thus my life :D).

vuillaut:
> @vuillaut This looks good to me, don't you want to write the sensitivity and process_mc tests before this gets merged?

When was I assigned the writing of sensitivity and process_mc tests??? 🤔 
It would be natural and easier for you and/or @misabelber who wrote the original code to write the related unit tests...

lstchain/tests/test_lstchain.py
```diff
@@ -75,58 +86,51 @@ def test_import_lstio():
...
-    infile = 'dl1_gamma_test_large.h5'
+    infile = dl1_file
 
     reg_energy, reg_disp, cls_gh = build_models(infile, infile, custom_config=custom_config, save_models=True)
 
     from sklearn.externals import joblib
-    joblib.dump(reg_energy, 'rf_energy.pkl')
-    joblib.dump(reg_disp, 'rf_disp.pkl')
-    joblib.dump(cls_gh, 'rf_cls_gh.pkl')
+    joblib.dump(reg_energy, os.path.join(test_dir, 'rf_energy.pkl'))
```
garciagenrique: Models' names could be defined also as variables.

lstchain/mc/tests/test_senstivity.py
```diff
@@ -1,7 +1,6 @@
 import numpy as np
 import pytest
 import pandas as pd
```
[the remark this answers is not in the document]
vuillaut: Right, thanks.

lstchain/tests/test_lstchain.py
vuillaut: Indeed, thanks for the suggestion.

PR #158, opened 2019-08-02
Fix def train_disp_sign

garciagenrique:
Replaced model and features variables in def train_disp_sign(). The R…F_model was set as a RandomForestRegressor, when it should be a RandomForesetClassifier, as the definition and documentation specifies. Also, the features were changed from regression_ to classification_features.


rlopezcoto:
Ups, looks good, but I believe we were not using the `train_disp_sign` anymore, but we were using the `train_disp_vector` function, right @misabelber?

vuillaut:
Hi @rlopezcoto 
`train_disp_vector` is indeed the method used by default in the scripts but I believe we want to keep both methods in the code.

rlopezcoto:
ok, thanks!

PR #161, opened 2019-08-09
remove unused code in utils

thomasgas:
This is a simple PR which removes some unused code that I introduced some time ago for the LST Analysis Bootcamp.
I think nobody is using this code in cta-lstchain
I can't understand if the entire folder is going to be removed too, which is my goal


vuillaut:
Thank you @thomasgas.
Yes, the entire folder should disappear.

PR #163, opened 2019-08-20
Fix ctapipe version to v0.7.0

vuillaut:
- Fixing ctapipe v0.7.0 in environment.yml
- Update travis config with two separate tests:
    - one with fixed ctapipe v0.7.0 and ctapipe_io_lst v.0.1
    - one testing ctapipe master and ctapipe_io_lst master.
This one is allowed to fail, meaning that it will not make the entire build fail but still warn us that the current code will not work with ctapipe master, allowing us to update it in the `ctapipe-master` branch.
- adding ctapipe_io_lst v0.1 to the install environment
- fixing python 3.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,26 @@
 language: python
 
+
+python:
+    - 3.6
+    - 3.7
+
+env:
+    - CTAPIPE_VERSION="v0.7.0" CTAPIPE_IO_LST_VERSION="v0.1"
+    - CTAPIPE_VERSION="master" CTAPIPE_IO_LST_VERSION="master"
+
 matrix:
-  include:
-    python: 3.7
-    os: linux
-    env:
-      - CTAPIPE_VERSION=v0.7.0
-      - PROTOZFITS_VERSION=v1.4.2
-      - CTAPIPE_IO_LST_VERSION=v0.1
-
   allow_failures:
-    python: 3.7
-    os: linux
-    env:
-      - CTAPIPE_VERSION=master
-      - PROTOZFITS_VERSION=v1.4.2
-      - CTAPIPE_IO_LST_VERSION=master
+      - python: 3.6
+        env: CTAPIPE_VERSION="master" CTAPIPE_IO_LST_VERSION="master"
+      - python: 3.7
+        env: CTAPIPE_VERSION="master" CTAPIPE_IO_LST_VERSION="master"
 
+  include:
+      - python: 3.6
+        env: CTAPIPE_VERSION="v0.7.0" CTAPIPE_IO_LST_VERSION="v0.1"
+      - python: 3.7
+        env: CTAPIPE_VERSION="v0.7.0" CTAPIPE_IO_LST_VERSION="v0.1"
 
 before_install:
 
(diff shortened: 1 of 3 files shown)


vuillaut:
Why has Travis not triggered?

vuillaut:
If someone could have a look at this PR before we can start/finish reviewing the pending ones that would be great.

rlopezcoto:
I'll do it in a couple of days, sorry but cannot make it before

vuillaut:
I'd like to merge this asap to avoid too much deviation and the potential integration of code not compliant with ctapipe v0.7.
If there is no problem, I'll do it this week.

.travis.yml
```diff
@@ -29,12 +35,10 @@ install:
     - conda env create --name cta --file environment.yml
     - conda activate cta
     - pip install https://github.com/cta-observatory/ctapipe/archive/$CTAPIPE_VERSION.tar.gz
```
rlopezcoto: Why eliminating PROTOZFITS from Travis?

README.md
```diff
@@ -6,41 +6,25 @@ The analysis is heavily based on [ctapipe](https://github.com/cta-observatory/ct
 master branch status: [![Build Status](https://travis-ci.org/cta-observatory/cta-lstchain.svg?branch=master)](https://travis-ci.org/cta-observatory/cta-lstchain)
 
 
-### Important message to lstchain users (May 4th 2019):
-*ctapipe* and *lstchain* are currently undergoing heavy and rapid changes.    
-The core developer team is trying to stay up-to-date with the master version of *ctapipe* before reaching *ctapipe v0.7* release.
-You might experience some issues if changes have been merged in *ctapipe* master before we could integrate these changes in *lstchain*. We are sorry for that. Do not hesitate to submit an issue or propose a patch through a pull request.
-
-- The basic functions of lstchain (reduction steps R0-->DL1 and DL1-->DL2) are unit tested and should be working as long as the build status is passing.
```
rlopezcoto: I would keep this sentence

.travis.yml
vuillaut: because it's not used

README.md
vuillaut: which part? 
if ctapipe=0.7 is a dependency of lstchain, users should not experience issues anymore

I agree for the last sentence though.

.travis.yml
rlopezcoto: where?

vuillaut: hu? nowhere 😄 

it's legitimate to have `PROTOZFITS` as a requirement of lstchain because most of the user will want to analyse real data (I suppose) but there is 0 unit test using real-data and thus using PROTOZFITS. So I don't see the point of installing it every time for no reason.

rlopezcoto:
We will have unit tests with real data written as soon as @FrancaCassol
gets the test data she requested... I would not remove it

vuillaut:
> We will have unit tests with real data written as soon as @FrancaCassol gets the test data she requested... I would not remove it

I would personally add packages when they are needed (so when the tests are added - tests will not pass otherwise anyway) rather than in a preventive way. But it's not really important I can revert this :)

.travis.yml
vuillaut: add PROTOZFITS back to travis

README.md
vuillaut: re-added last sentence

PR #165, opened 2019-08-22
applying cleaning to image before computing timing features

vuillaut:
Hi

Working on something different with @garciagenrique (thanks!), we discovered a small bug in the chain: the cleaning was not applied to the charge and pulse_time images before computing the timing parameters.

### DIFF
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -133,9 +133,13 @@ class DL1ParametersContainer(Container):
         self.disp_miss = disp.miss
 
     def set_timing_features(self, geom, image, pulse_time, hillas):
-        timepars = time.timing_parameters(geom, image, pulse_time, hillas)
-        self.time_gradient = timepars.slope.value
-        self.intercept = timepars.intercept
+        try:    # if np.polyfit fails (e.g. len(image) < deg + 1)
+            timepars = time.timing_parameters(geom, image, pulse_time, hillas)
+            self.time_gradient = timepars.slope.value
+            self.intercept = timepars.intercept
+        except ValueError:
+            self.time_gradient = np.nan
+            self.intercept = np.nan
 
     def set_leakage(self, geom, image, clean):
         leakage_c = leakage(geom, image, clean)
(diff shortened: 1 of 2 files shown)


vuillaut:
NB: that **should not** explain the current "too good results" of lstchain as not applying the cleaning should have worsened the reconstruction

kosack:
This seems to imply a documentation or design flaw in `timing_parameters()`.  I suggest we open an issue to change it's signature to be more similar to the `leakage()` calculation (e.g. to add cleaning_mask), and then you explicitly apply it inside:

```py3
def timing_parameters(geom, image, pulse_time, cleaning_mask, hillas_parameters):
```

Right now in the function, there is a line that basically re-applies the mask anyhow:
```py3
    greater_than_0 = image > 0
    pix_x = geom.pix_x[greater_than_0]
    pix_y = geom.pix_y[greater_than_0]
    image = image[greater_than_0]
    pulse_time = pulse_time[greater_than_0]
```

So that could just be replaced explicitly with `cleaning_mask` in place of `greater_than_zero`

vuillaut:
> This seems to imply a documentation or design flaw in `timing_parameters()`. I suggest we open an issue to change it's signature to be more similar to the `leakage()` calculation (e.g. to add cleaning_mask), and then you explicitly apply it inside:
> 
> ```python
> def timing_parameters(geom, image, pulse_time, cleaning_mask, hillas_parameters):
> ```
> 
> Right now in the function, there is a line that basically re-applies the mask anyhow:
> 
> ```python
>     greater_than_0 = image > 0
>     pix_x = geom.pix_x[greater_than_0]
>     pix_y = geom.pix_y[greater_than_0]
>     image = image[greater_than_0]
>     pulse_time = pulse_time[greater_than_0]
> ```
> 
> So that could just be replaced explicitly with `cleaning_mask` in place of `greater_than_zero`

Ha thanks for the heads-up @kosack, I should have checked the code indeed.
I agree with you that the mask should be passed, this would be more explicit than `greater_than_0` that is hidden to the user.
Do you want me to take care of it in ctapipe?

kosack:
@vuillaut yes that would be great!  I only noticed since I also made the same mistake when using it!

PR #166, opened 2019-08-26
Zero suppression

garciagenrique:


garciagenrique:
What it should have appeared on the description :
Implementation of a volume reduction module on lstchain:
 - tailcuts_clean(8/4) + 3 dilates 
 - and the - default - none volume reduction method.

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,146 @@
...
+    :param pixel_mask: [array]
+            A boolean mask (array) containing the zero suppressed pixels.
+    :param camera_geometry: [Camera geometry information]
+            `ctapipe.instrument.CameraGeometry` object.
+
+    :return: [array]
+        A boolean mask (array) containing the zero suppressed pixels.
+    """
+    for i in range(3):
+        if i == 0:
```
vuillaut: why this condition and why the deepcopy ?
I think you can simply do:
```
for i in range(number_of_dilation):
    pixel_mask = dilate(geom, pixel_mask)
return pixel_mask
```

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,146 @@
...
+                                        boundary_thresh=boundary_zs,
+                                        keep_isolated_pixels=keep_isolated_pix_zs,
+                                        min_number_picture_neighbors=min_num_pict_neigh_zs
+                                        )
+    mask_pixel_0_suppressed = dilate_pixel_mask(mask_0_suppression, camera_geometry)
+
+    return mask_pixel_0_suppressed
+
+
+def dilate_pixel_mask(pixel_mask, camera_geometry):
```
vuillaut: This is just a matter of taste but I'd have kept the order of the arguments of other functions (`dilate`, `tailcut_cleaning`...): `(geom, pixel_mask)`
(available for other functions)

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,146 @@
...
+    min_num_pict_neigh_zs = kwargs['min_number_picture_neighbors']
+
+    mask_0_suppression = tailcuts_clean(camera_geometry,
+                                        image,
+                                        picture_thresh=picture_zs,
+                                        boundary_thresh=boundary_zs,
+                                        keep_isolated_pixels=keep_isolated_pix_zs,
+                                        min_number_picture_neighbors=min_num_pict_neigh_zs
+                                        )
+    mask_pixel_0_suppressed = dilate_pixel_mask(mask_0_suppression, camera_geometry)
```
vuillaut: `mask_0_suppression = dilate_pixel_mask(mask_0_suppression, camera_geometry)` would avoid creating yet another unused array.

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,146 @@
...
+            Modifies the event containers by applying the computed mask to the image,
+            the waveform and the pulse_time objects, as:
+            image[~mask] = 0, ...
+    """
+    camera = event.inst.subarray.tel[tel_id].camera
+    image = event.dl1.tel[tel_id].image
+    pulse_time = event.dl1.tel[tel_id].pulse_time
+    waveform = event.r1.tel[tel_id].waveform
+
+    if config['volume_reducer']['algorithm'] is None:
```
vuillaut: I know I am the one who suggested to have the same function structure even for no volume reduction but that might have been a bad suggestion. It would be more efficient to just
```
if config['volume_reducer']['algorithm'] is None:
     pass
```

garciagenrique:
> Thanks for the implementation.
> There is a couple of things to change I think. Others comments are matters of taste.

Thanks for the comments @vuillaut ! Starting to implement/change them now.

vuillaut:
The code looks good to me.
However, in the meantime a [PR was proposed in ctapipe](https://github.com/cta-observatory/ctapipe/pull/1121) to handle volume reduction with this method proposed by Lenka Tomankova et al so I'd propose to wait for this PR to be accepted in ctapipe and use the method here. What do you think?

garciagenrique:
> The code looks good to me.
> However, in the meantime a [PR was proposed in ctapipe](https://github.com/cta-observatory/ctapipe/pull/1121) to handle volume reduction with this method proposed by Lenka Tomankova et al so I'd propose to wait for this PR to be accepted in ctapipe and use the method here. What do you think?

No problem. Let's wait until the PR is accepted.

rlopezcoto:
Hi, do we want to keep this opened as a PR or maybe just move to open an Issue? If you want to construct something different as that from ctapipe, then it should be ok, otherwise we may think on opening an issue to use the ctapipe module whenever it is merged.
Keep in mind that the ctapipe module will not be available for lstchain until a new ctapipe stable version is released.

garciagenrique:
Hi @rlopezcoto,
I would say to use the ctapipe module in the moment it is merged. However, it also depends on today's need. If there is (or very soon will be) the need of starting applying the zero suppression on data we can merge this simpler version.

vuillaut:
> Hi @rlopezcoto,
> I would say to use the ctapipe module in the moment it is merged. However, it also depends on today's need. If there is (or very soon will be) the need of starting applying the zero suppression on data we can merge this simpler version.

I agree. I thought the module in ctapipe would be implemented earlier.
We can implement a first version here I think and update with ctapipe in the future.

rlopezcoto:
Shall we then move on with the validation of this PR and leave an open Issue to move to use the ctapipe module whenever it is ready?

garciagenrique:
Sure ! Let me merge the master branch and solve the potential conflicts that will appear (the PR dates from few months...)

garciagenrique:
Ok, this PR is ready for re-revision !

In the moment it is approved, I will open an issue to note that the method is a temporal version and that we should use the ctapipe version when available.

Please note that for the moment the computed mask that is applied for the zero suppression is compute at the dl1 level (from the integrated image) and applied to dl0 and dl1 levels. So the 'first' dl0 data computed is overwritten.

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
+"""
+**TEMPORTAL** module with functions to perform a volume reduction to LST data:
```
vuillaut: I think you mean "temporary". Below too.

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
...
+
+NOTE : The volume reduction SHOULD BE performed at the dl0 data levels.
```
vuillaut: "Applied at xxx level" for this method is unclear imo.
I would simply write "the volume reduction produces reduced waveforms and write them in the DL0 container"

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
...
+"""
+from ctapipe.image.cleaning import tailcuts_clean, dilate
+
+__all__ = ['check_volume_reduction_method',
+           'apply_volume_reduction',
+           'zero_suppression_tailcut_dilation'
+           ]
+
+
+def check_volume_reduction_method(config_file):
```
vuillaut: `get_volume_reduction_method` would be more appropriate

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    if config_file['volume_reducer']['algorithm'] is None:
+        flag = False
+        algorithm = ''
+    else:
+        flag = True
+        algorithm = config_file['volume_reducer']['algorithm']
+        print(algorithm)
+
+    return flag, algorithm
```
vuillaut: This doubles the number of output with no clear benefit imo, since you will need to check the flag anyway afterwards instead of just checking if the algorithm is `None`

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
...
+
+    """
+    camera = event.inst.subarray.tel[tel_id].camera
+
+    image = event.dl1.tel[tel_id].image  # Volume reduction mask computed, to date, at dl1 level !
+
+    pulse_time = event.dl1.tel[tel_id].pulse_time
+    waveform = event.dl0.tel[tel_id].waveform
+
+    algorithm = globals()[config['volume_reducer']['algorithm']]
```
vuillaut: Here I think you should use the `get_volume_reduction_method` and if not `None` get the parameters and apply it.

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
...
+    parameters = config['volume_reducer']['parameters']
+
+    mask_volume_reduction = algorithm(camera, image, **parameters)
```
vuillaut: I would name the mask something like `good_pixels` or `pixels_to_keep` to be more explicit rather than `mask`.

lstchain/reco/volume_reducer.py
vuillaut: And maybe indicate that the event must have been calibrated before

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
...
+        flag = True
+        algorithm = config_file['volume_reducer']['algorithm']
+        print(algorithm)
+
+    return flag, algorithm
+
+
+def apply_volume_reduction(event, tel_id, config):
+    """
+    Apply to an event the volume reduction method specified in the configuration file.
```
vuillaut: Indicate "to a calibrated event"

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,145 @@
...
+        algorithm = ''
+    else:
+        flag = True
+        algorithm = config_file['volume_reducer']['algorithm']
+        print(algorithm)
+
+    return flag, algorithm
+
+
+def apply_volume_reduction(event, tel_id, config):
```
vuillaut: I would remove the `tel_id` option and apply the volume reduction to all  `tel_id` in `tels_with_data`,  same as when calibration is applied.

lstchain/reco/volume_reducer.py
garciagenrique: To get into this method, I have previously checked in the `dl0_to_dl1.py` script : 
`if volume_reduction_algorithm is not None:
                    apply_volume_reduction(...)`
So I will just pass the name of the algorithm (stored in the `volume_reduction_algorithm` variable rather than calling `get_volumen_reduction_method` that does the same).

lstchain/reco/volume_reducer.py
garciagenrique: Sure !

lstchain/reco/volume_reducer.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    if config_file['volume_reducer']['algorithm'] is None:
+        algorithm = None
+    else:
+        algorithm = config_file['volume_reducer']['algorithm']
+
+    return algorithm
+
+
+def apply_volume_reduction(event, vol_reduction_algorithm, config):
```
vuillaut: If you need to pass the config, then passing the `vol_reduction_algorithm` arguments is redundant (since it's given by the config), no?

vuillaut: Even more so if you pass only a string and not the function itself.

lstchain/reco/tests/test_volume_reducer.py
```diff
@@ -0,0 +1,59 @@
...
+def test_zero_suppression_tailcut_dilation():
+    source = event_source(get_dataset_path('gamma_test.simtel.gz'))
+    for i, event in enumerate(source):
+        for tel_id in list(event.r0.tels_with_data):
+            if tel_id <= 4:
+                break
+            else:
+                continue
+        if tel_id <= 4:
+            break
```
vuillaut: the volume reduction should work on any camera, no?

lstchain/reco/tests/test_volume_reducer.py
```diff
@@ -0,0 +1,59 @@
...
+    cal = CameraCalibrator()
+    cal(event)
+    camera = event.inst.subarray.tel[tel_id].camera
+    imag = event.dl1.tel[tel_id].image
+
+    pixels_to_keep = zero_suppression_tailcut_dilation(camera, imag)
+    reduced_imag = np.copy(imag)
+    reduced_imag[~pixels_to_keep] = 0
+
+    assert reduced_imag[pixels_to_keep].all() is imag[pixels_to_keep].all()
```
vuillaut: you are asserting that `True is True`

PR #167, opened 2019-09-09
Time lapse corr spike b

pawel21:
I have modified code according Seiya's Nozaki idea:

1. Updating times from slices -1 to 38 remove spike b (small jump of signal) without interpolation (Seiya Nozaki idea).

2. Interpolation spike A is only needed for even last capacitor (lc) in the first half of the DRS4 ring.


lstchain/calib/camera/r0.py
```diff
@@ -253,21 +253,20 @@ def interpolate_pseudo_pulses(waveform, expected_pixel_id, fc, fc_old, n_modules
...
-
-                    # looking for spike B
-                    spike_b_position = int(
-                        (fc_old[nr_module, gain, pix] - 1 - fc[nr_module, gain, pix] + 2 * size4drs) % size4drs)
-                    if spike_b_position < roisize - 1:
-                        pixel = expected_pixel_id[nr_module*7 + pix]
-                        interpolate_spike_B(waveform, gain, spike_b_position, pixel)
+                            # The correction is only needed for even last capacitor (lc) in the
+                            # first half of the DRS4 ring
+                            if (fc_old[nr_module, gain, pix] + 39) % 2 == 0 and (fc_old[nr_module, gain, pix] + 39) % 1024 <= 510:
```
rlopezcoto: numbers like 38, 39 could be `len_drs4_window` (or something like that)

lstchain/calib/camera/r0.py
```diff
@@ -341,15 +340,18 @@ def do_time_lapse_corr(waveform, expected_pixel_id, local_clock_list, fc, last_t
...
+                # updating times from slices -1 to 38 remove spike b (small jump of signal)
+                # without interpolation (Seiya Nozaki idea)
                 posads0 = int((0 + fc[nr_module, gain, pix]) % size4drs)
                 if posads0+40 < 4096:
-                    last_time_array[nr_module, gain, pix, posads0:(posads0+39)] = time_now
+                    last_time_array[nr_module, gain, pix, (posads0+4096-1)%4096:(posads0+39)] = time_now
                 else:
-                    for k in prange(0, 39):
+                    for k in prange(-1, 39):
                         posads = int((k + fc[nr_module, gain, pix]) % size4drs)
```
rlopezcoto: shouldn't this run to 38?

pawel21:
@rlopezcoto Thanks for comments. 
Now, I am at MAGIC shift, so I am a bit busy and need some time to improve code.

FrancaCassol:
Hi @pawel21,
is this PR valid only before run 1575?
I wonder, if the change of firmware implies only the change of the sample range (first and last slice to be considered), should we not simply define it using variables that are assigned as trailets so to keep only one corrector class for all the runs? (the main code will verify which value to use as function of the run). Or there are other changes necessary for the new firmware?

jsitarek:
Hi @FrancaCassol 
In the script to calculate pedestal coefficients we plan to do exactly what you are saying, there is is very simple and indeed it is better not to duplicate the functions. 

But the dt correction function and spike interpolation are messy, because multiple conditions change, and also things can be optimized differently (e.g. the old correction required updating times from one capacitor before the actual readout to 38th capacitor including, and the code was optimized to avoid too many loops. Now the special cases when you cross the ring look different.

lstchain/calib/camera/r0.py
```diff
@@ -341,15 +340,18 @@ def do_time_lapse_corr(waveform, expected_pixel_id, local_clock_list, fc, last_t
                             val = waveform[gain, pixel, k] - ped_time(time_diff_ms)
                             waveform[gain, pixel, k] = val
 
+                # updating times from slices -1 to 38 remove spike b (small jump of signal)
+                # without interpolation (Seiya Nozaki idea)
                 posads0 = int((0 + fc[nr_module, gain, pix]) % size4drs)
                 if posads0+40 < 4096:
```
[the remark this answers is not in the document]
pawel21: done

lstchain/calib/camera/r0.py
pawel21: done

pawel21:
Hi @rlopezcoto , @FrancaCassol , @SeiyaNozaki
I have updated code according to new DRAGON firmware, and I have created function with "old" code to allow using low level calibration on "old" data before 2019/11/05 (new firmware update date).
We (me and Julian) tested code on "old" data (Run 1532) and "new" data (Run 1579), seems work OK.  Code itself check `run id`, then according `run id` choosing function to low level calibration. 

I made some clean-up.
I know that the code still needs cleaning, but now before observing Crab I think that the most important thing is to have working code (for new firmware) in the main repository.

Please look at my new codes and  get this into the lstchain master as soon as possible.

lstchain/calib/camera/r0.py
```diff
@@ -252,24 +315,80 @@ def interpolate_pseudo_pulses(waveform, expected_pixel_id, fc, fc_old, n_modules
                 for pix in prange(0, n_pix):
                     for k in prange(0, 4):
                         # looking for spike A first case
-                        abspos = int(1024 - roisize - 2 - fc_old[nr_module, gain, pix] + k * 1024 + size4drs)
+                        abspos = int(1025 - roi_size - 2 - fc_old[nr_module, gain, pix] + k * 1024 + size4drs)
```
rlopezcoto: Can we define 1025 as `size1drs` + 1, with `size1drs` = 1024 or something like that?
Could you please also add a small explanation about this change as a comment? we had some e-mail interchange that may be lost for the future...

lstchain/calib/camera/r0.py
```diff
@@ -252,24 +315,80 @@ def interpolate_pseudo_pulses(waveform, expected_pixel_id, fc, fc_old, n_modules
...
                         spike_A_position = int((abspos - fc[nr_module, gain, pix] + size4drs) % size4drs)
                         if (spike_A_position > 2 and spike_A_position < 38):
-                            pixel = expected_pixel_id[nr_module*7 + pix]
-                            interpolate_spike_A(waveform, gain, spike_A_position, pixel)
+                            # The correction is only needed for even
+                            # last capacitor (lc) in the first half of the
+                            # DRS4 ring
+                            if ((fc_old[nr_module, gain, pix] + (roi_size-1)) % 2 == 0 and (fc_old[nr_module, gain, pix] + (roi_size-1)) % 1024 <= 511):
+                                pixel = expected_pixel_id[nr_module * 7 + pix]
```
rlopezcoto: same here, 511 can be `size1drs`/2 - 1

lstchain/calib/camera/r0.py
```diff
@@ -252,24 +315,80 @@ def interpolate_pseudo_pulses(waveform, expected_pixel_id, fc, fc_old, n_modules
...
+        roi_size = 40
+        size4drs = 4096
+        n_gain = 2
+        n_pix = 7
+        for nr_module in prange(0, n_modules):
+            for gain in prange(0, n_gain):
+                for pix in prange(0, n_pix):
+                    for k in prange(0, 4):
+                        # looking for spike A first case
+                        abspos = int(1024 - roi_size - 2 -fc_old[nr_module, gain, pix]+ k * 1024 + size4drs)
```
rlopezcoto: same comments as above

lstchain/calib/camera/r0.py
```diff
@@ -320,22 +441,72 @@ def subtract_pedestal_jit(event_waveform, expected_pixel_id, fc_cap, pedestal_va
...
+    """
+    Numba function for time lapse baseline correction.
+    This is function for data from 2018/10/10 to 2019/11/04 with old firmware.
+    Change waveform array.
+    """
+    size4drs = 4096
+    roi_size = 40
+    for nr_module in prange(0, number_of_modules):
+        time_now = local_clock_list[nr_module]
+        for gain in prange(0, 2):
```
rlopezcoto: (0, n_gain)

lstchain/calib/camera/r0.py
```diff
@@ -320,22 +441,72 @@ def subtract_pedestal_jit(event_waveform, expected_pixel_id, fc_cap, pedestal_va
...
+            for pix in prange(0, 7):
```
rlopezcoto: (0, n_pix)

lstchain/calib/camera/r0.py
pawel21: done

lstchain/calib/camera/r0.py
pawel21: done

lstchain/calib/camera/r0.py
pawel21: done

lstchain/calib/camera/r0.py
pawel21: done

lstchain/calib/camera/r0.py
pawel21: done

lstchain/calib/camera/r0.py
pawel21: done

lstchain/calib/camera/r0.py
```diff
@@ -170,25 +191,57 @@ def time_lapse_corr(self, event, tel_id):
         event : `ctapipe` event-container
         tel_id : id of the telescope
         """
+
+        run_id = event.lst.tel[1].svc.configuration_id
```
FrancaCassol: Hi @pawel21,
please, change tel[1] with tel_id

pawel21: Done

FrancaCassol:
Hi  @rlopezcoto ,

for me this code is good for merging

Issue #168, opened 2019-09-10
AtributeError when running create_pedestal_file.py


morcuended:
When trying to use create_pedestal_file.py I get this error:

```
Traceback (most recent call last):
  File "create_pedestal_file.py", line 60, in <module>
    pedestal = DragonPedestal(tel_id=telid, n_module=n_modules)
  File "/fefs/home/daniel.morcuende/software/cta-lstchain/lstchain/calib/camera/drs4.py", line 24, in __init__
    self.meanped = np.zeros((self.n_gain, self.n_pixels, self.size4drs))
AttributeError: 'DragonPedestal' object has no attribute 'size4drs'
```
Am I doing something wrong or the variable "size4drs" should be defined somewhere else?

rlopezcoto:
Thanks for reporting @morcuended. It seems that with the global definition of `size4drs` the call inside `DragonPedestal` was not updated. Could you make a quick fix it @pawel21 or should I do it?

pawel21:
Thank you @morcuended. I will fix it.

rlopezcoto:
Solved in #169 by @pawel21

rlopezcoto:
@morcuended can you confirm that the problem was solved? Thanks!

morcuended:
Yes, now it's working properly

PR #169, opened 2019-09-11
Fix atribute error when running script to create pedestal file

pawel21:
According to AtributeError when running create_pedestal_file.py #168
I have fixed error and add some comments.


rlopezcoto:
Hi @pawel21, wouldn't it be easier to leave a global definition for `sizedrs4` and change the call? I think that was the original intention of defining this outside of the `DragonPedestal` function

pawel21:
Hi @rlopezcoto .
Yes, you are right. Thank you.
I also have defined another parameters as global, I think, this is more readable.

rlopezcoto:
Yes, great! I'll merge as soon as Travis confirms everything is alright...

PR #180, opened 2019-09-27
params keys in io

vuillaut:
- Improve consistency across lstchain 
- easier for user that just want to load the param table

not related (but I was lazy to make another PR sorry):
- fix the ctaplot import that is not in the requirements with a try


vuillaut:
Thank you @misabelber !

Issue #182, opened 2019-10-02
Low level calibration code not compatible with missing modules


FrancaCassol:
Hi,

it seems that the present low level calibrator is not compatible with missing modules (as in the data since a week). In the loops over modules one should not use the index over the global number of modules "range(0, n_modules)" but the number of the modules stored in event.lst.tel[1].svc.module_ids.

@pawel21 could you take care of the correction? (otherwise I can try to do it) 
This is necessary to get low level calibration files for the present data.

pawel21:
Hi,
This is error in my code (works only for 265 modules)
I will fix it.

pawel21:
I have checked my code using data from Run 1413 and I have discovered that code should work for less than 265 modules.
I have already implemented (in r0.py) to use correct number of modules (event.lst.tel[telid].svc.num_modules).

What kind of problem did you find?

pawel21:
I found one possible scenario, when We can have problem with low level calib. 
This can happen when the number of module using to create pedestal file don't agree with number of modules from data run.

I have created PR #183

rlopezcoto:
Solved in #209

PR #183, opened 2019-10-03
Allow set number of modules

pawel21:
Issue #182 

I have modified code to allow set number of modules as parameter. 
This will allow use low level corrections, when modules as missing (works least than 265 modules)
The parameter `n_module` must agree with number of module use to create pedestal file.


rlopezcoto:
Hi @pawel21, I see a potential problem here. When we perform regular observations, the file used to produce a _pedestal_ will likely have the same number of modules as the one used for _data_. As we are working right now, we may use a file to calculate the _pedestal_ that has all the modules, but the _data_ one has less modules.
For me, ideally, if the file for the _pedestal_ covers all the modules and the one for the _data_ has *less* modules, the correction should still work. If the file used for the _pedestal_ has some missing modules that are present in the _data_ file, then an error should be issued. Would it be possible to program it that way?

FrancaCassol:
Dear @pawel21,
if I understand correctly what you are doing, you don't need any input parameters. As said in issue #182,  you must specifically use the module numbers contained in the event.lst.tel[tel_id].svc.module_ids array when you perform the loop on the modules. In this array you have the number of modules present during the run (for example at present the 118 is missing).  
Obviously, the pedestal run must have at least all the modules of the data-taking runs, that means if during the data-taking a module is switched on, the shifters must perform a new pedestal run.

pawel21:
Generally, my code to perform correction using number of pixels from current event `n_modules = event.lst.tel[self.tel_id].svc.num_modules`.

I see problem in function 
`def _load_calib(self):
        """
        Function to load pedestal file.
        """
        if self.pedestal_path:
            with fits.open(self.pedestal_path) as f:
                pedestal_data = np.int16(f[1].data)
                self.pedestal_value_array[:, :, :self.size4drs] = pedestal_data - self.offset
                self.pedestal_value_array[:, :, self.size4drs:self.size4drs + 40] \
                    = pedestal_data[:, :, 0:40] - self.offset
`
When the shape of `pedestal_data` from pedestal file not agree with shape of `pedestal_value_array` (array store pedestal value, which is created in `__init__`  [line 114] )

In case, when pedestal file covers all module, function works well.

pawel21:
Generally, my code to perform correction using number of pixels from current event `n_modules = event.lst.tel[self.tel_id].svc.num_modules`. 

@FrancaCassol Thank you for explanation. You are right. 
I will change it. 
I will try do as fast as I can, but I am at MAGIC shift, so I am a bit busy.

I see another problem in function 
```
def _load_calib(self):
        """
        Function to load pedestal file.
        """
        if self.pedestal_path:
            with fits.open(self.pedestal_path) as f:
                pedestal_data = np.int16(f[1].data)
                self.pedestal_value_array[:, :, :self.size4drs] = pedestal_data - self.offset
                self.pedestal_value_array[:, :, self.size4drs:self.size4drs + 40] \
                    = pedestal_data[:, :, 0:40] - self.offset
```
When the shape of `pedestal_data` from pedestal file not agree with shape of `pedestal_value_array` (array store pedestal value, which is created in `__init__`  [line 114] )
When we use pedestal file, which covers all pixels, correction should work well.

pawel21:
@FrancaCassol I am using `expected_pixel_id = ev.lst.tel[tel_id].svc.pixel_ids` to obtain pixel ids to perform correction and I think this field contain only "good" pixels (without pixel from "dead" module). 
What do you think about it?

pawel21:
I found bug in code to create pedestal file, when not all the modules are on.
I will fix it.

pawel21:
I closed.
I think, we don't need set number of modules as parameter.
About pedestal file: I have another PR #209 , where I solved problem with missing module.

PR #184, opened 2019-10-03
use pixels of only main island for hillas calculation

SeiyaNozaki:
Hello, this is the first pull request for my life :)

All survived pixels after image cleaning are used for hillas calculation with the current lstchain. If so, the estimation of shower axis could be bad when multiple islands exist, then all of the parameter could be useless.
With this branch, the number of island is calculated at first, then pixels in only main island which has maximum intensity is used for the parameter calculation later.
Please have a check.
Screenshot from 2019-07-24 04-25-33

### DIFF
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -108,6 +108,9 @@ def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}
             max_island_label = np.argmax(n_pixels_on_island)
             signal_pixels[island_labels!=max_island_label] = False
 
+        if np.sum(signal_pixels) < 3 :
+            return None
+
         hillas = hillas_parameters(camera[signal_pixels], image[signal_pixels])
 
         # Fill container
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -108,9 +108,6 @@ def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}
             max_island_label = np.argmax(n_pixels_on_island)
             signal_pixels[island_labels!=max_island_label] = False
 
-        if np.sum(signal_pixels) < 3 :
-            return None
-
         hillas = hillas_parameters(camera[signal_pixels], image[signal_pixels])
 
         # Fill container
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -27,7 +27,6 @@ from ..io.lstcontainers import ExtraImageInfo
 from ..calib.camera import lst_calibration, load_calibrator_from_config
 from ..io import DL1ParametersContainer, standard_config, replace_config
 from ctapipe.image.cleaning import number_of_islands
-from copy import deepcopy
 
 import tables
 from functools import partial
@@ -60,7 +59,7 @@ filters = tables.Filters(
 )
 
 
-def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}):
+def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}, use_main_island=True):
     """
     Return a DL1ParametersContainer of extracted features from a calibrated event.
     The DL1ParametersContainer can be passed to be filled if created outside the function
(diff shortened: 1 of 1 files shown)


rlopezcoto:
Hi Seiya,

Thanks for starting to contribute!

Regarding implementing this option, I agree it is probably the way to go when having images like the one you show, but I'm not sure if this will always be the case. We can implement it as an option that is True by default, if nobody opposes. Do you think that you could implement something like that?

lstchain/reco/dl0_to_dl1.py
```diff
@@ -92,19 +94,33 @@ def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}
...
+
+        # check the number of islands 
+        num_islands, island_labels = number_of_islands(camera, signal_pixels)
+        intensity_of_island = np.zeros(num_islands+1)
+
+        for iisland in range(1, num_islands+1):
+            intensity_of_island[iisland] = np.sum(image[island_labels==iisland])
+
+        max_island_label = np.argmax(intensity_of_island)
+        signal_pixels_main = deepcopy(signal_pixels)
```
vuillaut: I don't think this `deepcopy` is required and this is a time consuming operation.
`signal_pixels` is used later only by `set_n_islands`.
You can `set_n_islands` and then update `signal_pixels[island_labels!=max_island_label] = False`

lstchain/reco/dl0_to_dl1.py
```diff
@@ -92,19 +94,33 @@ def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}
...
-        hillas = hillas_parameters(camera[signal_pixels], image[signal_pixels])
+
+        # check the number of islands 
+        num_islands, island_labels = number_of_islands(camera, signal_pixels)
+        intensity_of_island = np.zeros(num_islands+1)
+
+        for iisland in range(1, num_islands+1):
+            intensity_of_island[iisland] = np.sum(image[island_labels==iisland])
+
+        max_island_label = np.argmax(intensity_of_island)
```
vuillaut: I am wondering if the sum of intensity is the best approach here rather than the number of pixels - or a combination of both, the mean.
If we have two pixels with high signal, is it more likely to be the shower compared to 4-5 pixels with average signal (but passed the cleaning of course).

lstchain/reco/dl0_to_dl1.py
```diff
@@ -92,19 +94,33 @@ def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}
...
-        dl1_container.set_timing_features(camera[signal_pixels],
-                                          image[signal_pixels],
-                                          pulse_time[signal_pixels],
+        dl1_container.set_timing_features(camera[signal_pixels_main],
+                                          image[signal_pixels_main],
+                                          pulse_time[signal_pixels_main],
                                           hillas)
-        dl1_container.set_leakage(camera, image, signal_pixels)
+        dl1_container.set_leakage(camera, image, signal_pixels_main)
         dl1_container.set_n_islands(camera, signal_pixels)
```
vuillaut: in response to my previous comment, to avoid recomputing the number of islands, you could set it manually :
`dl1_container.n_islands = num_islands`
But the operation cost is a low and you could argue that it is more clear so as you prefer.

vuillaut:
> Hi Seiya,
> 
> Thanks for starting to contribute!
> 
> Regarding implementing this option, I agree it is probably the way to go when having images like the one you show, but I'm not sure if this will always be the case. We can implement it as an option that is True by default, if nobody opposes. Do you think that you could implement something like that?

Hi.
I agree with @rlopezcoto. `True` by default seems fair to me - I expect the feature to improve results more often than it does not.

vuillaut:
And I want to add, thanks for the (first of many) pull-request and contribution !! :)

lstchain/reco/dl0_to_dl1.py
SeiyaNozaki: I see. I modified the code to use # of pixels on island to search main island following your comment for the moment. I will study which way is the best later.

SeiyaNozaki:
Ok, thank you for your comments!
I committed the code following your comments.
Any comments are welcome!

lstchain/reco/dl0_to_dl1.py
vuillaut: It was a suggestion, I have no idea which could be better.
But if you can have a look that'd be great.
In the meantime, I think it's a good implementation.

vuillaut:
It seems the build is failing because some parameters are now not well reconstructed (`NaN` values). Could you have a look pls?

SeiyaNozaki:
If the number of pixels on main island is less than three, width could be NAN.
I modified the code that "get_dl1" returns None if so.

vuillaut:
> If the number of pixels on main island is less than three, width could be NAN.
> I modified the code that "get_dl1" returns None if so.

There are many reasons why parameters could be `nan` or `inf`. This is dealt with at later stages, for example by the `filter_events` function. I would not add such a filter here for two reasons:
- the parameter might not be used by later stages, in which case the analysis can go on
- keeping the event anyway can help for debugging later on

Moreover, this case (n_pixels < 3) was already possible without taking into account the sub-islands and the pipeline was fine...

vuillaut:
I think the unit test is actually the problem there 😅

lstchain/reco/dl0_to_dl1.py
```diff
@@ -92,8 +93,26 @@ def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}
...
+            n_pixels_on_island = np.zeros(num_islands+1)
+
+            for iisland in range(1, num_islands+1):
+                n_pixels_on_island[iisland] = np.sum(island_labels==iisland)
+              
+            max_island_label = np.argmax(n_pixels_on_island)
+            signal_pixels[island_labels!=max_island_label] = False
+
+        if np.sum(signal_pixels) < 3 :
+            return None
```
vuillaut: Please remove and update (pull master), the tests should pass.

vuillaut: thanks!

PR #189, opened 2019-10-09
adding a release drafter file to keep track of changes

vuillaut:
See https://github.com/apps/release-drafter

That's what is currently used in ctapipe.


rlopezcoto:
yes, this sounds good, thanks

Issue #181, opened 2019-10-01
lstchain for real data analysis


vuillaut:
In its current state, a lot of MC information is used in several places to perform the reconstruction (e.g. telescope focal, pointing direction).
I will start looking more precisely into this.
In order to have a transparent use of the chain, the least bad option I see for now is to introduce `try/except` in required places depending on whether the data is MC or not.
To be discussed.

misabelber:
Great Thomas! I've been thinking that this issue is becoming very necessary so it's a good idea to start working on it right away. Let me know if you need any help. Probably we will need also some notebooks & scripts which work with real data.

rlopezcoto:
currently adapted and running on real data at all data levels. Probably some unification of scripts should be done, but since we would like to move them to `Tools` when the new ctapipe version comes out, it can be done at that time.

PR #190, opened 2019-10-10
Adding script to reco from dl1 files

vuillaut:
I often find myself starting again from DL1 instead of DL0.
This script does the reconstruction from DL1 (it's almost a copy of `lst-recopipe.py`).


rlopezcoto:
Hi, 
not sure if the almost duplication of a file is the best solution in this case. Wouldn't it be better to include a flag, that does not have any value by default (although I think that from dl0 to dl2 should be the default)?

vuillaut:
Hi.
Yes, it's a good question and I asked myself the same.
I finally prefered this approach because `dl1 to dl2` is still something different from `dl0 to dl2` (maybe the scripts should be named this way rather than recopipe!).
But it's just a conceptual difference, I have no strong opinion there.

vuillaut:
The best approach might be to have scripts for 
- `dl0_to_dl1`
- `dl1_to_dl2`
- `dl2_to_dl3`
- `dl0 to dl2` that just uses `dl0_to_dl1` and `dl1_to_dl2` 
- etc... 

?

rlopezcoto:
Sounds like a good compromise. Not to multiply too much the number of scripts I would go for:

- `dl0_to_dl1`
- `dl1_to_dl2`
- `dl2_to_dl3`
- `dl0_to_dl3` (since we do not have the `dl2_to_dl3` script yet, it is ok to leave it up to dl2 now, but just to keep it in mind).
In any case, I think that the `recopipe` naming had some "historical" reason (maybe @misabelber can give us some words here). Otherwise, I'm happy with the solution.

misabelber:
Hi! Yeah, I'm happy with the scripts Thomas is proposing regarding the needing to have access to each data level. The historical reason to have lst-recopipe its that it was a script destined to the user who wants to use already trained RFs over a raw data file (or simtel file in this case) without caring of the mid data levels. Apart from those scripts we will still need the one for training the RFs (either lst-trainpipe or lst-rfperformance, I personally alaways use lst-rfperformance because you can  actually check the performance plots of the RFs). We can also discuss the final name for those¿?. 
Also we should start adapting all these scripts to be used either in MC data or with real data as Thomas started in #181 .

vuillaut:
I am closing this PR in favor of #203 (plus I seem to have messed up when merging master).

Issue #174, opened 2019-09-20
Omission of attributes within lstchain/calib/camera/r0.py


garciagenrique:
Hi all,

Working in another stuff with (@devine73 and @vuillaut, thanks !) we have realised that there is  the use of the attribute self.tel_id instead of the variable tel_id, passed to the corresponding function. Specifically we have seen it on LSTR0Corrections.calibrate (lines 126-128, 130-131) and LSTR0Corrections.substract_pedestal, lines 137 and so on.

For the moment, this omission has no effect because tel_id and self.tel_id should be only 1 (working only with LST1, i.e., tel_id = 1. - (!) IF (!) .allowed_tels within the config is changed...). 

I can start having a look to the code and try and correct it, although it will be (likely) finished next week. 

Cheers,
Enrique

PR #192, opened 2019-10-10
Issue #174 self.tel_id

garciagenrique:
Pull request to correct Issue #174.
As mentioned on the issue, the attribute `self.tel_id` should have been `tel_id`. In this pull request it has been done : 

- Modification of the the attribute `self.tel_id` --> `tel_id` all along the file.
- Modification of the name of the attribute on the last function to be consistent with the name of the attributes/variables, i.e., `telid` --> `tel_id`
- Some PEP 8 cleaning

However; 

1. Because for the moment there is no more that a single telescope, the modifications of the code doesn't have any real effect. They would only prevent potential errors/problems on the future. And avoid real misunderstanding of the workflow.  
2. I have seen that this same problem (the use of `self.tel_id` when it should be use only the variable), also appear on the following files : `calib/pedestals.py`, `calib/drs4.py`, `calib/flatfield.py` and `tools/calc_camera_calibration.py` (In my opinion the `tel_id` defined at the beginning of each class should be removed). The issue gets a little bit bigger and to date quite complex for myself (as I have not contributed since the beginning), although I can take care of it if needed.

### DIFF
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -131,7 +131,7 @@ class LSTR0Corrections(CameraR0Calibrator):
             event.r1.tel[tel_id].trigger_time = event.r0.tel[tel_id].trigger_time
 
             samples = event.r1.tel[tel_id].waveform[:, :, self.r1_sample_start:self.r1_sample_end]
-            event.r1.tel[tel_id].waveform = samples.astype('float32') - self.offset
+            event.r1.tel[tel_id].waveform = samples.astype('int16') - self.offset
 
     def subtract_pedestal(self, event, tel_id):
         """
@@ -158,7 +158,7 @@ class LSTR0Corrections(CameraR0Calibrator):
             n_modules)
         event.r1.tel[tel_id].trigger_type = event.r0.tel[tel_id].trigger_type
         event.r1.tel[tel_id].trigger_time = event.r0.tel[tel_id].trigger_time
-        event.r1.tel[tel_id].waveform = samples[:, :, :]
+        event.r1.tel[tel_id].waveform = samples
 
     def time_lapse_corr(self, event, tel_id):
         """
@@ -174,7 +174,7 @@ class LSTR0Corrections(CameraR0Calibrator):
         local_clock_list = event.lst.tel[tel_id].evt.local_clock_counter
         n_modules = event.lst.tel[tel_id].svc.num_modules
         for nr_module in range(0, n_modules):
-            self.first_cap_time_lapse_array[nr_module, :, :] = self._get_first_capacitor(event, nr_module)
+            self.first_cap_time_lapse_array[nr_module, :, :] = self._get_first_capacitor(event, nr_module, tel_id)
 
         #If R1 container exist modifies it
         if isinstance(event.r1.tel[tel_id].waveform, np.ndarray):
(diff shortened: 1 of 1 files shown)


garciagenrique:
I would like to assign as referee @FrancaCassol and @pawel21 ( and @vuillaut @rlopezcoto too) as I have seen that they are the two main contributors, although I cannot assign them. 

I would like to rise couple of questions to them too:
Probably I am missing something, however I don't get why the following lines are repeated so many times inside `r0.py`:
`event.r1.tel[tel_id].trigger_type = event.r0.tel[tel_id].trigger_type`
`event.r1.tel[tel_id].trigger_time = event.r0.tel[tel_id].trigger_time` 
`event.r1.tel[tel_id].waveform = samples[:, :, :]`

Also `_get_first_capacitor` is repeated 3 times inside `calibrate`, as each defined method is going to call the first mentioned function.

lstchain/calib/camera/r0.py
```diff
@@ -123,99 +123,101 @@ def __init__(self, **kwargs):
...
             expected_pixel_id,
             self.first_cap_array,
             self.pedestal_value_array,
             n_modules)
-        event.r1.tel[self.tel_id].trigger_type = event.r0.tel[self.tel_id].trigger_type
-        event.r1.tel[self.tel_id].trigger_time = event.r1.tel[self.tel_id].trigger_time
-        event.r1.tel[self.tel_id].waveform = samples[:, :, :]
+        event.r1.tel[tel_id].trigger_type = event.r0.tel[tel_id].trigger_type
+        event.r1.tel[tel_id].trigger_time = event.r0.tel[tel_id].trigger_time
+        event.r1.tel[tel_id].waveform = samples[:, :, :]
```
vuillaut: Why `samples[:,:,:]` and not just `samples` ?
Nothing wrong there but the reading is not simplified

lstchain/calib/camera/r0.py
```diff
@@ -418,8 +421,8 @@ def __init__(self, **kwargs):
...
-        for telid in event.r0.tels_with_data:
-            event.r1.tel[telid].trigger_type = event.r0.tel[telid].trigger_type
-            event.r1.tel[telid].trigger_time = event.r0.tel[telid].trigger_time
-            samples = event.r0.tel[telid].waveform[:,:,self.r1_sample_start:self.r1_sample_end]
-            event.r1.tel[telid].waveform = samples.astype('float32')- self.offset
+        for tel_id in event.r0.tels_with_data:
+            event.r1.tel[tel_id].trigger_type = event.r0.tel[tel_id].trigger_type
+            event.r1.tel[tel_id].trigger_time = event.r0.tel[tel_id].trigger_time
+            samples = event.r0.tel[tel_id].waveform[:, :, self.r1_sample_start:self.r1_sample_end]
+            event.r1.tel[tel_id].waveform = samples.astype('float32') - self.offset
```
vuillaut: Why type suddenly becomes `float32` while it was `int16` before ?
R1 is stored as `int16` right?

pawel21: After call

```
lst_r0.subtract_pedestal(ev)
lst_r0.time_lapse_corr(ev)
lst_r0.interpolate_spikes(ev)
``` 
R1 is store as `int16`, but after call `lst_r0.calibrate(ev)` as `float32`.
This is mistake in my code. I am sorry.
Thanks @vuillaut for report.
Dear @garciagenrique could you change `samples.astype('float32')` to `samples.astype('int16')` or I can do it in next PR.

pawel21:
Dear @garciagenrique 
Thank you for your PR, very good job.
I think we need also modified example notebook [low level corrections](https://github.com/cta-observatory/cta-lstchain/blob/master/notebooks/low_level_corrections.ipynb) I can do it after merge this PR.

lstchain/calib/camera/r0.py
garciagenrique: Right, thanks Thomas

lstchain/calib/camera/r0.py
garciagenrique: Sure @pawel21, I can change it now.

garciagenrique:
Thanks for your comments @pawel21 !
If you need help changing the attribute in other files, let me know ! ;-)

vuillaut:
@garciagenrique , all suggested corrections have been made right?
@pawel21 proposed to correct the notebook in another PR.

PR #194, opened 2019-10-17
Using log in names when necessary

vuillaut:
- add `log_mc_energy`, `log_reco_energy` and `log_intensity` for clarity for the user (corresponds to previous `mc_energy`, `reco_energy` and `intensity` respectively)
- keep `mc_energy` and `reco_energy` in TeV
- keep `intensity` in linear scale


vuillaut:
> We need very badly to be compliance with astropy units...

I completely agree, it has been on the todo-list for a long time now...

PR #196, opened 2019-10-30
script to perform some tagging-related analysis on real muon events

pillera:
added a (not optimized) script to perform some tagging-related analysis on real muon events, full muon analysis content of the notebook analyze_real_muon_events is not yet included


rlopezcoto:
Thanks @vuillaut for reviewing the coding style thoroughly, but I would rather have a working code before starting to look into formatting.
@pillera regarding functionality, the code at the moment is:
- not working with new data because the tel_id tel[0] is not working anymore
- the threshold is currently applied to the maximum of the traces instead of number of phe
- only a few values of every muon are stored in the final data table, many needed values for the analysis are missing.
- no cuts on the muon parameters are applied and therefore when applied to a real data file a lot of background events enter into the analysis.
This is reducing a bit the functionality with respect to what was already in the [real muon analysis notebook](https://github.com/cta-observatory/cta-lstchain/blob/master/notebooks/Analyze_real_muon_data.ipynb). I've been working a bit on this and added some of the missing info, added the possibility of storing images and veto many of the events that are entering into the analysis right now. I can commit my changes directly to this PR or maybe open a new one, are you ok with this?

vuillaut:
> Thanks @vuillaut for reviewing the coding style thoroughly, but I would rather have a working code before starting to look into formatting.
> @pillera regarding functionality, the code at the moment is:
> 
> * not working with new data because the tel_id tel[0] is not working anymore
> * the threshold is currently applied to the maximum of the traces instead of number of phe
> * only a few values of every muon are stored in the final data table, many needed values for the analysis are missing.
> * no cuts on the muon parameters are applied and therefore when applied to a real data file a lot of background events enter into the analysis.
>   This is reducing a bit the functionality with respect to what was already in the [real muon analysis notebook](https://github.com/cta-observatory/cta-lstchain/blob/master/notebooks/Analyze_real_muon_data.ipynb). I've been working a bit on this and added some of the missing info, added the possibility of storing images and veto many of the events that are entering into the analysis right now. I can commit my changes directly to this PR or maybe open a new one, are you ok with this?

Not really a thorough review 😅 
Thanks for the much more useful review regarding code working as expected or not, I agree it should come first.
But remember that you will not be the only user of this code and currently, if I wanted to use it, I would have to look deeply into the code to understand what it does and how it works. Writing more explicit code and some docstring save a lot of time in the long run.

pillera:
@rlopezcoto and @vuillaut thank you for your comments. As I wrote when I did the pull request, this script would only do some tagging related analysis, it wasn't optimized and I didn't work at all on the style and user-friendlyness. If you want I can rewrite the script in order to do a more extensive analysis applying all your comments regarding functionality and formatting. 
In this case could you specify what changes have to be done regarding the quoted comment, or commit your changes directly if you prefer?
Thanks
> Thanks @vuillaut for reviewing the coding style thoroughly, but I would rather have a working code before starting to look into formatting.
> @pillera regarding functionality, the code at the moment is:
> 
> * not working with new data because the tel_id tel[0] is not working anymore
> * the threshold is currently applied to the maximum of the traces instead of number of phe
> * only a few values of every muon are stored in the final data table, many needed values for the analysis are missing.
> * no cuts on the muon parameters are applied and therefore when applied to a real data file a lot of background events enter into the analysis.
>   This is reducing a bit the functionality with respect to what was already in the [real muon analysis notebook](https://github.com/cta-observatory/cta-lstchain/blob/master/notebooks/Analyze_real_muon_data.ipynb). I've been working a bit on this and added some of the missing info, added the possibility of storing images and veto many of the events that are entering into the analysis right now. I can commit my changes directly to this PR or maybe open a new one, are you ok with this?

rlopezcoto:
Hi,

sorry if it sounded a bit harsh, but at the moment this script is doing the full muon analysis (with its corresponding long running time), but not really selecting useful muons, but many events.
I commited some changes that make it more useful for real data analysis, but it is still non-optimal and very dirty. If you want to write a script I would suggest to separate this into:
- a simple, quick script that tags muons based on a quick cut in size
- the analysis script that selects only good muon events and writes all their parameters to a table.

Right now the script looks more like the second option, but there is still some work to do. Regarding moving the functions to more general places, I agree with @vuillaut, but these are just some copied functions from (here)[https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/image/muon/muon_reco_functions.py] and I re-wrote some of them because they were not working with real data. Additionally there is right now discussion on re-factoring everything in the muon analysis, so not sure it is worth it to move it now or wait until things are settled...

pillera:
I added a function to tag muons according to a hardcoded threshold.

PR #198, opened 2019-11-02
writing pe image in dl1 files

vuillaut:


rlopezcoto:
wait, sorry, on some second thoughts... does it mean that we always store the information of the pe image when we write DL1?

vuillaut:
> wait, sorry, on some second thoughts... does it mean that we always store the information of the pe image when we write DL1?

If the information is available, yes.

rlopezcoto:
why do we want to write by default the information of the image of every event at a DL1 level? How much does this increase the size of the files?
In principle, for regular analysis, we should not need images at a higher level. I agree to add this option in case we would like to propagate this information for some kind of analysis, but not by default.

vuillaut:
> Looks good to me, that image is the true number of photoelectrons arriving to each pixel?

Yes. It's available only in specific MC files but is very useful for benchmarks.

vuillaut:
I don't know why all the recent history got included in my last commit...
I'll just close this PR and opened #205 instead.

Issue #197, opened 2019-10-31
Gain Selection


vuillaut:
gain selection is probably wrong.
See https://github.com/cta-observatory/ctapipe/issues/1166

Quick fix: go back to the manual gain selection that is still in lstchain but noted as deprecated.

rlopezcoto:
Current threshold from the [config file](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/data/lstchain_standard_config.json) should be 20.9 (phe) instead of 4094 (counts) until next ctapipe version is realeased (currently following v0.7.0)

rlopezcoto:
this should be solved in `lstchain >= v0.6` that uses `ctapipe v0.8`

PR #200, opened 2019-11-02
gain selector threshold in p.e.

vuillaut:


garciagenrique:
This should answer my comment in #201, I think.

rlopezcoto:
I think that we should add here some comments saying that we know that this is not correct, but it is a bugfix related to [this](https://github.com/cta-observatory/ctapipe/issues/1166) bug in ctapipe and that must be changed as soon as possible, whenever we follow the next ctapipe release. Maybe keep the old threshold commented out and add some explanation may do it.

vuillaut:
> I think that we should add here some comments saying that we know that this is not correct, but it is a bugfix related to [this](https://github.com/cta-observatory/ctapipe/issues/1166) bug in ctapipe and that must be changed as soon as possible, whenever we follow the next ctapipe release. Maybe keep the old threshold commented out and add some explanation may do it.

100% agree!
However, json format does not support commenting... 
Actually, I am not so sure anymore we want to merge this PR. As this is a configuration, users can change the number in the config they use...

kosack:
> However, json format does not support commenting...

If you are using JSON just because you are using the ctapipe configuration system, you should note that it also supports python-based config files that allow commenting.  See  the traitlets.config documentation, but they look like this:

```py
# config file
c = get_config()

c.Foo.name = u'bestname'
c.Bar.othervalue = 10
```

You can even auto-generate such a config file by calling `generate_config_file()` on a Tool instance (it will have all the options, and full comments as well).

vuillaut:
> > However, json format does not support commenting...
> 
> If you are using JSON just because you are using the ctapipe configuration system, you should note that it also supports python-based config files that allow commenting. See the traitlets.config documentation, but they look like this:
> 
> ```python
> # config file
> c = get_config()
> 
> c.Foo.name = u'bestname'
> c.Bar.othervalue = 10
> ```
> 
> You can even auto-generate such a config file by calling `generate_config_file()` on a Tool instance (it will have all the options, and full comments as well).

For now we just load the config as a dict.
+1 on using traitlets.

rlopezcoto:
Discussing with @FrancaCassol #241 and how to perform the gain selection for real data, we realized that the threshold from the [config file](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/data/lstchain_standard_config.json):

https://github.com/cta-observatory/cta-lstchain/blob/6a7a81dd2cc5ad419ef8e76935e467f15da476b3/lstchain/data/lstchain_standard_config.json#L104

is not applied to real data. The threshold to real data is actually applied by the `combine_channels` (deprecated?) function:

https://github.com/cta-observatory/cta-lstchain/blob/6a7a81dd2cc5ad419ef8e76935e467f15da476b3/lstchain/reco/dl0_to_dl1.py#L269

and the threshold properly applied at the waveform level. Did we get anything wrong @vuillaut?

vuillaut:
> Discussing with @FrancaCassol #241 and how to perform the gain selection for real data, we realized that the threshold from the [config file](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/data/lstchain_standard_config.json):
> 
> https://github.com/cta-observatory/cta-lstchain/blob/6a7a81dd2cc5ad419ef8e76935e467f15da476b3/lstchain/data/lstchain_standard_config.json#L104
> 
> is not applied to real data. The threshold to real data is actually applied by the `combine_channels` (deprecated?) function:
> 
> https://github.com/cta-observatory/cta-lstchain/blob/6a7a81dd2cc5ad419ef8e76935e467f15da476b3/lstchain/reco/dl0_to_dl1.py#L269
> 
> and the threshold properly applied at the waveform level. Did we get anything wrong @vuillaut?

Hi @rlopezcoto 
Indeed. I may have forgotten about that... 
@FrancaCassol, after `LSTCameraCalibrator`, the DL1 image still contains two channels?

FrancaCassol:
Hi @vuillaut  @rlopezcoto,

indeed, for the moment there is no gain selection in the LSTCameraCalibrator, actually I would find reasonable to put it there, where the dl1 level is calculated. Is there an argument against this?

vuillaut:
> Hi @vuillaut @rlopezcoto,
> 
> indeed, for the moment there is no gain selection in the LSTCameraCalibrator, actually I would find reasonable to put it there, where the dl1 level is calculated. Is there an argument against this?

Ok so this is different from the current MC schema where gain selection is done during calibration/integration. 
That's why I couldn't easily change the gain selection for MC but could for the real data.
@FrancaCassol Yes why not moving it in the calibration but then we have to remember this difference with MC (open an issue?).

rlopezcoto:
Ok, so since we do not want to merge this PR in the end, I suggest to close it and leave #197 open to remember that the gain selection is not applied properly in MC and needs to be changed with the next ctapipe release?
Also @FrancaCassol could open an issue after implementing gain selection in the `LSTCameraCalibrator ` to point out the difference between data and MC

FrancaCassol:
Hi @rlopezcoto,
> Ok, so since we do not want to merge this PR in the end, I suggest to close it and leave #197 open to remember that the gain selection is not applied properly in MC and needs to be changed with the next ctapipe release?

Could we not already create our LSTMCalibrator that makes add the gain selection also for the MC data?


> Also @FrancaCassol could open an issue after implementing gain selection in the `LSTCameraCalibrator ` to point out the difference between data and MC

Yes,  I try to finalise all this soon

vuillaut:
> Could we not already create our LSTMCalibrator that makes add the gain selection also for the MC data?

Do we really want to re-implement something existing and working in ctapipe?
The "20pe" fix should be good enough for the time being for MC and does not require any further development.

rlopezcoto:
Hi, I agree with @vuillaut that it is better to try to be compatible with ctapipe and use the functionalities existing there if possible

FrancaCassol:
Hi @rlopezcoto, @vuillaut,

here the idea is just to specialise (by inheritance) one ctapipe class to the need of LST (I did the same the class LSTCameraCalibrator which inherits from the ctapipe CameraCalibrator), we will remain perfectly compatible with ctapipe.

The import point here is that the gain selection in LST real data will be always performed on ADC values (am I correct?), so in order to be coherent we must have the same selection also for MC events . Don't forget that ctapipe has been create for MC data, so not always realistic for the first analysis steps. I can easily provide the calibrator, I think we should try to harmonise more data and MC.

vuillaut:
I am sorry I fail to see the point here.
Yes for real data there are specificities and a `LSTCameraCalibrator` was of course necessary.
However, LST MC should not be specific (and are not) so `CameraCalibrator` is sufficient.
Bug apart (fixed in next ctapipe release), the selection is done on ADC values yes, MC or real data.

FrancaCassol:
> Bug apart (fixed in next ctapipe release), the selection is done on ADC values yes, MC or real data.

Indeed the last version of ctapipe seems to solve completely the problem, do you know when we will have a v0.7.1 release?

rlopezcoto:
> Discussing with @FrancaCassol #241 and how to perform the gain selection for real data, we realized that the threshold from the [config file](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/data/lstchain_standard_config.json):
> 
> https://github.com/cta-observatory/cta-lstchain/blob/6a7a81dd2cc5ad419ef8e76935e467f15da476b3/lstchain/data/lstchain_standard_config.json#L104
> 
> is not applied to real data. The threshold to real data is actually applied by the `combine_channels` (deprecated?) function:
> 
> https://github.com/cta-observatory/cta-lstchain/blob/6a7a81dd2cc5ad419ef8e76935e467f15da476b3/lstchain/reco/dl0_to_dl1.py#L269
> 
> and the threshold properly applied at the waveform level. Did we get anything wrong @vuillaut?

According to #252, if we were applying the `gain_selection` using `combine_channels` and the 4095 value was hardcoded, it means that we were never using the low gain for the real data, right @vuillaut?

PR #209, opened 2019-11-09
Improve create pedestal script

pawel21:
Hi,
I have made few improvement in code to create pedestal file

1. Add optional argument to set start sample for waveform, which is using to create pedestal table. Default value is 11. 
The 2 first samples are more noise and the first 9 samples have occasionally increased signal due to Tsutomu pattern, hence we should skip first 11 samples. 
Average signal should be around zero after the low level corrections with new pedestal file  (so far is ~ 5/10 below 0).

2. Set constant value (1855) of number of pixels `n_pixels` to initialize pedestal array. (This solve problem when some modules are missing)


FrancaCassol:
Hi @pawel21, 

`self.n_pixels = 265*n_channel`

could you also assign the 265 to a constant variable on top of the module?

I have a question, if one module is missing is the first_capacitor array shorter? (I could check but right now I don't remember which run was with a missing module)

pawel21:
Hi @FrancaCassol ,

> could you also assign the 265 to a constant variable on top of the module?

Yes, I will do it.

> I have a question, if one module is missing is the first_capacitor array shorter? (I could check but right now I don't remember which run was with a missing module)

Yes, if one (or more) module is missing, the first_capacitor_array is shorter (this allow iterating over working modules)
Run 1408 is with missing module

pawel21:
Dear @rlopezcoto, @FrancaCassol 
Could you look on this PR?

lstchain/calib/camera/drs4.py
```diff
@@ -13,23 +16,32 @@
 roi_size = 40
 high_gain = 0
 low_gain = 1
+N_module_in_camera = 265
```
rlopezcoto: right now this number is hardcoded, if we have observations with less modules how can you change it?
For the naming:
N_module_in_camera -> n_modules_in_camera

scripts/lstchain_data_create_pedestal_file.py
```diff
@@ -15,24 +15,35 @@
...
 
 parser = argparse.ArgumentParser()
 
 # Required arguments
-parser.add_argument("--input_file", help="Path to fitz.fz file to create pedestal file.",
-                    type=str, default="")
+parser.add_argument("--input_file",
+                    help="Path to fitz.fz file to create pedestal file.",
+                    type=str,
+                    default="")
```
rlopezcoto: do you need a default here?

scripts/lstchain_data_create_pedestal_file.py
```diff
@@ -15,24 +15,35 @@
...
-                    type=int, default=5000)
-
-parser.add_argument('--deltaT', '-s', type=lambda x: bool(strtobool(x)),
+parser.add_argument("--max_events",
+                    help="Maximum numbers of events to read. Default = 5000",
+                    type=int,
+                    default=5000)
+
+parser.add_argument("--start_r0_waveform",
+                    help="Start sample for waveform. Default = 10",
```
rlopezcoto: default in the variable definition and the description are not coincident

lstchain/calib/camera/drs4.py
pawel21: This variable is used to create array with pedestal values. (I need it to avoid run out of range by numba function, when module is missing) To iterate over module code uses `self.n_module` read from data file.
Naming: I have changed.

scripts/lstchain_data_create_pedestal_file.py
pawel21: No, removed

scripts/lstchain_data_create_pedestal_file.py
pawel21: Yes, I have fixed

pawel21:
I have changed default ,,maximum numbers of events to read" from 5000 to 8000

FrancaCassol:
This code seems ok, could we merge it?

PR #210, opened 2019-11-13
Framework for the search for muons on real data

rlopezcoto:
In this PR there is the script for muon real analysis following some of the discussions in #196 and the suggestions by @vuillaut.
- It contains everything related to the muon analysis, the stored quantities needed for the analysis and a tag for good or bad rings to be used in the analysis.
- It also contains an additional quantity that is the size outside of the ring, suggested by @moralejo to find muons with the image of a hadron together.
- Right now a non-optimized muon tagging is used, but @pillera will work on the first of the bulletpoints discussed [here](https://github.com/cta-observatory/cta-lstchain/pull/196#issuecomment-550236451) to have some functions for the muon tagging.

Some of the functions of this PR are copied from [the muon analysis of ctapipe](https://github.com/cta-observatory/ctapipe/tree/master/ctapipe/image/muon). The code there is under discussion and [will be refactored](https://github.com/cta-observatory/ctapipe/pull/1154#issuecomment-548713583) at some point, but we need this now, so we will work on merging it with the general code later.

### DIFF
--- a/scripts/lstchain_data_muon_analysis_dl1.py
+++ b/scripts/lstchain_data_muon_analysis_dl1.py
@@ -1,0 +1,132 @@
+import argparse
+import sys, os
+import numpy as np
+from ctapipe.image.muon.features import ring_containment 
+from ctapipe.image.muon.features import ring_completeness
+from ctapipe.image.muon.features import npix_above_threshold
+from ctapipe.image.muon.features import npix_composing_ring
+from ctapipe.image.muon.muon_integrator import MuonLineIntegrate
+from ctapipe.image.cleaning import tailcuts_clean
+from ctapipe.instrument import CameraGeometry
+from ctapipe.io.hdf5tableio import HDF5TableReader
+from astropy import units as u
+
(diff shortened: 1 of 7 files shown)


scripts/lstchain_data_muon_analysis_dl1.py
```diff
@@ -0,0 +1,132 @@
...
+
+python lstchain_data_muon_analysis_dl1.py 
+--input_file dl1_Run01566_0322.h5 
+--output_file Data_table.fits 
+'''
+
+parser = argparse.ArgumentParser()
+
+# Required arguments
+parser.add_argument("--input_file", help = "Path to fits.fz data file.",
```
morcuended: Make clear in the help that this script takes dl1(a) .h5 files as input. The rest seems fine to me.

morcuended:
Why is travis check failing?

rlopezcoto:
Thanks for having a look @morcuended! I already implemented your changes

PR #211, opened 2019-11-15
Calibration onsite scripts

FrancaCassol:
Hi @vuillaut, 

here a first version of the onsite calibration scripts for creating the drs4 calibration file and the flat-field calibration file in the onsite directory tree. 

In the case of the flat-field calibration, a pdf file with some plots is also created (in the log directory). I would like have something similar for the drs4 calibration.

You can test everything using run 1571 for the pedestals and run 1572 for the calibrations.
This is work on going, I will probably add some improvements in the next, but I would like to have your opinion with this preliminary version

### DIFF
--- a/lstchain/visualization/plot_calib.py
+++ b/lstchain/visualization/plot_calib.py
@@ -1,0 +1,339 @@
+
+from matplotlib import pyplot as plt
+from ctapipe.visualization import CameraDisplay
+import numpy as np
+from matplotlib.backends.backend_pdf import PdfPages
+from ctapipe.instrument import  CameraGeometry
+
+# read back the monitoring containers written with the tool calc_camera_calibration.py
+from ctapipe.io.containers import FlatFieldContainer, WaveformCalibrationContainer, PedestalContainer, \
+    PixelStatusContainer
+
+from ctapipe.io.hdf5tableio import HDF5TableReader
+
+__all__ = ['read_file',
+           'plot_all'
+           ]
+
+ff_data = FlatFieldContainer()
+ped_data = PedestalContainer()
+calib_data = WaveformCalibrationContainer()
+status_data = PixelStatusContainer()
+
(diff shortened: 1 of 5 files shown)


pawel21:
Dear @FrancaCassol ,

Where you call r0 calibration?
I see one potential problem: when we have data after pedestal subtraction online (by EVB), we have to call only `time_lapse_corr` and `interpolate_spikes`

FrancaCassol:
Hi @Pawel,
as soon as your PR will be submitting I will call the script that you think it is correct correct.
Could you include it in your PR?

pawel21:
> Could you include it in your PR?
How can I do it?

FrancaCassol:
Dear @pawel21, 
now that I think it again, I understood that the chain of corrections remains the same, only the off-line baseline subtraction will be much less important if already a rough subtraction is done at EvB level. @moralejo could you please confirm?

pawel21:
I think we can call low corrections on the data after pedestal subtraction online in the two following ways:
1)

```
reader = LSTEventSource(input_url="LST-1.1.Run01578.0000.fits.fz", max_events=1000)
lst_r0 = LSTR0Corrections(tel_id=1)
for event in reader:
   event.r1.tel[1].waveform = event.r0.tel[1].waveform
   lst_r0.time_lapse_corr(event, 1)
   lst_r0.interpolate_spikes(event, 1)
```
2)

```
reader = LSTEventSource(input_url="LST-1.1.Run01578.0000.fits.fz", max_events=1000)
lst_r0 = LSTR0Corrections(tel_id=1)
for event in reader:
   lst_r0.calibrate(event)
```

FrancaCassol:
yes, I agree, for the moment I keep the second one (that is what we used till now).
To be verified if it brings something more than the first one

As agreed at the analysis meeting, I close this PR and open a new one to the master

Issue #212, opened 2019-11-15
`torch` call from gammalearn breaking all builds


rlopezcoto:
?
very strange, nothing has changed... @mikael10j, @vuillaut, any idea? 
Otherwise we will need to comment it out since it is not needed for the working of the repo.

vuillaut:
Funny (or not), it seems `pytest` behaviour has changed between v5.1 and v5.2 and is now finding modules that are not in the package... see #213

rlopezcoto:
Perfect, thanks!

rlopezcoto:
Solved in #214

PR #213, opened 2019-11-15
testing with pytest=5.1

vuillaut:
Test related to issue #212


vuillaut:
Ok this was to test pytest liability but #214 is a better solution I think.

PR #216, opened 2019-11-18
Improved condition in time lapse corr

pawel21:
Hi,

After discuss with @SeiyaNozaki, I realized that conditions in 'time_lapse_corr' should be a bit different.
So, I have changed conditions as I understood.

Dear @SeiyaNozaki,
Could you check if this version is correct?

Dear @rlopezcoto ,
I am sorry for mess, but now this is only small change in code in 3 places

### DIFF
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -493,11 +493,11 @@ def do_time_lapse_corr(waveform, expected_pixel_id, local_clock_list,
                 # code from Takayuki & Julian
                 if pix % 2 == 0:
                     first_cap = fc[nr_module, gain, pix]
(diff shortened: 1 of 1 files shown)


SeiyaNozaki:
Thank you @pawel21 for quick modification!
It looks fine for me!

rlopezcoto:
No problem @pawel21, thanks a lot!
It would anyways be great if in the near future all these "magic" numbers can be given a sense, either defining them as variables or adding comments.

pawel21:
> It would anyways be great if in the near future all these "magic" numbers can be given a sense, either defining them as variables or adding comments.

Exactly :) 
Let merge now and next step will be add comments and do code refactoring.

PR #217, opened 2019-11-18
Muon tagging

pillera:
I added an image size based muon tagging function in the muon module (muon_analysis.py).
It is called in the script lstchain_data_muon_analysis.py with the default values, i.e. no tagging.

### DIFF
--- a/lstchain/image/muon/muon_analysis.py
+++ b/lstchain/image/muon/muon_analysis.py
@@ -18,7 +18,7 @@ __all__ = ['get_muon_center',
            'fit_muon',
            'analyze_muon_event',
            'tag_muon',
-           'tag_muon_pix',
+           'tag_pix_thr',
            ]
 
 def get_muon_center(geom, equivalent_focal_length):
@@ -274,9 +274,9 @@ def tag_muon(image,thr_low=0,thr_up=1.e10):
     """
     return image.sum() > thr_low and image.sum() < thr_up
 
-def tag_muon_pix(image,thr_low=50,thr_up=300):
+def tag_pix_thr(image,thr_low=50,thr_up=300,pe_thr=10):
     """
-    Tag muon with a double threshold on the number of pixels above 10 photoelectrons 
+    Tag event with a double threshold on the number of pixels above 10 photoelectrons 
     Default values apply elimination of pedestal and calibration events
 
     Paramenters
@@ -284,11 +284,12 @@ def tag_muon_pix(image,thr_low=50,thr_up=300):
     image:      `np.ndarray` number of photoelectrons in each pixel
     thr_low: `int` lower threshold for number of pixel > 10 pe  
     thr_up: `int` upper threshold for number of pixel > 10 pe
+    pe_thr: 'float' minimum number of photoelectrons for a pixel to be counted
 
     Returns
     ---------
-    `bool` it determines whether a muon was tagged or not
+    `bool` it determines whether a the event is in the given nr of pixel range
 
     """
 
-    return ((np.size(image[0][image[0]>10.]) < thr_up) and (np.size(image[0][image[0]>10.]) > thr_low))
+    return ((np.size(image[0][image[0]>pe_thr]) < thr_up) and (np.size(image[0][image[0]>pe_thr]) > thr_low))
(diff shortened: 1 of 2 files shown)


pillera:
@rlopezcoto 
> Thanks @pillera for the PR!
> As I understand it, the tag_muon should substitute l.153 of `lstchain_data_muon_analysis.py`, right?
yes, the cuts at line 153 should be redundant if tag_muon is used.  I'll substitute those lines.

> Also, it would be great to have the same tag but in the number of pixels (at the moment, that cut is applied for < 300 pixels and >50 with more than 10 phe)
I will add a second function for this.

lstchain/image/muon/muon_analysis.py
```diff
@@ -271,4 +272,23 @@ def tag_muon(image,thr_low=0,thr_up=1.e10):
...
+    thr_low: `int` lower threshold for number of pixel > 10 pe  
+    thr_up: `int` upper threshold for number of pixel > 10 pe
+
+    Returns
+    ---------
+    `bool` it determines whether a muon was tagged or not
+
+    """
+
+    return ((np.size(image[0][image[0]>10.]) < thr_up) and (np.size(image[0][image[0]>10.]) > thr_low))
```
rlopezcoto: I would suggest to make the phe_threshold per pixel (10 in this case) another variable to be set on the function

pillera: I would use the tag_muon instead of the cut in size, since from a quick analysis the tagging using the nr of pe, is not so performing as the one in size.

scripts/lstchain_data_muon_analysis.py
```diff
@@ -150,9 +150,13 @@
 
         print("Event {}. Number of pixels above 10 phe: {}".format(event_id,
                                                                   np.size(image[0][image[0] > 10.])))
-        if((np.size(image[0][image[0]>10.]) > 300) or (np.size(image[0][image[0]>10.]) < 50)):
-            continue
+        
+        #if not tag_pix_thr(image): #default skipps pedestal and calibration events
```
rlopezcoto: at the moments no events are tagged with these lines commented out

pillera: That was because with default values no tagging is applied. I put the thresholds I found in the analysis presented at the INFN MC meeting, but they should be updated once the calibration is alright. 
Regarding your request to include a tagging with the nr of pixels above a given pe thr, it is not so performant as the previous one, so I added a more generic function, (not specifically for muon tagging), that can be used to substitute these lines
if((np.size(image[0][image[0]>10.]) > 300) or (np.size(image[0][image[0]>10.]) < 50)):
            continue
Since using the image size-based tagging makes those lines redundant I commented them out.

pillera: Now The script should analyze only events passing the image-size double thr conditions according to the values I found on run 442.

rlopezcoto: I'm not sure what do you mean by more performant, at the moment we would like to store images of events that may be coming from muons (or not), without applying any cuts assuming their size, width, radius, etc...
The filter that was applied (actually, `muon_filter` or something like that would probably be more appropriate naming) was assuring to get some pixels above threshold (>50), but getting rid of calibration events (<300, actually I think that this number could also be increased). For those images we would be performing a fit and storing values in the final table. I also added a tag that was evaluating whether the muon is good or not under standard analysis conditions, but this will not be the case until the telescope is in a good shape. Until then, we need a much wider range of events to pass the cuts, and therefore the cut in `size` should be introduced but not used until a later stage.

pillera: okay, thanks for the clarifications. So I will do the following changes:

1. change tag_muon -> muon_filter
2. leave default values i.e. no filtering is applied
3. uncomment lines skipping calibration and pedestal events leaving default threshold (50 and 300)

pillera: > I'm not sure what do you mean by more performant, at the moment we would like to store images of events that may be coming from muons (or not), without applying any cuts assuming their size, width, radius, etc...
a selection based on what you suggested earlier, would eliminate much more good muon candidates with respect to the selection in size

> The filter that was applied (actually, `muon_filter` or something like that would probably be more appropriate naming) was assuring to get some pixels above threshold (>50), but getting rid of calibration events (<300, actually I think that this number could also be increased). For those images we would be performing a fit and storing values in the final table. I also added a tag that was evaluating whether the muon is good or not under standard analysis conditions, but this will not be the case until the telescope is in a good shape. Until then, we need a much wider range of events to pass the cuts, and therefore the cut in `size` should be introduced but not used until a later stage.

PR #218, opened 2019-11-18
Script for processing real data r0 to dl1a+b

morcuended:
As discussed in today's call, I commit the code to process real data from raw r0 files up to dl1. **This is just a temporary solution** until the script of @vuillaut is ready and tested, which should be definitively the way to go. I did not have the time to work directly on that script during the past week. 

The code has been made out of the scripts to process MC files and the @FrancaCassol notebooks to apply the calibration coefficients. In the end, it became a Frankenstein script. Sorry for the very long PR in advance.

- **It cannot handle MC data.**
- It takes a calibration coefficient and pedestal files to perform the low and high-level calibration.
- Several functions as "get_dl1" from dl0_to_dl1.py script are redefined inside and modified to handle just real data.
- It creates two separated h5 files corresponding to dl1a & dl1b.
- It is possible two add pointing information from Drive reports if requested. This slows down quite a bit the performance of the script since it has to look for the precise timestamp in the (sometimes) huge drive reports. Besides that, there is no date convention for the name of the file so one has to check first which drive log should use. One drive log can contain several observation dates.
- It contains timing information from the ucts timestamps or alternatively from the TIB if the first is not available.
- Since TIB problems could arise and trigger type tags could be loose, the script processes all the events regardless of its trigger type.
- I will test it with the current master of lstchain.


rlopezcoto:
Thank you very much @morcuended for the PR. As discussed through e-mail, we will put this on hold and merge it only in case the script for real data analysis from #201 cannot be tested using real data by the onsite team. I would also suggest to rename the script: 
scripts/lstchain_data_r0_t0_dl1.py  -> scripts/lstchain_data_r0_t0_dl1_UCM.py

and add at the beginning of the script some docstring in which it is explained that this script will be deprecated whenever the above mentioned scripts are merged.

vuillaut:
Hi.
Sorry being on shift, I could not join yesterday's call.
My idea was to have merged contributions in a single branch (see branch production_scripts) or PR to benefit from everyone's work and merge a working stable version in master at once.
I am a bit busy on-site but I'll review this asap and see how we can proceed to merge works and have a single set of scripts.
Is that ok with you?

morcuended:
Hi @vuillaut,
Yes, that's totally fine with me. Right now I'm testing and reviewing your scripts to help to do their merge faster. I will include the timing information either from the UCTS timestamp or from the TIB counters. Afterward, I'll implement the possibility of adding the pointing information event-wise as we agreed a few weeks ago.

vuillaut:
Ok I could go through the code.
I have general comments rather than going through a detailed review:
- I don't think we want to go this way, duplicating a lot of existing code
(it makes also a difficult task to spot the changes/addition but I think I could spot them)
- The main and very important addition is the pointing information

I would propose to go forward by adding these changes to #201 :
- in `r0_to_dl1`, pass pointing filename as optional argument and:
```
if not is_simu:
    pointing_file = pointing_file_path
    PointingPosition()
    pointings.drive_path = (pointing_file)
    ucts_timestamp = event.lst.tel[telescope_id].evt.ucts_timestamp
    [...]
    azimuth, altitude = pointings.cal_pointingposition(utc_time.unix)
    event.pointing[telescope_id].azimuth = azimuth
    event.pointing[telescope_id].altitude = altitude
    dl1_container.alt_tel = altitude
    dl1_container.az_tel = azimuth
```

I haven't added all changes but you get the idea.     
Note the last line, I added in PR #201 alt_tel and az_tel and use them in dl1_to_dl2 if `mc_alt_tel` does not exist (then it means these are real data).

Then we update script `lstchain_data_r0_to_dl1` to pass pointing as you did there.

If you would prefer to go by step, we can review and merge #201 first that is working independently.

morcuended:
> Ok I could go through the code.
> I have general comments rather than going through a detailed review:
> 
> * I don't think we want to go this way, duplicating a lot of existing code
>   (it makes also a difficult task to spot the changes/addition but I think I could spot them)

I totally agree, I never intended to commit this script. It was just a temporal one to get already dl1 data to prototype the onsite analysis
> * The main and very important addition is the pointing information

Yes, basically get the time information and interpolate the pointing position
> 
> I would propose to go forward by adding these changes to #201 :
> 
> * in `r0_to_dl1`, pass pointing filename as optional argument and:
> 
> ```
> if not is_simu:
>     pointing_file = pointing_file_path
>     PointingPosition()
>     pointings.drive_path = (pointing_file)
>     ucts_timestamp = event.lst.tel[telescope_id].evt.ucts_timestamp
>     [...]
>     azimuth, altitude = pointings.cal_pointingposition(utc_time.unix)
>     event.pointing[telescope_id].azimuth = azimuth
>     event.pointing[telescope_id].altitude = altitude
>     dl1_container.alt_tel = altitude
>     dl1_container.az_tel = azimuth
> ```
> 
> I haven't added all changes but you get the idea.
> Note the last line, I added in PR #201 alt_tel and az_tel and use them in dl1_to_dl2 if `mc_alt_tel` does not exist (then it means these are real data).
> 
> Then we update script `lstchain_data_r0_to_dl1` to pass pointing as you did there.
> 
> If you would prefer to go by step, we can review and merge #201 first that is working independently.

I think it'd be better to have #201 merged and later on add these changes in a new PR

vuillaut:
> I think it'd be better to have #201 merged and later on add these changes in a new PR

Great, thanks Daniel.

rlopezcoto:
After #222, this one can be closed

PR #219, opened 2019-11-18
Onsite scripts for creating the drs4 pedestal file and the calibration file

FrancaCassol:
These onsite scripts can be used as command line (after installation of lstchain: pip install -e .) and are based on the offline scripts lstchain_data_create_pedestal_file.py and calc_camera_calibration.py

They suppose to work in a directory tree as defined in https://forge.in2p3.fr/issues/37938#note-5
(which contains a link to the real data). The root directory (base_dir) must be set correctly.
They create also some pdf files with cross-check plots .

They run simply giving the input run numbers:
```
onsite_create_drs4_pedestal_file -r 1571
onsite_create_calibration_file -r 1572 -p 1571 
```

To get help, use:
```
onsite_create_drs4_pedestal_file -h 
onsite_create_calibration_file -h  
```
Warning: I could test the whole chain only with runs taken with the old firmware. Hopefully a test with the new firmware will be possible soon.


rlopezcoto:
Thanks @FrancaCassol for the PR. As agreed yesterday,could you @pawel21 have a look, please? I'll also revise it as soon as possible

scripts/onsite/onsite_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,98 @@
+#!/usr//bin/env python
+
+"""
+ F. Cassol, 12/11/2019
```
rlopezcoto: same as above

setup.py
```diff
@@ -10,6 +10,14 @@ def list_scripts():
     script_list = [f'{os.path.join(script_dir, f)}' for f in os.listdir(script_dir) if f.startswith('lstchain_')]
     return script_list
 
+entry_points = {}
+entry_points['console_scripts'] = [
+    'calc_camera_calibration = lstchain.tools.calc_camera_calibration:main',
+    'onsite_create_drs4_pedestal_file = scripts.onsite.onsite_create_drs4_pedestal_file:main',
+    'onsite_create_calibration_file = scripts.onsite.onsite_create_calibration_file:main',
+    'lstchain_data_create_pedestal_file = scripts.lstchain_data_create_pedestal_file:main'
```
rlopezcoto: the scripts starting with "lstchain_" should be covered by l.10 of this file, right?

scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -0,0 +1,120 @@
+#!/usr//bin/env python
+
+"""
```
rlopezcoto: We do not usually do not include in the scripts dates/people names, we can keep track of these things through github

lstchain/visualization/plot_drs4.py
```diff
@@ -0,0 +1,88 @@
+
+from matplotlib import pyplot as plt
+
+import numpy as np
+from matplotlib.backends.backend_pdf import PdfPages
+from ctapipe.io import event_source
+from lstchain.calib.camera.r0 import LSTR0Corrections
+
+
+__all__ = ['plot_drs4',
```
rlopezcoto: should be 'plot_waveforms'

lstchain/visualization/plot_drs4.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # r0 calibrator
+    r0_calib = LSTR0Corrections(pedestal_path=pedestal_file, offset=offset_value,
+                                r1_sample_start=2, r1_sample_end=38, tel_id=tel_id )
+
+    # event_reader
+    reader = event_source(data_file, max_events=8)
+
+    pix = 0
+    pad = 420
```
pawel21: What is `pad`?

pawel21:
For me look very good.
I think important is tested this script in La Palma using data with new firmware.

setup.py
```diff
@@ -10,6 +10,14 @@ def list_scripts():
...
+]
+
```
vuillaut: `entry_points` are nice to transform functions into scripts but  I would suggest to not create an additional way of adding scripts (that requires adding each of them manually as entry point).
If you modify `list_scripts` this way, they will be automatically added:

```
def list_scripts():
    script_list = []
    script_dir = 'scripts'
    script_list.extend([f'{os.path.join(script_dir, f)}' for f in os.listdir(script_dir) if f.startswith('lstchain_')])
    script_dir = 'scripts/onsite'
    script_list.extend([f'{os.path.join(script_dir, f)}' for f in os.listdir(script_dir) if f.startswith('onsite')])
    return script_list
```

e.g. `lstchain_data_create_pedestal_file` is already a command line

lstchain/visualization/plot_drs4.py
```diff
@@ -0,0 +1,88 @@
...
+            plt.plot([0, 40], [offset_value, offset_value], 'k--',  label="offset")
+            plt.xlabel("time sample [ns]")
+            plt.ylabel("counts [ADC]")
+
+            plt.legend()
+            plt.ylim([-50, 500])
+
+        if plot_file != "none" and pad == 428:
+            pad = 420
+            pp.savefig()
```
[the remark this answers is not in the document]
FrancaCassol: ok

setup.py
FrancaCassol: Hi @vuillaut,
the point is that I want to have the scripts as command line, so I don't what to see the .py at the end of the name (I think this is confusing for the user/shifter). Could we keep at least the onsite scripts as real command lines?

lstchain/visualization/plot_drs4.py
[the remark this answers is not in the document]
FrancaCassol: Hi @pawel, in the new version I added the plot of the pedestal charges and std

lstchain/visualization/plot_drs4.py
FrancaCassol: pad is the pad number where I want to locate the plot

lstchain/visualization/plot_drs4.py
FrancaCassol: yes, actually I just modified this function adding the plot of other quantities (following the suggestions of Pawel and Julian)

scripts/onsite/onsite_create_calibration_file.py
FrancaCassol: yes, I took it out

setup.py
FrancaCassol: Hi @rlopezcoto, I don't understand the question. But, indeed one can take out from here the lstchain scripts and keep only the onsite as command line

scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -0,0 +1,120 @@
...
+                      type=int, default=0)
+optional.add_argument('-s', '--statistics', help="Number of events for the flat-field and pedestal statistics",
+                      type=int, default=10000)
+
+args = parser.parse_args()
+run = args.run_number
+ped_run = args.pedestal_run
+prod_id = 'v%02d'%args.version
+stat_events = args.statistics
+max_events = 1000000
```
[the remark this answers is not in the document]
FrancaCassol: I have problem to pass None, anyway onsite this number is not supposed to change, when I have time I try to understand the problem with the None

setup.py
rlopezcoto: I meant that they were included already in the script list, but I saw from one of your previous comments that your intention was actually to make them "runnable" by the users without including the `.py`

FrancaCassol:
@ruben, I didn't test the code in La Palma, I understood you wanted to install everything...
As said, I works in Marseille with the old firmware, I could not test with the new firmware, because I do not have calibration data.

FrancaCassol:
Hi @vuillaut, 
which change are you asking?

setup.py
vuillaut: Do you think users get this easily confused? 😄 
Either way, I don't mind, but I'd very much prefer to have a single implementation - that would require a bit of refactoring for the other scripts (that will need a main, and then be called as entry points) - do you think you can take care of it?
Also, I think you can automatize a bit to automatically add scripts following the schema `lstchain_***`and `onsite_***`

vuillaut: Btw. The current scripts **are** command lines. The command name indeed ends with `.py`

setup.py
vuillaut: They are already runnable, installed on your system when you install lstchain, and you don't need to call `python ***`. But I see that the `.py` naming really is confusing 😅

lstchain/visualization/plot_calib.py
```diff
@@ -0,0 +1,339 @@
...
+        next(h5_table.read(table, calib_data))
+        table = f"/tel_{tel_id}/pedestal"
+        next(h5_table.read(table, ped_data))
+        table = f"/tel_{tel_id}/pixel_status"
+        next(h5_table.read(table, status_data))
+
+    h5_table.close()
+
+
+def plot_all(ped_data, ff_data, calib_data, run=0, plot_file="none"):
```
vuillaut: why not `plot_file=None` if you don't want to save the figure?

scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -0,0 +1,120 @@
...
+                os.remove(output_file)
+            else:
+                exit(1)
+
+        print(f"\n--> Log file {log_file}")
+
+
+
+        # define config file
+        config_file = os.path.join(os.path.dirname(__file__), "../../lstchain/data/onsite_camera_calibration_param.json")
```
vuillaut: we need to find a better way to find default configs but that will do for now.

setup.py
FrancaCassol: > Do you think users get this easily confused? 😄

yes, definitely ;-)

> Either way, I don't mind, but I'd very much prefer to have a single implementation - that would require a bit of refactoring for the other scripts (that will need a main, and then be called as entry points) - do you think you can take care of it?

Ok, I can, I think entry_points  is also the way ctapipe is doing, so let's be coherent with it

> Also, I think you can automatize a bit to automatically add scripts following the schema `lstchain_***`and `onsite_***`

Ok, I  try. 

I have some questions:
-> If we agree to put the config files in the "data" directory, should we change its name to "configurations" or similar?
-> My scrips are tools, I will changed their name with lstchain*, but I will keep them the tools directory, ok?
-> Should we approve this PR and then I do a new one with all these agreed changes ?
  (not to mix too many things)

scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -0,0 +1,120 @@
...
+max_events = 1000000
+
+
+def main():
+
+    print(f"\n--> Start calculating calibration from run {run}")
+
+    try:
+        # verify input file
+        file_list=sorted(Path(f"{base_dir}/R0").rglob(f'*{run}.0000*'))
```
vuillaut: What about run `LST-1.2.Run01541.0040.fits.fz` for example ?

scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
vuillaut: https://stackoverflow.com/questions/48295246/how-to-pass-none-keyword-as-command-line-argument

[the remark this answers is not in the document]
vuillaut: wait, don't you want to have `max_events = stat_events`?
I don't understand the difference here since you state in doc "Number of events for the flat-field and pedestal statistics"

setup.py
vuillaut: Ok let's go for entry points then 👍 

Yes sure keep the tools they are great! I actually think we should only make tools.
If you don't want to do that in this PR, please open issues or add actions in the board so we keep track of it.

scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
FrancaCassol: max_event is the number of event to be read, stat_events are the one used for estimating the pedestals and flat-field parameters (mean, std, etc...). I can have then have more calibration events each corresponding to a stat_event statistics.

setup.py
FrancaCassol: I know @vuillaut, but believe me, it is better not to have the .py

vuillaut: No no it's ok I agree (see below) 👍

setup.py
FrancaCassol: It depends on how long this PR takes to be approved ;-)
If it is fast I will do the changes in a new one, otherwise I will everything in this one.

scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
vuillaut: So in any case you want `max_events > stat_events` right?

setup.py
vuillaut: It's ok let's divide the work in smaller bits, then more people can contribute and review is easier :)

FrancaCassol: Perfect, in this case, are there other changes to be done for this PR to be approved? (I will change the 100000 in the next PR, I think the problem is more complicated, in any case no sub-run will have more the 100000 events, so result is equivalent to None)

scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
vuillaut: I agree with @rlopezcoto, if 1000000 is just an arbitrarily large number, use None as default (it's not an input parameter anyway).

vuillaut:
> Hi @vuillaut,
> which change are you asking?

@FrancaCassol I have 2 more comments to be addressed.
And as usual, some unit tests would be awesome 😅

rlopezcoto:
@FrancaCassol, thanks! I'm happy with the answers, although I do not like the 100000000000000 :D, it is fine if it is modified afterwards. Regarding testing, I meant that it would be great if it can be tested that everything is alright and for that it is fine to use personal installations, for file production is for what I would like the common `lstanalyzer` installation to be used. 
Whenever you address @vuillaut comments, I'll merge the PR

scripts/onsite/onsite_create_calibration_file.py
FrancaCassol: The statistics of one subrun is  more than enough so we need to read only the 0000 run

scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
FrancaCassol: > So in any case you want `max_events > stat_events` right?

yes, but it is not a problem if max_events is less, the script will simply stop at the end of the file and collect a max_event statistics. Anyway, onsite the user is not supposed to change the default values. He has simply to give to run numbers

FrancaCassol:
Hi @vuillaut, 
it is not so easy to make tests for interactive scripts, could we keep it for a second phase?

scripts/onsite/onsite_create_calibration_file.py
vuillaut: Yes but then the script does not do the intended since user pass `--statistics` thinking it will use this value.
Could you add a warning if `max_events<stat_events` and if, at the end of the process, a number of events lower that `stat_events` have been used, please?

lstchain/visualization/plot_calib.py
vuillaut: I still think proper `None` would be better than a string but no big deal, as you prefer.

vuillaut:
> Hi @vuillaut,
> it is not so easy to make tests for interactive scripts, could we keep it for a second phase?

Ok, we'll need to get some updated test files also.

FrancaCassol:
Hi @vuillaut,
at the end I made the setup.py modifications in this PR, have a look if this is ok for you.

FrancaCassol:
For me it works, but I just tested the help of the scripts.

Wait, I just add the idea to set the base_dir as argument, so that @rlopezcoto do not need to touch  the scripts to run them in la Palma. I will make an other push soon

By the way, is the tree with the link to the data already available there?

lstchain/tools/pedestal_param.json
```diff
@@ -8,19 +8,19 @@
      "log_level":"DEBUG"
    },         
    "EventSource": {
-      "input_url": "/ctadata/franca/LST/LST-1.1.Run00472.0000.fits.fz",
+      "input_url": "/ctadata/franca/fefs/aswg/real/R0/20191107/LST-1.1.Run01579.0000.fits.fz",
```
vuillaut: change url please

FrancaCassol: This is an example, people must change it with their path... what you want I put?

vuillaut: There is no default in La Palma ?
If not it's fine.

vuillaut:
> By the way, is the tree with the link to the data already available there?

Yes.

rlopezcoto:
@FrancaCassol, I've seen that you have already updated the `base_dir`, thanks! There are still a couple of conflicts with the files, could you please work them out so we can go ahead and merge?

FrancaCassol:
Hi @rlopezcoto,
I can not see where the conflicts are, actually 

```
lstchain/tools/calc_camera_calibration.py
lstchain/tools/calc_pedestals.py
```

where renamed in this PR (so they are not more there) and for r0.py I just moved the "tel_id" trailet to the mother class...do you see the problem?

vuillaut:
> Hi @rlopezcoto,
> I can not see where the conflicts are, actually
> 
> ```
> lstchain/tools/calc_camera_calibration.py
> lstchain/tools/calc_pedestals.py
> ```
> 
> where renamed in this PR (so they are not more there) and for r0.py I just moved the "tel_id" trailet to the mother class...do you see the problem?

Hey @FrancaCassol - can you update and merge master to see conflicts on your side please?

rlopezcoto:
Thanks @FrancaCassol!

PR #223, opened 2019-11-21
Add pointing interpolated values

morcuended:
Here is the modification to add non-empty gps_time and interpolate the pointing values event-wise from the values contained in the Drive log files.

Something weird happened: I thought I had made a fresh installation of the lstchain repo yesterday, following the instructions written en the README. However, apparently the ctapipe-io-lst package was not pointing to the last version v0.2 as it is specified in environment.yml. So some names within the LSTDriveContainer were not properly set. I had to reinstall it manually to make everything work.

Disclaimer: When adding the pointing interpolated values, the code becomes substantially slower. Maybe this could be optimized. Perhaps it is due to the large file size of the Drive log.

Please have a look and let me know if you have any comment/suggestion

PR #224, opened 2019-11-21
Scripts fix

vuillaut:
Fixing tests in scripts.

I guess the cross merging of #222 and #219 messed things up and the tests were not properly done.
Anyway, the scripts are now part of the package, which solves the issue.
I added unit tests for some of them.


rlopezcoto:
Hi, thanks for moving the scripts @vuillaut, strange that this was now giving errors. I'll merge it to be able to merge #223 as well

PR #225, opened 2019-11-22
Update version

rlopezcoto:
Unfortunately, the version of the software was hardcoded and if you print it after 0.2.0 release, it will still print "0.1".
I copied the [version.py](https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/version.py) script from ctapipe (pointing back to the original one) and changed also the version.

### DIFF
--- a/lstchain/__init__.py
+++ b/lstchain/__init__.py
@@ -8,4 +8,5 @@ from . import image
 
 from .io import standard_config
 
-__version__ = 0.1
+from . import version
+__version__ = version.get_version(pep440=False)
--- a/lstchain/tests/test_lstchain.py
+++ b/lstchain/tests/test_lstchain.py
@@ -174,3 +174,10 @@ def test_polar_cartesian():
         for y in Y:
             p = cartesian_to_polar(x, y)
             np.testing.assert_almost_equal((x, y), polar_to_cartesian(*p))
+
+def test_version_not_unkown():
+    """
+    Test that lstchain.__version__ is not unkown
+    """
+    import lstchain
+    assert lstchain.__version__ is not 'unknown'
--- a/lstchain/version.py
+++ b/lstchain/version.py
@@ -1,0 +1,170 @@
+"""
+Get version identification from git.
+
+
+The update_release_version() function writes the current version to the
+VERSION file. This function should be called before packaging a release version.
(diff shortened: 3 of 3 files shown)


vuillaut:
Hi.
I am not sure to fully understand how this works 😅 
But on my laptop it gives version unkown.
```
In [1]: import lstchain
In [2]: lstchain.__version__
Out[2]: 'unknown'
```

rlopezcoto:
Hi,
mmm... strange, it is working in mine:
`In [1]: import lstchain`                                                                                                                                                            
`In [2]: lstchain.__version__`                                                                                                                                                       
`Out[2]: '0.2.0.post2+git3741767'`

vuillaut:
> Hi,
> mmm... strange, it is working in mine:
> `In [1]: import lstchain`
> `In [2]: lstchain.__version__`
> `Out[2]: '0.2.0.post2+git3741767'`

How do you install lstchain? `pip install .` ?

rlopezcoto:
`pip install -e .`

morcuended:
Hi @rlopezcoto, did you reinstall it in the container after fixing the version? Just checked it's still pointing to v 0.1

rlopezcoto:
Hi @morcuended, no, I realized we had the version hardcoded because I also printed the version after installing it in the IT container. 
In any case, I installed it after the release and everything should be ok (please let me know if there is any problem), it is just that the printed version is going to be 0.1

vuillaut:
Hi @rlopezcoto 
Could you add the following unit test in `test_lstchain.py` please:
```
def test_version_not_unkown():
    """
    Test that lstchain.__version__ is not unkown
    """
    import lstchain
    assert lstchain.__version__ is not 'unknown'
```

vuillaut:
Ok test is passing here.
I don't understand why but it seems that it is failing only for my install.... 🤔 
I'll approve.
Can you release 0.2.1 and reinstall please? 
lstchain version is copied in all metadata for provenance so that would be great to have the correct version.

rlopezcoto:
Thanks!
Just installed it in La Palma and it seems to work:
```
In [2]: lstchain.__version__                                                                                                                                                         
Out[2]: '0.2.1'
```

PR #226, opened 2019-11-23
Fix analysis

vuillaut:
Fixing some small errors in recent updates that I discovered running scripts tonight.


morcuended:
Hi @vuillaut, thanks for fixing it. Now I realize that with the hurry I did not test the script with recent data containing UCTS timestamps.

vuillaut:
> Hi @vuillaut, thanks for fixing it. Now I realize that with the hurry I did not test the script with recent data containing UCTS timestamps.

@morcuended I apologies, I suggested using astropy units because I thought the value would be used elsewhere later in the script and did not double-check in the hurry. It was good the first time.

PR #229, opened 2019-11-24
Get time info from TIB

morcuended:
Until the timing information from UCTS is well tested and trustable, we will use only TIB time information for the time being (see #227)

### DIFF
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -294,27 +294,18 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
                         dl1_container.fill_mc(event)
 
                     dl1_container.log_intensity = np.log10(dl1_container.intensity)
-                    # dl1_container.gps_time = event.trig.gps_time.value
-
-                    # GPS time is not available for the time being in the above field.
-                    # In the mean time, it is taken from UCTS timestamp (in unix time)
-                    # or alternatively from the TIB pps and 10 MHz counters. The first
-                    # one started to work around mid Nov 2019, so previous runs take
-                    # timing information from the TIB NTP counters. For the time being
-                    # we will stick with the TIB time since UCTS does not seem to be trustable.
-                    # This will be deprecated and modified back to the commented line
-                    # whenever GPS time is dumped to gps_time field.
+                    dl1_container.gps_time = event.trig.gps_time.value
 
                     if not is_simu:
-                        # ucts_timestamp = event.lst.tel[telescope_id].evt.ucts_timestamp
- 
-                        # if ucts_timestamp != 0:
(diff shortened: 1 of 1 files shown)


rlopezcoto:
thanks @morcuended! I'll wait for @FrancaCassol to have a look to be on the same page regarding [PR#15](https://github.com/cta-observatory/ctapipe_io_lst/pull/15) on ctapipe_io_lst. Then we need to 
* merge the codes 
* make a new ctapipe_io_lst release
* update ctapipe_io_lst version on cta-lstchain
* make a new cta-lstchain release
I can do that during the morning as soon as the codes are approved

lstchain/reco/dl0_to_dl1.py
```diff
@@ -306,15 +306,15 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
...
-                            utc_time = Time(datetime.utcfromtimestamp(ucts_timestamp / 1e9))
-                        else:
-                            start_ntp = event.lst.tel[telescope_id].svc.date
-                            ntp_time = start_ntp + event.r0.tel[telescope_id].trigger_time
-                            utc_time = Time(datetime.utcfromtimestamp(ntp_time))
+                        # if ucts_timestamp != 0:
+                        #     # UCTS timestamps are nanoseconds in UNIX timestamp format and must be given in seconds
+                        #     utc_time = Time(datetime.utcfromtimestamp(ucts_timestamp / 1e9))
+                        # else:
+                        start_ntp = event.lst.tel[telescope_id].svc.date
```
labsaha: So, now it avoids UCTS time and uses TIB time. This is what you were using before you incorporated UCTS time, right? If so, it seems ok to me.

morcuended: Yes, this is how I was first taking time info

labsaha: Ok, it is then already tested. It should be fine then. Let's then wait for what @FrancaCassol's comments

FrancaCassol:
After the merging of [`https://github.com/cta-observatory/ctapipe_io_lst/pull/15](url) one can simplify the code just using the R0/R1 event trigger_time which is now the TIB time (TAI time in s). 
So when we will move to UCTS time this part of code must not be updated

morcuended:
Yes @FrancaCassol , I will fix it

morcuended:
> After the merging of [`https://github.com/cta-observatory/ctapipe_io_lst/pull/15](url) one can simplify the code just using the R0/R1 event trigger_time which is now the TIB time (TAI time in s).
> So when we will move to UCTS time this part of code must not be updated

@FrancaCassol  @rlopezcoto I just modified the code to take directly the time info from the trigger_time field. Please have a look.

FrancaCassol:
Yes! That's fine

PR #230, opened 2019-11-25
update ctapipe-io-lst version

rlopezcoto:


rlopezcoto:
I'll merge this, since it is only the update of the ctapipe-io-lst version

Issue #231, opened 2019-11-25
Processing time dramatically increased 


vuillaut:
Hi.

The processing time of real data is much longer (by about 15) than the one of MC data.
It seems that all the time goes into `pointings.py`.
@morcuended can you have a look please?
I can send you the profiling file if you want.
![image](https://user-images.githubusercontent.com/4263646/69579319-2d893a80-0fca-11ea-8cf7-25de308de97b.png)

labsaha:
Hi @vuillaut,
Thanks for reporting. Yes, we have also seen that it is taking time. @morcuended told us about this. Now I checked the code and found misplacement of a line in the code. That is definitely a reason for that, although not sure if it is fully responsible unless we check it.

morcuended:
Hi @vuillaut, we just went through this. We were actually opening the drive report file for each event. We fixed it so that the file is opened only once and store it in a table to find later the proper coordinates that are used to interpolate. Here is a profiling plot with the changes we made @labsaha and me (just for 100 events).

Adding pointing
![profiling_withpointing](https://user-images.githubusercontent.com/32680689/69627628-4b4ab400-104b-11ea-82bf-5f8ccc4bae4c.png)

So here the pointing part now is that pink rectangle I'm pointing at.

Without pointing
![profiling_nopointing](https://user-images.githubusercontent.com/32680689/69627266-99ab8300-104a-11ea-82a8-d812e46caf51.png)

I will open a PR with the modifications

vuillaut:
Solved by #233

PR #233, opened 2019-11-26
Optimize pointing scripts

morcuended:
As @vuillaut reported in #231, the code related to the addition of pointing information to the data stream was terribly slowing down the performance. It was basically reading the drive report file for every event.

@labsaha and I modified the code in order for the drive file to be open only once at the beginning of the script and buffer it to be used later on.

@labsaha, @vuillaut and @rlopezcoto can you please take a look at the modifications?

### DIFF
--- a/lstchain/pointing/pointings.py
+++ b/lstchain/pointing/pointings.py
@@ -52,7 +52,7 @@ class PointingPosition(Component):
         else:
             raise Exception("No drive report file found")
 
-    def cal_pointingposition(self, ev_time):
+    def cal_pointingposition(self, ev_time, drive_data):
         """
         Calculating pointing positions by interpolation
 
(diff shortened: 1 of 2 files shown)


lstchain/pointing/pointings.py
```diff
@@ -65,10 +65,9 @@ def cal_pointingposition(self, ev_time):
             a container filled with drive information
         """
         drive_container = LSTDriveContainer()
-        data = self._read_drive_report()
-        drive_container.time = data['time'].data
-        drive_container.azimuth_avg = data['azimuth_avg'].data
-        drive_container.altitude_avg = data['altitude_avg'].data
+        drive_container.time = drive_data['time'].data
+        drive_container.azimuth_avg = drive_data['azimuth_avg'].data
+        drive_container.altitude_avg = drive_data['altitude_avg'].data
```
vuillaut: do you think it would be worth reading the other data as well?
It could be used later to check that the pointing was not crazy... What do you think?

vuillaut:
Looks much better :)
Thank you!

lstchain/pointing/pointings.py
labsaha: > do you think it would be worth reading the other data as well?
It could be used later to check that the pointing was not crazy... What do you think?

Yes, off course that would be good. In fact, the `drive_data` stores all information from the drive report. Only some of them are tagged as e,g. `time`, `azimuth_avg`, etc. Today we saw that the RA and DEC are also present in the Drive report. At present, there are no RA and DEC information in the DL1 files. They are indeed required for  finer analysis and can be added as metadata in the DL1 file.

rlopezcoto:
Hi all, thanks a lot. I think the one @vuillaut proposed is a nice improvement and you can maybe add an issue to keep track of it. I'll merge this now and make a bugfix release for you to include this information in the real data analysis.

PR #240, opened 2019-12-02
Obtain time stamps from three different sources (UCTS, TIB and Dragon modules)

morcuended:
As mentioned in Issue #227 here I include the possibility of getting the timestamps from three different sources: UCTS, TIB counters and Dragon modules counters. **The three timestamps will be stored in the dl1 parameter tables.** Even if it is redundant information, for the time being, it will give us continuity in the time values since Dragon counters are quite stable and every now and then UCTS timestamp and TIB counters stop working fine.

Still, there is the question of which time info we should take by default. Right now it is TIB. 

What do you think @FrancaCassol @rlopezcoto @vuillaut @labsaha?

### DIFF
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -312,11 +312,27 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
                         # TAI time in s taken from TIB
                         tai_time = event.r0.tel[telescope_id].trigger_time
                         utc_time = Time(datetime.utcfromtimestamp(tai_time))
-
                         gps_time = utc_time.gps
-                        dl1_container.gps_time = gps_time
 
-                        if pointing_file_path:
+                        ucts_time = event.lst.tel[telescope_id].evt.ucts_timestamp * 1e-9 # nsecs
+
+                        # Get counters of module 0
+                        dragon_time = (
+                                event.lst.tel[telescope_id].svc.date + 
+                                event.lst.tel[telescope_id].evt.pps_counter[0] + 
+                                event.lst.tel[telescope_id].evt.tenMHz_counter[0] * 10**(-7)) 
+
+                        tib_time = (
+                                event.lst.tel[telescope_id].svc.date + 
+                                event.lst.tel[telescope_id].evt.tib_pps_counter + 
+                                event.lst.tel[telescope_id].evt.tib_tenMHz_counter * 10**(-7)) 
+
+                        dl1_container.gps_time = gps_time
+                        dl1_container.tib_time = tib_time
+                        dl1_container.ucts_time = ucts_time
+                        dl1_container.dragon_time = dragon_time
+
+                        if pointing_file_path and tai_time > 0:
                             azimuth, altitude = pointings.cal_pointingposition(utc_time.unix, drive_data)
                             event.pointing[telescope_id].azimuth = azimuth
                             event.pointing[telescope_id].altitude = altitude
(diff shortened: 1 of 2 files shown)


rlopezcoto:
what about NTP as discussed in #235?

morcuended:
> what about NTP as discussed in #235?

NTP, taken as `event.lst.tel[telescope_id].svc.date`, is the same for all the subruns of each run.

FrancaCassol:
Hi @morcuended, @rlopezcoto 

what about dragon time as said yesterday? (I will just take the module zero)
I will can do change in the reader I need anyway to make PR for changing the geometry file

morcuended:
> Hi @morcuended, @rlopezcoto
> 
> what about dragon time as said yesterday? (I will just take the module zero)
> I will can do change in the reader I need anyway to make PR for changing the geometry file

Yes, @FrancaCassol, timestamps can already be calculated in the [ctapipe_io_lst](https://github.com/cta-observatory/ctapipe_io_lst) reader. However, I just included them here to keep the three of them in the output dl1 files in case we need to compare them. Maybe it would be convenient for the time being to not select any of them in `ctapipe_io_lst` but just calculate them and dump them here.

morcuended:
Hi again @FrancaCassol, Isidro suggested me to take the counters from the module placed at the center of the camera, i.e. index number 82, instead of taking the module with index 0 which corresponds to one at the edge (is this correct?). In principle, all the modules counters are quite stable up to a precision of ~100-200 ns

FrancaCassol:
Hi @morcuended,
Actually the central module is the 132 (starting from zero). No problem, I can use it.

FrancaCassol:
> However, I just included them here to keep the three of them in the output dl1 files in case we need to compare them. Maybe it would be convenient for the time being to not select any of them in `ctapipe_io_lst` but just calculate them and dump them here.

I fill uncomfortable to have zero time in the r0 container. The dragon time is always there, let use it for the moment. Unfortunately, I can not dump the information in your container because it is not defined at ctapipe_io_lst level. I suggest for the moment that I provide the the dragon time and you calculate and store all the other one. What do you think?

morcuended:
> > However, I just included them here to keep the three of them in the output dl1 files in case we need to compare them. Maybe it would be convenient for the time being to not select any of them in `ctapipe_io_lst` but just calculate them and dump them here.
> 
> I fill uncomfortable to have zero time in the r0 container. The dragon time is always there, let use it for the moment. Unfortunately, I can not dump the information in your container because it is not defined at ctapipe_io_lst level. I suggest for the moment that I provide the the dragon time and you calculate and store all the other one. What do you think?

That is fine with me. I defined them already in lstcontainer and can store them in dl1 files.

IsidroAguadoRuesga:
> Hi @morcuended,
> Actually the central module is the 132 (starting from zero). No problem, I can use it.

Hi @FrancaCassol, you are right that the central module has module ID 132 but has Module Index 82.

morcuended:
I remove the lines that were converting timestamps already in UNIX format (1.57...e9) to GPS format since the latter is not used anywhere else. Besides, timestamps in the drive report are also in UNIX format.

lstchain/reco/dl0_to_dl1.py
```diff
@@ -340,7 +355,7 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
                     writer.write(table_name=f'telescope/image/{tel_name}',
                                  containers=[event.r0, tel, extra_im])
                     writer.write(table_name=f'telescope/parameters/{tel_name}',
-                                 containers=[dl1_container])
+                            containers=[dl1_container])
```
rlopezcoto: please align

morcuended: done

morcuended:
> @morcuended @FrancaCassol this PR (after my minor comment) should be good to go, right?

In principle, this is working already as it is. I just wanted to avoid redundant code between this script and the __init__.py script of ctapipe_io_lst where the counters are defined. I think is fine like this.

Issue #208, opened 2019-11-06
Single config file ?


vuillaut:
Shall we merge config files (calib/analysis) to use a single one, regardless of usage (MC or real data)?

morcuended:
Each config file is used by a different script, right? The camera_calibration_param.json by calc_camera_calibration and the standard config file for lstchain_data_r0_to_dl1. Maybe for clarity is more convenient if we have them separated. It's not a strong opinion though.

FrancaCassol:
Hi @vuillaut, @morcuended 
yes, I think it is better to to separate the config files for each step. The important thing is that they are written out inside the output hdf5 file (as meta data). Also the config file will be different for data and MC (at least for the calibration part)

misabelber:
I also think is good to have separate config files but perhaps we could create a directory with a descriptive name (like "config_files" :'D) where to store all of them? Now they are distrubuted between "lstchain/tools" and "utils" and its a bit inconvenient.

vuillaut:
There are at least redundant parameters between configs that should be merged.
We are working reorganising the scripts with @FrancaCassol, let's clean a bit the config at the same time.

FrancaCassol:
Hi @vuillaut,
I am at a good point with the integration of the time correction. I think it is time to re-open this issue. 
We need indeed to combine the configuration values at least for all the calibration part.

garciagenrique:
Hello @FrancaCassol,
Would you need help with this issue ? Unless the task is very advanced (or @morcuended or @rlopezcoto had started doing something, as I have seen some ongoing discussion), I can take care of the issue.

FrancaCassol:
Hi @garciagenrique,
thanks, actually together with @rlopezcoto we are rather advanced on that, but I will let you know if we need help.

PR #241, opened 2019-12-06
Update muon script for running with old data

FrancaCassol:


lstchain/scripts/lstchain_data_muon_analysis.py
```diff
@@ -62,6 +55,11 @@
                                          "Default = 442",
                     type = int, default = 442)
 
+parser.add_argument("--tel_id", help = "telescope id"
+                                         "Default = 1",
+                    type = int, default = 1)
```
rlopezcoto: `tel_id` should be 0 for run number < xxx, right?
We should either specify the run number and select the tel_id with it or specify the tel_id without mentioning the run number

lstchain/scripts/lstchain_data_muon_analysis.py
```diff
@@ -74,7 +72,7 @@ def main():
     print("max events: {}".format(args.max_events))
 
     # Camera geometry
-    geom = CameraGeometry.from_name("LSTCam-002")
+    geom = CameraGeometry.from_name("LSTCam-003")
```
rlopezcoto: new camera geometry?

lstchain/scripts/lstchain_data_muon_analysis.py
```diff
@@ -91,78 +89,57 @@ def main():
...
-    cal_data =  WaveformCalibrationContainer()
-    ped_data =  PedestalContainer()
+        r1_sample_start=2,r1_sample_end=38, tel_id=tel_id)
 
-    dc_to_pe = []
-    ped_median = []
+    charge_config = Config({
+        "LocalPeakWindowSum": {
+            "window_shift": 4,
+            "window_width": 8
```
rlopezcoto: why using here different parameters as those used by default in the r0_to_dl1 analysis?

vuillaut: Linked to #208

lstchain/scripts/lstchain_data_muon_analysis.py
FrancaCassol: Yes, indeed, what do you prefer?

lstchain/scripts/lstchain_data_muon_analysis.py
FrancaCassol: this should be the correct one see 
[https://github.com/cta-observatory/ctapipe-extra/pull/25](https://github.com/cta-observatory/ctapipe-extra/pull/25)
I am going to change it also in ctapipe_io_lst

lstchain/scripts/lstchain_data_muon_analysis.py
FrancaCassol: Sorry, actually I think I changed them looking at the waveforms (they seemed the best one).
We should indeed used the values inside a configuration file. Which value are used in the r0_to_dl1 (where are the configuration files?)
But the way, are muons pulses as shower pulses?

vuillaut: Hi Franca.
Maybe we can take this opportunity to work on the merging of config files (in another PR) to get a single one?
The one used for `r0_to_dl1` is there: https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/data/lstchain_standard_config.json

FrancaCassol: Hi @vuillaut,
ok, I will have a look, but I am going before to insert the time pulse correction in the calibration chain. Then, let do a bit of cleaning also in this part.

FrancaCassol:
The content of this PR has moved to [https://github.com/cta-observatory/cta-lstchain/pull/263](https://github.com/cta-observatory/cta-lstchain/pull/263)

PR #247, opened 2019-12-12
Source dependent ana

SeiyaNozaki:
Here I implemented the source-dependent analysis in dl1_to_dl2.py and some scripts.
- prepare source-dependent parameters(time_gradient, skewness(with correct sign), dist) using the expected source position
-  add 'dist' to the regression_features/classification_features on the codes

I wanted to prepare json file for source-dependent analysis(including 'dist' and removing 'x','y' for poing-gamma MC) at first, but didn't manage to assign correct path of the json file.... 
So here I just added 'dist' directly on the codes
Do you have any ideas...?


rlopezcoto:
Great @SeiyaNozaki, thank you very much for this PR, very necessary to improve single-telescope analysis! We will review it as soon as possible.
For the json files, you can find them in [this path](https://github.com/cta-observatory/cta-lstchain/tree/master/lstchain/data), although it would be great if you could merge any of the additional options in the same file, since we are moving towards using a unique configuration file (#208)

SeiyaNozaki:
Thanks!
I included RF regression/classification features for source-dependent analysis in standard configuration file.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -294,28 +317,48 @@ def build_models(filegammas, fileprotons,
...
 
     df_gamma = utils.filter_events(df_gamma, filters=events_filters)
     df_proton = utils.filter_events(df_proton, filters=events_filters)
 
+    #Set source-dependent paramters
+    if src_dependent:
+        print("source-dependent analysis is activated (source position = {}deg)".format(expected_src_pos))
+        
+        focal_length = 28 * u.m
+        expected_src_pos = focal_length.value * np.tan(np.deg2rad(expected_src_pos))
```
rlopezcoto: I would explicitly change the name `expected_src_pos` to something like `expected_src_pos_m`, or better, define `expected_src_pos` in l.274 as `expected_src_pos_deg`

rlopezcoto: Also, this is not the usual way we use in lstchain to transform from `deg` to `m` in camera coordinates, could you have a look to the functions defined in [utils](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/reco/utils.py)?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -369,10 +412,27 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
-    classification_features = config["classification_features"]
 
     dl2 = dl1.copy()
+
+    #Set source-dependent paramters
+    if src_dependent:
+        print("source-dependent analysis is activated (source position = {}deg)".format(expected_src_pos))
+
+        focal_length = 28 * u.m
+        expected_src_pos = focal_length.value * np.tan(np.deg2rad(expected_src_pos))
```
rlopezcoto: same comments as above

lstchain/scripts/lstchain_mc_dl1_to_dl2.py
```diff
@@ -47,6 +48,24 @@
                     default=None
                     )
 
+parser.add_argument('--source_dependent', '-srcdep', action='store', type=lambda x: bool(strtobool(x)),
+                    dest='src_dependent',
+                    help='Boolean. True for source-dependent analysis. Default = False (source-independent analysis)',
+                    default=False)
+
+parser.add_argument('--expected_src_pos_x', '-srcposx', action='store', type=str,
```
rlopezcoto: at the moment, if I'm not wrong, the source position simulated in the MC is always taken as one point (for the point-like one). What will happen when instead of a point source we simulate a ring (same distance from the center of the camera, but a circle around it)? The expected source position is probably better read from the file since it will be different for every event (at the moment it is not, but just because of the characteristics of the simulation we are using right now)

lstchain/reco/dl1_to_dl2.py
```diff
@@ -294,28 +317,48 @@ def build_models(filegammas, fileprotons,
...
     df_proton = pd.read_hdf(fileprotons, key=dl1_params_lstcam_key)
 
     df_gamma = utils.filter_events(df_gamma, filters=events_filters)
     df_proton = utils.filter_events(df_proton, filters=events_filters)
 
+    #Set source-dependent paramters
+    if src_dependent:
+        print("source-dependent analysis is activated (source position = {}deg)".format(expected_src_pos))
+        
+        focal_length = 28 * u.m
```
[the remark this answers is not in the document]
SeiyaNozaki: focal length is needed to convert sky frame to camera frame, so this is needed for real data.
I will update it at the same time when I will add the function for real data.

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: for the moment I deleted it

SeiyaNozaki: Is sky_to_camera(alt, az, focal, pointing_alt, pointing_az) for the transform from deg to m?

lstchain/scripts/lstchain_mc_dl1_to_dl2.py
SeiyaNozaki: You're right. I modified to read expected source position from the file for MC data, only for gamma. For proton MC/real data dist is defined as just the distance from the center of the camera.
But for real data we need pointing information including wobble information(wobble offset/angle) to know the expected source position. I saw telescope alt/az information in drive report, but there was no information about the pointing offset if I didn't miss it.
So for the moment dist is defined only for ON observation mode, then I will modify later.
(At that time I will also modify the code following the above comments)

lstchain/data/lstchain_standard_config.json
```diff
@@ -72,6 +72,19 @@
...
+    "length",
+    "wl",
+    "skewness",
+    "kurtosis",
+    "time_gradient",
+    "leakage",
+    "n_islands",
+    "dist"
+  ],
+
```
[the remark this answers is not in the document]
SeiyaNozaki: ok, I just added "dist" to the default features in the script.
If there will be any problems, I'll prepare another config file.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -32,7 +32,7 @@
 
 
 
-def train_energy(train, custom_config={}):
+def train_energy(train, custom_config={}, src_dependent=False):
```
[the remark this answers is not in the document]
SeiyaNozaki: ok! done.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -294,28 +316,44 @@ def build_models(filegammas, fileprotons,
...
 
     df_gamma = utils.filter_events(df_gamma, filters=events_filters)
     df_proton = utils.filter_events(df_proton, filters=events_filters)
 
+    #Set source-dependent paramters
+    if src_dependent:
+        print("source-dependent analysis is activated")
+        
+        #For gamma data(MC), 'dist' is calculated as the distance between source position defined in MC and C.O.G. of shower images
+        df_gamma['dist'] = np.sqrt((df_gamma['x'] - df_gamma['src_x'])**2 + (df_gamma['y'] - df_gamma['src_y'])**2)
```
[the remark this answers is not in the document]
SeiyaNozaki: Right... done.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -294,28 +316,44 @@ def build_models(filegammas, fileprotons,
...
+        df_gamma['time_gradient'] = df_gamma['time_gradient'] * np.sign(df_gamma['x'] - df_gamma['src_x'])
```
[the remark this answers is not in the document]
SeiyaNozaki: We can get better correlation between time gradient/skewness and impact parameter(two components -> single component), which makes better regression, I think.
(But I have never checked...)
If we know the source position, there is no reason not to set the right sign in my opinion.

image

But you're right, overwriting is not good. Should I prepare another parameter, 'time_gradient_from_source' or something else? Do you have any ideas?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -294,28 +316,44 @@ def build_models(filegammas, fileprotons,
...
+        df_gamma['skewness'] = df_gamma['skewness'] * np.sign(df_gamma['x'] - df_gamma['src_x'])
+        
+        #For proton data(MC/Real), 'dist' is calculated as the distance between center of the camera and C.O.G. of shower images
+        df_proton['dist'] = np.sqrt(df_proton['x']**2 +  df_proton['y']**2)
```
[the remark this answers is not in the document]
SeiyaNozaki: thanks!

SeiyaNozaki:
Thanks @vuillaut for the review!! :)
Now I added the flag for source-dependent analysis in standard config file, and also dedicated function to set the source-dependent parameters, removing the duplicated re-definition. 

Regarding to the fixed_source option as you suggested above, it means to set given fixed source position instead of reco_source_position for the gamma/hadron separation, right? If so, should I set this fixed position as 'reco_src_x/y' for this analysis or another parameter?

moralejo:
On Mon, 3 Feb 2020 at 13:03, Seiya Nozaki <notifications@github.com> wrote:

> Thanks @vuillaut <https://github.com/vuillaut> for the review!! :)
> Now I added the flag for source-dependent analysis in standard config
> file, and also dedicated function to set the source-dependent parameters,
> removing the duplicated re-definition.
>
> Regarding to the fixed_source option as you suggested above, it means to
> set given fixed source position instead of reco_source_position for the
> gamma/hadron separation, right? If so, should I set this fixed position as
> 'reco_src_x/y' for this analysis or another parameter?
>
Not in reco_*, since that would be extremely misleading. Those should have
NaNs, or whatever error values, in case of source-dependent analysis. The
target used in the analysis should be stored (in RA Dec coordinates) in
some "analysis run header", and event-wise you should keep its position in
camera coordinates (or horizontal Alt,Az coordinates, whatever is more
convenient).

-- 

Abelardo Moralejo Olaizola
Institut de Física d'Altes Energies
Tel : +34 931641662
Fax: +34 935811938
Avís - Aviso - Legal Notice - (LOPD) - http://legal.ifae.es

-- 
Avís -
Aviso - Legal Notice - (LOPD) - http://legal.ifae.es 
<http://legal.ifae.es/>

SeiyaNozaki:
Thanks @moralejo . I agree with you.

About the source position, I saw the issue about source information #270 (Thanks @labsaha !), so for the moment I want to implement
1) input source RA/Dec by user (until such a information is included in the run header)
2) convert source RA/Dec to alt/az and the position in camera coordinate

lstchain/reco/dl1_to_dl2.py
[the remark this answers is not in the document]
vuillaut: Ha yes, you are right.
In this case, I'd suggest a new parameter indeed.
Maybe simply `time_gradient_times_sign` ? -  or `from_source` as you  suggest.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -391,20 +392,33 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
         config["classification_features"].append("dist")   
 
     regression_features = config["regression_features"]
     classification_features = config["classification_features"]
-      
+    
+    if source_dependent_analysis:
+        rf_data = temp_dl2
+    else:
+        rf_data = dl2
```
vuillaut: Hi.
I am lost here.
What is the purpose of the `temp_dl2` and `rf_data`?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -436,32 +450,46 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
-    dl2['reco_type'] = classifier.predict(dl2[classification_features]).astype(int)
-    probs = classifier.predict_proba(dl2[classification_features])[0:, 0]
+    dl2['reco_type'] = classifier.predict(rf_data[classification_features]).astype(int)
+    probs = classifier.predict_proba(rf_data[classification_features])[0:, 0]
     dl2['gammaness'] = probs
     return dl2
 
 
-def set_source_dependent_parameters(data):
+def set_source_dependent_parameters(data, src_pos_camera=[0.195, 0.]):
```
vuillaut: I would not set a default value to `src_pos_camera` here, the value should be explicitly and all the time passed.

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: If another parameter which will be used for random forest is defined,e.g. 'time_gradient_times_sign', another config file is needed for the regression_features.
But it is unfavorable to overwrite the parameters with the same parameter name.

So here I prepared(copied) the data file only for the random forest(temp_dl2), then the data file for random forest can be chosen depending on the source-dependent/independent analysis.

But there are some other possibilities:
1) save the source-dependent parameters with the different name(e.g. 'time_gradient_times_sign'), and also prepare another parameter for random forest input(like 'time_gradient_times_rf '), then this value depends on the source-dependent/independent.

2) save the source-dependent parameters with the different name(e.g. 'time_gradient_times_sign'), and prepare another config file(or another definition of 'regression_features')

3) like the current one

do you have any ideas...?
 for random forest input

SeiyaNozaki:
Hi,
About the definition of the source-dependent parameters, the following flow is what I'm thinking now

source-dependent-ana

About the protonMC, there are two options for me.
1) config file in which source position(camera coordinate) is written
2) input also gammaMC as a reference, then get the info about source position from this MC

Any comments are welcome!

moralejo:
> About the protonMC, there are two options for me.
> 
> 1. config file in which source position(camera coordinate) is written
> 2. input also gammaMC as a reference, then get the info about source position from this MC
> 
> Any comments are welcome!

I think option 1 is the way to go for proton MC, gamma MC, and also for real data, the only difference being that in the real data the input should be RA/Dec whereas for the MC it should be (x,y) on the camera or Alt/Az.

The only advantage I see from taking it from the file would be for the special case of diffuse gamma MC (with true gamma direction changing event-wise), in case you want to evaluate the performance for a point-like source at a certain offset from the FoV center. In that case, one would still use the config file for protons, using a fixed point at the desired distance from the FoV center (and this would have to be repeated for each distance at which the performance has to be evaluated).

SeiyaNozaki:
Following the comments, I refined the codes like below:
- Source-dependent parameters are "dist", "time_gradient_from_source", "skewness_from_source"
(with different name(adding "from_source") for "time_gradient" and "skewness" to avoid it overwritten)
- If analyzers select those parameters in config file, those parameters are calculated
- Concerning about the expected source position of protomMC, "mc_wobble_offset" should be written in config file, then that one in camera coordinate is calculated

rlopezcoto:
we should move on with this PR to be able to include this analysis into the standard chain.
@SeiyaNozaki could you please merge the master so we can solve the conflicts currently open?
@moralejo, @vuillaut could you have a look at the changes he implemented to see if your questions/comments were addressed?

moralejo:
> * Concerning about the expected source position of protomMC, "mc_wobble_offset" should be written in config file, then that one in camera coordinate is calculated

Apologies for long silence in this. I think one would need two coordinates, not just mc_wobble_ofset. At large zeniths it may make a significant difference in which direction the offset goes. Something like  mc_nominal_source_x, mc_nominal_source_y

SeiyaNozaki:
@moralejo  Thanks for the comments, done!

MaxNoe:
How do you deal with background estimation here? 

With on / off observations with the source in the camera center, you don't have to do anything differently.

But with wobble observations, you have to apply models using source dependent parameters for each of the off positions with the correct source dependent feature set for that off position.

We had this implemented in https://github.com/fact-project/aict-tools, but it actually did not perform better than a source independent analysis and was much more complicated so we took the decision to completely remove that possibility and our lives got much better after this.

The maintenance burden and all the trip wires for getting source dependent wobble analysis right is very high. Also the storage and computational costs are higher, as you have to apply the models for each off position and store `n_off_posititions` sets of source dependent variables and predictions.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -417,3 +429,68 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
     dl2['gammaness'] = probs
     return dl2
 
+
+def set_source_dependent_parameters(data, config):
```
MaxNoe, line 433: This is a very large and complex (lot's of branching) function. Better split in multiple smaller functions for each of the cases you handle in the different if/else statements. This will also make it easier to write tests for this./

lstchain/reco/dl1_to_dl2.py
```diff
@@ -417,3 +429,68 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
+
+            # disp_sign = source_position - c.o.g. position, so need to be multiplied by "-1"
+            if ('time_gradient_from_source' in config['regression_features']) or ('time_gradient_from_source' in config['classification_features']):
+                data['time_gradient_from_source'] = data['time_gradient'] * data['disp_sign'] * -1
+
+            if ( 'skewness_from_source' in config['regression_features']) or ('skewness_from_source' in config['classification_features']):
+                data['skewness_from_source'] = data['skewness'] * data['disp_sign'] * -1
+
+        #For proton MC, nominal source position is one written in config file
+        if all(data['mc_type'])!=0:
```
MaxNoe, line 459: This looks wrong. Do you mean `if np.all(data['mc_type'] != 0)` ?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -417,3 +429,68 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
+
+            if( ('dist' in config['regression_features']) or ('dist' in config['classification_features'])
```
MaxNoe, line 461: You could use set operations here.

```
source_features = {'dist', 'time_gradient_from_source', 'skewness_from_source', 'alpha'}

if source_features.intersection(config['regression_features']):
```

SeiyaNozaki:
Hi @MaxNoe  thanks for the valuable comments based on your experiences!

> How do you deal with background estimation here?
With on / off observations with the source in the camera center, you don't have to do anything differently.
But with wobble observations, you have to apply models using source dependent parameters for each of the off positions with the correct source dependent feature set for that off position.

Here I assume only the case of on / off observation, not wobble observation as you saw.
And yes, we need to discuss how to proceed for the analysis of wobble data after this PR.
As you said the code for wobble could make complicated. 
But if this can give us better results than source-independent analysis, it worth living with this source-dependent analysis.

Now I'm wondering why you didn't get better results with source-dependent analysis.
I don't have so much experiences in this kind of analysis, so just I wanted to know your experiences.
I found the following issue, now somebody wants to understand the decision :)
fact-project/aict-tools#22

MaxNoe:
Ok, let me see if I can get the story together.

At the starting point, we had a pretty bad reconstruction of origin for the gamma rays based on a very simplistic disp formula. 
The disp was something like `param * (length / width)` and the head/teal disambiguation was done with source dependent features (e.g. alpha).

In this state, the source dependent analysis performed much better than the first version of the source independent analysis I developed.
After implementing a machine learning based disp estimation (the one in the aict-tools), performance of the source independent analysis was now much better than the old source dependent analysis.

We never really tried to combine a source dependent analysis with the newer disp regression, as then DL3 format and the open crab paper came a long and this required a source independent analysis anyways.

So clearly it's a combination of source dependent analysis vs. independent analysis and the introduction of the new disp regression. But implementing source dependent analysis correctly for wobble observations is imho not worth the hassle (after having done so, I might add).

moralejo:
I agree source-dependent analysis is messy because of the need of processing each event with different assumptions on the source position (the true one, and control ones for background estimation). Compared to the clean approach of "one event <=> one entry in the DL3 file", it is certainly a pain. But I think it also provides some performance improvement (in energy estimation, and in g/h separation), and for MAGIC-1 it was the standard approach. Perhaps a difference with FACT is that already at threshold FACT deals with relatively well-formed showers, unlike MAGIC or LST. I think the source-dependent approach is worth pursuing. Note also that in the two physics cases in which LST standalone is most competitive (pulsars & GRBs) one can also determine the background by off-pulse or post-burst observations without needing to reconstruct the events w.r.t. an off sky position

vuillaut:
I like https://github.com/cta-observatory/cta-lstchain/issues/321#issuecomment-609633813 proposal to have a clear separation between source independent and source dependent analysis.
To implement it, we should have a `DL1SourceDependentParameters` container.
@SeiyaNozaki I can help you with implementing the changes and moving on this PR that has been stalled for quite some time if you need.

SeiyaNozaki:
Hi @vuillaut 
Actually it was one of the solution which I thought before, so I also like it!
Could you please implement this containers? Then after that I can adapt it!

vuillaut:
I think this is ready to be reviewed.
The analysis and parameters are now clearly independent and can be used easily in a separate pipeline.

rlopezcoto:
Thanks @SeiyaNozaki and @vuillaut. Let's wait a bit if anybody else has any comments since this generated some discussion in #321 and then we can merge

SeiyaNozaki:
Hi @vuillaut  @rlopezcoto 
I just modified the json config file, now everything that I wanted to implement was done.

lstchain/_version_cache.py
```diff
@@ -1 +1 @@
-version='0.4.5.post23+gite9b21b4'
+version='0.4.5.post189+git1dc69fb'
```
MaxNoe, line 1: Is this file supposed to be under version control?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -417,3 +429,85 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
+            data_type = 'mc_proton'
+    else:
+        data_type = 'real_data'
+    
+    expected_src_pos_x_m, expected_src_pos_y_m = get_expected_source_pos(data, data_type, config)
+
+    src_dep_params['expected_src_x'] = expected_src_pos_x_m
+    src_dep_params['expected_src_y'] = expected_src_pos_y_m
+    
+    src_dep_params['dist'] = np.sqrt((data['x'] - expected_src_pos_x_m)**2 + (data['y'] - expected_src_pos_y_m)**2)
```
MaxNoe, line 462: What is x here? Shouldn't be the relevant sign be the sign of the source in the shower coordinates? As in `ctapipe.image.hillas.pixel_to_shower_coordinates`?

MaxNoe: I.e. `np.sign(disp)` from below.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -417,3 +429,85 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
+    src_dep_params['time_gradient_from_source'] = data['time_gradient'] * np.sign(data['x'] - expected_src_pos_x_m)
+    src_dep_params['skewness_from_source'] = data['skewness'] * np.sign(data['x'] - expected_src_pos_x_m)
+    
+    disp, miss = camera_to_shower_coordinates(expected_src_pos_x_m, expected_src_pos_y_m, data['x'], data['y'], data['psi'])
+    alpha = np.rad2deg(np.arctan2(miss, disp))
```
MaxNoe, line 467: I think you can get around these different cases here by doing `np.rad2deg(np.arctan(np.abs(miss / disp))`

MaxNoe: So deliberately throwing away any sign information should give you the angle in [0, 90°]

lstchain/reco/dl1_to_dl2.py
```diff
@@ -417,3 +429,85 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
+    if data_type == 'mc_gamma':
+        expected_src_pos_x_m = data['src_x'].values
+        expected_src_pos_y_m = data['src_y'].values
+
+    #For proton MC, nominal source position is one written in config file
+    if data_type == 'mc_proton':
+        
+        focal_length = OpticsDescription.from_name('LST').equivalent_focal_length
+        
+        expected_src_pos = utils.sky_to_camera((data['mc_alt_tel'].values + config['mc_nominal_source_x_deg']) * u.deg,
```
MaxNoe, line 497: Use `u.Quantity(data['mc_alt_tel'].values + config['mc_nominal_source_x_deg'], u.deg, copy=False)` to be more efficient here.

Also linebreak after the first ( and then just indend by 4 spaces fixes the line length issue, like this:

```
function(
    arg1,
    arg2,
)
```

lstchain/_version_cache.py
vuillaut: no, it shoud be removed.

rlopezcoto: it is there:
https://github.com/cta-observatory/cta-lstchain/blob/f4c8df9d59c170b8863fa9dd74864fef6a43fd83/.gitignore#L2

actually you added it :D

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: `x` is x-position of c.o.g in camera coordinate, so `dist` is the distance between cog and expected source position also in camera coordinate.

About the `time_gradient` and `skewness`, the sign of those parameters is already defined in shower coordinate. At that time, its sign depends on `psi` angle. Here the sign is corrected according to the `psi` (the relation between the position of cog and expected source)

SeiyaNozaki: `np.abs(skewness) * np.sign(disp)` is also one of the option

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: thanks, done!

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: thanks, done!

lstchain/_version_cache.py
SeiyaNozaki: removed!

rlopezcoto:
@SeiyaNozaki there are some conflicts, could you resolve them before merging?

@MaxNoe, @moralejo, @vuillaut are you happy with this PR?

MaxNoe:
There is a conflict that needs resolution

SeiyaNozaki:
Hi, I resolved the conflict!

MaxNoe:
@SeiyaNozaki Basically only 17 % of the code you've writte is covered by unit tests. Could you maybe add some tests to make surethis works and stays working?

lstchain/reco/dl1_to_dl2.py
MaxNoe: Are you sure taking the x camera position works for all possible source location? I didn't think it through.

SeiyaNozaki: You're right... I misunderstood...
Let me modify the codes.

SeiyaNozaki: done!

lstchain/reco/dl1_to_dl2.py
```diff
@@ -417,3 +429,92 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, custom_config={})
...
+    #For proton MC, nominal source position is one written in config file
+    if data_type == 'mc_proton':
+        
+        focal_length = OpticsDescription.from_name('LST').equivalent_focal_length
+        
+        expected_src_pos = utils.sky_to_camera(
+            u.Quantity(data['mc_alt_tel'].values + config['mc_nominal_source_x_deg'], u.deg, copy=False),
+            u.Quantity(data['mc_az_tel'].values + config['mc_nominal_source_y_deg'], u.deg, copy=False),
+            focal_length,
+            data['mc_alt_tel'].values * u.deg,
```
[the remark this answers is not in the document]

lstchain/scripts/lstchain_add_source_dependent_parameters.py
```diff
@@ -0,0 +1,40 @@
+# script to add the source dependent parameters to a DL1 file
```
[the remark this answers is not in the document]
SeiyaNozaki: I already implemented the unit test in `lstchain/tests/test_lstchain.py`.
Is it fine?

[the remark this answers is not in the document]
MaxNoe: Adding the test for the script should be very simple, just like the other scripts are tested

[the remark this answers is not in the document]
rlopezcoto: @SeiyaNozaki I think @MaxNoe refers to an implementation of a test unit for the script `lstchain_add_source_dependent_parameters.py`

as he is suggesting, you can do it the same way as it is done for the other scripts in  `lstchain/scripts/`. You should implement it in the test file `lstchain/scripts/tests/test_scripts.py`, where you can find some examples of the implementations for the other scripts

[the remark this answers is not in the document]
SeiyaNozaki: Ok, thanks. I implemented the test in lstchain/scripts/test_scripts.py

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -9,13 +11,23 @@
...
     input_file = mc_gamma_testfile
     cmd = f'lstchain_mc_r0_to_dl1 -f {input_file} -o {output_dir}'
     os.system(cmd)
     assert os.path.exists(dl1_file)
 
 
+@pytest.mark.run(after='test_lstchain_mc_r0_to_dl1')
+def test_add_source_dependent_parameters():
+    cmd = f'lstchain_add_source_dependent_parameters -f {dl1_file}'
+    os.system(cmd)
```
MaxNoe, line 26: I don't think pytest realizes that this counts towards coverage. So better keep the other test ;)

SeiyaNozaki: ok, done:)
Now all checks have passed!

PR #248, opened 2019-12-13
Update calibrator with pulse correction and gain selection

FrancaCassol:
In this PR it has been added:
- in calibrator.py: 
-- the pulse time correction: if the calibrator find the time calibration file it applies the drs4 corrections otherwise  it just applies the flat-field time corrections
-- the gain selection based on the threshold method: if the threshold is set to None, no selection is performed.  I add the new notebook apply_calibration_with_gain_selection.ipynb to show an example of application
- in onsite_create_calibration_file.py script: is added the calculation of the drs4 time corrections that are written in a hdf5 file in calibration directory of the onsite tree
- in lstchain_data_create_calibration_file.py: I add new cuts to avoid magic lidar events

The configuration for all the calibration is in onsite_camera_calibration_param.json
Next step is to better coordinate all the configuration files


FrancaCassol:
Hi @vuillaut, @rlopezcoto,

I think that at this point we should fast review this code

FrancaCassol:
To make things more clear I  will commit the correction of the reconstruction script in a different PR

lstchain/calib/camera/time_correction_calculate.py
```diff
@@ -48,15 +49,10 @@ class TimeCorrectionCalculate(Component):
...
-                       help='Define the width of the integration window'
-                      ).tag(config=True)
-
-    window_shift = Int(
-                      default_value=3,
-                      help='Define the shift of the integration window'
-                      'from the peak_index (peak_index - shift)'
-                      ).tag(config=True)
+    charge_product = Unicode(
+        'LocalPeakWindowSum',
```
rlopezcoto: Shall this be hardcoded?

FrancaCassol: This extractor is not suppose to change often, but why do you want to have it hardcoded?

lstchain/calib/camera/calibrator.py
```diff
@@ -128,9 +154,25 @@ def _calibrate_dl1(self, event, telid):
...
+        if self.time_corrector:
+            pulse_corr_array = self.time_corrector.get_corr_pulse(event, pulse_time)
+
+        # otherwise use the ff time correction (not drs4 corrected)
+        else:
+            pulse_corr_array = pulse_time + self.mon_data.tel[telid].calibration.time_correction
+
+        # perform the gain selection if the threshold is defined
+        if self.gain_threshold:
+            waveforms, gain_mask = self.gain_selector(event.r1.tel[telid].waveform)
```
vuillaut: shouldn't this be calculated on r0?
r1 are in p.e. at this point?

lstchain/calib/camera/calibrator.py
```diff
@@ -128,9 +154,25 @@ def _calibrate_dl1(self, event, telid):
...
+            event.dl1.tel[telid].image = charge[gain_mask, np.arange(charge.shape[1])]
+            event.dl1.tel[telid].pulse_time = pulse_corr_array[gain_mask, np.arange(pulse_corr_array.shape[1])]
+            # remember the mask in the pixel_status array (this info  is missing for the moment in the
+            # r1 cantainer)
+            event.lst.tel[telid].evt.pixel_status = gain_mask
+
+        else:
+            event.dl1.tel[telid].image = charge
+            event.dl1.tel[telid].pulse_time = pulse_corr_array
```
vuillaut: In this case, the image and pulse_time will be of shape `(1855,2)` instead of `(1855,)`.
I don't think this is what we want.
Instead, I'd recommend a ManualGainSelection to select high gain by default.

vuillaut: And then the notebook will need to be modified to plot images with single dimension only.

FrancaCassol: Hi @vuillaut, 
actually this is what I want: I would like to have the possibility to calibrate without performing any gain selection (mainly for debug purposes). The HG by default can be obtained just putting the default threshold to 4095 ;-)
By the way, here I fixed the gain selector to the Threshold one, should we give the possibility to choose other selectors?

lstchain/calib/camera/calibrator.py
FrancaCassol: The r1 waveform is low-level corrected but still in ADC

lstchain/calib/camera/calibrator.py
vuillaut: I understand the advantage of having this for debugging but here the default option is to not select gains (since `None` is the default for `gain_threshold`). Maybe the default should be to have a gain selection to make sure that image and pulse_time have the correct dimension by default? And keeping both gains be a "hidden feature" for expert and debug?

To your question, yes I think it would be better to have the gain selector be the one passed in the config file.

lstchain/calib/camera/calibrator.py
vuillaut: Isn't this different from CTA data model?

lstchain/calib/camera/calibrator.py
FrancaCassol: > I understand the advantage of having this for debugging but here the default option is to not select gains (since `None` is the default for `gain_threshold`). Maybe the default should be to have a gain selection to make sure that image and pulse_time have the correct dimension by default? And keeping both gains be a "hidden feature" for expert and debug?

No problem, I can make 4094 as default
> 
> To your question, yes I think it would be better to have the gain selector be the one passed in the config file.

Then I must take out the threshold trailet and pass the gain selector and one must not forget to pass the selector Config. @rlopezcoto what do you think?

FrancaCassol: > And then the notebook will need to be modified to plot images with single dimension only.

which notebook?

vuillaut: The one in this PR

FrancaCassol: This is already working with images with single dimansion

lstchain/calib/camera/calibrator.py
FrancaCassol: Hi @vuillaut,
 thinking gain to this point I agree that  it is better to use the r0 waveform (the fact that it is not low level corrected should not affect the result). I am going to change it.
You are right, our r1 is different than in the ctpipe data model. This is due to the fact that we have a level more in the data (ctapipe r0 is supposed to be already low-level corrected), we should solve this issue in the next (but not in this PR ;-))

FrancaCassol:
Hi,
are there other comments on this PR? 
Otherwise I think we can merge it.

lstchain/scripts/lstchain_data_muon_analysis.py
```diff
@@ -29,6 +21,7 @@
 --output_file Data_table.fits --pedestal_file pedestal_file_run446_0000.fits 
 --calibration_file calibration.hdf5
 --max_events 1000
+--tel_id 0
```
vuillaut: shouldn't this be 1 by default?
ok it is just an example...

lstchain/scripts/lstchain_data_muon_analysis.py
```diff
@@ -74,7 +72,7 @@ def main():
     print("max events: {}".format(args.max_events))
 
     # Camera geometry
-    geom = CameraGeometry.from_name("LSTCam-002")
+    geom = CameraGeometry.from_name("LSTCam-003")
```
vuillaut: Is `LSTCam-003` already part of current environment?
ctapipe v0.7 comes with what? ctapipe_resources v0.2.18 ?

morcuended: LSTCam-003 was recently included in ctapipe_io_lst (see [New time and geometry #19](https://github.com/cta-observatory/ctapipe_io_lst/pull/19))

rlopezcoto:
@FrancaCassol it seems that the latest commits made the code break on the calibration part, could you please check it out? Thanks

rlopezcoto:
Hi @FrancaCassol, did you have time to have a look at this? It would be great if we could move on with this one to get #241 also merged

FrancaCassol:
Hi @rlopezcoto,
yes I will have a look

rlopezcoto:
Thanks @FrancaCassol!
Regarding the discussion we had in #252, and just for the record, now for the gain selection, the *threshold=3500* and is set on *R1* waveforms, right?

FrancaCassol:
Hi @ruben,
yes, the threshold in lstchian_standard_config.json is 3500 and the selection is done on the R1 waveform :

`waveforms, gain_mask = self.gain_selector(event.r1.tel[telid].waveform)`

Issue #246, opened 2019-12-12
Write only images table (DL1a) if cleaning levels are set to 0


morcuended:
Related to #237 

Provided cleaning levels are set to 0 in order to keep all the events, it does not make sense to calculate Hillas parameters. Thus DL1 output file should contain only images and pulse times tables in these cases.

vuillaut:
Part of  https://github.com/cta-observatory/cta-lstchain/projects/1#card-30110590

Once this refactoring is done, DL1a and b will be done in different steps and thus the event selection by cleaning will be done after DL1a .
Although I was inclined to wait a bit to see things settling in ctapipe concerning DL1 processing tools and DL1 format to start this.

morcuended:
Thanks for the clarification @vuillaut. Then, for the time being, should we implement in the `dl0_to_dl1.py` script the possibility of not storing the parameters table? Or do we just wait for ctapipe?
I thought we could add just a line before writing the table checking whether cleaning levels are 0 or not as we discussed in the past call.

PR #249, opened 2019-12-13
Not writing DL1b if cleaning levels are set to 0

morcuended:
Following #246 I implement here the possibility of not storing parameter tables if cleaning levels are set to 0 in order to keep all the events.

This will not be necessary once the refactoring is done as @vuillaut comments in #246

### DIFF
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -56,7 +56,7 @@ cleaning_method = tailcuts_clean
 
 filters = tables.Filters(
     complevel=5,    # enable compression, with level 0=disabled, 9=max
-    complib='blosc:zstd',   #  compression using blosc
+    complib='blosc:zstd',   # compression using blosc
     fletcher32=True,    # attach a checksum to each chunk for error correction
     bitshuffle=False,   # for BLOSC, shuffle bits for better compression
 )
@@ -98,17 +98,17 @@ def get_dl1(calibrated_event, telescope_id, dl1_container=None, custom_config={}
 
     if image[signal_pixels].sum() > 0:
 
-        # check the number of islands 
+        # check the number of islands
         num_islands, island_labels = number_of_islands(camera, signal_pixels)
 
         if use_main_island:
             n_pixels_on_island = np.zeros(num_islands+1)
 
             for iisland in range(1, num_islands+1):
-                n_pixels_on_island[iisland] = np.sum(island_labels==iisland)
-              
+                n_pixels_on_island[iisland] = np.sum(island_labels == iisland)
+
             max_island_label = np.argmax(n_pixels_on_island)
-            signal_pixels[island_labels!=max_island_label] = False
+            signal_pixels[island_labels != max_island_label] = False
 
         hillas = hillas_parameters(camera[signal_pixels], image[signal_pixels])
 
@@ -183,7 +183,10 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
 
     if not is_simu:
         # TODO : add calibration config in config file, read it and pass it here
-        charge_config = Config({"LocalPeakWindowSum":{"window_shift": 5,"window_width":12}})
+        charge_config = Config({"LocalPeakWindowSum":
+                                {"window_shift": 5,
+                                 "window_width": 12}
+                                })
 
         r0_r1_calibrator = LSTR0Corrections(pedestal_path=pedestal_path,
                                             r1_sample_start=2,  # numbers in config ?
@@ -311,7 +314,7 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
 
                         # gps_time = event.r0.tel[telescope_id].trigger_time
 
-                        ucts_time = event.lst.tel[telescope_id].evt.ucts_timestamp * 1e-9 # nsecs
+                        ucts_time = event.lst.tel[telescope_id].evt.ucts_timestamp * 1e-9  # nsecs
 
                         # Get counters from the central Dragon module
                         module_id = 82
@@ -357,8 +360,7 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
 
                     if config['tailcut']['picture_thresh'] > 0 \
                             and config['tailcut']['boundary_thresh'] > 0 \
-                            and config['tailcut']['min_number_picture_neighbors'] > 0 \
-                            and not config['tailcut']['keep_isolated_pixels']:
+                            and config['tailcut']['min_number_picture_neighbors'] > 0:
                         writer.write(table_name=f'telescope/parameters/{tel_name}',
                                      containers=[dl1_container])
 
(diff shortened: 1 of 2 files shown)


morcuended:
I removed one of the conditions (on `keep_isolated_pixels`) in order to make the code pass the test. 

By default, the values in the config file are:
```
  "tailcut": {
    "picture_thresh": 6,
    "boundary_thresh": 3,
    "keep_isolated_pixels": false,
    "min_number_picture_neighbors": 1
  },
```
Setting them to:
```
  "tailcut": {
    "picture_thresh": 0,
    "boundary_thresh": 0,
    "keep_isolated_pixels":true,
    "min_number_picture_neighbors": 0
  },
```
keeps all the events and prevents from writing the parameters table.

rlopezcoto:
not sure if we would like to leave as default configuration file this option, isn't there any other way of bypassing this test?

morcuended:
> not sure if we would like to leave as default configuration file this option, isn't there any other way of bypassing this test?

Do you mean having a flag selecting whether one wants to write parameters table or not instead of using the tailcut cleaning values?

rlopezcoto:
no, sorry, I thought that you had changed the default config file (to the config to keep all events) to make the code pass the tests. 
I saw the changes in the last commit and they are ok, but we should make sure this information propagates somehow... I know @vuillaut wanted to unify all config files, but what do you think about creating one config file `lstchain..._keep_all_events.json` so we can use it as a default to keep all events?

morcuended:
Sure, I can add a duplicated version of the existing standard config file just changing the tailcut cleaning levels to:
```
  "tailcut": {
    "picture_thresh": 0,
    "boundary_thresh": 0,
    "keep_isolated_pixels":true,
    "min_number_picture_neighbors": 0
  },
```
Is that OK? What about the gain selection threshold? Should I set it to 20.9 phe?

morcuended:
Concerning the gain selection, I just saw the discussion within PR #248, so right now the threshold value written in config file is actually not taken into account

rlopezcoto:
> Concerning the gain selection, I just saw the discussion within PR #248, so right now the threshold value written in config file is actually not taken into account

The threshold value from the config file is used for MC (and should be given in phe, so 20.9 as you said), but it is not used for data because it was hardcoded and actually never exceeded (see #252)

rlopezcoto:
Hi, I'm happy with this, but let's see what others think.

lstchain/reco/dl0_to_dl1.py
```diff
@@ -354,8 +357,12 @@ def r0_to_dl1(input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
 
                     writer.write(table_name=f'telescope/image/{tel_name}',
                                  containers=[event.r0, tel, extra_im])
-                    writer.write(table_name=f'telescope/parameters/{tel_name}',
-                                 containers=[dl1_container])
+
+                    if config['tailcut']['picture_thresh'] > 0 \
```
rlopezcoto: Just re-checking, if the parameter table is not written here, the code would break when using simulations and trying to access the parameter table to write the disp, right?
https://github.com/cta-observatory/cta-lstchain/blob/b330595571dd0becb5fbc243b4ce6b4b85197bbd/lstchain/reco/dl0_to_dl1.py#L373

garciagenrique:
> This will not be necessary once the refactoring is done as @vuillaut comments in #246

The PR seems good to me too (except for the problem that Ruben just rised up). 

However, once the refactoring would be done, we will need to modify again the code. Wouldn't be easier to create a local branch for this task and wait for the implementation of the refactoring ?

lstchain/reco/dl0_to_dl1.py
morcuended: You are right @rlopezcoto, actually, I did not test the script with MC data. Does it make sense to include also the condition `and not is_simu` to the others one as a workaround?

morcuended:
> > This will not be necessary once the refactoring is done as @vuillaut comments in #246
> 
> The PR seems good to me too (except for the problem that Ruben just rised up).
> 
> However, once the refactoring would be done, we will need to modify again the code. Wouldn't be easier to create a local branch for this task and wait for the implementation of the refactoring ?

Yes, so how do we proceed then?

lstchain/reco/dl0_to_dl1.py
rlopezcoto: I would say that in the case of simulations we may also want to write all events for some reason, so you could include the same `if` condition to write the disp information in the parameter table only in case some cleaning > 0 is given

rlopezcoto:
> Yes, so how do we proceed then?
I would say we correct what was reported above and merge the code (there are a couple of bugs (#252 ) that _must_ be corrected and @FrancaCassol is waiting for this and #248 to be merged to correct them). We should make a release with all these changes for the reprocessing of the November data.
Later we can think on the refactoring, do you agree?

vuillaut:
Hi all.
Sorry I could not look at this PR earlier.
I agree with @garciagenrique.
I am not very fond of the approach (introducing conditions for one specific case, maybe breaking some code later on in the analysis) to deal in a circumvent way with the production of the DL1a without DL1b... The advantages/drawbacks balance is not positive in my humble opinion.
Is it really an issue to have the parameters in the file at the moment until we progress to have clearly separated steps (DL0 - DL1a - DL1b - DL2 ... )?
I am sorry I don't like to push back on on-going work, this is just my opinion there.

morcuended:
I guess that having the parameters in the files is not a big deal. I just pushed it because we agreed to do it like this. But yes, I agree that this is just a patch. We can leave it as it is, it's OK with me. Maybe keeping just the new config file to keep all the events?

rlopezcoto:
Hi, I agree with keeping the config file to keep all events.

Also regarding DL1a/b, I think we need to have this discussion and the code refactored to be able to write either only images (dl1a, in cases like this one here in which parameters do not make sense) and only parameters (dl1b, in case we do not want to have big volumes). Even if this is not absolutely fulfilling with the final CTA Data model, it is very much needed in the test stage we are at the moment.
For the time being, if no errors are issued when writing dl1b when applying 0 threshold, then it is ok to leave the parameters written for the time being.

Btw, @morcuended, did you check that a threshold set at 0 is enough to keep all events? We may want to put a negative value just to make sure.

FrancaCassol:
Hi,
I think also that it is better to set a clear DL1a - DL1b  definition before changing the writing model. In any case, if an "if" should be used, better to put it before the parameter calculation (it is a waste of CPU to calculate something that you don't use afterwards)

morcuended:
> Btw, @morcuended, did you check that a threshold set at 0 is enough to keep all events? We may want to put a negative value just to make sure.

Setting them as negative was giving me errors, but will check it out again. When setting them to zero I get 52999 events/subrun. I guess these are all the events/subrun contained in the raw files

rlopezcoto:
the pep8 corrections were nice 😄
in any case, I suggest we move ahead with this and keep in mind the possibility of separating DL1ab (Open an issue?), do you agree @morcuended ?

morcuended:
> the pep8 corrections were nice 
> in any case, I suggest we move ahead with this and keep in mind the possibility of separating DL1ab (Open an issue?), do you agree @morcuended ?

Yes, I did them in the same commit and I reverted it. My bad.
If it is worth to keep this config file let's go forward then. We can open an issue reminding that this refactoring is on the project board indeed.

Issue #22, opened 2018-10-17
Use uproot for reading root files


MaxNoe:
You don't need root for that, just `pip install uproot`.

See https://github.com/scikit-hep/uproot/ and https://github.com/cta-observatory/mars_dl2_to_json/blob/master/mars_to_hdf5.py

rlopezcoto:
I guess you are referring to the magic-lst scripts by @labsaha, right?

labsaha:
@MaxNoe Good to know that `uproot` can be used to read mars classes.

MaxNoe:
I just wanted to drop the info here that uproot can read MARS root trees efficiently into numpy arrrays and is a pure-python library.

That would remove the need for any conversion, you could just create a MAGIC `EventSource`

PR #253, opened 2019-12-17
Update ctapipe_io_lst version and __version__ of lstchain

rlopezcoto:
- Updated `ctapipe_io_lst` version

- After [PR #22 from ctapipe_io_lst](https://github.com/cta-observatory/ctapipe_io_lst/pull/22), I have updated the way the version from lstchain is taken following @MaxNoe's solution

### DIFF
--- a/lstchain/__init__.py
+++ b/lstchain/__init__.py
@@ -8,5 +8,5 @@ from . import image
 
 from .io import standard_config
 
-from . import version
-__version__ = version.get_version(pep440=False)
+from .version import get_version
+__version__ = get_version(pep440=False)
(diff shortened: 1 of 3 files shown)


MaxNoe:
The path needs to be adapted. Or better be made as an argument for `update_release_version`. Otherwise the installed version will be `'unknown'`.

See https://github.com/cta-observatory/ctapipe_io_lst/blob/9c96587164e86d8be1c5e38c568e37a9e1143704/version.py#L36

A better solution would be to define a base dir for the version file.

morcuended:
Hi @rlopezcoto, `environment.yml` is still pointing to ctapipe_io_lst v0.3.1 instead of the new release v0.4.0

PR #254, opened 2019-12-18
update version

rlopezcoto:
@morcuended thanks for spotting this in #253 and sorry, I got distracted with the update in the version of `lstchain` and forgot to commit the change of version of `ctapipe_io_lst`...


rlopezcoto:
Thanks @garciagenrique!

PR #263, opened 2020-01-20
Update muon script and calibration scripts

FrancaCassol:
This PR include
- the muon script is updated with the new gain selection and with tel_id information in order to be able to run it on old data.
- the calibration script get some new trailets which permit to select only the time or only the charge calibration 
- some other minors changes


rlopezcoto:
So this one cancels #241?

FrancaCassol:
Ops, yes, I will close it

FrancaCassol:
Concerning the time calibration file, after discussion with @rlopezcoto  the idea is to create for each night a link (in the calibration directory)  to the correct time calibration file. This is performed by the onsite_create_calibration_script. 
The action is steered by the trailet "defaut_time_run".  If this trailet is set to zero, the script calculates the time calibration from the data of that night otherwise perform the link. The default value of the trailet is now 1625 with is the run suggested by Pawel as a good time run.

rlopezcoto:
Hi @FrancaCassol, changes look good to me. This needs to be merged before the next release, so I'll leave it here for a while in case somebody has a comment, otherwise I'll merge it tonight.

FrancaCassol:
ok!

PR #264, opened 2020-01-20
Updating the scripts to adapt zenith from the drive log

labsaha:
Regarding the pointing information. This PR now considers "Elevation" as  "zenith" angle. Then we calculate Altitude  which is equal to `"90 - zenith"`
We could have changed everything in the `pointings.py` file and get the altitude from that one. However, I want to keep is explicit that we get `Azimuth` and `Zenith` from the drive report. Hence I changed also to `dl0_to_dl1` file


labsaha:
@rlopezcoto Yes, this is based on that solution. I'll write to them so that they just rename it.

PR #268, opened 2020-01-22
Load geometry from ctapipe_io_lst

FrancaCassol:


lstchain/visualization/plot_calib.py
```diff
@@ -3,7 +3,8 @@
 from ctapipe.visualization import CameraDisplay
 import numpy as np
 from matplotlib.backends.backend_pdf import PdfPages
-from ctapipe.instrument import  CameraGeometry
+from ctapipe.instrument import CameraGeometry
```
rlopezcoto: do we still need this line?

FrancaCassol: no

PR #274, opened 2020-01-28
Improve time calib

pawel21:
Hi,
I added two changes to improve time calibration:

1. `calib_pulse_time_jit` as static method to avoid numba warning. (@FrancaCassol )
2. Update notebook to according last changes


rlopezcoto:
it looks good to me, but maybe @FrancaCassol, as she was the one suggesting it, has some comments

rlopezcoto:
@FrancaCassol are you ok with these changes?

PR #276, opened 2020-01-30
adding trigger info in dl1 files

### LINKED ISSUE #275: Missing trigger info in DL1?
Hi,
I would like to play with interleaved events using the DL1 files, but I could not find the trigger info in the event parameters, is it missing or am I  just looking in the wrong place? If it is missing, I would strongly suggest to add it so as to able to easily select the correct type of events.

vuillaut:
Info are added both in image and params tables.
Fixes #275

### DIFF
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -194,6 +194,10 @@ class ExtraMCInfo(Container):
 class ExtraImageInfo(Container):
     """ attach the tel_id """
     tel_id = Field(0, "Telescope ID")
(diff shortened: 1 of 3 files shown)


FrancaCassol:
Hi @morcuended, @rlopezcoto,
would  be possible to perform the next DL1 production with this PR included? 
thanks

rlopezcoto:
Hi @FrancaCassol, as far as I know, the v.0.4.3 production should be almost done by now. This PR should be included in the next reprocessing.

Issue #278, opened 2020-02-04
Wrong argument gammaness for plot ROC Curve on lstchain.visualization.plot_dl2 


andres-baquero:
When I'm trying to get a ROC Curve using lstchain.visualization.plot_dl2

I get this error message:

```
ValueErrorTraceback (most recent call last)
<ipython-input-111-57d9c62c3bcc> in <module>
----> 1 plot_ROC(cls_gh,test,classification_features,-1)

<ipython-input-110-e0248c5b64a4> in plot_ROC(clf, data, features, Energy_cut)
      7     print(check)
      8 
----> 9     fpr_rf, tpr_rf, _ = roc_curve(1-data['gammaness'],check)
     10 
     11     plt.plot(fpr_rf, tpr_rf,

/fefs/aswg/software/virtual_env/anaconda3/envs/cta/lib/python3.7/site-packages/sklearn/metrics/_ranking.py in roc_curve(y_true, y_score, pos_label, sample_weight, drop_intermediate)
    769     """
    770     fps, tps, thresholds = _binary_clf_curve(
--> 771         y_true, y_score, pos_label=pos_label, sample_weight=sample_weight)
    772 
    773     # Attempt to drop thresholds corresponding to points in between and

/fefs/aswg/software/virtual_env/anaconda3/envs/cta/lib/python3.7/site-packages/sklearn/metrics/_ranking.py in _binary_clf_curve(y_true, y_score, pos_label, sample_weight)
    534     if not (y_type == "binary" or
    535             (y_type == "multiclass" and pos_label is not None)):
--> 536         raise ValueError("{0} format is not supported".format(y_type))
    537 
    538     check_consistent_length(y_true, y_score, sample_weight)

ValueError: continuous format is not supported
```
Apparently this is because `gammaness` is not a binary value.

Looking in sklearn.metrics.roc_curve parameters this must be like this:

```
Parameters

    y_true:array, shape = [n_samples]
        True binary labels. If labels are not either {-1, 1} or {0, 1}, then pos_label should be explicitly given.
    y_score:array, shape = [n_samples]
        Target scores, can either be probability estimates of the positive class, confidence values, or non-thresholded measure of decisions (as returned by “decision_function” on some classifiers).

```

vuillaut:
Hi @andres-baquero 
Thank you for reporting this issue.
[...]

PR #279, opened 2020-02-04
cleanup and bugfix in viz dl2

vuillaut:
Fixes issue #278 

I took this as an opportunity to do some cleanup in viz functions .


kosack:
> Again, nothing against ctaplot, but for me it is an overkill using external libraries to plot either roc curves, energy/angular resolutions or sensitivities. Since this is a personal package, I would like to hear the opinion from @kosack whether it is recommended or not to use it (why is it not used anywhere in ctapipe?)

I disagree there - the way in which sensitivity curves, IRFs, etc are computed and plotted, is quite complex and can be done in many ways with various assumptions that hide what the user is doing and make it impossible to do a true comparison. We've had many problems with that in the past - not everyone agrees even what "PSF" is (90% containment? RMS? etc).  

 However, we've been discussing with @vuillaut and others that sensitivity (and other IRFs) must be in a common package, and will start to move toward that (and perhaps ctaplot is not the right place).

The idea of ctaplot is to have a nice toolbox for making benchmark plots for code generated by ctapipe-based or other pipelines (that use a common data format and model), in a common way, with common assumptions, so that we can generate a set of notebooks with all the required benchmarks, without having a ton of boilerplate code pasted in each one.

vuillaut:
> Hi @vuillaut, thanks for the fixes.
> Again, nothing against `ctaplot`, but for me it is an overkill using external libraries to plot either roc curves, energy/angular resolutions or sensitivities. Since this is a personal package, I would like to hear the opinion from @kosack whether it is recommended or not to use it (why is it not used anywhere in ctapipe?)

I am actually in the process of moving `ctaplot` under cta-observatory from @kosack request.
The whole purpose of `ctaplot` is actually to *not* reimplant such simple but specific plots in each reconstruction pipeline of parallel projects. It would also ensure that we all use the same definitions of metrics everywhere (clearly not obvious) and also reduce a  lot code maintenance.
`ctaplot` is already used in `lstchain` (and I was not the one to introduce it) but if you don't see the use of it, I won't be pushing for it (but won't reimplant these plots either in `lstchain` and make double maintenance  when there are bugs in things  already existing as the example above)
Concerning the  use in `ctapipe`, I think the answer is simple, `ctapipe` does not provide a pipeline.

kosack:
We will also use ctaplot in the Protopipe benchmarks

rlopezcoto:
@kosack:
> sensitivity (and other IRFs) must be in a common package, and will start to move toward that...

> The idea of ctaplot is to have a nice toolbox for making benchmark plots

I think these are two different things: shouldn't the "toolbox for making benchmark plots"  be the purpose of `cta-benchmarks`? `ctaplot` is in principlye a library to plot energy/angular resolution and sensitivity with the outputs (the points) that have been given to you.

@vuillaut: 
> ctaplot is already used in lstchain (and I was not the one to introduce it)

and I had already discussed about it at least in 3/4 PRs. We are calculating everything in `lstchain` and a simple plot should take 3 lines.  My proposal has always been that we write the output (the points) that can be handled by ctaplot or other plotting library.
> ctapipe does not provide a pipeline.

the definition of ctapipe is "CTA Low-level Data Processing **Pipeline** Framework Prototype"

vuillaut:
> @vuillaut:
> 
> > ctaplot is already used in lstchain (and I was not the one to introduce it)
> 
> and I had already discussed about it at least in 3/4 PRs. We are calculating everything in `lstchain` and a simple plot should take 3 lines. My proposal has always been that we write the output (the points) that can be handled by ctaplot or other plotting library.

I thought the problem was adding it as a dependency but was fine to have it for control plots, which is exactly the case here.
But no problem, I'll close the PR. 

> > ctapipe does not provide a pipeline.
> 
> the definition of ctapipe is "CTA Low-level Data Processing **Pipeline** Framework Prototype"

And yet... 🙄

vuillaut:
> I think these are two different things: shouldn't the "toolbox for making benchmark plots" be the purpose of `cta-benchmarks`? `ctaplot` is in principlye a library to plot energy/angular resolution and sensitivity with the outputs (the points) that have been given to you.

For the record, `cta-benchmarks` is aiming at providing a framework to automatically produce benchmarks, version them and control the evolution of pipelines. It does *not* aim at being used by users, nor at providing a way to compute IRFs, but will use an external library able to do so.

kosack:
> the definition of ctapipe is "CTA Low-level Data Processing Pipeline Framework Prototype"

I think you're ignoring the word "framework" - its a library for making pipelines, not a pipeline

rlopezcoto:
Ok, sorry it took so long to get back to this. If we want to use `ctaplot` to perform these comparisons, I would suggest that that it is clearly explained what are the inputs of every of the functions that are being called on lstchain. Additionally, I would suggest that on output should be in any case written with the final results (points and errors), so they can be handled with other plotting options if desired, does that sound ok?

rlopezcoto:
any news on this @vuillaut? do you agree with the latest proposal?

vuillaut:
Sorry, I completely over-regarded this issue and then forgot.

> If we want to use `ctaplot` to perform these comparisons, I would suggest that that it is clearly explained what are the inputs of every of the functions that are being called on lstchain. 

I am not sure to understand what you mean here, apart from improving the docstring?

> Additionally, I would suggest that on output should be in any case written with the final results (points and errors), so they can be handled with other plotting options if desired, does that sound ok?

The computation and the plotting can be separated for some of the metrics (actually ctaplot is built that way and the functions exist).
But to write them, what you are asking is basically to build the DL3 files. As there is an on-going task to do exactly that outside of lstchain, I think that would effectively double the work and/or define a different output dataformat, which would be counter-effective.

rlopezcoto:
Connected to #391, ctaplot was already moved into the `cta-observatory` organization:
https://github.com/cta-observatory/ctaplot

> I am not sure to understand what you mean here, apart from improving the docstring?

I think there have been some documentation updates since this PR was opened, I'll have a look at that.

> But to write them, what you are asking is basically to build the DL3 files. As there is an on-going task to do exactly that outside of lstchain, I think that would effectively double the work and/or define a different output dataformat, which would be counter-effective.

what I was proposing here is that this is actually written by `ctaplot`. If I do not understand wrong, `DL3` files should be photon lists, what I'm saying is that the plot that is outputted by `ctaplot` comes together with a file (.txt?) with the points that are plotted (angular/energy resolution) or the histograms.

Final question, are we compatible with the latest ctaplot release (v0.5.0)? We should probably follow some released version to avoid breaking of the code in the future.

MaxNoe:
> what I was proposing here is that this is actually written by ctaplot. If I do not understand wrong, DL3 files should be photon lists, what I'm saying is that the plot that is outputted by ctaplot comes together with a file (.txt?) with the points that are plotted (angular/energy resolution) or the histograms

No. These file will be the input to ctaplot. Not its output.

DL3 is photons lists AND the IRFs already included. See https://gamma-astro-data-formats.readthedocs.io/en/latest/

Reconstructed event lists used as input for the IRF calculation would be considered DL2, I think.

Work as also started on a common package for calculation of IRFs: https://github.com/cta-observatory/pyrf

ctaplot should read input and plot, not make irf calculations or complicated IO.
If you are talking about serialization plots to something other than image formats, this would probably be nice. But the important information is stored in the DL3 files, not in something ctaplot produces.

vuillaut:
> > what I was proposing here is that this is actually written by ctaplot. If I do not understand wrong, DL3 files should be photon lists, what I'm saying is that the plot that is outputted by ctaplot comes together with a file (.txt?) with the points that are plotted (angular/energy resolution) or the histograms
> 
> No. These file will be the input to ctaplot. Not its output.
> 
> DL3 is photon lists AND the IRFs already included. See https://gamma-astro-data-formats.readthedocs.io/en/latest/
> 
> Reconstructed event lists used as input for the IRF calculation would be considered DL2.
> 
> Work as also started on a common package for calculation of IRFs: https://github.com/cta-observatory/pyrf
> 
> ctaplot should read input and plot, not make irf calculations or complicated IO.
> If you are talking about serialization of plots to something other than image formats, this would probably be nice. But the important information is stored in the DL3 files, not in something ctaplot produces.

Sorry catching up with all the weekend exchanges.
That is probably not the right place for this conversation but let's clarify some things.
ctaplot provides metrics, benchmarks and plot for DL0, DL1 and DL2 data.
I think this is what we want here.

The definitive resolution should come at level 3 (DL3), once all cuts have been optimised, with the complete IRFs. pyrf is an attempt to do that (very preliminary) and the output should be DL3 FITS IRFs as defined in the gamma astro data format.


>  I'm saying is that the plot that is outputted by ctaplot comes together with a file (.txt?) with the points that are plotted (angular/energy resolution) or the histograms.

Yes, no problem to have the values and plots (maybe not for all).

> Final question, are we compatible with the latest ctaplot release (v0.5.0)? We should probably follow some released version to avoid breaking of the code in the future.

That I think was the subject of this PR.
If used, stating it as a dependency with a version would definitely help.

Let me know, I can either update this PR or close it.

rlopezcoto:
btw, we agreed in #391 to add `ctaplot` as a dependency, but without any fixed version because it is likely not working with the latest one. I would prefer not to keep it like this in the setup.py because ctaplot is quickly evolving and the compatibility is likely to break.
Ideally, here or in a new PR, the code should be adapted to the latest release and freeze that version.

lstchain/scripts/lstchain_mc_rfperformance.py
```diff
@@ -111,80 +111,33 @@ def main():
...
-    focal_length = 28 * u.m
-    src_pos_reco = utils.reco_source_position_sky(gammas.x.values * u.m,
-                                                  gammas.y.values * u.m,
-                                                  gammas.reco_disp_dx.values * u.m,
-                                                  gammas.reco_disp_dy.values * u.m,
-                                                  focal_length,
-                                                  gammas.mc_alt_tel.values * u.rad,
-                                                  gammas.mc_az_tel.values * u.rad)
-
+    selected_gammas = dl2.query('reco_type==0 & mc_type==0')
```
rlopezcoto, line 114: does this imply you are selecting only events that have `reco_type == 0`?

lstchain/scripts/lstchain_mc_rfperformance.py
```diff
@@ -111,80 +111,33 @@ def main():
 
     ####PLOT SOME RESULTS#####
 
-    gammas = dl2[dl2.gammaness >= 0.5]
-    protons = dl2[dl2.gammaness < 0.5]
-    gammas.reco_type = 0
-    protons.reco_type = 1
-
-    focal_length = 28 * u.m
-    src_pos_reco = utils.reco_source_position_sky(gammas.x.values * u.m,
```
rlopezcoto, line 120: is this now calculated within some of the ctaplot functions? If so, where exactly?

lstchain/scripts/lstchain_mc_rfperformance.py
vuillaut: Yes, this is equivalent to the previous `dl2.gammaness >= 0.5` but ensuring consistent in the threshold applied in the reconstruction

lstchain/scripts/lstchain_mc_rfperformance.py
vuillaut: the `src_pos_reco` you mean?
No, it just does not make sense to compute here anymore, `reco_alt` and `reco_az` are now computed in `apply_models`

lstchain/scripts/lstchain_mc_rfperformance.py
rlopezcoto: I forgot to ask... I guess that within this query we can request different gammaness thresholds, e.g. `dl2.query('dl2.gammaness >= 0.9')`? or how would that call be in this case?

vuillaut: That would be just `'gammaness >= 0.9'` but yes sure. It works like masks.

Issue #285, opened 2020-02-14
Error in dl1 to dl2 step when pointing info is missing


morcuended:
Sometimes UCTS timestamp is missing in the latest runs, hence pointing coordinates cannot be interpolated nor added to the data stream. In these cases, `alt_tel` and `az_tel` are set to `NaN`. Later on, when running `lstchain_mc_dl1_to_dl2.py` over those runs with missing pointing info I get this error:
```
Traceback (most recent call last):
  File "/home/daniel.morcuende/software/cta-lstchain/cta-lstchain/lstchain/scripts/lstchain_mc_dl1_to_dl2.py", line 89, in <module>
    main()
  File "/home/daniel.morcuende/software/cta-lstchain/cta-lstchain/lstchain/scripts/lstchain_mc_dl1_to_dl2.py", line 79, in main
    dl2 = dl1_to_dl2.apply_models(data, cls_gh, reg_energy, reg_disp_vector, custom_config=config)
  File "/fefs/home/daniel.morcuende/software/cta-lstchain/cta-lstchain/lstchain/reco/dl1_to_dl2.py", line 396, in apply_models
    alt_tel = dl2['alt_tel'].values
  File "/home/daniel.morcuende/.conda/envs/mycta/lib/python3.7/site-packages/pandas/core/frame.py", line 2800, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/daniel.morcuende/.conda/envs/mycta/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 2648, in get_loc
    return self._engine.get_loc(self._maybe_cast_indexer(key))
  File "pandas/_libs/index.pyx", line 111, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 138, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 1618, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 1626, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'alt_tel'
```
How can we overcome this? Should we throw away those problematic events in this second step by checking first whether dl2['alt_tel'].values are NaN?
Any suggestion?

morcuended:
Maybe just filter out those events when producing DL1 files?

morcuended:
[...]

PR #286, opened 2020-02-14
alt_tel missing

vuillaut:
Partly fix issue #285 
- avoid issue when alt_tel is not in the dl2 parameters
- load LST focal length from ctapipe (rather than having it hardcoded)


rlopezcoto:
> Partly fix issue #285
Why only partly?

+1 for the usage of the optical description instead of hardcoding values

vuillaut:
Partly because it does not solve the root cause that should be addressed and because this patch might cause other problems down the line with pointings to (0,0). I'll comment in #285 that should be left opened.

PR #291, opened 2020-02-21
set unit of energy to TeV in sensitivity code

SeiyaNozaki:
Now the unit of e_true & e_reco is set to TeV because the unit of mc_energy & reco_energy in DL2 is TeV. Almost all the codes are working well, but I found following the only one part which have a problem.

e_aftercuts_w = np.sum(np.power(e_aftercuts, crab_par['alpha'] - mc_par_g['sp_idx']))
eff_area[i] = e_aftercuts_w.to_value() / n_sim_bin[i] * mc_par_g['area_sim'].to(u.m ** 2).to_value()

e_aftercuts_w is calculated with TeV unit, and n_sim_bin is with GeV.
So here I'd like to propose to set e_true amd e_reco to GeV at first.

### DIFF
--- a/lstchain/mc/sensitivity.py
+++ b/lstchain/mc/sensitivity.py
@@ -330,23 +330,23 @@ def find_best_cuts_sensitivity(simtelfile_gammas, simtelfile_protons,
     mc_par_g['sim_ev'] = mc_par_g['sim_ev'] * nfiles_gammas
     mc_par_p['sim_ev'] = mc_par_p['sim_ev'] * nfiles_protons
 
-    # Pass units to GeV and cm2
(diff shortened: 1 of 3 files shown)


vuillaut:
> So here I'd like to propose to set e_true amd e_reco to GeV at first.

On the contrary, I'd propose to make bins in TeV. This is now the default unit in simtel, ctapipe...

rlopezcoto:
I agree to follow compatibility with simtel, ctapipe and so on. @SeiyaNozaki could you modify the sensitivity code to make it work with `TeV` units everywhere? Thanks

SeiyaNozaki:
thanks for the comments!
I modified the codes to make it work with TeV units.

Issue #294, opened 2020-02-21
Error when reading a simtel file with no defined telescope or camera name


yrenier:
Hello,
I produced simtel files where the name of some of the telescopes, some of the cameras was not defined. When I tried to convert them to dl1 I got several errors such as
`OSError: Table /instrument/telescope/camera/UNKNOWN already exists`

adding in the function lstchain.io.io.write_array_info() the following lines before the call to camera.to_table().write() close to the line 360 fixes the issue. 
```
if str(camera) is 'UNKNOWN':
    continue
```

rlopezcoto:
Hi @yrenier, thanks for reporting! Was this error appearing the first time that you run your script? This looks like an overwriting error. In any case, let us have a look and maybe you could open a PR working it out as you proposed.

yrenier:
The problems comes that it writes the info of the first camera in /instrument/telescope/camera/UNKNOWN
As the second camera has also no name defined, it tries to put the info in the same place and crashes.

rlopezcoto:
Solved in #296

PR #296, opened 2020-02-21
Error when reading a simtel file with no defined telescope or camera name

yrenier:
Hello,
I produced simtel files where the name of some of the telescopes, some of the cameras was not defined. When I tried to convert them to dl1 I the following error:
`OSError: Table /instrument/telescope/camera/UNKNOWN already exists`
This PR fixes the issue (which was reported here): https://github.com/cta-observatory/cta-lstchain/issues/294
I also raised an explicit exception when trying to create an DL1 file which already exist. Otherwise it would fail later with some obscure error similar to the one above.

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -361,7 +361,7 @@ def write_array_info(event, output_filename):
         if len(ids) > 0:  # only write if there is a telescope with this camera
             tel_id = list(ids)[0]
             camera = sub.tel[tel_id].camera
-            camera_name = str(sub.tel[tel_id])
+            camera_name = str(camera)
             with tables.open_file(output_filename, mode='r') as f:
                 telescope_chidren = f.root['instrument/telescope']._v_children.keys()
                 if 'camera' in telescope_chidren:
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -361,9 +361,17 @@ def write_array_info(event, output_filename):
         if len(ids) > 0:  # only write if there is a telescope with this camera
             tel_id = list(ids)[0]
             camera = sub.tel[tel_id].camera
-            camera_name = str(camera)
-            if camera_name is 'UNKNOWN':
-                continue
+            camera_name = str(sub.tel[tel_id])
+            with tables.open_file(output_filename, mode='r') as f:
+                telescope_chidren = f.root['instrument/telescope']._v_children.keys()
+                if 'camera' in telescope_chidren:
+                    cameras_name = f.root['instrument/telescope/camera']._v_children.keys()
+                    if camera_name in cameras_name:
+                        print(
+                            f'WARNING during lstchain.io.write_array_info():',
+                            f'camera {camera_name} seems to be already present in the h5 file.'
+                        )
+                        continue
             camera.to_table().write(
                 output_filename,
                 path=f'/instrument/telescope/camera/{camera_name}',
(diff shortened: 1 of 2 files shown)


lstchain/io/io.py
```diff
@@ -361,9 +361,12 @@ def write_array_info(event, output_filename):
         if len(ids) > 0:  # only write if there is a telescope with this camera
             tel_id = list(ids)[0]
             camera = sub.tel[tel_id].camera
+            camera_name = str(camera)
+            if camera_name is 'UNKNOWN':
```
rlopezcoto: with this line, what you are doing is _not_ to write any of the camera information into the file. I guess this is ok because when the camera name is UNKNOWN, this information does not exist, right?

yrenier: exactly, that happens when ctapipe does not recognize the telescope.

yrenier:
I just merged with master. 
I also modified the code so instead of just checking for 'UNKNOWN' telescopes, it looks if a camera with that name exists already in the hdf5 file.

lstchain/io/io.py
kosack: since SimTel files don't contain camera names, there is just a lookup table in ctapipe, which returns UNKNOWN if no matches are found.  Perhaps you shold open an issue for that - we could for example expose an API to allow you to add your own in the case they are not defined.  In fact, right now, you can just modify `ctapipe.instrument.guess.TELESCOPE_NAMES` inside your code, before you try to load the SimTel file. Just add an entry for your new telescope.  The GuessingKey is just a tuple of `(num_pixels, focal_length)`

kosack: To be clearer, just add the line:
```py3
from ctapipe.instrument.guess import TELESCOPE_NAMES, GuessingKey, GuessingResult

TELESCOPE_NAMES[GuessingKey(2048, 2.28)] =  GuessingResult("LST", "Custom", "MyCamera", 2)
```

(filling in the correct number of pix and focal length), and then you will get a telescope named "LST_Custom_MyCamera".

yrenier: Thank you for the trick Karl, I had to modify ctapipe code to get the same result. 
Defining the telescope works well but on the LST + Magic production the 2 magic telescopes are slightly different but share the same focal length and the same number of pixel and so trigger the same bug.
The latest commits I pushed handle exactly that.

moralejo: > Defining the telescope works well but on the LST + Magic production the 2 magic telescopes are slightly different but share the same focal length and the same number of pixel and so trigger the same bug.
> The latest commits I pushed handle exactly that.

@yrenier I think camera-geometry wise the two telescopes should be identical - they have slightly different characteristicas in other respects (is that what you meant?), but that should not affect the camera geometry selection.

kosack: The purpose of the TelescopeDescription is really to describe "classes" of telescope, so if the MAGIC telescopes have similar mirrors, cameras, and focal length, I would still call them both "MAGIC", and then you get 2 tel-ids, (1 and 2) that let you distiguish between the two telescopes.   For the combined LST-Magic array, we might need to define what tels get which id number (e..g [1,2] = MAGIC, and 3=LST?), but other than that I think the reconstruciton should still work fine.

yrenier: what I noticed was that the iteration 
`for telescope_type in sub.telescope_types:`
was iterating on each of the the MAGIC telescope. 
If I add the MAGIC telescope as suggested by @kosack both get the same name.
If i don't add them they get the same name: UNKNONW
Either way the  write_array_info() function crash when trying to write the 2nd camera with  camera.to_table().write() as the /instrument/telescope/camera/{camera_name} key already exists.
I can provide a small sim_tearrayl file so you can reproduce the problem if you want.

Issue #293, opened 2020-02-21
Bug in image extractor handling


rlopezcoto:
Hi all,
I open this issue for discussion, I will open a separate PR for the fixing. @FrancaCassol, @moralejo and I have been investigating the light collection efficiency derived using muon analysis. It seems that we have found a bug on how the selected image extractor is passed to the analysis chain. At the moment, the image extractor is defined in:
https://github.com/cta-observatory/cta-lstchain/blob/921c5cc65d4577a08fa4db7bbe197705ae42bdbb/lstchain/data/lstchain_standard_config.json#L100

the script of data analysis calls:
https://github.com/cta-observatory/cta-lstchain/blob/921c5cc65d4577a08fa4db7bbe197705ae42bdbb/lstchain/scripts/lstchain_data_r0_to_dl1.py#L67

that calls the `LSTCameraCalibrator`
https://github.com/cta-observatory/cta-lstchain/blob/921c5cc65d4577a08fa4db7bbe197705ae42bdbb/lstchain/reco/dl0_to_dl1.py#L192

here we pass the variable `image_extractor`
https://github.com/cta-observatory/cta-lstchain/blob/921c5cc65d4577a08fa4db7bbe197705ae42bdbb/lstchain/reco/dl0_to_dl1.py#L194

but the one `LSTCameraCalibrator` expects is `extractor_product`:
https://github.com/cta-observatory/cta-lstchain/blob/921c5cc65d4577a08fa4db7bbe197705ae42bdbb/lstchain/calib/camera/calibrator.py#L20

The fix is just to change `image_extractor` -> `extractor_product` in https://github.com/cta-observatory/cta-lstchain/blob/921c5cc65d4577a08fa4db7bbe197705ae42bdbb/lstchain/reco/dl0_to_dl1.py#L194

I'm currently running tests,  I'll make the PR afterwards. Any opinions?

### Consequences:
- All the *real data* that has been processed up to now are suffering from this bug, so any signal extractor that you have been using was not passed to the analysis. I checked this using the data processed by @morcuended using different signal extractors and results are identical.
- The data was processed using the default image extractor defined in `extractor_product`: `NeighborPeakWindowSum`, with its default initialization ( window_shift: 3, window_width: 7). This is specially bad because of time misalignment in neighbor modules (@FrancaCassol can comment further on that).

kosack:
[...]

PR #298, opened 2020-02-27
Change default integrator

rlopezcoto:
To avoid problems as #293, in spite that we can now select a signal extractor thanks to #297, I would like to change the default signal extractor in the LSTCameraCalibrator. The reason is that the `NeighborPeakWindowSum` has not been working very well for real data, so I would rather have as a default `LocalPeakWindowSum` in case we do not set any in the configuration file.
L.21 is the main change, the rest is only code cleanup.


rlopezcoto:
thanks @vuillaut! I just wanted to reduce to the minimum the bug-possibilities as it happened in the past. In any case and as soon as we have a new ctapipe release, we should try refactoring some of the scripts we have at the moment, as suggested in #293 by @kosack

PR #304, opened 2020-03-10
Speeding-up of muon ring analysis, and incorporation into the standard r0 to dl1 analysis script

moralejo:
Problematic pixels, as read from a calibration file provided by the user, are now zeroed. TBD: interpolation would be better!


lstchain/scripts/lstchain_data_muon_analysis_dl1.py
```diff
@@ -13,6 +13,7 @@
 
 from lstchain.image.muon import analyze_muon_event, muon_filter, tag_pix_thr
 from lstchain.io.io import dl1_params_lstcam_key, dl1_images_lstcam_key
+from lstchain.visualization import plot_calib as calib
```
vuillaut: Since there is a module called `calib` in lstchain, I would recommend removing the `as calib` part and use `plot_calib` as is to avoid confusion

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
```diff
@@ -69,7 +74,12 @@ def main():
                          'impact_x_array': [],
                          'impact_y_array': [],
                          }
-    
+
+    calib.read_file(args.calib_file)
+    bad_pixels = calib.calib_data.unusable_pixels[0]
+    print("Found a total of", np.sum(bad_pixels), "pixels.")
+    good_pixels = np.logical_not(bad_pixels)
```
vuillaut: Just as a minor comment (can be ignored):
- you can write `print(f"Found a total of {np.sum(bad_pixels)} pixels.")`
- `good_pixels = ~bad_pixels` is equivalent. This way you can avoid defining `good_pixels` and later just do `image = full_image*(~bad_pixels)`

moralejo:
> Thanks @moralejo!
> With this PR you are making compulsory the usage of a calibration file for the muon analysis. I'm nor sure if this is the best choice, or we could simply put it as an option.

Well I introduced it out of necessity after seeing how badly spoiled muon ring fits (about all of them) were by a group of just three bad neighboring pixels. Certainly treatment of those can be improved, but simply ignoring them should not be an option I think.

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
moralejo: Done

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
moralejo: Done, thx

moralejo:
I have introduced a few more changes:
- added two more columns in the output to understand better data - MC differences
- Cleaned up a bit the "good muon ring" selection
- Now the detailed (and very time consuming) minuit fit is only done for good candidates

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
```diff
@@ -43,13 +48,19 @@
 parser.add_argument("--plots_path", help = "Path to the plots",
                     default = None, type = str)
 
+parser.add_argument("--max_muons", help = "Maximum number of processed muon ring candidates",
```
rlopezcoto: I do not think we would like to set as a default `max_muons = -1`

lstchain/image/muon/muon_analysis.py
```diff
@@ -87,19 +89,15 @@ def fit_muon(x, y, image, geom, tailcuts):
...
-    ring_dist = np.abs(dist - muonringparam.ring_radius)
-    muonringparam = muonring.fit(
-            x, y, image_clean * (ring_dist < muonringparam.ring_radius * 0.4)
-        )
-
-    dist = np.sqrt(np.power(x - muonringparam.ring_center_x, 2) +
-                       np.power(y - muonringparam.ring_center_y, 2))
-    ring_dist = np.abs(dist - muonringparam.ring_radius)
+    # Do an iterative fit removing pixels which are beyond 0.4*ring_radius of the ring (along the radial direction):
+    # (Q: is this effective? Seems anyway unnecessary for clean rings which are those we want in the end)
```
rlopezcoto: this was done for the fit to converge

lstchain/image/muon/muon_analysis.py
```diff
@@ -87,19 +89,15 @@ def fit_muon(x, y, image, geom, tailcuts):
...
 
-    muonringparam = muonring.fit(
-            x, y, image_clean * (ring_dist < muonringparam.ring_radius * 0.4)
+    for _ in [0]*2:  # just to iterate the fit twice more
+        dist = np.sqrt(np.power(x - muonringparam.ring_center_x, 2)
+                       + np.power(y - muonringparam.ring_center_y, 2))
+        ring_dist = np.abs(dist - muonringparam.ring_radius)
+        muonringparam = muonring.fit(
+        x, y, image_clean * (ring_dist < muonringparam.ring_radius * 0.4)
```
rlopezcoto: as we've been discussing, the 40% of the ring_radius may be a bit too much, specially in the case of using biased integrators. Maybe it is the time to change it?
One of your proposals was to relate it to the `ring_width`

rlopezcoto: Also, in L.142 you define the `ring_integration_width`. If you change it later there, we will forget that here it was changed

lstchain/image/muon/muon_analysis.py
```diff
@@ -131,90 +129,135 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
+    min_pix = 148                              # (8%) minimum number of pixels in the ring with >0 signal
+    min_pix_fraction_after_cleaning = 0.1      # minimum fraction of the ring pixels that must be above tailcuts[0] 
+    min_ring_radius = 0.8*u.deg                # minimum ring radius
+    max_ring_radius = 1.5*u.deg                # maximum ring radius 
+    max_radial_stdev = 0.1*u.deg               # maximum standard deviation of the light distribution along ring radius
+    max_radial_excess_kurtosis = 1.            # maximum excess kurtosis
+    min_impact_parameter = 0.2                 # in fraction of mirror radius
+    max_impact_parameter = 0.9                 # in fraction of mirror radius
+    ring_integration_width = 0.4               # +/- integration range along ring radius, in fraction of ring radius
+    outer_ring_width = 0.2                     # in fraction of ring radius, width of ring just outside the integrated muon ring, used to check pedestal bias
```
rlopezcoto: is this condition implying:
ring_width < ring_radius * outer_ring_width ?

lstchain/image/muon/muon_analysis.py
```diff
@@ -131,90 +129,135 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
-                    pix_ringwidth_im[idx_ringwidth], tailcuts[0]) / len(
-                    pix_ring[idx_ringwidth])
-
-    print("Impact parameter = %s"
-                             "ring_width=%s, ring radius=%s, ring completeness=%s"% (
-                             muonintensityoutput.impact_parameter,
-                             muonintensityoutput.ring_width,
-                             muonringparam.ring_radius,
-                             muonintensityoutput.ring_completeness))
+    size_outside_ring = np.sum(pix_outside_ring * clean_mask)
```
rlopezcoto: be careful here, because
`ring_size ` is calculated _without_ applying cleaning
`size_outside_ring ` is calculated applying cleaning

therefore the total 

Also, I suggest we move `size` -> `intensity` to be compatible with the rest of the naming. `ring_size` should also be renamed, but this is a ctapipe inherited variable...

lstchain/image/muon/muon_analysis.py
```diff
@@ -291,3 +340,44 @@ def tag_pix_thr(image, thr_low = 50, thr_up = 500, pe_thr = 10):
...
+    """
+    # Convert everything to degrees:
+    x0 = centre_x.to(u.deg).value
+    y0 = centre_y.to(u.deg).value
+    pix_x = pixel_x.to(u.deg).value
+    pix_y = pixel_y.to(u.deg).value
+
+    pix_r = np.hypot(pix_x-x0, pix_y-y0)
+
+    # meanm, standard deviation & skewness of light distribution along ring radius.
```
rlopezcoto: meanm -> mean

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
```diff
@@ -43,13 +48,19 @@
...
 
-    print("input file: {}".format(args.input_file))
+    print("input files: {}".format(args.input_file))
+    print("calib file: {}".format(args.calib_file))
     print("output file: {}".format(args.output_file))
+    
+    max_muons = args.max_muons
 
     # Camera geometry
     geom = CameraGeometry.from_name("LSTCam-002")
```
rlopezcoto: geometry needs to be updated, could you do it here, please?

lstchain/image/muon/muon_analysis.py
moralejo: > this was done for the fit to converge

Yes, I was just wondering if it is needed for the good rings, and effective for the shower-contaminated ones. I think it will work at least for cases in which there are very few bad pixels, so fine for me to keep it. I modified the comment.

lstchain/image/muon/muon_analysis.py
moralejo: > as we've been discussing, the 40% of the ring_radius may be a bit too much, specially in the case of using biased integrators. Maybe it is the time to change it?

I would leave 40% here, because this is just to identify really far outliers.

> One of your proposals was to relate it to the `ring_width`

Now I am not sure that is the best for light integration, because then it will be somewhat dependent on the PSF. I would simply reduce it. I am setting it to 25% as default now.

lstchain/image/muon/muon_analysis.py
rlopezcoto: then I do not understand the `outer` part of the name

lstchain/image/muon/muon_analysis.py
rlopezcoto: sorry, the complete answer was:
"therefore the total size of the _cleaned_ or _uncleaned_ image will not be equal to `ring_size + size_outside_ring"

ok for the `size` renaming

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
rlopezcoto: `LSTCam-004`
from 002 to 003, there was a wrong camera rotation:
https://github.com/cta-observatory/ctapipe-extra/pull/25

from 003 to 004 there was only a small pixel rotation:
https://github.com/cta-observatory/ctapipe_io_lst/pull/23

moralejo: > `LSTCam-004`
> from 002 to 003, there was a wrong camera rotation:
> [cta-observatory/ctapipe-extra#25](https://github.com/cta-observatory/ctapipe-extra/pull/25)
> 
> from 003 to 004 there was only a small wrong pixel rotation:
> [cta-observatory/ctapipe_io_lst#23](https://github.com/cta-observatory/ctapipe_io_lst/pull/23)

Ok, done now

lstchain/image/muon/muon_analysis.py
moralejo: > sorry, the complete answer was:
> "therefore the total size of the _cleaned_ or _uncleaned_ image will not be equal to `ring_size + size_outside_ring"
> 
Ok, that is clear, but I see no problem with it.

lstchain/image/muon/muon_analysis.py
moralejo: > then I do not understand the `outer` part of the name

"outer" refers to the fact that this is a ring _just outside_ the muon ring in which the muon-related charge is integrated.

lstchain/reco/dl0_to_dl1.py
```diff
@@ -182,13 +188,24 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
...
+    
+    # Camera geometry TBD: read from config file?
+    geom = load_camera_geometry()
 
+    # Dictionary to store muon ring parameters
+    muon_parameters  = create_muon_table()
+
+    plot_calib.read_file(calibration_path)
+    bad_pixels = plot_calib.calib_data.unusable_pixels[0]
+    print(f"Found a total of {np.sum(bad_pixels)} bad pixels.")
```
FrancaCassol: Hi @moralejo,
you do not need to read the calibration file. You find the unusable pixels in each event monitoring container:
event.mon.tel[telid].calibration.unusable_pixels

Also, bad pixels are now getting gain=0, so you do not need anymore to  correct for them

FrancaCassol: Sorry, I realise here you are at DL1 level, my first comment is not correct. You do not have the monitoring data here.
But the second comment is correct, you should not more need it...

lstchain/image/muon/muon_analysis.py
MaxNoe: Are hard coded cuts inline in the code really a good idea?

lstchain/image/muon/muon_analysis.py
```diff
@@ -131,90 +131,135 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
+            threshold=30,
+            bins=30)
+
+        pix_ringwidth_im = image * dist_ringwidth_mask
+        idx_ringwidth = np.nonzero(pix_ringwidth_im)
+        muonintensityoutput.ring_pix_completeness = npix_above_threshold(
+            pix_ringwidth_im[idx_ringwidth], tailcuts[0]) / len(
+                pix_ring[idx_ringwidth])
+
+    else:
```
MaxNoe, line 219: Indentation is a level off here

lstchain/image/muon/muon_analysis.py
```diff
@@ -131,90 +131,135 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
-                             muonintensityoutput.impact_parameter,
-                             muonintensityoutput.ring_width,
-                             muonringparam.ring_radius,
-                             muonintensityoutput.ring_completeness))
+    size_outside_ring = np.sum(pix_outside_ring * clean_mask)
 
+    # This is just mean charge per pixel in pixels just around the ring (on the outer side):
+    mean_pixel_charge_around_ring = np.sum(pix_ring_2)/len(pix_ring_2) 
+
+    if candidate_clean_ring:
```
MaxNoe, line 237: print should not be used here. Either logging or print somewhere else.

lstchain/image/muon/muon_analysis.py
```diff
@@ -131,90 +131,135 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
+        print("Impact parameter=%s, ring_width=%s, ring radius=%s, ring completeness=%s"%
```
MaxNoe, line 238: This is a very strange mixture of the very old python string formatting and the newer method. Please replace with:

```
print("Impact parameter={:.3f}, ring_width={:.3f}, ring radius={:.3f}, ring completeness={:.3f}".format(
   muonintensityoutput.impact_parameter, muonintensityoutput.ring_width, 
   muonringparam.ring_radius, muonintensityoutput.ring_completeness,
))
```

lstchain/image/muon/muon_analysis.py
```diff
@@ -247,12 +292,18 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
         plot_muon_event(ax, geom, image * clean_mask, centroid, ringrad_camcoord, 
                         ringrad_inner, ringrad_outer, event_id)
 
+        plt.figtext(0.15, 0.20, 'radial std dev: {0:.3f}'.format(radial_distribution['standard_dev']))
```
MaxNoe, line 295: should this additional code go directlly into the plotting function?

lstchain/image/muon/muon_analysis.py
```diff
@@ -291,3 +342,98 @@ def tag_pix_thr(image, thr_low = 50, thr_up = 500, pe_thr = 10):
     """
 
     return ((np.size(image[image > pe_thr]) < thr_up) and (np.size(image[image > pe_thr]) > thr_low))
+
+
+def radial_light_distribution(centre_x, centre_y, pixel_x, pixel_y, image):
```
MaxNoe, line 347: mixture of british and american english. code should be american english

lstchain/image/muon/muon_analysis.py
```diff
@@ -291,3 +342,98 @@ def tag_pix_thr(image, thr_low = 50, thr_up = 500, pe_thr = 10):
...
+        Y position of pixel in image
+    image: ndarray
+        Amplitude of image pixels
+
+    Returns
+    -------
+    standard_dev, skewness
+    """
+    # Convert everything to degrees:
+    x0 = centre_x.to(u.deg).value
```
MaxNoe, line 368: Use `to_value(u.deg)` instead of `to(u.deg).value`. This avoids a copy if the input is already in deg`

lstchain/image/muon/muon_analysis.py
```diff
@@ -291,3 +342,98 @@ def tag_pix_thr(image, thr_low = 50, thr_up = 500, pe_thr = 10):
...
+    y0 = centre_y.to(u.deg).value
+    pix_x = pixel_x.to(u.deg).value
+    pix_y = pixel_y.to(u.deg).value
+
+    pix_r = np.hypot(pix_x-x0, pix_y-y0)
```
MaxNoe, line 373: hypot is pretty slow as it does all sorts of edge case detection for very small  float values. If you don't expect this to happen, just use `np.sqrt((pix_x - x0)**2 + (pix_y - y0)**2)`

lstchain/image/muon/muon_analysis.py
```diff
@@ -291,3 +342,98 @@ def tag_pix_thr(image, thr_low = 50, thr_up = 500, pe_thr = 10):
...
+
+    # mean, standard deviation & skewness of light distribution along ring radius.
+    # ring_radius calculated elsewhere is approximately equal to "mean", but not exactly, so we recalculate it here:
+    mean = np.sum(np.multiply(image,pix_r))/np.sum(image)
```
MaxNoe, line 377: use numpy average: `np.average(pix_r, weights=image)`, see e.g. the hillas parameter implementation in ctapipe:
https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/image/hillas.py

lstchain/image/muon/muon_analysis.py
```diff
@@ -291,3 +342,98 @@ def tag_pix_thr(image, thr_low = 50, thr_up = 500, pe_thr = 10):
...
+    standard_dev = np.sqrt(np.sum(np.multiply(image,np.square(pix_r-mean)))/np.sum(image))
```
MaxNoe, line 378: Same as above. numpy and scipy provide weighted averages and standard deviations. Which will both be more readable and faster than this.

MaxNoe:
@moralejo if you ok with this, I'll push to this branch implementing some of the suggestions I have made

MaxNoe:
As a side node: it's better not to use the master branch of a fork to implement changes. Always uses feature branches like this:

```
$ git fetch upstream  # update from the cta-observatory repo
# create a new branch starting at the current cta-observatory master
$ git checkout -b new_branch_name upstream/master  
# do your work...
```

Your forked master branch should be up do date with the upstream master. You can do this using
```
$ git fetch upstream
$ git checkout master
# ALERT: danger: will remove all changes you made to your local master branch
$ git reset --hard upstream/master
```

lstchain/reco/dl0_to_dl1.py
```diff
@@ -182,14 +188,28 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
...
 
     metadata = global_metadata(source)
     write_metadata(metadata, output_filename)
 
-    cal = load_calibrator_from_config(config)
+    cal_mc = load_calibrator_from_config(config)
+
+    
+    # Camera geometry TBD: read from config file?
+    geom = load_camera_geometry()
```
vuillaut, line 201: The camera geometry is already part of each event
`geom = event.inst.subarray.tel[tel_id].camera`

lstchain/scripts/lstchain_data_r0_to_dl1.py
```diff
@@ -9,8 +9,8 @@
 
 parser.add_argument('--infile', '-f', type=str,
                     dest='infile',
-                    help='path to the file with simtelarray events',
-                    default=get_dataset_path('gamma_test_large.simtel.gz'))
+                    help='path to the .fits.fz file with the raw events',
+                    default=None, required=True)
```
vuillaut, line 13: well spoted

lstchain/scripts/lstchain_data_r0_to_dl1.py
```diff
@@ -47,6 +47,12 @@
                     default=None
                     )
 
+parser.add_argument('--max_events', '-maxevts', action='store', type=int,
+                    dest='max_events',
+                    help='Maximum number of events to be processed.',
+                    default=int(1e15)
+                    )
```
vuillaut, line 54: `max_events`  is already an entry in the config file, is it really necessary to add another argument to the script?

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
```diff
@@ -99,7 +100,10 @@ def main():
             mirror_area = telescope_description['mirror_area'].values * pow(u.m,2)
             
             for full_image, event_id, dragon_time in zip(images, parameters['event_id'], parameters['dragon_time']):
-                image = full_image*(~bad_pixels)
+                if args.calib_file is not None:
+                    image = full_image*(~bad_pixels)
```
MaxNoe, line 104: `image[bad_pixels] = 0`

lstchain/reco/dl0_to_dl1.py
moralejo: But is it there already for raw data? When is it introduced?

vuillaut: Sure.
It is the reader's responsibility to load such information.

MaxNoe: The lst event source is doing it.

moralejo: > The lst event source is doing it.

Ok, thx. But how does it work when reading a dl1 file? In the dl1 there is the camera description as a table, how do I read it in a geom object?

By the way, @vuillaut , could that be why lstchain_mc_dl1ab.py had issues? It is using LSTCam-002 hardcoded, perhaps it is not the right one?

MaxNoe: If the order of the pixels did not change, this should not be a problem. I think the only differences were the coordinate frames between 2 → 3 and the pixel rotation in 3 → 4. Coordinate frames would be important for event reconstructoin, pixel rotation is only important for displaying events at the moment.

moralejo: > If the order of the pixels did not change, this should not be a problem. I think the only differences were the coordinate frames between 2 → 3 and the pixel rotation in 3 → 4. Coordinate frames would be important for event reconstructoin, pixel rotation is only important for displaying events at the moment.

True, but I am now dealing with some (muon) MC in which there seems to be a different ordering of pixels w.r.t. LST-002 to LST-004. So I was wondering if that could be it.

moralejo: > > The lst event source is doing it.
> 
> Ok, thx. But how does it work when reading a dl1 file? In the dl1 there is the camera description as a table, how do I read it in a geom object?
> 
> By the way, @vuillaut , could that be why lstchain_mc_dl1ab.py had issues? It is using LSTCam-002 hardcoded, perhaps it is not the right one?

Ok, I figured out how to do it, should have checked before asking!

moralejo:
> As a general comment, I see you are introducing the muon analysis in `dl0_to_dl1`.
> How much speed-up improvement are we talking about here? I am concerned about blowing up the analysis time of an already way too slow analysis...

There are two different things: 
- speeding up the muon ring analysis: I did this just by avoiding to do the expensive Minuit fit for all but the most promising rings. The speed-up, when starting from DL1, was a factor ~10 (and the minuit fit was anyway useless for all those bad rings). Further speeding-up is possible via the ctapipe code, I think one can remove all astropy.units stuff from the low-level functions passed to Minuit, and also eliminate a lot of function nesting in the calculation of the predicted muon image.
- avoiding to do again the whole r0->dl1 step with a different pulse integrator for all events, just to get the few interesting rings. As of now, we have to re-calibrate and re-parametrize all events, also doubling the dl1 data volume, just because of the rings. 
- By having the ring analysis integrated in the r0->dl1 step we can include more easily data checks based on the muons. This was actually the main trigger for this PR.

Ok, that was _three_ things.

lstchain/image/muon/muon_analysis.py
moralejo: > Same as above. numpy and scipy provide weighted averages and standard deviations. Which will both be more readable and faster than this.

First thing I intended to do, but failed to find it. Can you point me to those functions?

moralejo: > > Same as above. numpy and scipy provide weighted averages and standard deviations. Which will both be more readable and faster than this.
> 
> First thing I intended to do, but failed to find it. Can you point me to those functions?

ok, saw your previous comment now.

MaxNoe: I already implemented this in the PR I opened in your fork, I sent you the link by mail

MaxNoe: https://github.com/moralejo/cta-lstchain/pull/1

lstchain/image/muon/muon_analysis.py
moralejo: > use numpy average: `np.average(pix_r, weights=image)`, see e.g. the hillas parameter implementation in ctapipe:
> https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/image/hillas.py

Ok, using numpy.average in all cases (use of cov does not seem clearer to me)

lstchain/image/muon/muon_analysis.py
moralejo: > Are hard coded cuts inline in the code really a good idea?

Certainly to be improved... but I'd prefer to close this first (hard-coded cuts were there already,  I just added some, and tried to place them together for clarity)

lstchain/reco/dl0_to_dl1.py
moralejo: > Hi @moralejo,
> you do not need to read the calibration file. You find the unusable pixels in each event monitoring container:
> event.mon.tel[telid].calibration.unusable_pixels

ok, good, thanks!

> Also, bad pixels are now getting gain=0, so you do not need anymore to correct for them

Ok, unless one does the muon analysis with pre-existing DL1 files in which this was not yet the case. I guess I can keep it with a warning to eventually remove it.

lstchain/image/muon/muon_analysis.py
moralejo: I agree, but the use of print here predates this PR, I'd prefer to address this at a later point.

lstchain/image/muon/muon_analysis.py
moralejo: Ok, thanks, done.

lstchain/image/muon/muon_analysis.py
moralejo: Not sure it is nicer, it means passing three additional arguments.

lstchain/image/muon/muon_analysis.py
moralejo: Done

lstchain/image/muon/muon_analysis.py
moralejo: Done, thx

lstchain/image/muon/muon_analysis.py
moralejo: Done

lstchain/image/muon/muon_analysis.py
moralejo: Done

lstchain/image/muon/muon_analysis.py
moralejo: Thx, done

moralejo:
> @moralejo if you ok with this, I'll push to this branch implementing some of the suggestions I have made

Saw this in the wrong order, I had introduced nearly all of those changes, so I'll not merge, ok?

lstchain/scripts/lstchain_data_muon_analysis_dl1.py
moralejo: > `image[bad_pixels] = 0`

It is not the same... and I would have to take care later that the pixel coordinates' arrays are also masked, passing the bad pixels to other functions. I prefer not to take the risk now.

moralejo:
> As a side note: it's better not to use the master branch of a fork to implement changes. Always uses feature branches like this:
> 
> ```
> $ git fetch upstream  # update from the cta-observatory repo
> # create a new branch starting at the current cta-observatory master
> $ git checkout -b new_branch_name upstream/master  
> # do your work...
> ```
> 
> To move theses changes into their own branch:
> 
> ```
> # create new branch `muon_improvements` from the current master
> $ git branch muon_improvements master
> ```
> 
> Your forked master branch should be up do date with the upstream master. You can do this using
> 
> ```
> $ git fetch upstream
> $ git checkout master
> # ALERT: danger: will remove all changes you made to your local master branch
> $ git reset --hard upstream/master
> ```

Thanks for the detailed tips!  However I am very scared of messing things up with git (I did before), so I'll use a branch next time - sorry"

lstchain/image/muon/muon_analysis.py
```diff
@@ -212,9 +212,8 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
 
         pix_ringwidth_im = image * dist_ringwidth_mask
         idx_ringwidth = np.nonzero(pix_ringwidth_im)
```
MaxNoe, line 214: Why are indices needed here? The mask could be directly used and that would be faster?

lstchain/image/muon/muon_analysis.py
```diff
@@ -365,23 +362,21 @@ def radial_light_distribution(centre_x, centre_y, pixel_x, pixel_y, image):
...
     # mean, standard deviation & skewness of light distribution along ring radius.
     # ring_radius calculated elsewhere is approximately equal to "mean", but not exactly, so we recalculate it here:
-    mean = np.sum(np.multiply(image,pix_r))/np.sum(image)
-    standard_dev = np.sqrt(np.sum(np.multiply(image,np.square(pix_r-mean)))/np.sum(image))
-    skewness =  np.sum(np.multiply(image,np.power((pix_r-mean)/standard_dev,3.)))/np.sum(image)
-    excess_kurtosis = np.sum(np.multiply(image,np.power((pix_r-mean)/standard_dev,4.)))/np.sum(image) - 3.
+    mean = np.average(pix_r, weights=image)
+    delta_r = pix_r - mean
+    standard_dev = np.sqrt(np.average(delta_r**2, weights=image))
+    skewness = np.average((delta_r/standard_dev)**3, weights=image)
```
MaxNoe, line 377: Dividing by standard dev after averaging will divide a number by a number instead of an array by a number, improving speed

lstchain/image/muon/muon_analysis.py
```diff
@@ -87,19 +91,15 @@ def fit_muon(x, y, image, geom, tailcuts):
...
-    dist = np.sqrt(np.power(x - muonringparam.ring_center_x, 2) +
-                       np.power(y - muonringparam.ring_center_y, 2))
-    ring_dist = np.abs(dist - muonringparam.ring_radius)
+    # Do an iterative fit removing pixels which are beyond 0.4*ring_radius of the ring (along the radial direction):
+    # The goal is to improve fit for good rings with very few additional non-ring bright pixels.
 
-    muonringparam = muonring.fit(
-            x, y, image_clean * (ring_dist < muonringparam.ring_radius * 0.4)
+    for _ in [0]*2:  # just to iterate the fit twice more
+        dist = np.sqrt(np.power(x - muonringparam.ring_center_x, 2)
```
MaxNoe, line 98: use the power operator, not the power function: `np.sqrt((x - c_x)**2 + (y - c_y)**2)`, see discussion in https://github.com/cta-observatory/ctapipe/issues/1028

lstchain/image/muon/muon_analysis.py
MaxNoe: ok

lstchain/image/muon/muon_analysis.py
```diff
@@ -131,90 +131,132 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
+            bins=30)
+
+        pix_ringwidth_im = image * dist_ringwidth_mask
+        idx_ringwidth = np.nonzero(pix_ringwidth_im)
+        muonintensityoutput.ring_pix_completeness = npix_above_threshold(pix_ringwidth_im[idx_ringwidth], tailcuts[0]) / len(
+            pix_ring[idx_ringwidth])
+
+    else:
+            muonintensityoutput = MuonIntensityParameter()
+            # Set default values for cases in which the muon intensity fit (with MuonLineIntegrate) is not done:
```
MaxNoe, line 220: Passing the container fields as keyword arguments to the constructor is faster, as the container won't initialize defaults and then override them.

However, I will also open a PR in ctapipe, that the default values are actually `np.nan`

lstchain/image/muon/muon_analysis.py
moralejo: Done, thx

lstchain/image/muon/muon_analysis.py
moralejo: Indeed, done.

lstchain/image/muon/muon_analysis.py
moralejo: Right, replaced by:

> pix_ringwidth_im = image[dist_ringwidth_mask]
> muonintensityoutput.ring_pix_completeness =  npix_above_threshold(pix_ringwidth_im, tailcuts[0]) / len(pix_ringwidth_im)

MaxNoe:
@moralejo you need to resolve conflicts,

```
$ git fetch upstream
$ git merge upstream/master
```

lstchain/scripts/lstchain_data_r0_to_dl1.py
moralejo: Not absolutely necessary, but I find it easier for tests to have it also as a command-line option. I guess it does not hurt...

moralejo:
> @moralejo you need to resolve conflicts,
> 
> ```
> $ git fetch upstream
> $ git merge upstream/master
> ```

Done

moralejo:
Please do not merge this yet.
After solving the conflict and merging other changes from the master I get some errors that were not there before...

MaxNoe:
You mean more than the test errors due to #318 ?

moralejo:
> You mean more than the test errors due to #318 ?

Yes, nothing to do with that. I noticed slight changes in muon ring parameters after the merge. I am now looking into it. Besides noticing that I was not using the extractor window width & shift  I intended for the muon ring analysis, I see a strange behaviour: the extracted signals seem to be different for a given event, depending on whether or not I skip all events before it. Any idea whether that is possible? How can that be "history-dependent"?
Is it ok to re-extract the pixel signals for an event which has already been processed with a different extractor?

moralejo:
> Yes, nothing to do with that. I noticed slight changes in muon ring parameters after the merge. I am now looking into it. Besides noticing that I was not using the extractor window width & shift I intended for the muon ring analysis, I see a strange behaviour: the extracted signals seem to be different for a given event, depending on whether or not I skip all events before it. Any idea whether that is possible? How can that be "history-dependent"?

Stupid me... of course, these are data for which time-lapse pedestal correction is applied offline, so it indeed matters whether one skips events. The slight changes I saw in muon ring parameters may come from anywhere, I'll just check that everything looks still healthy.

moralejo:
>  The slight changes I saw in muon ring parameters may come from anywhere, [...]

Most likely #307, the change of time-lapse pedestal correction curve

MaxNoe:
@moralejo to get travis running again, you need to `git fetch upstream / git merge upstream/master` again.

moralejo:
> @moralejo to get travis running again, you need to `git fetch upstream / git merge upstream/master` again.

Done!

lstchain/image/muon/muon_analysis.py
```diff
@@ -87,19 +91,17 @@ def fit_muon(x, y, image, geom, tailcuts):
...
-    ring_dist = np.abs(dist - muonringparam.ring_radius)
-
-    muonringparam = muonring.fit(
-            x, y, image_clean * (ring_dist < muonringparam.ring_radius * 0.4)
+    max_allowed_outliers_distance = 0.4
+    # Do an iterative fit removing pixels which are beyond max_allowed_outliers_distance*ring_radius
+    # of the ring (along the radial direction):
+    # The goal is to improve fit for good rings with very few additional non-ring bright pixels.
+
+    for _ in [0]*2:  # just to iterate the fit twice more
```
MaxNoe, line 99:
```

In [1]: %%timeit 
   ...: for i in range(2): 
   ...:     pass 
   ...:                                                                                                                                       
153 ns ± 1.83 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [2]: %%timeit 
   ...: for i in [0] * 2: 
   ...:     pass 
   ...:                                                                                                                                       
91 ns ± 1.21 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %%timeit 
   ...: for i in (0, 0): 
   ...:     pass 
   ...:                                                                                                                                       
38 ns ± 0.59 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

However, this will not play any role in the overall runtime of this analysis chain.

vuillaut:
> > As a general comment, I see you are introducing the muon analysis in `dl0_to_dl1`.
> > How much speed-up improvement are we talking about here? I am concerned about blowing up the analysis time of an already way too slow analysis...
> 
> There are two different things:
> 
> * speeding up the muon ring analysis: I did this just by avoiding to do the expensive Minuit fit for all but the most promising rings. The speed-up, when starting from DL1, was a factor ~10 (and the minuit fit was anyway useless for all those bad rings). Further speeding-up is possible via the ctapipe code, I think one can remove all astropy.units stuff from the low-level functions passed to Minuit, and also eliminate a lot of function nesting in the calculation of the predicted muon image.
> * avoiding to do again the whole r0->dl1 step with a different pulse integrator for all events, just to get the few interesting rings. As of now, we have to re-calibrate and re-parametrize all events, also doubling the dl1 data volume, just because of the rings.
> * By having the ring analysis integrated in the r0->dl1 step we can include more easily data checks based on the muons. This was actually the main trigger for this PR.
> 
> Ok, that was _three_ things.

Hi @moralejo 
Thanks for all the updates!
I checked quickly on a file.
Over ~13.000 events, the analysis found ~30 muons, increasing the analysis time of about ~10% - which I think is acceptable! :)

So since the muon analysis will now be part of the standard r0_to_dl1 - why not including these events and parameters in the DL1 file? (I can indicate you by PM how or push to your branch or do that in another PR)

rlopezcoto:
I agree with @vuillaut that it seems a good idea to include the muon parameters inside the DL1, but shall we try to stick to the [DL1 data writer](https://github.com/cta-observatory/ctapipe/pull/1163) whenever it is out?

In any case, I would very much like to merge this PR without proceeding with further changes, is that ok?

MaxNoe:
I agree with @rlopezcoto, we should not make this PR larger than it is now.

moralejo:
> > > As a general comment, I see you are introducing the muon analysis in `dl0_to_dl1`.
> > > How much speed-up improvement are we talking about here? I am concerned about blowing up the analysis time of an already way too slow analysis...
> > 
> > 
> > There are two different things:
> > 
> > * speeding up the muon ring analysis: I did this just by avoiding to do the expensive Minuit fit for all but the most promising rings. The speed-up, when starting from DL1, was a factor ~10 (and the minuit fit was anyway useless for all those bad rings). Further speeding-up is possible via the ctapipe code, I think one can remove all astropy.units stuff from the low-level functions passed to Minuit, and also eliminate a lot of function nesting in the calculation of the predicted muon image.
> > * avoiding to do again the whole r0->dl1 step with a different pulse integrator for all events, just to get the few interesting rings. As of now, we have to re-calibrate and re-parametrize all events, also doubling the dl1 data volume, just because of the rings.
> > * By having the ring analysis integrated in the r0->dl1 step we can include more easily data checks based on the muons. This was actually the main trigger for this PR.
> > 
> > Ok, that was _three_ things.
> 
> Hi @moralejo
> Thanks for all the updates!
> I checked quickly on a file.
> Over ~13.000 events, the analysis found ~30 muons, increasing the analysis time of about ~10% - which I think is acceptable! :)
> 
> So since the muon analysis will now be part of the standard r0_to_dl1 - why not including these events and parameters in the DL1 file? (I can indicate you by PM how or push to your branch or do that in another PR)

Hi,

they are now as you see in a separate fits file. Eventually they may be included in the DL1 .h5 but it is not urgent I think.
Also, for looking into the results "interactively" and with a graphical interface, I think FITS readers are ahead of hdf5 ones. So I'd prefer to keep it as a separate output for now. It is also more convenient from the point of view of moving files around.

moralejo:
From my side I'm done. I do not understand though what are the "requested changes" which block automatic merging.

MaxNoe:
The tests are failing because you are testing vs. an older version of ctapipe_io_lst

Is that intentional? Do you really want to still support `ctapipe_io_lst` v0.2.0?

vuillaut:
> The tests are failing because you are testing vs. an older version of ctapipe_io_lst
> 
> Is that intentional? Do you really want to still support `ctapipe_io_lst` v0.2.0?

The travis config seems wrong.

rlopezcoto:
The only one not passing is that of `ctapipe_io_lst == v0.2.0`, shall we merge and correct travis later?

vuillaut:
@moralejo you can merge master and the tests should pass.
Then we can merge (I think) this most reviewed PR ever in lstchain :)

MaxNoe:
Let's just merge this, the one test that fails is the one that is now removed.

rlopezcoto:
you also have to approve! 😄

PR #305, opened 2020-03-12
Merge files run-wise and sorted

morcuended:
I propose to include these lines in order to be able to merge all the files corresponding to a given run. If no run number is specified, the merging is performed as it was before.

### DIFF
--- a/lstchain/scripts/lstchain_merge_hdf5_files.py
+++ b/lstchain/scripts/lstchain_merge_hdf5_files.py
@@ -28,13 +28,22 @@ parser.add_argument('--no-image', action='store', type=lambda x: bool(strtobool(
                     dest='noimage',
                     help='Boolean. True to remove the images',
                     default=False)
-
-args = parser.parse_args()
+                                                                                    
+parser.add_argument('--run-number', '-r', action='store', type=str,                                                           
+                    dest='run_number',                                                                                  
+                    help='Merge files run-wise',                                                                        
+                    default=None)                                                                                       
+                                                                                                                        
+args = parser.parse_args()                                                                                              
 
 
 def main():
 
-    file_list = [args.srcdir + '/' + f for f in os.listdir(args.srcdir) if f.endswith('.h5')]
+    if args.run_number:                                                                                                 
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_merge_hdf5_files.py
```diff
@@ -28,13 +28,22 @@
...
+                    default=None)                                                                                       
+                                                                                                                        
+args = parser.parse_args()                                                                                              
 
 
 def main():
 
-    file_list = [args.srcdir + '/' + f for f in os.listdir(args.srcdir) if f.endswith('.h5')]
+    if args.run_number:                                                                                                 
+        file_list = sorted([args.srcdir + '/' + f for f in os.listdir(args.srcdir) if (f.endswith('.h5') and
```
vuillaut: Could you improve the code by using `os.path.join` rather than + '/' at the same time, please?

lstchain/scripts/lstchain_merge_hdf5_files.py
```diff
@@ -28,13 +28,22 @@
                     dest='noimage',
                     help='Boolean. True to remove the images',
                     default=False)
-
-args = parser.parse_args()
+                                                                                    
+parser.add_argument('--run-number', '-r', action='store', type=str,                                                           
+                    dest='run_number',                                                                                  
+                    help='Merge files run-wise',                                                                        
+                    default=None)
```
vuillaut: you could also set `default=''` and avoid a condition check (`'' in f` is always True) but it's a matter of preference here

vuillaut: Maybe `None` is more clear to the user. I would just add a comment about default in the help anyway. "Merge files run-wise if a run number is passed, otherwise merge all files in the directory".

morcuended: Done

lstchain/scripts/lstchain_merge_hdf5_files.py
morcuended: Done

morcuended:
I just created a new PR #310 because I messed this one using different branches.

PR #306, opened 2020-03-18
Check plots of DL1b parameters

morcuended:
Here is the proposal of the script to plot the distribution of image parameters from DL1b files as a possible datacheck for DL1b real data. In principle, it is though to work subrun and/or run-wise since involves `event_id` values which are repeated in every run. It also plots the altitude vs time and the three different timestamps.

The script can be either simply executed or called from any other part of the lstchain code.

Several things still need to be discussed/improved:
- Use it only for real data? So far it is not compatible with MC DL1b data since it involves variables only present in real data.
- Represent timestamps in a readable format, something like YYYY/MM/DD - HH:MM:SS
- Add a unit test (real DL1b dataset needed)

I uploaded this as the first step to create a check plot script at DL1 level. Any comment, suggestion, modification is very welcome.

Example:
image
image


morcuended:
#382 overcomes this, so I close it

PR #307, opened 2020-03-19
Changes for drs4 dt corrections, gain flat fielding and other

FrancaCassol:
Dear all,
this PR contains the following topics:
- new drs4 dt correction values at 20 degrees
- flat-fielding of the F-factor gains which was not added yet.
- the gain of declared unusable pixels is now set to zero by the calibrator
- there are then several changes (with no effect on data) performed in order to run the script on MC events (even if  this is possible only with a new version of ctapipe, the changes will be useful as soon we move to the ctapipe master)


lstchain/calib/camera/calibrator.py
```diff
@@ -119,11 +119,16 @@ def _initialize_correction(self):
...
 
+                    dc_to_pe=self.mon_data.tel[telid].calibration.dc_to_pe
+                    # put to zero unusable pixels
+                    dc_to_pe[self.mon_data.tel[telid].calibration.unusable_pixels] = 0
+                    # eliminate inf values id any (should be done probably before)
                     dc_to_pe[np.isinf(dc_to_pe)] = 0
-                    self.log.info(f"read {self.mon_data.tel[telid].calibration.dc_to_pe}")
+
+                    # read the pixel_status container
+                    table = '/tel_' + str(telid) + '/pixel_status'
```
rlopezcoto: Sorry @FrancaCassol, where is the `pixel_status` being filled and with which values?

lstchain/calib/camera/flatfield.py
```diff
@@ -118,20 +118,20 @@ def calculate_relative_gain(self, event):
         if self.num_events_seen == self.sample_size:
             self.num_events_seen = 0
 
+        pixel_mask = np.logical_or(
+            event.mon.tel[self.tel_id].pixel_status.hardware_failing_pixels,
```
rlopezcoto: how do you identify `hardware_failing_pixels `?

lstchain/calib/camera/r0.py
```diff
@@ -562,7 +562,11 @@ def ped_time(timediff):
     Coefficients from curve fitting to dragon test data
     at temperature 30 degC
     """
-    return 27.33 * np.power(timediff, -0.24) - 10.4
+    #return 27.33 * np.power(timediff, -0.24) - 10.4
+
+    # values at 20 degC
+    return 32.99 * np.power(timediff, -0.22) - 11.9
```
rlopezcoto: great! Can you add a comment saying who provide these values to you and in which date?

lstchain/tools/lstchain_data_create_calibration_file.py
```diff
@@ -192,7 +203,13 @@ def start(self):
                     continue
 
                 # if pedestal event
-                if event.r1.tel[self.tel_id].trigger_type == 32:
+                if event.r1.tel[self.tel_id].trigger_type == 32 or (
+                    self.simulation and
+                    np.median(np.sum(event.r1.tel[self.tel_id].waveform[0], axis=1))
+                    < self.minimum_hg_charge_median):
+
+                    #self.log.debug(f"ped {event.r1.tel[self.tel_id].waveform[0,1]} "
```
rlopezcoto: shall this comment be eliminated?

lstchain/tools/lstchain_data_create_calibration_file.py
```diff
@@ -192,7 +203,13 @@ def start(self):
                     continue
 
                 # if pedestal event
-                if event.r1.tel[self.tel_id].trigger_type == 32:
+                if event.r1.tel[self.tel_id].trigger_type == 32 or (
+                    self.simulation and
+                    np.median(np.sum(event.r1.tel[self.tel_id].waveform[0], axis=1))
+                    < self.minimum_hg_charge_median):
```
rlopezcoto: what is this condition for?

lstchain/calib/camera/calibrator.py
FrancaCassol: Hi @rlopezcoto,
The pixel status container is filled by the calibration script on the base of the inputs from hardware (R0), pedestals and flat fields

lstchain/calib/camera/flatfield.py
FrancaCassol: no, I get them from the EvB

lstchain/calib/camera/r0.py
FrancaCassol: ok

lstchain/tools/lstchain_data_create_calibration_file.py
FrancaCassol: yes

lstchain/tools/lstchain_data_create_calibration_file.py
FrancaCassol: This is for the MC events for which I do not have a trigger

FrancaCassol:
Hi @ruben, 
is this PR ok? It would be nice to have a new production with improved dt corrections and removed bad pixels  ...

rlopezcoto:
Yes @FrancaCassol, ok from my side, although there are still some open discussions I would like to close before the new release/production. Maybe we can add a point next Monday on the pending issues.

jsitarek:
Hi @FrancaCassol,
We are discussing with @rlopezcoto  in an e-mail thread that setting bad pixels based on the pedestal RMS would be a good way to reject fake islands in the image due to star. 
Does the new code also do something like this, or are the bad pixels set only when the pedestal subtaction completely fails?

FrancaCassol:
Hi @jsitarek ,
yes, at present the pedestal std is used to find bad pixels at the beginning of the night with the calibration run, but I am working in order to have the check continuously during the data taking by exploiting the interleaved pedestals. Hope to have it ready soon.

PR #308, opened 2020-03-23
First UCTS timestamp as reference to calculate Dragon and TIB timestamps 

morcuended:
(Discussion started from the e-mail from Taka)

> I (and probably many other people) hadn't understood precisely what is the origin of the clock. When GPS is not implemented, the PPS and 10MHz clock originate from the oscillator of the WR switch. So, just counting the clocks does not provide a reliable time stamp. It drifts.
> But when GPS is connected to the WR switch, not only the absolute time stamp is provided every second to UCTS, but also (our) PPS and 10 MHz clock base on the GPS PPS. So, just counting the clocks together with a single reference timestamp gives ~100ns precision time stamp, and it does not drift. 
> Each Dragon as well as TIB counts PPS and 10 MHz all the time. If there is a reference timestamp, then we can build a ~100ns precision time stamp with them.
> Jan, 18th, 2020, GPS was properly implemented.
UCTS does not work stably yet, but in the beginning of the run, it works well (since long ago).
> So, by using the first ucts time stamp of a given run as a reference point and using Dragon counters for the following events, we can have perfect timestamps in all events taken after Jan 18th, 2020. (I checked it with only one run, when ucts was stable).
> If we want good timestamps right now or for the previous data taken after Jan18th, it could be a good idea to change the code of lstchain (r0todl1). But UCTS hardware/firmware upgrade activity is on-going. Once UCTS becomes stable, then using the ucts timestamp directly is wiser (?).

As suggested by Taka, we can use the first UCTS timestamp as the reference point to build the TIB and Dragon-based timestamps out of their PPS and 10 MHz clocks.

I took into account the initial PPS and 10 MHz counters to calculate both TIB and Dragon-based TS.


moralejo:
Changes look ok to me. Just one thing: when the timestamp is used to interpolate pointing, the conversion TAI => UTC is not yet done, right? Where is it done?

morcuended:
> Changes look ok to me. Just one thing: when the timestamp is used to interpolate pointing, the conversion TAI => UTC is not yet done, right? Where is it done?

The conversion TAI to UTC is not done yet. The discussion on how properly take into account those 37 secs is still open (see #288). After the last check you proposed there I did not see clearly how to proceed further.

morcuended:
Discussing with José Luis and Isidro we see several drawbacks here:

- In order to get `first_ucts_time` value, we may need to first check whether `ucts_time` has a correct time (presence flag is OK) and get the first timestamp that fulfills this condition, which can be can be easily done.
- In the current scheme, the analysis is run on a subrun basis and these changes rely on the fact that the first event of every subrun has a correct timestamp value which is usually not the case. We might use the timestamp of the first event of the first subrun and share this value among all the runs for example. However, this correction would imply that the subruns cannot be analyzed in parallel and independently of each other as it is now.

Any suggestion?

lstchain/reco/dl0_to_dl1.py
```diff
@@ -312,33 +312,38 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
...
-                        ucts_time = event.lst.tel[telescope_id].evt.ucts_timestamp * 1e-9 # nsecs
+                        ucts_time = event.lst.tel[telescope_id].evt.ucts_timestamp * 1e-9  # nsecs
 
-                        # Get counters from the central Dragon module
-                        module_id = 82
+                        module_id = 82  # Get counters from the central Dragon module
 
-                        dragon_time = (
-                                event.lst.tel[telescope_id].svc.date +
+                        if i == 0:
```
rlopezcoto: Hi, at the moment this will be looking into the first event of the file you are providing, that will only be the first event of the run in case you are processing the first subrun.

rlopezcoto:
>- In order to properly get `first_ucts_time` value, we may need to first check whether `ucts_time` has a correct time (presence flag is OK) and get the first timestamp that fulfills this condition, which can be can be easily done.
>- In the current scheme, the analysis is run on a subrun basis and these changes rely on the fact that the first event of every subrun has a correct timestamp value which is usually not the case. We might use the timestamp of the first event of the first subrun and share this value among all the runs for example. However, this correction would imply that the subruns cannot be analyzed in parallel and independently of each other as it is now.

(Answer by Taka): Exactly. I just read the code and got the same opinion. 
Actually "presence flag" of UCTS is not good enough. When UCTS is in "unreliable" state (typically after ~million events), 20% of the time stamp is wrong, even if the presence flag is positive.
Options to handle this may be:
1) wait until UCTS is fixed.
2) change the scheme of dl0->dl1 to run-by-run basis (don't we need to update the ADC conversion factor anyway?)
3) keep the dragon timestamp as a relative value w.r.t. the first event of the subrun, and re-compute the absolute timestamp in higher level analysis where it is really needed (merging step, for example?)

rlopezcoto:
> 1. wait until UCTS is fixed.

we still need UCTS for all the data that have been taken up to now.
> 2. change the scheme of dl0->dl1 to run-by-run basis (don't we need to update the ADC conversion factor anyway?)

if we cannot parallelize the data processing subrun-wise, it will make it impossible from the computing-time aspect.
> 3. keep the dragon timestamp as a relative value w.r.t. the first event of the subrun, and re-compute the absolute timestamp in higher level analysis where it is really needed (merging step, for example?)

we can also keeping the first ucts time stamp of the run? It becomes a not so trivial change in the data processing, but it may do the trick (already something similar suggested by @morcuended, sorry that I missed it when I replied to the mail).

tysaito2008:
> > 1. change the scheme of dl0->dl1 to run-by-run basis (don't we need to update the ADC conversion factor anyway?)
> 
> if we cannot parallelize the data processing subrun-wise, it will make it impossible from the computing-time aspect.

 Then, how do you plan to use interleaved calibration events?

> 
> > 1. keep the dragon timestamp as a relative value w.r.t. the first event of the subrun, and re-compute the absolute timestamp in higher level analysis where it is really needed (merging step, for example?)
> 
> we can also keeping the first ucts time stamp of the run? It becomes a not so trivial change in the data processing, but it may do the trick (already something similar suggested by @morcuended, sorry that I missed it when I replied to the mail).

I am fine with any solution. There is no difference in the obtained timestamp.

moralejo:
We can keep the parallelization, and get the time in the "to_dl1" step if each job opens first the first subrun, takes the needed starting values, and then opens the subrun to be processed. I am assuming here that the Dragon counters keep increasing through the whole run and are never reset (is that the case?)

rlopezcoto:
this implies opening 200+ times the first subrun (one per subrun analyzed) more than needed...

DirkHoffmann:
> this implies opening 200+ times the first subrun (one per subrun analyzed) more than needed...

Why don't you generate a database with runwise information of:
- run number
- start time 
   - "human timescale" time, according to EVB start / ZFITS creation time
   - UCTS timestamp according to first event
   - DRAGON counter position at first event
   - (others?)
- number of events/files
- run characteristics
   - enabled triggers
   - trigger rate per TTYPE
- etc. ?

I looks to me like this would answer to many of the problems that arose in the last weeks.

vuillaut:
Hi.
Sorry for dropping in so late.
Wouldn't it be easier to just merge the runs and apply time correction at the end on the entire run at DL1 level as we did for the pointing correction?

morcuended:
> Hi.
> Sorry for dropping in so late.
> Wouldn't it be easier to just merge the runs and apply time correction at the end on the entire run at DL1 level as we did for the pointing correction?
> 
> Note that this implies again messing at higher level than lstchain and will induce other issues but to be considered anyway.

Actually we thought of doing the correction after the merge of DL2 files since for the production of these files we still stick with the sub-run basis.

moralejo:
> this implies opening 200+ times the first subrun (one per subrun analyzed) more than needed...

Of course, to avoid that we could build a database like Dirk proposed, but that might be overkill to solve a problem that will eventually be solved when UCTS information is 100% reliable.

To me, opening 200+ times a file to read just the first event does not seem a big deal, though.

Alternative: can the EVB do this somehow? i.e. calculate a Dragon-based timestamp through the whole run.

Alternative: take it out of lstchain: LSTOSA opens first file, gets the two relevant Dragon counter & UCTS time values, and uses them as arguments passed in the launching of all subruns.

tysaito2008:
>  I am assuming here that the Dragon counters keep increasing through the whole run and are never reset (is that the case?)
 Yes, that is the case.

morcuended:
> Alternative: take it out of lstchain: LSTOSA opens first file, gets the two relevant Dragon counter & UCTS time values, and uses them as arguments passed in the launching of all subruns.

We could do something like this using the NightSummary files produces by Isidro. First UCTS TS could be included there and read somehow when processing all the subrun files

moralejo:
> > Alternative: take it out of lstchain: LSTOSA opens first file, gets the two relevant Dragon counter & UCTS time values, and uses them as arguments passed in the launching of all subruns.
> 
> We could do something like this using the NightSummary files produces by Isidro. First UCTS TS could be included there and read somehow when processing all the subrun files

Good! Also the corresponding Dragon counter is needed, right?

morcuended:
> > > Alternative: take it out of lstchain: LSTOSA opens first file, gets the two relevant Dragon counter & UCTS time values, and uses them as arguments passed in the launching of all subruns.
> > 
> > 
> > We could do something like this using the NightSummary files produces by Isidro. First UCTS TS could be included there and read somehow when processing all the subrun files
> 
> Good! Also the corresponding Dragon counter is needed, right?

Yes, right. Isidro can include all the values needed to properly build the Dragon/TIB TS

rlopezcoto:
ok, I think that @moralejo's suggestion is the one that would include less overhead either in the processing, and also in the changes needed to be implemented. This will imply that we will *not* change lstchain and that the only runs with proper timestamp will be the ones processed by LSTOSA.
As discussed, this is only a temporary solution until UCTS is 100% reliable.
@morcuended: in which container will you fill this time info?

morcuended:
>ok, I think that @moralejo's suggestion is the one that would include less overhead either in the processing, and also in the changes needed to be implemented. This will imply that we will not change lstchain and that the only runs with proper timestamp will be the ones processed by LSTOSA.
>As discussed, this is only a temporary solution until UCTS is 100% reliable.
> @morcuended: in which container will you fill this time info?

Do you mean the time info needed to correct all the Dragon time stamps or where will we store the corrected time stamps themselves. 

Thinking more about this possibility we would need a modified version of lstchain capable of at least reading this info from the NightSummary, right?

rlopezcoto:
> Do you mean the time info needed to correct all the Dragon time stamps or where will we store the corrected time stamps themselves.

I meant the corrected timestamps
> Thinking more about this possibility we would need a modified version of lstchain capable of at least reading this info from the NightSummary, right?

right... 
if this is the case, then we should probably implement it as an option: if you give the NightSummary as an input on the processing of the data, then this correction is applied.
If the modification needs to go into lstchain anyways, I'm starting to think that maybe the suggestion by @moralejo of reading always, together with the subrun we are analyzing, the first subrun of the run, may be the easiest option. 

How much time of the total of the analysis of one subrun do we spend on I/O?

moralejo:
> 
> Thinking more about this possibility we would need a modified version of lstchain capable of at least reading this info from the NightSummary, right?

My suggestion was to have a command-line option in the script which creates the DL1, like --ucts_t0 xxx --Dragon_counter0 yyy. The values could be retrieved by LSTOSA from the NightSummary.

morcuended:
> @morcuended: in which container will you fill this time info?

@rlopezcoto, should we then create new containers for these corrected time stamps or use the existing ones `dragon_time` and `tib_time` instead. I would say use simply the existing ones.

rlopezcoto:
> @rlopezcoto, should we then create new containers for these corrected time stamps or use the existing ones dragon_time and tib_time instead. I would say use simply the existing ones.

I agree with this. So will you take care of implementing these changes @morcuended ?

moralejo:
On Tue, 24 Mar 2020 at 13:57, Ruben Lopez-Coto <notifications@github.com>
wrote:

> Do you mean the time info needed to correct all the Dragon time stamps or
> where will we store the corrected time stamps themselves.
>
> I meant the corrected timestamps
>
> Thinking more about this possibility we would need a modified version of
> lstchain capable of at least reading this info from the NightSummary, right?
>
> right...
>
My suggestion is that in lstchain the needed starting values are just
arguments of the script. The reading of the values from the Night summary,
or directly from the first subrun, would be the task of *lstosa*, which
then uses it to launch the lstchain jobs. Does that sound ok to you?
Fine for me to re-use existing fields to store the results.


-- 

Abelardo Moralejo Olaizola
Institut de Física d'Altes Energies
Tel : +34 931641662
Fax: +34 935811938
Avís - Aviso - Legal Notice - (LOPD) - http://legal.ifae.es

-- 
Avís -
Aviso - Legal Notice - (LOPD) - http://legal.ifae.es 
<http://legal.ifae.es/>

morcuended:
> > @rlopezcoto, should we then create new containers for these corrected time stamps or use the existing ones dragon_time and tib_time instead. I would say use simply the existing ones.
> 
> I agree with this. So will you take care of implementing these changes @morcuended ?

Yes, I will implement them.

> My suggestion is that in lstchain the needed starting values are just arguments of the script. The reading of the values from the Night summary, or directly from the first subrun, would be the task of *lstosa*, which then uses it to launch the lstchain jobs. Does that sound ok to you? Fine for me to re-use existing fields to store the results.

In order to read these values directly from the first subrun, we have to process it first. This idea does not quite suit the current scheme of processing all the subruns in parallel. However, we could handle more easily the parallel processing when reading them from the NightSummary instead. So, in principle, I would go for the second option. Is that OK?

rlopezcoto:
> In order to read these values directly from the first subrun, we have to process it first. This idea does not quite suit the current scheme of processing all the subruns in parallel. However, we could handle more easily the parallel processing when reading them from the NightSummary instead. So, in principle, I would go for the second option. Is that OK?

sounds good, but how will you (or @Isidro) fill this information in the NightSummary?

morcuended:
Since  @IsidroAguadoRuesga has to go through the R0 files in order to make the low-level check plots, he can easily store the UCTS time stamp and the Dragon pps + 10MHz counters of each run in seconds. So we would go from the current NightSummary file like this:

```
 01830    4 DRS4  2020-01-18 19:37:07  
 01831    4 CALI  2020-01-18 19:57:15  
 01832   70 DATA  2020-01-18 20:05:14  
 01833   73 DATA  2020-01-18 20:26:27  
 01834   79 DATA  2020-01-18 20:47:42  
[...]
```

to have the two additional time parameters required for each run:
```
[...]
 01832   70 DATA  2020-01-18 20:05:14  <ucts_t0> <Dragon_counter0>
[...]
```
They can be retrieved later on when running `lstchain_data_r0_to_dl1.py` by using the new command-line options that @moralejo proposed: `--ucts_t0 xxx --Dragon_counter0 yyy`

morcuended:
> In order to read these values directly from the first subrun, we have to process it first. This idea does not quite suit the current scheme of processing all the subruns in parallel. However, we could handle more easily the parallel processing when reading them from the NightSummary instead. So, in principle, I would go for the second option. Is that OK?

I retract myself. There is no need for having processed in advance the first sub run in order to be able to process the rest. It would be just reading the information time and Dragon counters from the raw data file. So this scheme would be feasible as well.

rlopezcoto:
@morcuended and @IsidroAguadoRuesga, remember that the ucts flag should also be ok. @IsidroAguadoRuesga is it ok to rerun back all the NightSummary to include this information?

morcuended:
> @morcuended and @IsidroAguadoRuesga, remember that the ucts flag should also be ok. @IsidroAguadoRuesga is it ok to rerun back all the NightSummary to include this information?

Yes, it won't be a problem to get the NightSummary files again. Isidro will get these values from the first event whose UCTS flag is OK. According to @IsidroAguadoRuesga, in some (rare) cases, the first event with correct UCTS timestamp is not within the first 10^4 events and possibly not even within the whole subrun. In those cases, we will not have the Dragon TS properly calculated.

Just for the record, here is the profile when I normally run `lstchain_data_r0_to_dl1.py` for a given sub run. Around 20% of the time is spent in the `event_source` function. I haven't checked yet how this would increase if we always open the first sub run as well.

image

IsidroAguadoRuesga:
El lun, 30-03-2020 a las 06:42 -0700, Ruben Lopez-Coto escribió:
> @morcuended and @IsidroAguadoRuesga, remember that the ucts flag
> should also be ok. @IsidroAguadoRuesga is it ok to rerun back all the
> NightSummary to include this information?
Yes. I hope it will be ready in a few days.

morcuended:
@rlopezcoto do we still want to keep TIB timestamp? In principle, I'm just recalculating the Dragon-based one using the first ucts time.

rlopezcoto:
I think that the Dragon-based should be enough.

DirkHoffmann:
> @rlopezcoto do we still want to keep TIB timestamp? In principle, I'm just recalculating the Dragon-based one using the first ucts time.

If you want my advice, from the experience with newly deployed hardware, I would keep all possible information, in particular if it only represents a tiny fraction (32 bits? Or even less, if you decided to store all "other timestamps" as "delta with respect to the reference timestamp of this event"). Things may evolve, break or just have a temporary failure. You do not want to depend on hardwired code and complete reprocessing in such cases.
At least until the reference timestamp has been proven to be correct and useful, for instance by detecting or confirming pulsar behaviour of an object, I would recommend to stay on the safe side.

tysaito2008:
In principle, TIB clock counter and Dragon clock counter should be the same.

But, there is a possibility that the central Backplane produce the 10 MHz clock and distribute  over Dragons. In that case, TIB counts 10 MHz from WR and Dragon counts 10MHz from the central BP. In such a case, TIB counter is useful while Dragon's not.  There was a certain period (anyway before GPS) when our camera was configured like that by mistake.

morcuended:
Thanks @DirkHoffmann and @tysaito2008, then we'll try to store also TIB timestamps

morcuended:
Just to illustrate what we'll have by using the first valid `ucts_time` of each run to build `dragon_time`

Whenever ucts is right:
image

differences being:
image


However, provided `ucts_time` values go wrong after some time:
image

differences being this time:
image

morcuended:
@rlopezcoto, @tysaito2008, @moralejo I just implemented the calculation of Dragon timestamps based on first valid `ucts_time` and its corresponding Dragon counter provided by @IsidroAguadoRuesga in the NightSummary files.

Here is an example, last two numbers are passed to `lstchain_data_r0_t0_dl1.py` as `ucts_t0` and `dragon_counter0`
```
 01915    4 DRS4  2020-01-31 21:33:34 0001 1580506462208626706 2208626500  
 01916    5 CALI  2020-01-31 22:09:15 0001 1580508610052979006 9052978700  
 01917    4 CALI  2020-02-01 01:00:32 0001 1580518883322614305 5322614000  
 01918  271 DATA  2020-02-01 01:12:35 0001 1580519604210915532 4210915300  
 01919  299 DATA  2020-02-01 01:44:05 0003 1580521498164948130 6164947900  
 01920  296 DATA  2020-02-01 02:19:23 0001 1580523618574809957 8574809700  
 01921  342 DATA  2020-02-01 02:54:34 0001 1580525726240811910 5240811700
```

The calculation of TIB timestamp in the same fashion is still missing.

lstchain/reco/dl0_to_dl1.py
```diff
@@ -140,7 +140,9 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
               pedestal_path = None,
               calibration_path = None,
               time_calibration_path = None,
-              pointing_file_path = None
+              pointing_file_path = None,
+              ucts_t0 = None,
```
rlopezcoto, line 144: Since these are floats, they should be `nan`, right?

lstchain/reco/dl0_to_dl1.py
```diff
@@ -312,37 +314,35 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
...
 
-                        # Get counters from the central Dragon module
-                        module_id = 82
-
-                        dragon_time = (
-                                event.lst.tel[telescope_id].svc.date +
+                        if ucts_t0 and dragon_counter0:
+                            dragon_time = ((ucts_t0 - dragon_counter0) * 1e-9 +  # secs
+                                event.lst.tel[telescope_id].evt.pps_counter[module_id] +
+                                event.lst.tel[telescope_id].evt.tenMHz_counter[module_id] * 10**(-7))
```
rlopezcoto, line 331: at the moment, no `tib_time` if `ucts_t0` and `dragon_counter0` are given

lstchain/reco/dl0_to_dl1.py
morcuended: Right. Then should I defined them as `nan` here?
```
              ucts_t0 = nan,
              dragon_counter0 = nan
```

However, with those variables defined as `nan` I would need to change the condition `if ucts_t0 and dragon_counter0:` since it won't work as it is now.

```python
a = float('nan')
if a:
    print('yes')
else:
    print('no')
>>> yes

b = None
if b:
    print('yes')
else:
    print('no')
>>> no
```

rlopezcoto: You can use `math.isnan(a)` or `np.isnan(a)` (first one is faster)

lstchain/reco/dl0_to_dl1.py
morcuended: Right now yes. Isidro already provided me the values to calculate TIB TS so I will implement it.

```
 01915    4 DRS4  2020-01-31 21:33:34 0001 1580506462208626706 2208626500 0001 1580506462208626706 2208626500  
 01916    5 CALI  2020-01-31 22:09:15 0001 1580508610052979006 9052978700 0001 1580508610052979006 9052978600  
 01917    4 CALI  2020-02-01 01:00:32 0001 1580518883322614305 5322614000 0001 1580518883322614305 5322614000  
 01918  271 DATA  2020-02-01 01:12:35 0001 1580519604210915532 4210915300 0001 1580519604210915532 4210915400  
 01919  299 DATA  2020-02-01 01:44:05 0003 1580521498164948130 6164947900 0003 1580521498164948130 6164947800  
 01920  296 DATA  2020-02-01 02:19:23 0001 1580523618574809957 8574809700 0001 1580523618574809957 8574809800  
 01921  342 DATA  2020-02-01 02:54:34 0001 1580525726240811910 5240811700 0001 1580525726240811910 5240811800 
```

 The drawback here is that I need to add two new extra `ucts_t0_tib` and `tib_counter0` variables since we need to find an event (ucts_t0_tib and its corresponding counter) with valid UCTS and TIB flag which does not have to be necessarily the same event used to build Dragon TS. Do you see a problem here having so many options?

rlopezcoto: ok, since this is temporary...

morcuended:
> Thanks @morcuended, just a couple of comments.
> Now, regarding #288, are you planning to implement a patch as discussed there?

My idea was to manually define a variable like `LEAP_SECONDS = 37` and subtract it to all of the timestamps. In addition, we could define these timestamps as `Time` astropy object in Unix format (with UTC scale).

lstchain/reco/dl0_to_dl1.py
morcuended: I've defined those variables as `math.nan` in the call to `r0_to_dl1` function inside `dl0_to_dl1.py`. Should I define them by default as `math.nan` in the argparser of `lstchain_data_r0_to_dl1.py` script? Right now they are set to `None`.

lstchain/reco/dl0_to_dl1.py
morcuended: implemented the calculation of TIB timestamp base on the first valid TIB counter and ucts_time

PR #309, opened 2020-03-23
removing intercept from configs

vuillaut:


morcuended:
Looking at the `lstchain_standard_config_keep_all_events.json` file I realize that it's no longer up to date. It needs to be either updated or entirely removed.

vuillaut:
> Looking at the `lstchain_standard_config_keep_all_events.json` file I realize that it's no longer up to date. It needs to be either updated or entirely removed.

Ok, thanks for the check.
Is it used at all? (if not I'd remove it to keep things cleaner)

morcuended:
> Ok, thanks for the check.
> Is it used at all? (if not I'd remove it to keep things cleaner)

I would say that not anymore

rlopezcoto:
the config to keep all events, is it really unused @morcuended?

morcuended:
> the config to keep all events, is it really unused @morcuended?

I'm not using it at least. Just the standard config file or a modified one if a different cleaning setting is required. But not exactly this one that became outdated quite some time ago.

rlopezcoto:
ok, so I agree to remove it to avoid confusion, thanks @vuillaut

PR #317, opened 2020-04-01
Add pyproject.toml, move static options to setup.cfg, style

MaxNoe:
Reference for pyproject.toml and why we should have one:
https://www.python.org/dev/peps/pep-0518/
https://www.python.org/dev/peps/pep-0517/
https://snarky.ca/what-the-heck-is-pyproject-toml/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,41 +3,53 @@
 # import sys
 from setuptools import setup, find_packages
 import os
-from version import get_version, update_release_version
+import sys
+
+# pep 517 builds do not have pwd in PATH
+sys.path.insert(0, os.path.dirname(__file__))
+from version import get_version, update_release_version  # noqa
+
 
 update_release_version()
 version = get_version()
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -3,41 +3,53 @@
...
-      tests_require=['pytest', 'pytest-ordering'],
-      author='LST collaboration',
-      author_email='',
-      license='',
-      url='https://github.com/cta-observatory/cta-lstchain',
-      long_description='',
-      classifiers=[],
-      entry_points=entry_points
-      )
+setup(
```
rlopezcoto, line 38: why removing the `name`?

MaxNoe: it's in setup.cfg now with all the other options I moved

vuillaut:
Related discussion:
https://github.com/cta-observatory/ctapipe/pull/1246

PR #318, opened 2020-04-02
Update gammapy stat

rlopezcoto:
Due to the new version of gammapy (v0.17), `excess_matching_significance_on_off` function disappeared and now it is substituted by the following lines

### DIFF
--- a/lstchain/mc/sensitivity.py
+++ b/lstchain/mc/sensitivity.py
@@ -207,9 +207,9 @@ def calculate_sensitivity_lima_ebin(n_excesses, n_background, alpha, n_bins_ener
     Parameters
     ---------
     n_excesses:   `numpy.ndarray` number of excess events in the signal region
-    n_background:   `numpy.ndarray` number of events in the background region
-    alpha: `float` inverse of the number of off positions
-    n_bins_energy: `int` number of bins in energy
+    n_background: `numpy.ndarray` number of events in the background region
+    alpha:        `float` inverse of the number of off positions
+    n_bins_energy:`int` number of bins in energy
 
     Returns
     ---------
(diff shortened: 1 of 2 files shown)


rlopezcoto:
tests will not pass until this is merged...

lstchain/mc/sensitivity.py
```diff
@@ -215,8 +218,19 @@ def calculate_sensitivity_lima_ebin(n_excesses, n_background, alpha, n_bins_ener
...
 
     """
-    n_excesses_5sigma = excess_matching_significance_on_off(\
-        n_off = n_background, alpha = alpha, significance = 5, method = 'lima')
+
+    if not (len(n_excesses) == n_bins_energy and
+        len(n_background) == n_bins_energy and
+        len(alpha) == n_bins_energy):
+        print("Excess, background and alpha arrays must have the same length")
+        return
```
MaxNoe, line 226: `raise ValueError` with that message, not print and return

MaxNoe: Maybe `if any(len(a) != n_bins_energy for a in (n_excesses, n_background, alpha))` as a more concise formulation

rlopezcoto: Thanks, already implemented. The only problem with this (also present with my formulation) is that `(n_excesses, n_background, alpha)` must be input as lists/arrays, and an error is raised if only single `floats` are given as input

MaxNoe: Then apply `np.asanyarray(n_on, ndmin=1)` to the input. If you also want to return a single number when the input were numbers, you can do sometehing like this:

```
scalar = np.isscalar(n_on)
n_on = np.asanyarray(n_on, ndmin=1)
n_off = ...
alpha = ...

...

if scalar:
     return result[0]
return result
```

MaxNoe: But let's get this merged first so the other PRs can run again

Issue #320, opened 2020-04-03
DL2: units mismatch between reco_energy and log_reco_energy


dipierr:
I'm looking at some preliminary DL2 (run 1835) and I have noticed that reco_energy is in TeV, while log_reco_energy was calculated with energies in GeV (see plots).

![reco_energy](https://user-images.githubusercontent.com/17765148/78365312-39f85200-75bf-11ea-8e57-64cb4556f004.png)

![log_reco_energy](https://user-images.githubusercontent.com/17765148/78365247-1c2aed00-75bf-11ea-8855-fae74b9dd894.png)

vuillaut:
Yes, that's historical but we should probably change that...

rlopezcoto:
Solved in #325

PR #325, opened 2020-04-06
Change unit of log_e_reco from GeV to TeV

rlopezcoto:
Solves #320

### DIFF
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -343,8 +343,8 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
                     dl1_container.wl = dl1_container.width / dl1_container.length
                     # Log10(Energy) in TeV
                     if is_simu:
-                        dl1_container.mc_energy = event.mc.energy.value
-                        dl1_container.log_mc_energy = np.log10(event.mc.energy.value)
+                        dl1_container.mc_energy = event.mc.energy.to_value(u.TeV)
+                        dl1_container.log_mc_energy = np.log10(event.mc.energy.to_value(u.TeV))
                         dl1_container.fill_mc(event)
 
                     dl1_container.log_intensity = np.log10(dl1_container.intensity)
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -341,7 +341,7 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
 
                     # Some custom def
                     dl1_container.wl = dl1_container.width / dl1_container.length
-                    # Log10(Energy) in GeV
+                    # Log10(Energy) in TeV
                     if is_simu:
                         dl1_container.mc_energy = event.mc.energy.value
                         dl1_container.log_mc_energy = np.log10(event.mc.energy.value)
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -344,7 +344,7 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
                     # Log10(Energy) in GeV
                     if is_simu:
                         dl1_container.mc_energy = event.mc.energy.value
-                        dl1_container.log_mc_energy = np.log10(event.mc.energy.value * 1e3)
+                        dl1_container.log_mc_energy = np.log10(event.mc.energy.value)
                         dl1_container.fill_mc(event)
 
                     dl1_container.log_intensity = np.log10(dl1_container.intensity)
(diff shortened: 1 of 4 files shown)


rlopezcoto:
that is going to make a bit messy with the models already produced when applied with newer lstchain versions, right?

My proposal was to change only the `log_e_reco` written in the DL2 file so that it is consistent with the unit of `e_reco`

vuillaut:
> that is going to make a bit messy with the models already produced when applied with newer lstchain versions, right?
> 
> My proposal was to change only the `log_e_reco` written in the DL2 file so that it is consistent with the unit of `e_reco`

Yes it will, but models should not be used between two main releases of lstchain anyway.
So we will just produce new models and IRF with the new release.
Having an inconsistency between `mc` and `reco` will be even worse than the current situation in my opinion.

rlopezcoto:
ok, thanks @vuillaut! 
So we should keep this in mind when re-processing new data (shall the next one be a major release then?)

vuillaut:
> ok, thanks @vuillaut!
> So we should keep this in mind when re-processing new data (shall the next one be a major release then?)

Yes, I think it should - we have introduced major changes recently, no?

I think there are more places where changes should be introduced concerning this PR, for example (but not exhaustive):
- filtering on `log_reco_energy` in build_models by default should now be `-np.inf`
- some labels in plots

lstchain/visualization/plot_dl2.py
```diff
@@ -200,7 +200,7 @@ def plot_e(data, n_bins, emin, emax, true_hadroness=False):
     plt.plot(bin_centers,standard_deviations,
              marker='X',linestyle='None')
     plt.ylabel('STD',fontsize=24)
-    plt.xlabel('$log_{10}E_{true}(GeV)$',fontsize=24)
+    plt.xlabel('$log_{10}E_{true}[TeV]$',fontsize=24)
```
MaxNoe, line 203: Should be `r'$\log_{10}(E_\mathrm{true} / TeV)$`

lstchain/visualization/plot_dl2.py
```diff
@@ -49,7 +49,7 @@ def plot_features(data, true_hadroness=False):
              histtype=u'step',bins=100,
              label="Protons")
     plt.ylabel(r'# of events',fontsize=15)
-    plt.xlabel(r"$log_{10}E$(GeV)")
+    plt.xlabel(r"$log_{10}E$[TeV]")
```
MaxNoe, line 52: Should be `r'$\log_{10}(E_\mathrm{true} / TeV)$'`

lstchain/visualization/plot_dl2.py
```diff
@@ -432,7 +432,7 @@ def plot_e_resolution(data, n_bins, emin, emax):
     plot1 = ax1.plot(bin_centers,standard_deviations,
                      marker='X',linestyle='None')
     plt.ylabel('STD',fontsize=24)
-    plt.xlabel('$log_{10}E_{true}(GeV)$',fontsize=24)
+    plt.xlabel('$log_{10}E_{true}[TeV]$',fontsize=24)
```
MaxNoe, line 435: see above

lstchain/visualization/plot_dl2.py
rlopezcoto: done

lstchain/visualization/plot_dl2.py
rlopezcoto: done

lstchain/visualization/plot_dl2.py
rlopezcoto: done

lstchain/reco/dl0_to_dl1.py
```diff
@@ -344,7 +344,7 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
                     # Log10(Energy) in GeV
                     if is_simu:
                         dl1_container.mc_energy = event.mc.energy.value
-                        dl1_container.log_mc_energy = np.log10(event.mc.energy.value * 1e3)
+                        dl1_container.log_mc_energy = np.log10(event.mc.energy.value)
```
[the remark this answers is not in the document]
rlopezcoto: changed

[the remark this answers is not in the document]
MaxNoe: I don't see it yet here, forgot to push?

lstchain/visualization/plot_dl2.py
MaxNoe: also not seeing this yet

MaxNoe:
@vuillaut @rlopezcoto there were no new commits since I reviewed, did you forget to push?

vuillaut:
> @vuillaut @rlopezcoto there were no new commits since I reviewed, did you forget to push?

I saw that @rlopezcoto said "done" to each comment and assumed (too quickly) they have been addressed.

rlopezcoto:
sorry, weird... I was left with no space on my computer and files were modified but commit/push did not work...

Issue #324, opened 2020-04-03
Consider astropy 4.1 in order to use the new `unix_tai` format


morcuended:
As discussed in #288, I suggest considering astropy 4.1 whenever it is out. It will allow us to use the newly added class that defines the `unix_tai` format which takes into account the leap seconds to easily transform Time objects between `tai` and `utc` scales.

In the astropy version 3.2.3 we currently use in lstchain, `unix` format is always considering `utc` scale without taking into account the leap seconds.

rlopezcoto:
I was testing the installation of astropy 4.0 together with @MaxNoe as discussed in #333, but I found that the astropy version is [hardcoded](https://github.com/cta-observatory/ctapipe/blob/71622417f4b80d625562c08df32da5c6f1d6808c/setup.py#L56) in ctapipe 0.7.0,so we cannot update astropy version until ctapipe does not get updated as well

kosack:
yes, we had to make some fixes for astropy 4.0, so it's only supported in the dev for now.  We will release a new version soon, though.

kosack:
One of the big changes with astropy 4 and also the latest numpy is that now all numpy functions support astropy units (which breaks things where we had assumed units were ignored).  That may cause you some problems as well.  So just upgrading may not be so simple.

rlopezcoto:
thanks @kosack. I saw your PRs and that is why I was willing to test astropy 4 and see which errors popped up, but Travis already throws conflicts in the environment creation due to the hardcoding in ctapipe

MaxNoe:
>  hardcoded in ctapipe 0.7.0,

Just to clarify: it is not hardcoded. It is required to be an astropy 3.x, as we require 3 and 4.0 broke some stuff. It's  not hard coded, it specifies which versions work with ctapipe.

maxnoe:
We found out that the definitions of `unix_tai` in astropy 4.1 differ from the actual definition UCTS uses by 8 seconds because of using a different epoch.

UCTS uses: `1970-01-01T00:00:00 TAI = 1969-12-31T23:59:52.000 UTC`

astropy `unix_tai` uses:  `1970-01-01T00:00:08 TAI = 1969-12-31 23:59:59.999918 UTC`

morcuended:
[...]

PR #326, opened 2020-04-06
Transform unix timestamps from TAI to UTC scale

morcuended:
As decided in #288, here is the workaround to transform Unix timestamps from TAI to UTC scale using `jd` format as an intermediate step. Whenever astropy v4.1 is ready we should simply use the new `unix_tai` format that automatically considers the leap seconds (current `unix` format does not). See #324.

As of now, with this transformation:
- event timestamps will be compared to the ones in the drive logs using UTC scale in both cases.
- UCTS, Dragon and TIB- based event timestamps will be stored in UTC scale (37 secs behind TAI scale) from now on.

### DIFF
--- a/lstchain/reco/dl0_to_dl1.py
+++ b/lstchain/reco/dl0_to_dl1.py
@@ -43,6 +43,7 @@ import pandas as pd
 from . import disp
 import astropy.units as u
 from astropy.table import Table
+from astropy.time import Time
 from .utils import sky_to_camera
 from ctapipe.instrument import OpticsDescription
 from traitlets.config.loader import Config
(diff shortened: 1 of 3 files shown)


lstchain/reco/dl0_to_dl1.py
```diff
@@ -383,24 +384,30 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
                                 event.lst.tel[telescope_id].evt.tib_pps_counter +
                                 event.lst.tel[telescope_id].evt.tib_tenMHz_counter * 10**(-7))
 
-                        dl1_container.ucts_time = ucts_time
-                        dl1_container.dragon_time = dragon_time
-                        dl1_container.tib_time = tib_time
+                        # FIXME: directly use unix_tai format whenever astropy v4.1 is out
```
MaxNoe, line 387: Could you make these functions and add a test, that the expected number of seconds are there?

Something like:
```
def unix_tai_to_utc(timestamp):
    '''Workaround for unix time in scale TAI''' 
    tai = Time(timestamp, format='unix', scale='tai')
    return Time(tai.utc.jd, format='jd', scale='tai')
```

morcuended: @MaxNoe, the test function should be placed in `lstchain/reco/tests/test_utils.py` script?

MaxNoe: If the functions lives in `lstchain/reco/utils.py`, yes

morcuended: Should be the test function something like the following?

```python
def test_unix_tai_to_utc():
    timestamp = 1579376359.3225002
    tai = Time(timestamp, format='unix', scale='tai')
    utc = Time(tai.utc.jd, format='jd', scale='tai')
    np.testing.assert_allclose(tai.unix, utc.unix) 
```

or should explicitly consider the 37 leap seconds somehow?

MaxNoe: If these two are `allclose`, isn't the conversion doing nothing?

MaxNoe: Let me understand, what the input timestamp represents. Is this si seconds since `1970-01-01T00:00:00 TAI`?

morcuended: Actually `rtol` was too relaxed and tai and utc were indeed `all_closed`.

This is at least considering those 37 secs:
```python
def test_unix_tai_to_utc():
    timestamp = 1579376359.3225002
    leap_seconds = 37
    tai = Time(timestamp, format='unix', scale='tai')
    utc = Time(tai.utc.jd, format='jd', scale='tai')
    np.testing.assert_allclose(tai.unix, utc.unix, rtol=1e-12, atol=leap_seconds) 
```

MaxNoe: You should define a function and test that function. A test that tests lines of code that are not used somewhere else, is not really helpful.

So define the function and test if it is what you expect.

morcuended: True... I was not aware of that
It is the first time I implement a test.

Provided the function is:

```python
def unix_tai_to_utc(timestamp):
    '''Workaround for unix time in scale TAI'''
    tai = Time(timestamp, format='unix', scale='tai')
    return Time(tai.utc.jd, format='jd', scale='tai')
```

This could serve as a test?
```python
def test_unix_tai_to_utc():
    timestamp_tai = 1579376359.3225002
    leap_seconds = 37
    utc_time = unix_tai_to_utc(timestamp_tai)
    np.testing.assert_allclose(utc_time.unix, timestamp_tai - leap_seconds, rtol=1e-12)
```

MaxNoe: Yes, but I still think is is a bit of a convoluted, unintuitive solution. Adding the timestamp to the epoch is much clearer I think.

morcuended: I agree. I implemented it.

lstchain/reco/utils.py
```diff
@@ -470,6 +470,9 @@ def clip_alt(alt):
...
 
 def unix_tai_to_utc(timestamp):
-    '''Workaround for unix time in scale TAI'''
-    tai = Time(timestamp, format='unix', scale='tai')
-    return Time(tai.utc.jd, format='jd', scale='tai')
+    """
+    Transform unix time from TAI to UTC scale considering the leap seconds
+    by adding the timestamp in seconds to the epoch value.
+    """
+    UCTS_EPOCH = Time('1970-01-01T00:00:00', scale='tai', format='isot')
```
MaxNoe, line 477: You could this define as a constant outside of the function

PR #331, opened 2020-04-07
WIP: R1 calibration

yrenier:
In the context of looking at stars in LST1 data ( https://indico.cta-observatory.org/event/2700/contributions/23482/attachments/17239/22795/pointing_300320_upload.pdf ), I have been looking at the calibration from R0 to R1.
I'm working with data from the 28th of February, and I corrected some small issues and warnings I found during the conversion of R0 to DL1.
To do the correction, I do:
```
lstchain_data_create_drs4_pedestal_file --input_file LST-1.1.Run02034.0000.fits.fz --output_file drs4_pedestal.fits
lstchain_data_create_calibration_file --input_file LST-1.1.Run02035.0000.fits.fz --output_file calibration.h5 --pedestal_file drs4_pedestal.fits --config lstchain/data/onsite_camera_calibration_param.json
lstchain_data_create_time_calibration_file --input_file LST-1.1.Run02035.0000.fits.fz -ped drs4_pedestal.fits --output_file time_calibration.fits
```
then I can finally do:
```
lstchain_data_r0_to_dl1 -f LST-1.1.Run02036.0000.fits.fz -o . -pedestal drs4_pedestal.fits -calib calibration.h5 -time_calib time_calibration.fits
```

I'm not sure which config to use for the calibration step as there are 2 which seems to be done for that: 
- lstchain/data/onsite_camera_calibration_param.json
- lstchain/tools/camera_calibration_param.json

Both gave error when I was using them and I fixed both, but it would be less confusing to have only one.

That PR is still a work in progress as I intend also to move the flat-fielding in the R0 to R1 step since for the study on the star we need calibrated wave-forms.  (I also believe R1 has to be flat-fielded as a CTA requirement)

### DIFF
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -148,7 +148,7 @@ class LSTR0Corrections(CameraR0Calibrator):
 
             event.r1.tel[tel_id].trigger_time = event.r0.tel[tel_id].trigger_time
 
-            samples = event.r0.tel[tel_id].waveform[:, :, self.r1_sample_start:self.r1_sample_end]
+            samples = event.r1.tel[tel_id].waveform[:, :, self.r1_sample_start:self.r1_sample_end]
 
             event.r1.tel[tel_id].waveform = samples.astype('int16') - self.offset
 
@@ -178,7 +178,7 @@ class LSTR0Corrections(CameraR0Calibrator):
                                         n_modules)
 
         event.r1.tel[self.tel_id].trigger_type = event.r0.tel[self.tel_id].trigger_type
-        event.r1.tel[self.tel_id].trigger_time = event.r0.tel[self.tel_id].trigger_time
+        event.r1.tel[self.tel_id].trigger_time = event.r1.tel[self.tel_id].trigger_time
         event.r1.tel[self.tel_id].waveform = samples[:, :, :]
 
 
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -315,15 +315,14 @@ class LSTR0Corrections(CameraR0Calibrator):
         if not isinstance(event.r1.tel[tel_id].waveform, np.ndarray):
             return
 
-        waveforms = event.r1.tel[tel_id].waveform
-        event.mon.tel[tel_id].calibration = self.mon_data.tel[tel_id].calibration
-        event.mon.tel[tel_id].pixel_status = self.mon_data.tel[tel_id].pixel_status
+        mon_data_tel = self.mon_data.tel[tel_id]
+        event.mon.tel[tel_id].calibration = mon_data_tel.calibration
+        event.mon.tel[tel_id].pixel_status = mon_data_tel.pixel_status
 
         # subtract the pedestal per sample (should we do it?) and multiply for the calibration coefficients
         #
-        pedestal_per_sample = self.mon_data.tel[tel_id].calibration.pedestal_per_sample[:, :, np.newaxis]
-        dc_to_pe = self.mon_data.tel[tel_id].calibration.dc_to_pe[:, :, np.newaxis]
-        event.r1.tel[tel_id].waveform = (waveforms - pedestal_per_sample) * dc_to_pe
+        event.r1.tel[tel_id].waveform -= mon_data_tel.calibration.pedestal_per_sample[:, :, np.newaxis]
+        event.r1.tel[tel_id].waveform *= mon_data_tel.calibration.dc_to_pe[:, :, np.newaxis]
 
 
     @staticmethod
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -474,7 +474,8 @@ class LSTR0Corrections(CameraR0Calibrator):
                     table = '/tel_' + str(telid) + '/pixel_status'
                     next(h5_table.read(table, self.mon_data.tel[telid].pixel_status))
         except:
-            self.log.error(f"Problem in reading calibration file {self.calibration_path}")
+            self.log.exception(f"Problem in reading calibration file {self.calibration_path}")
+            raise FileNotFoundError(f"Problem in reading calibration file {self.calibration_path}")
 
 
     def _get_first_capacitor(self, event, nr_module, tel_id):
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -148,7 +148,7 @@ class LSTR0Corrections(CameraR0Calibrator):
 
             event.r1.tel[tel_id].trigger_time = event.r0.tel[tel_id].trigger_time
 
-            samples = event.r1.tel[tel_id].waveform[:, :, self.r1_sample_start:self.r1_sample_end]
+            samples = event.r0.tel[tel_id].waveform[:, :, self.r1_sample_start:self.r1_sample_end]
 
             event.r1.tel[tel_id].waveform = samples.astype('int16') - self.offset
 
@@ -178,7 +178,7 @@ class LSTR0Corrections(CameraR0Calibrator):
                                         n_modules)
 
         event.r1.tel[self.tel_id].trigger_type = event.r0.tel[self.tel_id].trigger_type
-        event.r1.tel[self.tel_id].trigger_time = event.r1.tel[self.tel_id].trigger_time
+        event.r1.tel[self.tel_id].trigger_time = event.r0.tel[self.tel_id].trigger_time
         event.r1.tel[self.tel_id].waveform = samples[:, :, :]
 
 
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -1,7 +1,9 @@
 import numpy as np
 from astropy.io import fits
 from ctapipe.core import Component
-from ctapipe.core.traits import Unicode, Int
+from ctapipe.core.traits import Unicode, Int, List
+from ctapipe.io.containers import MonitoringContainer
+from ctapipe.io.hdf5tableio import HDF5TableReader
 from abc import abstractmethod
 from numba import jit, prange
 
@@ -84,6 +86,17 @@ class LSTR0Corrections(CameraR0Calibrator):
                             allow_none=True,
                             help='Path to the LST pedestal binary file'
                             ).tag(config=True)
+    calibration_path = Unicode(
+        '',
+        allow_none=True,
+        help='Path to LST calibration file'
+    ).tag(config=True)
+
+    allowed_tels = List(
+        [1],
+        help = 'List of telescope to be calibrated'
+    ).tag(config = True)
+
 
     def __init__(self, **kwargs):
         """
@@ -99,6 +112,9 @@ class LSTR0Corrections(CameraR0Calibrator):
             Tool executable that is calling this component.
             Passes the correct logger to the component.
             Set to None if no Tool to pass.
+        calibration_path :
+            Path to LST calibration file to get the pedestal and flat-field corrections
+
         kwargs
         """
         super().__init__(**kwargs)
@@ -135,8 +151,10 @@ class LSTR0Corrections(CameraR0Calibrator):
         self.first_cap_old_array = np.zeros((self.n_module,
                                              self.n_gain,
                                              self.n_pix))
+        self.mon_data = MonitoringContainer()
 
         self._load_calib()
+
 
     def calibrate(self, event):
         for tel_id in event.r0.tels_with_data:
@@ -151,6 +169,9 @@ class LSTR0Corrections(CameraR0Calibrator):
             samples = event.r1.tel[tel_id].waveform[:, :, self.r1_sample_start:self.r1_sample_end]
 
             event.r1.tel[tel_id].waveform = samples.astype('int16') - self.offset
+
+            if self.calibration_path:
+                self.convert_dc_to_pe(event, tel_id)
 
 
     def subtract_pedestal(self, event, tel_id):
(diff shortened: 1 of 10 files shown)


yrenier:
by the way I still have few warning in the R0 -> DL1 step that I did not correct as I did not find the origin:
```
/home/yves/.local/lib/python3.7/site-packages/ctapipe/coordinates/__init__.py:50: MissingFrameAttributeWarning: AltAz coordinate has no obstime, assuming same frame
/home/yves/.local/lib/python3.7/site-packages/ctapipe/coordinates/__init__.py:58: MissingFrameAttributeWarning: Horizontal coordinate has no location, assuming same frame
```

morcuended:
> then I can finally do:
> 
> ```
> lstchain_data_r0_to_dl1 -f LST-1.1.Run02036.0000.fits.fz -o . -pedestal drs4_pedestal.fits -calib calibration.h5 -time_calib time_calibration.fits
> ```

Are you using also the information from drive_logs with `-pointing` option to interpolate the pointing coordinates event-wise?

morcuended:
Isn't it possible to look for stars at DL1 level?

yrenier:
I did not use it yet, but you are right I should try it (as it might well be the reason for the warnings above)

yrenier:
At DL1 we would have much less photons (only the one falling in one integration window).
So we want to use R1 (to keep photons during the full readout window), that's why I wish to move the flat fielding in the R0->R1 step.
I just got that part running, I need to check now that the new results are compatible with the original ones

yrenier:
the dc -> pe conversion is now moved in r1. I checked with a data file for LST-1 and there are no difference at the DL1 output between the original lstchain implementation and this one.

lstchain/mc/sensitivity.py
```diff
@@ -5,11 +5,15 @@
...
 from lstchain.spectra.proton import proton_bess
-from gammapy.stats import WStatCountsStatistic
 from lstchain.reco.utils import reco_source_position_sky
 from astropy.coordinates.angle_utilities import angular_separation
 from lstchain.io import read_simu_info_merged_hdf5
 from lstchain.io.io import dl2_params_lstcam_key
+try:
+    from gammapy.stats import WStatCountsStatistic
+except ImportError:
+    from gammapy.stats import wstat as WStatCountsStatistic
```
rlopezcoto, line 15: where is this import error coming from? I tested it with gammapy 0.17 and did not get it

lstchain/data/onsite_camera_calibration_param.json
```diff
@@ -1,45 +1,48 @@
...
+  "CalibrationHDF5Writer": {
+    "minimum_hg_charge_median": 5000,
+    "maximum_lg_charge_std": 300,
+    "one_event": true,
+    "flatfield_product": "FlasherFlatFieldCalculator",
+    "pedestal_product": "PedestalIntegrator",
+    "r0calibrator_product": "LSTR0Corrections",
+    "log_level":"DEBUG"
+  },
+  "EventSource": {
```
rlopezcoto, line 13: what was exactly the problem with this configuration file, that it was not containing `max_events`? I do not think that including by default `max_events = 1000` is a good solution here...

lstchain/mc/sensitivity.py
MaxNoe: Yes, please just update gammapy using `conda install -c conda-forge gammapy=0.17`

lstchain/calib/camera/r0.py
```diff
@@ -412,6 +450,33 @@ def _load_calib(self):
...
+                    # put to zero unusable pixels
+                    dc_to_pe[self.mon_data.tel[telid].calibration.unusable_pixels] = 0
+                    # eliminate inf values id any (should be done probably before)
+                    dc_to_pe[np.isinf(dc_to_pe)] = 0
+
+                    # read the pixel_status container
+                    table = '/tel_' + str(telid) + '/pixel_status'
+                    next(h5_table.read(table, self.mon_data.tel[telid].pixel_status))
+        except:
+            self.log.error(f"Problem in reading calibration file {self.calibration_path}")
```
MaxNoe, line 477: Please don't ignore exceptions, especially as so crucial ones as that a calibration file could not be read.
Things like this should fail loud and early.

So at least replace `log.error` with `log.exception` so that the traceback of the actual exception is loggend and add a `raise` after that so the exception is propagated upwards.

lstchain/calib/camera/r0.py
```diff
@@ -412,6 +450,33 @@ def _load_calib(self):
                 self.pedestal_value_array[:, :, self.size4drs:self.size4drs + 40] \
                     = pedestal_data[:, :, 0:40] - self.offset
 
+        if not self.calibration_path:
```
MaxNoe, line 453: I think this should fail if no calibration path is give, not just silently return None. raise ValueError or a config error.

lstchain/calib/camera/r0.py
```diff
@@ -289,6 +310,22 @@ def interpolate_spikes(self, event, tel_id):
             event.r1.tel[self.tel_id].trigger_type = event.r0.tel[self.tel_id].trigger_type
             event.r1.tel[self.tel_id].trigger_time = event.r0.tel[self.tel_id].trigger_time
 
+
+    def convert_dc_to_pe(self, event, tel_id):
+        if not isinstance(event.r1.tel[tel_id].waveform, np.ndarray):
```
MaxNoe, line 315: What are the cases when waveform is not an ndarray? Why do you want to return nan if that happens?

lstchain/calib/camera/r0.py
```diff
@@ -289,6 +310,22 @@ def interpolate_spikes(self, event, tel_id):
...
+
+        waveforms = event.r1.tel[tel_id].waveform
+        event.mon.tel[tel_id].calibration = self.mon_data.tel[tel_id].calibration
+        event.mon.tel[tel_id].pixel_status = self.mon_data.tel[tel_id].pixel_status
+
+        # subtract the pedestal per sample (should we do it?) and multiply for the calibration coefficients
+        #
+        pedestal_per_sample = self.mon_data.tel[tel_id].calibration.pedestal_per_sample[:, :, np.newaxis]
+        dc_to_pe = self.mon_data.tel[tel_id].calibration.dc_to_pe[:, :, np.newaxis]
+        event.r1.tel[tel_id].waveform = (waveforms - pedestal_per_sample) * dc_to_pe
```
MaxNoe, line 326: If you do:
```
event.r1.tel[tel_id].waveform -= pedestal_per_sample
event.r1.tel[tel_id].waveform *= dc_to_pe
```
this will be much more efficient, as no new arrays will have to be created. (a - b) * c creates 2 intermediate new arrays (for the result of a - c and the result of (a - c) * b

FrancaCassol:
Dear @yrenier ,
I am afraid your changes are not ctapipe compliant.
I like to add that Improvements in the code are always welcome (and surely the present calibrator can be improved), but in order to proceed in a collaborative manner (that is what we are used to do in lstchain) I strongly suggest you to contact  people responsible of a task  before proposing major changes in the code related to that task (that helps also to avoid useless loss of time). So, I suggest to have a direct exchange in order to understand which are your needs and problems.

lstchain/calib/camera/r0.py
```diff
@@ -84,6 +86,17 @@ class LSTR0Corrections(CameraR0Calibrator):
                             allow_none=True,
                             help='Path to the LST pedestal binary file'
                             ).tag(config=True)
+    calibration_path = Unicode(
+        '',
```
MaxNoe, line 90: If you allow_none, the default should be `None` as well.

lstchain/calib/camera/r0.py
```diff
@@ -84,6 +86,17 @@ class LSTR0Corrections(CameraR0Calibrator):
...
+        allow_none=True,
+        help='Path to LST calibration file'
+    ).tag(config=True)
+
+    allowed_tels = List(
```
MaxNoe, line 95: Is this `allowed_tels` for the EventSource? If yes, this should be an alias to that config, not it's own option

lstchain/calib/camera/r0.py
```diff
@@ -412,6 +450,33 @@ def _load_calib(self):
...
 
+        if not self.calibration_path:
+            self.log.info("no dc to pe conversion done as no calibration file as been given")
+            return
+
+        self.mon_data.tels_with_data = self.allowed_tels
+        self.log.info(f"read {self.calibration_path}")
+        try:
+            with HDF5TableReader(self.calibration_path) as h5_table:
+                assert h5_table._h5file.isopen == True
```
MaxNoe, line 461: I don't think you need that assert here. The first line should raise an exception if opening the file does not work. Also, iy might be good to specify `mode=r`

lstchain/mc/sensitivity.py
yrenier: My bad, I had gammapy 0.16 installed, with gammapy 0.17 it works. I will revert the change and add 0.17 as a minimum version in setup.py

MaxNoe:
> I am afraid your changes are not ctapipe compliant.

What do you mean with that?

>  I strongly suggest you to contact people responsible of a task before proposing major  changes

Actually, this is not how open developement works. Yes, opening an issue when a problem exists would probably been better, but opening a PR with change proposals.

lstchain/mc/sensitivity.py
MaxNoe: I already did in #333

lstchain/data/onsite_camera_calibration_param.json
yrenier: without `max_event` I get the following error:
```
  File "/usr/local/bin/lstchain_data_create_calibration_file", line 11, in <module>
    load_entry_point('lstchain', 'console_scripts', 'lstchain_data_create_calibration_file')()
  File "/home/yves/ctasoft/cta-lstchain/lstchain/tools/lstchain_data_create_calibration_file.py", line 336, in main
    exe.run()
  File "/home/yves/.local/lib/python3.7/site-packages/ctapipe/core/tool.py", line 199, in run
    self.start()
  File "/home/yves/ctasoft/cta-lstchain/lstchain/tools/lstchain_data_create_calibration_file.py", line 175, in start
    if count == self.tot_events-1 or count == self.eventsource.max_events-1:
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```
An other option would to change the code in lstchain_data_create_calibration_file.py to use all event if `max_event` is not present

lstchain/calib/camera/r0.py
yrenier: You are right, I wll implement the changes you propose.
It is code from LSTCameraCalibrator._initialize_correction() I just moved here.

lstchain/data/onsite_camera_calibration_param.json
MaxNoe: Add a `self.eventsource.max_events is not None` as first check.

lstchain/calib/camera/r0.py
yrenier: That was what I though first too, but during the creation of the calibration file itself the LSTR0Corrections is used.

rlopezcoto:
> Actually, this is not how open developement works. Yes, opening an issue when a problem exists would probably been better, but opening a PR with change proposals is also totally fine.

It is certainly ok to do a PR, but when it implies a major code refactor, it may be better to discuss the general picture before. At the moment details are being discussed and it may be that the moving of the code is not necessary

lstchain/calib/camera/r0.py
MaxNoe: I don't understand how this relates to the fact that a function that **loads** a file should do nothing.

MaxNoe: ah, ok

MaxNoe: Maybe split this in `load_pedestal` and `load_calibration` then? So each function only does one thing?

lstchain/calib/camera/r0.py
yrenier: `waveform` is not an array when it has not be filled yet.
That part from the code came from LSTCameraCalibrator._calibrate_dl0() and was using CameraCalibrator._check_dl0_empty(). I could not use it as LSTR0Corrections does not herrit from CameraCalibrator, so I used the same test as in LSTR0Corrections.time_lapse_corr() and LSTR0Corrections.interpolate_spikes() instead.

lstchain/data/onsite_camera_calibration_param.json
MaxNoe:
```
max_events_reached = self.eventsource.max_events is not None and count == self.eventsource.max_events - 1
if count == self.tot_events - 1 or max_events_reached
```

lstchain/calib/camera/r0.py
yrenier: It makes sense, but that is not what is done elsewhere in the code.
Should I change that as well ?

yrenier:
> I strongly suggest you to contact people responsible of a task before proposing major changes

I understand that this change may be controversial and I thought that a PR would be a reasonable place to discuss it. Sorry if it is not the case :-/ 
I will discuss it by e-mail/call. If anyone is interested,let me know that you want to be included.

> At the moment details are being discussed and it may be that the moving of the code is not necessary
I will put the WIP tag back waiting for a conclusion to be reached

MaxNoe:
@yrenier please open an issue describing what you are trying to solve here

yrenier:
I just opened #335

yrenier:
As indicated in issue #335 , the calibrated waveforms are already available in dl0, which mean this PR is not really needed anymore. 
Should I create an other PR to include the various improvements suggested here (but without moving the calibration into the R0->R1 step) ?
I will close the PR once that is clarified.

FrancaCassol:
Hi @yrenier ,
yes, if you have improvements not related to the calibration movement, please go ahead with an other PR

PR #342, opened 2020-04-11
Apply volume reduction to non gain selected waveforms

vuillaut:


lstchain/reco/volume_reducer.py
```diff
@@ -75,7 +75,6 @@ def check_and_apply_volume_reduction(event, config):
         pass
 
     else:
-        print("Volume reduction algorithm being used: {}".format(volume_reduction_algorithm))
```
garciagenrique, line 78: Not really against or in favor (just matter of tastes). But why taking out the stdout message ? It is true that to apply the vol_reduc, you must had consciously modified the config file... But this would be just a double check of what are you doing/applying int the analysis

MaxNoe: There shouldn't be and prints in the code base. If you want to keep this message, use the logging module (debug or info for this)

lstchain/reco/volume_reducer.py
```diff
@@ -92,7 +91,12 @@ def check_and_apply_volume_reduction(event, config):
 
             image[~pixels_to_keep] = 0
             pulse_time[~pixels_to_keep] = 0
-            waveform[~pixels_to_keep, :] = 0
+            if waveform.ndim == 2:
```
rlopezcoto, line 94: do we still have cases in which we do not apply gain selection or this is just for testing purposes?

vuillaut: We actually do not apply gain selection in real data currently (in the sense that we currently keep both channels).

rlopezcoto: ok, sorry this is for R0 data

vuillaut: Hum, no DL0

rlopezcoto: yeah, I meant before we apply gain selection (for DL1)

vuillaut: So if it's fine with you, I'd like to merge this

PR #344, opened 2020-04-16
Fix dl1 file naming to get rid of the stream number and "fits" extension

### LINKED ISSUE #301: Naming of DL1 files and dealing with data streams
We are producing DL1 files that are called:
`dl1_LST-1.1.Run02032.0192.fits.h5`
from files called:
`LST-1.*.Run02032.0192.fits.fz`
Do we want to keep the `fits` part in the dl1 output? 
Also, the `.1.` part is a bit misleading. I understand that when we call the code for the input file `LST-1.1.Run02032.0192.fits.fz`, all four streams (`LST-1.1.*`, `LST-1.2.*` `LST-1.3.*` and `LST-1.4.*`) are included in the analysis, but it would be great to distinguish between the processing of all four streams `LST-1.Run02032.0192.fits.fz` and individual ones `LST-1.1.Run02032.0192.fits.fz`

rlopezcoto:
Fixes #301 

To get rid of the `.1.` and `fits` I had to include two different paths depending on the name of the file. Another option is to read the input file and treat it differently if it is `mc` or `real_data`, but this one will work as long as we do not modify the naming of the files.


rlopezcoto:
Also, I had some troubles with the `__init__` because the file `r0_to_dl1.py` contains the function `r0_to_dl1` (amongst others), then it cannot be included in the init as:
`from .r0_to_dl1 import *`

lstchain/reco/__init__.py
```diff
@@ -1,4 +1,6 @@
-from .utils import get_event_pos_in_camera
-from .r0_to_dl1 import get_dl1
-from . import dl1_to_dl2
+from .utils import *
+from .r0_to_dl1 import get_dl1, add_disp_to_parameters_table
+from .dl1_to_dl2 import *
+from .volume_reducer import *
+from .disp import *
```
MaxNoe, line 5: Please no star imports without providing `__all__`. Neither editors nor code linters will be able to work properly (autocompletion / linting)

lstchain/reco/r0_to_dl1.py
```diff
@@ -182,9 +179,16 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
 
     """
     if output_filename is None:
-        output_filename = (
-            'dl1_' + os.path.basename(input_filename).rsplit('.',1)[0] + '.h5'
-        )
+        if (input_filename.startswith('LST')):
+            output_filename = (
+                'dl1_' + os.path.basename(input_filename).split('.',5)[0] + '.' \
```
MaxNoe, line 184: you don't need line continuations inside parenthesis

lstchain/reco/r0_to_dl1.py
```diff
@@ -182,9 +179,16 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
...
+                    + os.path.basename(input_filename).split('.',5)[2] + '.' \
+                    + os.path.basename(input_filename).split('.',5)[3] + '.h5'
+                )
+        else:
+             output_filename = 'dl1_' \
+                 + os.path.basename(input_filename).rsplit('.',1)[0] + '.h5'
```
MaxNoe, line 190: os.path.splitext

lstchain/reco/r0_to_dl1.py
```diff
@@ -182,9 +179,16 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
 
     """
     if output_filename is None:
-        output_filename = (
-            'dl1_' + os.path.basename(input_filename).rsplit('.',1)[0] + '.h5'
-        )
+        if (input_filename.startswith('LST')):
```
MaxNoe, line 182: this might be a lot more elegant with using `pathlib.Path`.

e.g.

```python
>>>from pathlib import Path
>>>p = Path('run_001.simtel.gz')
>>>p.with_name('dl1_' + p.stem).with_suffix('.hdf5')  
```

lstchain/reco/__init__.py
rlopezcoto: I'm not sure what do you mean, all files provide `__all__`, but some functions were missing though, this is what you mean?

lstchain/reco/r0_to_dl1.py
rlopezcoto: actually the `simtel` part is at the moment kept for the MC. I'm in favor of getting rid of it as well, anybody thinks we should keep it?

lstchain/reco/r0_to_dl1.py
rlopezcoto: Done

lstchain/reco/r0_to_dl1.py
rlopezcoto: done

lstchain/reco/r0_to_dl1.py
MaxNoe: Then you Can use p.name instead of p.stem

rlopezcoto: done. 
I kept the `simtel` part for the time being, we can come back to this in the future if people are not happy.

rlopezcoto:
@MaxNoe are you ok with the changes?

lstchain/reco/__init__.py
MaxNoe: This file does not have an `__all__`.

rlopezcoto: ok, I get it. We need to modify all the `__init__` files because none have `__all__` and are making wildcard imports. Also modify some of the calls to avoid
> Also, I had some troubles with the __init__ because the file r0_to_dl1.py contains the function r0_to_dl1 (amongst others), then it cannot be included in the init as:
from .r0_to_dl1 import *

can we merge this one and I open an Issue/PR right afterwards to solve that in a different thread?

rlopezcoto: so @MaxNoe shall we move on with this PR and I'll correct the different `__init__` files afterwards?

MaxNoe: ok

lstchain/reco/r0_to_dl1.py
```diff
@@ -188,9 +187,16 @@ def r0_to_dl1(
 
     """
     if output_filename is None:
-        output_filename = (
-            'dl1_' + os.path.basename(input_filename).rsplit('.', 1)[0] + '.h5'
-        )
+        if (input_filename.startswith('LST')):
+            output_filename = (
```
MaxNoe, line 191: This looks like you are expecting a certain filename pattern here. Not just that the file starts with `LST`.

What is the format?

rlopezcoto: Yes, names are e.g.:
`LST-1.4.Run01975.0005.fits.fz`
so:
'LST-' + `tel_id` + '.' + `stream_number` + '.Run' + `run_number` + '.' + `subrun_number` + '.fits.fz'

MaxNoe: So I think we should use a regex for this and check if this matches, I'll prepare a PR.

MaxNoe: And you want to remove the stream here? So dl1_LST-{tel_id}.{run_number}.{stream_number}'?

rlopezcoto: No, that is the misleading thing about streams, see #301 . 
The reader, if all 4 streams are in the same folder, reads the 4 of them and aligns the event numbers from the four files. So when you process the data in the standard way, event though the input string may be `LST-1.1.Run01975.0005.fits.fz`, the `dl1` output has all 4 streams, that was the reason to get rid of the `stream_number`

rlopezcoto: I agree that this may be problematic if you process only one of the streams, because you would lose the stream number in the output, but this is related to how the reader processes the data and in any case, you process all the 4 streams in most of the cases

MaxNoe: THis is what I wrote here, or not? The intention of the code is to add the `dl1_` prefix, replace the extension with `.h5` and to remove the stream  number, correct?

rlopezcoto: yes, that is right, but this is what is done in the code right now, right?
sorry, I read the original message that contained some `stream_number` in the output name

morcuended:
Is the stream number already removed in the master? From my latest test I think I'm still getting them:
`dl1_LST-1.1.Run01844.0131.fits.h5`

rlopezcoto:
Grrr, I think that I know what happened there. The output file name is actually changed in:
https://github.com/cta-observatory/cta-lstchain/blob/6e3770e51003ddd337b7fa46a388961649a4b8de/lstchain/scripts/lstchain_data_r0_to_dl1.py#L125
(addint the `dl1_` already in the `lstchain_r0_to_dl1` script). Then when it checks this:
https://github.com/cta-observatory/cta-lstchain/blob/6e3770e51003ddd337b7fa46a388961649a4b8de/lstchain/reco/r0_to_dl1.py#L193

the file name does not start with 'LST' anymore... Will this get solved with #391 @MaxNoe ? Otherwise I can work it out.

MaxNoe:
It should be, yes.

rlopezcoto:
then let's wait for the merge

PR #345, opened 2020-04-16
Fix dl2 naming

rlopezcoto:
I open this one apart from #344 to see if everybody agrees. I'm not sure why in the `dl2` file naming we kept `dl1`, as in:
`dl2_dl1_LST-1...`

I propose to eliminate the `dl1` part for the future naming of dl2 files


rlopezcoto:
Thanks for the approval @morcuended and sorry, but I had to solve some conflicts before merging (need another approval)

PR #347, opened 2020-04-17
Fix dl1ab script

vuillaut:
This fixes the script dl1ab to be in line with the r0 to dl1 one.
The equality of recalculated parameters is now unit tested.

As stated before, the approach in this script is **not** what we want in the longer term: the part dl1a to dl1b should be defined in common - but again I don't want to refactor to much the chain at the moment - so at least this is working.


vuillaut:
Note: as stated and tested, this works for MC data.
To be tested on real data, but I expect a discrepancy due to the camera geometry.

rlopezcoto:
for the camera geometry, would a solution like this one work?
https://github.com/cta-observatory/cta-lstchain/blob/5c8914a51ab8c9eff842634367cd31e907530edf/lstchain/scripts/lstchain_dl1_muon_analysis.py#L95

vuillaut:
> for the camera geometry, would a solution like this one work?
> https://github.com/cta-observatory/cta-lstchain/blob/5c8914a51ab8c9eff842634367cd31e907530edf/lstchain/scripts/lstchain_dl1_muon_analysis.py#L95

Thanks for the pointer!
That would work indeed, but if I understand correctly, stating the path implies that one already knows what is the desired camera geometry (`LSTCam` here)?

rlopezcoto:
I think this is tracing back to something else...
I checked and either in MC or Read Data files, we call the geometry table:
`instrument/telescope/camera/LSTCam`, no matter if it is really the `LSTCam` geometry (SimTel) or the `LSTCam-004` (current real data), was that table naming done on purpose like that?

vuillaut:
> I think this is tracing back to something else...
> I checked and either in MC or Read Data files, we call the geometry table:
> `instrument/telescope/camera/LSTCam`, no matter if it is really the `LSTCam` geometry (SimTel) or the `LSTCam-004` (current real data), was that table naming done on purpose like that?

exactly, which led me to https://github.com/cta-observatory/ctapipe_io_lst/issues/27

rlopezcoto:
ok, sorry that I missed the discussion, actually related to this, we are (were) also stripping some part of the telescope name here:
https://github.com/cta-observatory/cta-lstchain/blob/2873c089fbad19f1f762b7e5f0bdc0f550bcf40a/lstchain/reco/r0_to_dl1.py#L325

could you please correct it here? (or maybe we can open a new PR)

vuillaut:
> ok, sorry that I missed the discussion, actually related to this, we are (were) also stripping some part of the telescope name here:
> https://github.com/cta-observatory/cta-lstchain/blob/2873c089fbad19f1f762b7e5f0bdc0f550bcf40a/lstchain/reco/r0_to_dl1.py#L325
> 
> could you please correct it here? (or maybe we can open a new PR)

Actually stripping the tel name there was on purpose to always access the parameters at the same place (in my opinion, the tel name should not even be `LSTCam` but `LST1` or just `1` there - that's what we are moving to with the CTA DL1 data model anyhow).
But the geometry stripped of the version, that got me confused.
Anyway, this is actually non-issue and I will update to use the code you propose here.

rlopezcoto:
no, for me it is ok to strip the tel name, it is just that it should not be hardcoded to `-003`

lstchain/reco/r0_to_dl1.py
```diff
@@ -322,7 +322,6 @@ def r0_to_dl1(input_filename = get_dataset_path('gamma_test_large.simtel.gz'),
                 tel.prefix = ''  # don't really need one
                 # remove the first part of the tel_name which is the type 'LST', 'MST' or 'SST'
                 tel_name = str(event.inst.subarray.tel[telescope_id])[4:]
-                tel_name = tel_name.replace('-003', '')
```
rlopezcoto, line 325: sorry @vuillaut I'm a bit confused now with the other discussion in cta-observatory/ctapipe_io_lst#27, what would be the `tel_name` that is arriving at this point when using the `LSTCam-004` geometry?

vuillaut: `tel_name` is always `LSTCam` whatever the geometry, the version is already stripped

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -54,12 +56,18 @@ def test_mc_dl1ab():
...
+def test_mc_dl1ab_validity():
+    dl1 = pd.read_hdf(os.path.join(output_dir, 'dl1_gamma_test_large.simtel.h5'), key=dl1_params_lstcam_key)
+    dl1ab = pd.read_hdf(os.path.join(output_dir, 'dl1ab.h5'), key=dl1_params_lstcam_key)
+    np.testing.assert_allclose(dl1, dl1ab, rtol=1e-4)
+
+
 @pytest.mark.run(after='test_lstchain_dl1_to_dl2')
 def test_mc_r0_to_dl2():
-    cmd = f'lstchain_mc_r0_to_dl2 -f {mc_gamma_testfile} -p {output_dir} -s1 False -o {output_dir}'
+    cmd = f'lstchain_mc_r0_to_dl2 -f {mc_gamma_testfile} -p {output_dir} -s1 True -o {output_dir}'
```
rlopezcoto, line 68: why is the DL1 file now written?

vuillaut: To be used in `test_mc_dl1ab_validity` to be checked vs `dl1ab.h5`.

lstchain/scripts/lstchain_mc_dl1ab.py
```diff
@@ -84,21 +86,32 @@ def main():
...
+
+                signal_pixels = tailcuts_clean(camera_geom, image, **config['tailcut'])
                 n_pixels = np.count_nonzero(signal_pixels)
                 if n_pixels > 0:
-                    num_islands, island_labels = number_of_islands(geom, signal_pixels)
-                    hillas = hillas_parameters(geom[signal_pixels], image[signal_pixels])
+                    num_islands, island_labels = number_of_islands(camera_geom, signal_pixels)
+                    n_pixels_on_island = np.zeros(num_islands + 1)
+
+                    for iisland in range(1, num_islands + 1):
```
MaxNoe, line 100: You can use np.bincount to avoid this loop.
```
n_pixels_on_island = np.bincount(island_labels)
```

should be what you want.

vuillaut:
> Some conflicts need to be resolved... but it looks good to me otherwise

Sorry, can you review again?
There were indeed some conflicts.

lstchain/scripts/lstchain_mc_dl1ab.py
```diff
@@ -88,27 +90,35 @@ def main():
...
-                    dl1_container.set_timing_features(geom[signal_pixels],
+                    dl1_container.set_timing_features(camera_geom[signal_pixels],
                                                       image[signal_pixels],
                                                       pulse_time[signal_pixels],
                                                       hillas)
-                    dl1_container.set_leakage(geom, image, signal_pixels)
-                    dl1_container.set_concentration(geom, image, hillas)
+
+                    dl1_container.set_leakage(camera_geom, image, signal_pixels)
+                    dl1_container.set_leakage(camera_geom, image, signal_pixels)
```
rlopezcoto, line 112: repeated?

PR #348, opened 2020-04-17
Features

MaxNoe:
Adds
* leakage1_intensity
* leakage1_pixel
* leakage2_pixel
* n_pixels
* concentration_cog
* concentration_core
* concentration_pixel

Renames
* leakage → leakage2_intensity

I also fixed an issue concerned with backseeking.
The script used the first event twice to get subarray information, I moved this inside the event loop and used the first event there.

This should make reading of gz files slightly (20%) faster and reading zstd compressed files possible.

### DIFF
--- a/lstchain/_version_cache.py
+++ b/lstchain/_version_cache.py
@@ -1,1 +1,1 @@
-version='0.4.5.post23+gite9b21b4'
+version='0.4.5.post185+gitcf2101d'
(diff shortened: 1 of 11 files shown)


rlopezcoto:
The `leakage` variables have already changed name in the ctapipe master, did any other of the variables here?

MaxNoe:
Yes, the concentration members lost the `concentration_` prefix as this is now handled with the prefix of the container itself.

rlopezcoto:
ok, let's keep all of it in mind whenever we update to the new ctapipe release

PR #351, opened 2020-04-17
Use tmp dir for tests, fixes #350

### LINKED ISSUE #350: Use temporary directory for unit tests
I had issues concerning clean up of files when a test failed.

We should use a temporary directory for the unit tests, not a fixed one.

maxnoe:


rlopezcoto:
No strong opinion here, @vuillaut would you still prefer to keep non-tmp directories?

lstchain/tests/test_lstchain.py
```diff
@@ -1,24 +1,22 @@
...
+    with tempfile.TemporaryDirectory() as d:
+        yield Path(d)
 
-mc_gamma_testfile = get_dataset_path('gamma_test_large.simtel.gz')
-dl1_file = os.path.join(test_dir, 'dl1_gamma_test_large.simtel.h5')
-dl2_file = os.path.join(test_dir, 'dl2_gamma_test_large.simtel.h5')
-fake_dl2_proton_file = os.path.join(test_dir, 'dl2_fake_proton.simtel.h5')
-file_model_energy = os.path.join(test_dir, 'reg_energy.sav')
-file_model_disp = os.path.join(test_dir, 'reg_disp_vector.sav')
-file_model_gh_sep = os.path.join(test_dir, 'cls_gh.sav')
```
vuillaut, line 21: Was it really an issue to have these files defined globally?
It allows other tests to import the paths and define them in a single place.

vuillaut:
> No strong opinion here, @vuillaut would you still prefer to keep non-tmp directories?

I do personally prefer having a non-tmp directory but it's just a personal preference and is probably not the best practice so no problem there.

lstchain/tests/test_lstchain.py
MaxNoe: The whole point of this was to get to a state where tests don't depend on each other, so that single tests can be run and stuff like `pytest --last-failed` works. I'll change this to a draft PR and come back as soon as I have something that works to this regard.

vuillaut:
Close? Leave open?

morcuended:
I did a refactoring of the temporal directories used in lstchain tests in #587 using fixtures. Then it is possible to run tests separately.

PR #354, opened 2020-04-20
Update r0 sample start and end

pawel21:
Hi,

I have changed default value of start and end sample in waveform in 'r0.py' and updated notebook according to the change in EVB in November 2019.


rlopezcoto:
This PR follows from this discussion by e-mail by @jsitarek:

> During the call today we discussed the removal of slices and if it can affect the pulse shape correction.
Before the firmware upgrade the first 2 and last 2 samples had some extra noise and had to be removed from the analysis. After the upgrade it shifted by one and now first 3 and last 1 have to be removed. Currently in the code still 2+2 are removed.
When you use the method of Taka on the R1 files you should be aware of this shift, so e.g. if you skip the first 2 samples, and first capacitor is 111 then in fact the first sample that is recorded is 111+2 = 113. If you make the tables from files that have this skipping done and apply it to a file with the same format it might even work fine, but if you apply to files without skipping, or with different number of skipped samples this can completely screw up the method.

PR #356, opened 2020-04-21
WIP modify __init__ files

rlopezcoto:
As discussed in #344, @MaxNoe requested to add `__all__` functions to the `__init__` files with wildcards for editors to be able to autocomplete these calls.

I explicitly added all the functions to the calls of `reco/__init__` and added `__all__` as well to the `__init__` file. If you are happy with this, I'll do it for the rest of the `__init__` files


rlopezcoto:
Already addressed in #369

PR #358, opened 2020-04-21
Add processing of Interleaved events

FrancaCassol:
This PR contains the following changes:
- Introduction of the new Component  CalibrationCalculator which collects all calculations for  extracting  the calibration coefficients 
- Processing of interleaved events inside the r0_to_dl1.py module and writing of the results in the DL1 h5 files inside the table  /dl1/event/telescope/monitoring. The first table line contains the calibration read from the input calibration file. These initialisation values are used for all events till an interleaved calibration event is calculated. The calculation is steered with cards inside the list_standard_config.json file. If the requested statistic for  interleaved events is put to a high value (e.g. 10000), it will be collected over the whole sub-run and the interleaved results will be written down at the end of the event loop. As a result, all events will be calibrated with the  calibration run values and each h5 file will contain only one interleaved calibration event obtained with the full interleaved statistics of the sub-run (this is a provisory solution)
- Time flat-fielding is no more based on the time calibration file (which is updated each some months), but on the charge calibration file (which is updated each night)


lstchain/calib/camera/calibration_calculator.py
```diff
@@ -0,0 +1,258 @@
...
+
+        flatfield_calculator: lstchain.calib.camera.flatfield
+             The flatfield to use. If None, then FlatFieldCalculator
+                will be used by default.
+
+        pedestal_calculator: lstchain.calib.camera.pedestal
+             The pedestal to use. If None, then
+               PedestalCalculator will be used by default.
+
+        kwargs
```
MaxNoe, line 82: What are kwargs used here for? And is it safe to assume that both the FlatField and the PedestalCalculator get the same kwargs?

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -0,0 +1,258 @@
...
+        calib_data = event.mon.tel[self.tel_id].calibration
+
+        # mask from pedestal and flat-field data
+        monitoring_unusable_pixels = np.logical_or(status_data.pedestal_failing_pixels,
+                                                   status_data.flatfield_failing_pixels)
+        # calibration unusable pixels are an OR of all masks
+        calib_data.unusable_pixels = np.logical_or(monitoring_unusable_pixels,
+                                                   status_data.hardware_failing_pixels)
+        # Extract calibration coefficients with F-factor method
+        # Assume fix F2 factor, F2=1+Var(gain)/Mean(Gain)**2 must be known from elsewhere
```
MaxNoe, line 176: So this should be a configration option, not hard coded here

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -0,0 +1,258 @@
...
+        F2 = 1.2
+
+        # calculate photon-electrons
+        n_pe = F2 * (ff_data.charge_median - ped_data.charge_median) ** 2 / (
+                ff_data.charge_std ** 2 - ped_data.charge_std ** 2)
+
+        # fill WaveformCalibrationContainer (this is an example)
```
MaxNoe, line 183: What do you mean with "this is an example"?

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        Process interleaved calibration events (pedestals and FF)
+        Parameters
+        ----------
+        """
+        new_ped = False
+        new_ff = False
+
+        # if pedestal event
+        if event.r1.tel[self.tel_id].trigger_type == 32:
```
MaxNoe, line 224: It would probably be usefull to define a  dictionary with trigger types centrally somewhere instead of hard coding values here.

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+            # if new ff, calculate new calibration coefficients
+            if new_ff:
+                self.calculate_calibration_coefficients(event)
+
+
+        return new_ped, new_ff
+
+    def force_interleaved_results(self, event):
```
MaxNoe, line 247: What has this to do with force?

lstchain/calib/camera/flatfield.py
```diff
@@ -43,6 +43,11 @@ class FlasherFlatFieldCalculator(FlatFieldCalculator):
         [-3, 3],
         help='Interval (number of std) of accepted charge standard deviation around camera median value'
     ).tag(config=True)
+    time_calibration_path = Unicode(
```
MaxNoe, line 46: Use `ctapipe.core.traits.Path` here, it also supports checking if the file exists.

lstchain/calib/camera/flatfield.py
```diff
@@ -74,6 +79,15 @@ def __init__(self, **kwargs):
         self.arrival_times = None  # arrival time per event in sample
         self.sample_masked_pixels = None  # masked pixels per event in sample
 
+        # declare time calibrator if correction file exist
+        if os.path.exists(self.time_calibration_path):
+            self.time_corrector = PulseTimeCorrection(
+                calib_file_path = self.time_calibration_path)
+        else:
+            self.time_corrector = None
+            self.log.info(f"File {self.time_calibration_path} not found. No drs4 time corrections")
```
MaxNoe, line 88: I think this should not be silently ignored and for sure not be "info".

If the user gave a calibration path that does not exist, I think it would be better to fail loudly and immediately.

So this should be `raise IOError('Time calibration file {self.time_calibration_path' not found')`, or just use the `exists=True` option of the `Path` traitlet, see above.

lstchain/io/io.py
```diff
@@ -670,3 +671,38 @@ def recursive_copy_node(src_file, dir_file, path):
...
+def write_calibration_data(writer,mon_index, mon_event, new_ped=False, new_ff=False):
+    mon_event.pedestal.prefix = ''
+    mon_event.flatfield.prefix = ''
+    mon_event.calibration.prefix = ''
+    mon_index.prefix = ''
+
+
+    # update index
+    if new_ped:
+        mon_index.pedestal_id += 1
```
MaxNoe, line 683: I find it a bit surprising that a `write_` function mutates its input. This should not happen!

MaxNoe: Also, This function depends on two booleans that make ifs inside this. This is a clear sign that this function should just be split in two, one to write a new pedestal and one to write a new flat fielding.

lstchain/io/lstcontainers.py
```diff
@@ -89,6 +93,14 @@ class DL1ParametersContainer(Container):
     tel_pos_y = Field(None, "Telescope y position in the ground")
     tel_pos_z = Field(None, "Telescope z position in the ground")
 
+    trigger_type = Field(None, "trigger type")
+    ucts_trigger_type = Field(None, "UCTS trigger type")
+    trigger_time = Field(None, "trigger time")
```
MaxNoe, line 98: How is this trigger time different from the `gps_time`, `dragon_time` or `ucts_time` from above?

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: It is all the trailers passed the CalibrationCalculator. Each Component will look for each own trailets

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: historical comment ;-)

lstchain/calib/camera/calibration_calculator.py
MaxNoe: But raise errors for unknown ones.

You should explictly pass `config` and `parent` to this class and pass `parent=self` to all subcomponents.

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -0,0 +1,258 @@
...
+        5000,
+        help='Temporary cut on HG charge till the calibox TIB do not work (default for filter 5.2)'
+    ).tag(config=True)
+
+    maximum_lg_charge_std = Float(
+        300,
+        help='Temporary cut on LG std against Lidar events till the calibox TIB do not work (default for filter 5.2) '
+    ).tag(config=True)
+
+    def __init__(self, **kwargs):
```
MaxNoe, line 135: You don't need this __init__ method if it just calls `super.__init__`, can be removed.

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: yes, you are right, actually I promised @moralejo to do it. Perhaps the best place is in ctapipe_io_lst? (but that will ask to create a new release)

MaxNoe: Yes, this should be done then.

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: I don't understand, what should be an option?

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: changed

lstchain/calib/camera/flatfield.py
FrancaCassol: Ok

lstchain/calib/camera/calibration_calculator.py
MaxNoe: The F2 could be a `Float` traitlet or (in the next ctapipe version) a `FloatTelescopeParamete.` 

You say here this F2 factor depends on external knowledge, so it should not be hard coded here as `1.2` but be a configuration option, maybe with default 1.2

lstchain/calib/camera/flatfield.py
FrancaCassol: This is because at the beginning not all nights had the time calibration file. Now we can indeed always ask for it. I will add the raise IOError because I get problems with the Path

MaxNoe: What kind of problems? You should use Path. If you have problems with it, let's solve them instead of doing something else

MaxNoe: Does `timecalibration_path = Path(None, exists=True, directory_ok=False)` not work for you?

lstchain/io/io.py
FrancaCassol: yes, all this is ugly, but I expect this module will be soon replaced by a clean tool. I just try to reduce the number of function calls inside  it (it is already long enough)

lstchain/io/lstcontainers.py
FrancaCassol: From above, where precisely?

MaxNoe: The same container has these fields, a couple of lines above the ones you added here.

MaxNoe: If you view this comments in the "changes" tab of github, you can see them

lstchain/calib/camera/flatfield.py
FrancaCassol: I want to use it as trailet, how do you add the help? (it didn't like it)

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: ok, it can be done

lstchain/calib/camera/flatfield.py
MaxNoe: Yes, sorry. Support for the help and other standard traitlets args was only added after 0.7.

I would to suggest to leave it as unicode then und wait with `Path` for the ctapipe 0.8 switch

lstchain/io/lstcontainers.py
FrancaCassol: yes, now a remember :-) I didn't  add them, I simply  took them out from the ExtraImageInfo because they better belong to the event parameters. Don't ask me to explain why we have so many times and trigger types, that would be a pain for both ;-)

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: mmh, I had often problem with this stuff. For the moment, I used the same way of proceeding with all the calibration Components and it works. Let keep the suggested change as a coming improvement for  the lstchain update to the ctapipe master, ok?

lstchain/io/lstcontainers.py
MaxNoe: But how exactly is the generic `trigger_time` related to the 3 specific (gps, ucts, dragon) and why is it needed as additional field here? This should be in the description, otherwise no one can actually understand those descriptions...

lstchain/calib/camera/calibration_calculator.py
MaxNoe: No, please let's do it right here. It is actually pretty simple. If you have a component consisting of subcomponents, pass `parent=self` to the subcomponents.

MaxNoe: No, please let's do it right here. It is actually pretty simple. If you have a component consisting of subcomponents, pass `parent=self` to the subcomponents.

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: I preferred to add a class in the lstcontainers.py module. I hope the trigger bit description is still correct

lstchain/io/lstcontainers.py
```diff
@@ -266,3 +267,43 @@ class DL1MonitoringEventIndexContainer(Container):
...
+    bit 1: stereo
+    bit 2: Calibration
+    bit 3: Single Phe
+    bit 4: Softrig(from the UCTS)
+    bit 5: Pedestal
+    bit 6: slow control
+    bit 7: busy
+    """
+
+    def is_mono(trigger_type):
```
MaxNoe, line 284: These can all be staticmethods.

I see that this makes sense to have in a common namespace, but this class is not a valid python class (even if it seems to work for you at the moment). 

The first argument to any normal method should be `self`, the reference to the instance, which is automatically inserted.

If you have free functions (that don't refer to any data in the class, nor to the class itself, use `@staticmethod`
So at least add `@staticmethod` above each method here.

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: Hi @MaxNoe, 
I still have problems to get the all the trailets in the Components without kargs. I don't think
this change is fundamental for the coming release (which is still based on a "bastard" code). 
Actually the change should be performed also in the ctapipe calibration Components. Let do everything properly everywhere in the successive release (both in lstchain and in ctapipe) where also trailets will be correctly passed through a tool script

lstchain/reco/r0_to_dl1.py
```diff
@@ -457,14 +509,11 @@ def r0_to_dl1(
...
-
                     dl1_container.prefix = tel.prefix
 
+                    # extra info for the image table
                     extra_im.tel_id = telescope_id
-                    extra_im.num_trig_pix = event.r0.tel[telescope_id].num_trig_pix
-                    extra_im.trigger_time = event.r0.tel[telescope_id].trigger_time
-                    extra_im.trigger_type = event.r0.tel[telescope_id].trigger_type
-                    extra_im.trig_pix_id = event.r0.tel[telescope_id].trig_pix_id
+                    extra_im.low_gain_mask = event.lst.tel[telescope_id].evt.pixel_status >> 2 & 1
```
MaxNoe, line 516: I think this line broke the tests for the mc data, because mc does not have the `lst` member of `event`

lstchain/calib/camera/flatfield.py
```diff
@@ -78,12 +78,19 @@ def __init__(self, **kwargs):
         self.arrival_times = None  # arrival time per event in sample
         self.sample_masked_pixels = None  # masked pixels per event in sample
 
-        # declare time calibrator if correction file exist
-        if os.path.exists(self.time_calibration_path):
-            self.time_corrector = PulseTimeCorrection(
-                calib_file_path = self.time_calibration_path)
+        # declare time calibrator None of MC events
+        if self.time_calibration_path ==  "None":
```
MaxNoe, line 82: This should not happen! Either it's a string or it's None, but it should not be a string containing `None`.

Why does this happen?

rlopezcoto:
Why is this PR not analyzed by Travis?
Also @FrancaCassol could you merge the master to solve the current conflict? Thanks

FrancaCassol:
Hi @ruben,
I am testing the code with the new model MC data. I will do soon some further commit with corrections

FrancaCassol:
Hi @rlopezcoto,
I am testing the code with the new model MC data. I will do soon some further commit with corrections

FrancaCassol:
@rlopezcoto for me this PR is ok

lstchain/calib/camera/flatfield.py
rlopezcoto: Solved:
https://github.com/cta-observatory/cta-lstchain/pull/358/commits/ffc249c8a7723295835e9a2f10b9014f63c9a676

lstchain/reco/r0_to_dl1.py
rlopezcoto: Solved:
https://github.com/cta-observatory/cta-lstchain/pull/358/commits/4de59e4a57ae62261ff45bcde67f116481ada4ba

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: Done

lstchain/calib/camera/flatfield.py
```diff
@@ -74,6 +79,20 @@ def __init__(self, **kwargs):
         self.arrival_times = None  # arrival time per event in sample
         self.sample_masked_pixels = None  # masked pixels per event in sample
 
+        # declare time calibrator None if MC events
+        if self.time_calibration_path is None:
```
MaxNoe, line 83: This comment is misleading. It says time_calibration is None for mc, but it does not test if it's mc, it tests if a file was given.

Is a timecalibration path required for observed data? If yes, this script should fail if observed data is analyzed and no time calibration path was given.

If it is not required, the comment should just be something like "don't do time calibration if no calibration file was provided"

lstchain/calib/camera/flatfield.py
```diff
@@ -43,6 +43,11 @@ class FlasherFlatFieldCalculator(FlatFieldCalculator):
         [-3, 3],
         help='Interval (number of std) of accepted charge standard deviation around camera median value'
     ).tag(config=True)
+    time_calibration_path = Unicode(
+        '',
```
MaxNoe, line 47: The `allow_none` only makes sense if the default value is also None. I would recommend to use it this way.

So `Unicode(None, allow_none=True, help='...')`

lstchain/calib/camera/pulse_time_correction.py
```diff
@@ -152,7 +152,9 @@ def get_first_capacitor(self, event, nr):
 
 @njit()
 def get_corr_time_jit(first_cap, fan, fbn, n_harmonics, n_cap):
-    time = fan[0] / 2.
+
+    #time = fan[0] / 2. #commented because time flat-fielding is performed in charge calibrator
+    time = 0
```
pawel21, line 157: Why here time is set to 0?

FrancaCassol: Hi @pawel21,
it is written in the commented line above. As far as i understood this is the flat-field coefficient (can you confirm this?) and we decided to perform it later on at charge calibration level.

lstchain/calib/camera/flatfield.py
FrancaCassol: I took out the comment

lstchain/calib/camera/flatfield.py
FrancaCassol: I put None as default

lstchain/calib/camera/pulse_time_correction.py
pawel21: This is flat-field coefficient in time domain.

rlopezcoto: so is it correct or not?

pawel21: It is correct for me, `fan[0] / 2`is constant factor, so this will only change value of pulse time, but shouldn't affect time correction.

PR #361, opened 2020-04-22
Add pypi upload to travis config

MaxNoe:


MaxNoe:
As discussed with @rlopezcoto, this does not really help until also the protozfitsreader and the ctapipe_io_lst packages are on pypi.

rlopezcoto:
thanks @MaxNoe! what is very important is that [lstchain](https://pypi.org/project/lstchain/) is now accessible via pip:
`pip install lstchain`

MaxNoe:
Let's try this with a test tag after merge

PR #367, opened 2020-04-22
Only test vs ctapipe master and ctapipe_io_lst master

MaxNoe:
Update the travis config so that only tested vs the master branch of ctapipe and that it is required to pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
     # against future changes
     - PYTHONIOENCODING=UTF8
     - MPLBACKEND=Agg
+    - PYTEST_ADDOPTS='--color=yes'
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


MaxNoe:
Why doesn't the travis build show up here?

MaxNoe:
> Why changing to a fixed version of protozfits but keeping ctapipe_io_lst master ?

I only removed an unused environment variable. The protozfits dependency is handled by `ctapipe_io_lst` automatically.

MaxNoe:
Ah, sorry, wrong PR, can be removed here as well.

MaxNoe:
Also please note that this PR is expected to fail on travis, as we do not yet support the current ctapipe master.

.travis.yml
```diff
@@ -1,26 +1,29 @@
 ---
 language: python
 
-python:
-    - 3.6
-    - 3.7
-
 env:
-    - CTAPIPE_VERSION="v0.7.0" CTAPIPE_IO_LST_VERSION="v0.4.1" PROTOZFITS_VERSION=v1.4.2
```
rlopezcoto, line 9: I have here the same comment as in https://github.com/cta-observatory/ctapipe_io_lst/pull/32, could you add the ctapipe v0.8 test as well?

PR #368, opened 2020-04-22
Fix python versions on travis

MaxNoe:
Travis was not actually testing different python versions, because we install anaconda and don't make sure which python version was installed using anaconda, so it was always what came with miniconda.

Currently 3.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ before_install:
 install:
     - conda env create --file environment.yml
     - conda activate lst-dev
+    - python --version
     - |
       pip install \
       https://github.com/cta-observatory/ctapipe/archive/$CTAPIPE_VERSION.tar.gz \
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ before_install:
     - conda info -a  # Useful for debugging any issues with conda
 
 install:
+    - sed -i -e "s/- python=.*/python=$TRAVIS_PYTHON_VERSION/g" environment.yaml
     - conda env create --file environment.yml
     - conda activate lst-dev
     - python --version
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,25 +1,45 @@
...
+    - name: "python=3.7, ctapipe=0.7"
+      python: 3.7
+      env:
+        - CTAPIPE_VERSION="v0.7.0"
+        - CTAPIPE_IO_LST_VERSION="v0.4.1"
+
+    - name: "python=3.6, ctapipe=master"
+      python: 3.6
+      env:
+        - CTAPIPE_VERSION="master"
```
rlopezcoto, line 28: do we really want to test against ctapipe `master`?

environment.yml
```diff
@@ -4,6 +4,7 @@ channels:
   - cta-observatory
   - conda-forge
 dependencies:
+  - python=3.7
```
rlopezcoto, line 7: why would we be testing against python 3.6 and requiring 3.7?

MaxNoe: Again, what is in this environment is not what you require. It is an environment specification.
You can only have a single version of python in an environment.

Which python version you require is set in setup.py (`python_requires` I think), and ctapipe currently supports `>=3.6`, so I thought here it is the same.

.travis.yml
MaxNoe: This was what was there before. I think it is good. It is under allowed failures.

But this gives a nice heads up, that as soon as we are compatible to the next release, when we diverged from master again.

rlopezcoto: ok, but it will start failing very soon after we are even with ctapipe master

PR #370, opened 2020-04-23
Use subprocess to test scripts

MaxNoe:
Here I use subprocess to check if a script ran successfully, not only if the output file is produced.

In case an error happens, the output of the script is reported.

This immediately brought up a test failure in the dl1ab script. The script currently does not finish successfully but fails mit run with this error (I added a print of the number that makes `u.Quantity(` fail:

```
lstchain/scripts/tests/test_lstchain_scripts.py:77: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = ('lstchain_mc_dl1ab', 'testfiles/scripts/dl1_gamma_test_large.simtel.h5', 'testfiles/scripts/dl1ab.h5')
result = CompletedProcess(args=('lstchain_mc_dl1ab', 'testfiles/scripts/dl1_gamma_test_large.simtel.h5', 'testfiles/scripts/dl1...ypeError("The value must be a valid Python or "\nTypeError: The value must be a valid Python or Numpy numeric type.\n')

    def run_program(*args):
        result = sp.run(
            args,
            stdout=sp.PIPE, stderr=sp.STDOUT, encoding='utf-8'
        )
    
        if result.returncode != 0:
            raise ValueError(
>               f'Running {args[0]} failed with return code {result.returncode}'
                f', output: \n {result.stdout}'
            )
E           ValueError: Running lstchain_mc_dl1ab failed with return code 1, output: 
E            /home/maxnoe/.local/anaconda3/envs/lst/lib/python3.7/site-packages/corsikaio/subblocks/dtypes.py:20: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
E             return np.dtype(dict(**dt))
E           /home/maxnoe/.local/anaconda3/envs/lst/lib/python3.7/site-packages/tables/path.py:155: NaturalNameWarning: object name is not a valid Python identifier: 'ASTRICam.__table_column_meta__'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
E             check_attribute_name(name)
E           /home/maxnoe/.local/anaconda3/envs/lst/lib/python3.7/site-packages/tables/path.py:155: NaturalNameWarning: object name is not a valid Python identifier: 'LSTCam.__table_column_meta__'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
E             check_attribute_name(name)
E           /home/maxnoe/.local/anaconda3/envs/lst/lib/python3.7/site-packages/tables/path.py:155: NaturalNameWarning: object name is not a valid Python identifier: 'optics.__table_column_meta__'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
E             check_attribute_name(name)
E           /home/maxnoe/.local/anaconda3/envs/lst/lib/python3.7/site-packages/tables/path.py:155: NaturalNameWarning: object name is not a valid Python identifier: 'FlashCam.__table_column_meta__'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
E             check_attribute_name(name)
E           /home/maxnoe/.local/anaconda3/envs/lst/lib/python3.7/site-packages/tables/path.py:155: NaturalNameWarning: object name is not a valid Python identifier: 'layout.__table_column_meta__'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
E             check_attribute_name(name)
E           {'picture_thresh': 6, 'boundary_thresh': 3, 'keep_isolated_pixels': False, 'min_number_picture_neighbors': 1}
E           0
E           545.1665679253376
E           0.1881637865687579 m
E           0.035333557155097006 m
E           0.19145252893896592 m
E           0.185619rad
E           1.535555735882352
E           1.0973243621378959
E           0.91809rad
E           -0.6749770158117206
E           1.6709347786790794
E           0.7145268337122826
E           0.19145252893896592 m
E           0.14887171941662342
E           0.35243380860840584
E           0.011859838274932614
E           None
E           Traceback (most recent call last):
E             File "/home/maxnoe/.local/anaconda3/envs/lst/bin/lstchain_mc_dl1ab", line 11, in <module>
E               load_entry_point('lstchain', 'console_scripts', 'lstchain_mc_dl1ab')()
E             File "/home/maxnoe/Uni/CTA/cta-lstchain/lstchain/scripts/lstchain_mc_dl1ab.py", line 115, in main
E               params[ii][p] = Quantity(dl1_container[p]).value
E             File "/home/maxnoe/.local/anaconda3/envs/lst/lib/python3.7/site-packages/astropy/units/quantity.py", line 370, in __new__
E               raise TypeError("The value must be a valid Python or "
E           TypeError: The value must be a valid Python or Numpy numeric type.
```


MaxNoe:
To make it clear: the script fails on current master branch, but the tests did not detect the failure because the output file was produced.

MaxNoe:
Ok, I found the error. The `leakage2_pixel` was never filled. But this should demonstrate, how usefull it is to have a test fail if the script is not successful and have the output of the script reported.

vuillaut:
> 
> This immediately brought up a test failure in the dl1ab script. The script currently does not finish successfully but fails mit run with this error (I added a print of the number that makes `u.Quantity(` fail:

See #347

MaxNoe:
@vuillaut looking at the diff, the changes seem unrelated.

MaxNoe:
@vuillaut @rlopezcoto merge? This should provide better tests and error messages for every other PR.

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -12,54 +13,81 @@
 file_model_gh_sep = os.path.join(output_dir, 'cls_gh.sav')
 
 
+def run_program(*args):
```
rlopezcoto, line 16: can this `run_program` function be useful somewhere else so it does not get defined inside a test file?

MaxNoe: Not really. Running this executables like this should only happen in tests of the scripts.

PR #371, opened 2020-04-23
bugfix in source-dependent analysis

SeiyaNozaki:
I found some small bugs in the recent pull request for source dependent analysis.
So this PR fixed them.


lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -67,7 +67,9 @@ def main():
     data = pd.read_hdf(args.datafile, key=dl1_params_lstcam_key)
 
     if config['source_dependent']:
-        data = pd.concat([data, pd.read_hdf(data, key=dl1_params_src_dep_lstcam_key)], axis=1)
+        data = pd.concat([data, pd.read_hdf(args.datafile, key=dl1_params_src_dep_lstcam_key)], axis=1)
+        data = data.drop("index", axis=1)
```
MaxNoe, line 71: Why do you need to drop "index" here?

I think if there is a column "index" in the file, that is not recognized as **the** index when reading it with pandas, either the thing writing it should make something different.

SeiyaNozaki: After adding the source parameters using `lstchain_add_source_dependent_parameters.py`, `index` column is apparently included in `dl1_params_src_dep_lstcam_key`.
Actually `write_dataframe` saved this `index` column.

dl1_src_dep

Maybe due to this `index` column, I got the following error  during ` write_dl2_dataframe` in `lstchain_dl1_to_dl2.py`
```
ValueError: appended items do not match existing items in table!
```

Do you have any suggestions...?

MaxNoe: Most pandas write methods have an `index` keyword argument, that defines if the index should be stored to file. 

Normally you want that, and normally pandas detects that again on reading the file and automatically assigns that column as the `index` of the `DataFrame` you get. 

Somehow, it seems to be duplicated here (observe how the column `index` is identical to the actual index of the DataFrame.

A question would be, if the DataFrame you pass to `write_dataframe` already has the `index` column or if the `write_dataframe` function calls `reset_index`.

MaxNoe: In any case, that column should not be in the file as a column.

SeiyaNozaki: I fixed it by using `set_index` when reading.

rlopezcoto:
This one looks good now, is it ok @MaxNoe or are you planning to work on a better solution also related to #375? In any case I would accept this fix by now

MaxNoe:
Let's merge, I'll make a PR adressing the hdf5 stuff also fixing the index stuff

MaxNoe:
Actually, there is 0 % test coverage here.

Can we test dl1_to_dl2 with `source_dependent=True`?

rlopezcoto:
> Can we test dl1_to_dl2 with source_dependent=True?

Do you want to include the test in this PR? We definitely need to improve the test coverage in general...

PR #375, opened 2020-04-24
Write DL2 files without images

### LINKED ISSUE #372: DL2 files including images and DL1 parameters and tables
By processing some DL1 files, I realized that the size was too large, when I checked their containment, they had the following datasets:
```
'dl1/event/subarray/mc_shower',
 'dl1/event/subarray/trigger',
 'dl1/event/telescope/image/LST_LSTCam',
 'dl2/event/telescope/parameters/LST_LSTCam',
...
```
All the `dl1/*` should have been removed at DL2 level, specially the images that are the ones increasing the size of these files by a large amount. I reproduced this with `v0.4.5` and `master`. I'll look into solving it.

rlopezcoto:
Stripped the images and the copy of the DL1 parameters from the DL2 file. Fixes #372.


lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -96,8 +108,19 @@ def main():
...
     outfile = os.path.join(args.outdir, os.path.basename(args.datafile).replace('dl1','dl2'))
 
+    if os.path.exists(outfile):
+        raise IOError(outfile + ' exists, exiting.')    
+
+    dl1_keys = get_dataset_keys(args.datafile)
+    dl1_keys.remove(dl1_params_lstcam_key)
+    dl1_keys.remove(dl1_images_lstcam_key)
+
+    file = open_file(args.datafile)
```
vuillaut, line 118: Nice.
Maybe just a context manager here.

rlopezcoto: Thanks, done.

MaxNoe:
Why is there 0% diff hit? Isn't the dl1_to_dl2 script tested in `lstchain/scripts/tests_test_scripts`?

rlopezcoto:
> Why is there 0% diff hit? Isn't the dl1_to_dl2 script tested in lstchain/scripts/tests_test_scripts?

Yes, it is:
https://github.com/cta-observatory/cta-lstchain/blob/b27f4d91737bc1acff69d12b9c10c3b8c31154b4/lstchain/scripts/tests/test_lstchain_scripts.py#L64

MaxNoe:
Ah, sorry. probably related to #370 

If you merge master, you should see coverage for the script.

rlopezcoto:
There is something wrong with the addition of source dependent parameters that is not related to #371, now for DL1 files I get, apart from the usual keys:
```
'dl1/event/telescope/parameters_src_dependent/LST_LSTCam',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/abounds',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/bounds',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/indices',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/indicesLR',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/mbounds',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/mranges',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/ranges',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/sorted',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/sortedLR',
 'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/zbounds',
.....
``` 

and hdf reading gets blocked when reaches `'dl1/event/telescope/parameters_src_dependent/_i_LST_LSTCam/alpha/indices'`
I moved the test to add the src_dependent parameters to the end, @SeiyaNozaki any idea what is going on?

rlopezcoto:
The problem with the dataframe containing all these subindex seems to be related to:
https://github.com/cta-observatory/cta-lstchain/blob/b27f4d91737bc1acff69d12b9c10c3b8c31154b4/lstchain/io/io.py#L576

because it is also there when we write DL2 files using:
https://github.com/cta-observatory/cta-lstchain/blob/b27f4d91737bc1acff69d12b9c10c3b8c31154b4/lstchain/io/io.py#L594

(that calls `write_dataframe`), maybe the dataframe is not properly written this way?

rlopezcoto:
I tried to get rid of these indexed keys with the help of @MaxNoe, but I was not able to read later the dataframe that I had written through that key. 
I included a fix directly in `lstchain_dl1_to_dl2` not to read/write these keys.

vuillaut:
@rlopezcoto I think you can replace the `write_dl2_dataframe` function by a simple
```
from astropy.io import Table
tab = Table.from_pandas(dataframe)
tab.write(outfile, path=dl2_params_lstcam_key)
```

rlopezcoto:
@vuillaut the problem I was reporting was not caused by the DL2 writing (although the dataframe written also contain these index keys `_i_*`). In any case, the proposed solution I think it does not work to append that dataframe to the current file.
The solution may be to modify `write_dataframe`, @MaxNoe had some ideas about this, maybe he can look into it. In any case I think this PR can be accepted as it is now because it will work even if we keep writing dataframes as we had done until now.

rlopezcoto:
shall we accept this one before reviewing #381 ?

MaxNoe:
I don't mind the order

MaxNoe:
Tests are still failing though due to the changed command line options

rlopezcoto:
> Tests are still failing though due to the changed command line options

Should be fixed now

rlopezcoto:
I removed the check on tables starting with `_i_` since there is a test checking that now:
https://github.com/cta-observatory/cta-lstchain/blob/24d81bcf7615d8aa3c0589d5b0fa6be41634652a/lstchain/io/tests/test_io.py#L27

rlopezcoto:
I messed with the git history... working it out

rlopezcoto:
Re-opened in #384

Issue #143, opened 2019-06-20
Repository size


vuillaut:
The repository size is currently ~50MB.
This is way too much and probably due to some old data file in the history.

vuillaut:
notebooks are way oversized, probably due to a lot of images dumped there.
```
1,7M	notebooks/Calculate_sensitivity.ipynb
296K	notebooks/Calculate_sensitivity_eventio.ipynb
648K	notebooks/Plot_muon_parameters.ipynb
800K	notebooks/RF-performance.ipynb
4,8M	notebooks/RFcls_GH.sav
1,6M	notebooks/RFreg_Disp.sav
1,6M	notebooks/RFreg_Energy.sav
7,6M	notebooks/Read_and_plot_LST_Data.ipynb
964K	notebooks/Step_by_Step_dl0_to_dl1.ipynb
424K	notebooks/Tests_on_Impact_parameter.ipynb
2,3M	notebooks/calculate_calibration_coefficients.ipynb
1,8M	notebooks/calculate_pedestals.ipynb
216K	notebooks/data_vis.ipynb
172K	notebooks/disp.ipynb
 68K	notebooks/example_container.ipynb
688K	notebooks/gain_estimation.ipynb
188K	notebooks/low_level_corrections.ipynb
476K	notebooks/lstchain.ipynb
 56K	notebooks/mc_energy_distr.ipynb
304K	notebooks/read_LST1_data.ipynb
8,0K	notebooks/tutorial_gammalearn.ipynb
 20K	notebooks/utils_coords
```

A possible solution would be to accept only notebooks with striped output as done in another repo.

rlopezcoto:
I agree, we could do also some cleanup on the already existing notebooks, I'm sure we can reduce their size more than one order of magnitude...

rlopezcoto:
Reduced notebook size in #377. Please do not upload notebooks with size > 1 MB unless there is a good reason for it.

vuillaut:
Sorry but the repo is still 90MB (try a git clone right now).
Cleaning the notebooks was required but does not erase the git history.

rlopezcoto:
yeah, we discussed that in #377. To reduce the repo size we would need to delete the history and one would have to be very selective and careful about that... not sure if it is the way to go.

MaxNoe:
> Sorry but the repo is still 90MB (try a git clone right now).

[...]

PR #377, opened 2020-04-28
Reduce notebook sizes, addressing #143 

rlopezcoto:
As discussed in #143, the repository size is too large due to many images stored in notebooks. I took the following actions:

- I got rid of `Analyze_real_muon_data.ipynb` because it was illustrating how to analyze a muon from DL0 level, while we do it right now from DL1 level and have written a script for that (`lstchain_dl1_muon_analysis.py`)
- I eliminated most of the images from the notebooks that had a size > 1 MB. I propose that from now on, notebooks with a size larger than that, should be justified.
- Some of the notebooks are outdated/not working. I tried to update some, but we also need to update the data files in [cta-lstchain-extra](https://github.com/cta-observatory/cta-lstchain-extra). It would be great if the people that wrote these notebooks could update them.


MaxNoe:
This will actually not reduce repository size, as the notebooks including images will still be present in the history. There is no real solution here, as removing them from the history is not an option for an important project like this.

Notebooks should be stripped of all output before committing.

rlopezcoto:
> Notebooks should be stripped of all output before committing.

Yeah, so let's stick to this: 
> I eliminated most of the images from the notebooks that had a size > 1 MB. I propose that from now on, notebooks with a size larger than that, should be justified.

for the future. I just left a few output images in every notebook and got rid of the rest

Issue #378, opened 2020-04-28
Unify scripts input naming


rlopezcoto:
At the moment we have `infile`, `datafile`, `input_file`... for the input file, I propose to move to:

- `input_file` for the inputs
- `output_file/output_dir` for the outputs
- change all calibration/pedestal/pointing_`path` to `file` if they are indicating a file or `dir` if they are indicating a folder
- any other inconsistencies

Opinions?

rlopezcoto:
closed via #380

PR #380, opened 2020-04-28
Rename variables inside scripts

rlopezcoto:
Working out #378, this will break several scripts calling these ones, but some unification was needed... 
If you have other suggestions for the variable, please let me know, otherwise I'll continue for the rest of them.

### DIFF
--- a/lstchain/scripts/lstchain_add_source_dependent_parameters.py
+++ b/lstchain/scripts/lstchain_add_source_dependent_parameters.py
@@ -29,7 +29,7 @@ parser.add_argument('--input-file', '-f', type=str,
                     )
 
 # Optional arguments
-parser.add_argument('--config-file', '-c', action='store', type=str,
+parser.add_argument('--config', '-c', action='store', type=str,
                     dest='config_file',
                     help='Path to a configuration file for source dependent analysis',
                     default=None
--- a/lstchain/scripts/lstchain_add_source_dependent_parameters.py
+++ b/lstchain/scripts/lstchain_add_source_dependent_parameters.py
@@ -24,7 +24,7 @@ parser = argparse.ArgumentParser(description="Add the source dependent parameter
 
 # Required arguments
 parser.add_argument('--input_file', '-f', type=str,
-                    dest='inpu_file',
+                    dest='input_file',
                     help='path to a DL1 HDF5 file',
                     )
 
--- a/lstchain/scripts/lstchain_add_source_dependent_parameters.py
+++ b/lstchain/scripts/lstchain_add_source_dependent_parameters.py
@@ -1,4 +1,16 @@
-# script to add the source dependent parameters to a DL1 file
+#!/usr/bin/env python3
+
+"""
+Script to add the source dependent parameters to a DL1 file
+
+- Input: DL1 data file.
+- Output: DL1 data file.
+
+Usage: 
+
+$> python lstchain_add_source_dependent_parameters.py --input_file dl1_file.h5
+
+"""
 
 import os
 import argparse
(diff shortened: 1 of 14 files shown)


lstchain/scripts/lstchain_add_source_dependent_parameters.py
```diff
@@ -1,4 +1,16 @@
-# script to add the source dependent parameters to a DL1 file
+#!/usr/bin/env python3
+
+"""
+Script to add the source dependent parameters to a DL1 file
+
+- Input: DL1 data file.
+- Output: DL1 data file.
```
MaxNoe, line 7: technically no output right? It mutates the input

MaxNoe:
It's very unusual for command line arguments to have underscores. It should be minuses.

This will also be cleaned up in ctapipe to conform to POSIX. 

Also, a single minus is only for single letter options, and this is combinable, so `-fd` is supposed to be the same as `-f -d`.

lstchain/scripts/lstchain_add_source_dependent_parameters.py
rlopezcoto: right, I changed the description

rlopezcoto:
@MaxNoe thanks for the tips, I changed underscores for hyphens. I'll modify the other scripts the same way

lstchain/scripts/lstchain_add_source_dependent_parameters.py
```diff
@@ -1,14 +1,14 @@
 #!/usr/bin/env python3
 
 """
-Script to add the source dependent parameters to a DL1 file
+Script to add the source dependent parameters to a DL1 file.
 
-- Input: DL1 data file.
-- Output: DL1 data file.
+- Input: DL1 data file. The program returns the same DL1 data file with
```
MaxNoe, line 6: It does not return. It modifies in place. 

Maybe: `Input: DL1 data file. Source dependent parameters will be added to this file.`

lstchain/scripts/lstchain_mc_dl1ab.py
```diff
@@ -19,31 +32,36 @@
...
+                    dest='input_file',
+                    help='path to the DL1a file ',
+                    default=None, required=True)
+
+parser.add_argument('--output-file', '-o', action='store', type=str,
+                    dest='output_file',
+                    help='key for the table of new parameters',
+                    default=None, required=True)
+# Optional arguments
+parser.add_argument('--config_file', '--conf', action='store', type=str,
```
garciagenrique, line 49: I think this argument should be `--config-file` to be in agreement with the rest of the changes

lstchain/scripts/lstchain_data_create_time_calibration_file.py
```diff
@@ -7,35 +22,34 @@
...
 
-parser.add_argument('--config_file', '-conf', action='store', type=str,
+parser.add_argument('--config_file', '-c', action='store', type=str,
                     dest='config_file',
                     help='Path to a configuration file. If none is given, a standard configuration is applied',
                     default=None
                     )
 
-parser.add_argument('--pedestal_file', '-ped', action='store', type=str,
+parser.add_argument('--pedestal_file', '-p', action='store', type=str,
```
garciagenrique, line 52: Same here (and 6 lines above - line 46) ?. 'argument variables' with `-` and destination variables with `_` ?

lstchain/scripts/lstchain_mc_dl1ab.py
MaxNoe: ctapipe `Tool` uses `--config`, maybe we can bring this in line already

lstchain/scripts/lstchain_data_create_time_calibration_file.py
rlopezcoto: I did not want to modify the destination variables because they are called later in the code and I wanted to modify the written code the minimum possible.

lstchain/scripts/lstchain_add_source_dependent_parameters.py
rlopezcoto: done

lstchain/scripts/lstchain_mc_dl1ab.py
rlopezcoto: well spotted. I'm also moving all `--config-file` to `--config`

rlopezcoto:
Thanks for all the comments, I implemented all suggestions and I'm done with the changes now

PR #381, opened 2020-04-29
Use pytables to write dataframes

MaxNoe:
This should fix some issues @rlopezcoto experienced.

Pandas uses some not-so-straightforward layouts when writing dataframes into hdf5  tables.

This uses pytables directly, creating a simple hdf5 dataset that is still easily readable by pandas, h5py, tables, astropy and even cpp


MaxNoe:
I added an explicit test for that

vuillaut:
@MaxNoe the base branch was in lstchain repo. It can be deleted?

MaxNoe:
sure

PR #383, opened 2020-04-29
Add functions to parse and convert filenames

MaxNoe:


lstchain/paths.py
```diff
@@ -0,0 +1,52 @@
...
+    '''Parse an raw data file name and return a nametuple of it's components'''
+    m = DL1_RE.match(os.path.basename(filename))
+
+    if m is None:
+        raise ValueError(f'Filename {filename} does not match pattern {R0_RE}')
+
+    return Run(*map(parse_int, m.groups()))
+
+
+def r0_run_to_filename(run):
```
vuillaut, line 45: If the function should be used by users, at least state what is run.

MaxNoe:
> These path solutions are very specific for the real data format that we currently have, wouldn't it be better if they go into io?

Munging file names is not really io, right? Why not just leave it in their own namespace.

Yes, I will look through the code (probably only tomorrow), and look for places where it would be appropriate to use these functions.

lstchain/paths.py
MaxNoe: I'll add the docstrings

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -10,8 +10,8 @@
 import subprocess as sp
 
 output_dir = os.path.join(test_dir, 'scripts')
-dl1_file = os.path.join(output_dir, 'dl1_gamma_test_large.simtel.h5')
```
rlopezcoto, line 13: Just for people to know, we are also getting rid of the `simtel` string:
https://github.com/cta-observatory/cta-lstchain/pull/344#discussion_r410127986

MaxNoe: I can put it back in, if you want. There was some previous discussions on it. The behaviour before was not well-defined. It would be removed for uncompressed simtel files (without .gz or .zst) but left in for compressed simtel files.

rlopezcoto: no, I'm fine with this, as I said in the comment from the previous PR, but just wanted for people dealing with simulations to be aware

PR #384, opened 2020-04-30
Write DL2 files without images

### LINKED ISSUE #372: DL2 files including images and DL1 parameters and tables

rlopezcoto:
Substitutes #375 (that fixes #372) in which I messed with the history of the repo...

### DIFF
--- a/lstchain/scripts/lstchain_dl1_to_dl2.py
+++ b/lstchain/scripts/lstchain_dl1_to_dl2.py
@@ -20,7 +20,6 @@ import astropy.units as u
 import numpy as np
 import os
 import pandas as pd
-import shutil
 
 from tables import open_file
 from astropy.table import Table
@@ -28,15 +27,14 @@ from sklearn.externals import joblib
 from lstchain.reco.utils import filter_events, impute_pointing
 from lstchain.reco import dl1_to_dl2
 from lstchain.io import (
-    read_configuration_file, 
-    standard_config, 
+    read_configuration_file,
+    standard_config,
     replace_config,
     write_dl2_dataframe,
     get_dataset_keys,
-    auto_merge_h5files,
 )
 from lstchain.io.io import (
-    dl1_params_lstcam_key, 
+    dl1_params_lstcam_key,
     dl1_params_src_dep_lstcam_key,
     dl1_images_lstcam_key
 )
@@ -119,7 +117,7 @@ def main():
     output_file = os.path.join(args.output_dir, os.path.basename(args.input_file).replace('dl1','dl2'))
 
     if os.path.exists(outfile):
-        raise IOError(outfile + ' exists, exiting.')    
+        raise IOError(outfile + ' exists, exiting.')
 
     dl1_keys = get_dataset_keys(args.datafile)
     dl1_keys.remove(dl1_params_lstcam_key)
--- a/lstchain/scripts/lstchain_dl1_to_dl2.py
+++ b/lstchain/scripts/lstchain_dl1_to_dl2.py
@@ -120,7 +120,9 @@ def main():
     dl1_keys = get_dataset_keys(args.input_file)
     dl1_keys.remove(dl1_images_lstcam_key)
     dl1_keys.remove(dl1_params_lstcam_key)
-    dl1_keys.remove(dl1_params_src_dep_lstcam_key)
+
+    if config['source_dependent']:
+        dl1_keys.remove(dl1_params_src_dep_lstcam_key)
 
     with open_file(args.input_file, 'r') as h5in:
         with open_file(output_file, 'a') as h5out:
--- a/lstchain/scripts/lstchain_dl1_to_dl2.py
+++ b/lstchain/scripts/lstchain_dl1_to_dl2.py
@@ -21,7 +21,6 @@ import os
 import pandas as pd
 
 from tables import open_file
-from astropy.table import Table
 from sklearn.externals import joblib
 from lstchain.reco.utils import filter_events, impute_pointing
 from lstchain.reco import dl1_to_dl2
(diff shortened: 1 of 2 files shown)


rlopezcoto:
can we accept this one?

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -106,8 +116,20 @@ def main():
     os.makedirs(args.output_dir, exist_ok=True)
     output_file = os.path.join(args.output_dir, os.path.basename(args.input_file).replace('dl1','dl2'))
 
+    if os.path.exists(output_file):
+        raise IOError(output_file + ' exists, exiting.')
+
+    dl1_keys = get_dataset_keys(args.input_file)
```
MaxNoe, line 122: Wouldn't it be better to define what we want to have in DL2 instead of copying and removing certain things from DL1?

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -106,8 +116,20 @@ def main():
...
+    dl1_keys.remove(dl1_params_lstcam_key)
+    dl1_keys.remove(dl1_images_lstcam_key)
+    if config['source_dependent']:
+        dl1_keys.remove(dl1_params_src_dep_lstcam_key)
+
+    with open_file(args.input_file) as file:
```
MaxNoe, line 128: If you open both files using tables, you can copy datasets:

http://www.pytables.org/usersguide/libref/file_class.html?highlight=copy#tables.File.copy_children

This will be much more efficient than parsing into a table and writing again.
This will also directly preserve metadata, which might be lost with this approach here, right?

MaxNoe:
```
In [7]: with tables.open_file('test.hdf5', 'r') as infile, tables.open_file('out.hdf5', 'w') as out: 
   ...:     g = out.create_group('/', 'data') 
   ...:     infile.copy_node('/data/foo', g) 
```

rlopezcoto: why would metadata be lost?

lstchain/scripts/lstchain_dl1_to_dl2.py
rlopezcoto: At the moment in DL1 we have:
```
 'dl1/event/subarray/mc_shower',
 'dl1/event/subarray/trigger',
 'dl1/event/telescope/image/LST_LSTCam',
 'dl1/event/telescope/parameters_src_dependent/LST_LSTCam',
 'instrument/subarray/layout',
 'instrument/subarray/layout.__table_column_meta__',
 'instrument/telescope/camera/ASTRICam',
 'instrument/telescope/camera/ASTRICam.__table_column_meta__',
 'instrument/telescope/camera/FlashCam',
 'instrument/telescope/camera/FlashCam.__table_column_meta__',
 'instrument/telescope/camera/LSTCam',
 'instrument/telescope/camera/LSTCam.__table_column_meta__',
 'instrument/telescope/optics',
 'instrument/telescope/optics.__table_column_meta__',
 'simulation/run_config',
 'simulation/thrown_event_distribution'
```
and we are getting rid of:
```
 'dl1/event/telescope/image/LST_LSTCam',
 'dl1/event/telescope/parameters_src_dependent/LST_LSTCam',
```
anything else that should not be written (this is the gamma_test file, that is why we have information from several cameras written)?

lstchain/scripts/lstchain_dl1_to_dl2.py
MaxNoe: Because `file.root[k][:]` is a numpy structures array, so where should `Table` get the hdf5 attributes from?

rlopezcoto: ok, that sounds a good solution, but maybe we can accept this one for the time being and then propose a general solution to substituted `write_dataframes`, that is used not only here, but also on `lstchain_add_source_dependent_parameters.py` and probably somewhere else, does that sound ok?

lstchain/scripts/lstchain_dl1_to_dl2.py
vuillaut: > Wouldn't it be better to define what we want to have in DL2 instead of copying and removing certain things from DL1?

I disagree on that point. Right now the definition of the data model is going in a bottom-up approach and things are constantly added at the R1/DL0/DL1 levels. In my opinion, at the moment we should keep all the information downward.

lstchain/scripts/lstchain_dl1_to_dl2.py
MaxNoe: I gave you the solution above that directly preserves metadata.

Is there a problem with that code? If yes, let's merge and I'll try to address that, if not, let's just use it directly.

lstchain/scripts/lstchain_dl1_to_dl2.py
MaxNoe: Sounds good.

lstchain/scripts/lstchain_dl1_to_dl2.py
MaxNoe: Does this work?

```
with tables.open_file(args.input_file, 'r') as h5in:
    with tables.open_file(output_file, 'a') as h5out:

        for k in dl1_keys:
            path, name = k.rsplit('/', 1)

            if path not in h5out:
                grouppath, groupname = path.rsplit('/', 1)
                g = h5out.create_group(
                    grouppath, groupname, createparents=True
                )
            else:
                g = h5out.get_node(path)

            h5in.copy_node(k, g, overwrite=True)
```

MaxNoe: I checked, and this preserves hdf5 attributes in the datasets

rlopezcoto: Thanks for the solution @MaxNoe. Juste needed to add:
```
if not k.startswith('/'):
    k = '/' + k
```
because pytables was complaining

Issue #386, opened 2020-04-30
Error running lstchain_data_r0_to_dl1


rlopezcoto:
@FrancaCassol was doing some tests after the last merge of master and found a problem. I could not reproduced it using the latest data:
`lstchain_data_r0_to_dl1 -f ~/Downloads/LST-1.1.Run02034.0000.fits.fz --pedestal-file ~/Downloads/drs4_pedestal.Run2005.0000.fits --calibration-file ~/Downloads/calibration.Run2006.0000.hdf5 --time-calibration ~/Downloads/time_calibration.Run2006.0000.hdf5`

but it was there when reading older data:
```
lstchain_data_r0_to_dl1 -f ~/Downloads/LST-1.4.Run01566.0326.fits.fz --pedestal-file ~/Downloads/drs4_pedestal.Run2005.0000.fits --calibration-file ~/Downloads/calibration.Run2006.0000.hdf5 --time-calibration ~/Downloads/time_calibration.Run2006.0000.hdf5 
/Users/rubenlopez/anaconda/envs/cta-dev/lib/python3.6/site-packages/corsikaio/subblocks/dtypes.py:20: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  return np.dtype(dict(**dt))
EXTRACTOR <ctapipe.image.extractor.LocalPeakWindowSum object at 0x1a2ac22470>
EXTRACTOR <ctapipe.image.extractor.GlobalPeakWindowSum object at 0x1088385f8>
USING FILTERS:  Filters(complevel=5, complib='blosc:zstd', shuffle=True, bitshuffle=False, fletcher32=True, least_significant_digit=None)
0
Traceback (most recent call last):
  File "/Users/rubenlopez/anaconda/envs/cta-dev/lib/python3.6/site-packages/astropy/units/quantity_helper/converters.py", line 181, in converters_and_unit
    if can_have_arbitrary_unit(args[i]):
  File "/Users/rubenlopez/anaconda/envs/cta-dev/lib/python3.6/site-packages/astropy/units/quantity_helper/converters.py", line 122, in can_have_arbitrary_unit
    return np.all(np.logical_or(np.equal(value, 0.), ~np.isfinite(value)))
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

[...]

PR #388, opened 2020-04-30
Nan timestamp

MaxNoe:
As discussed in #386, adding support for nan values in `unix_tai_to_utc`

I also renamed the function `unix_tai_to_time` as it does not actually convert to utc.
It correctly creates an `astropy.Time` object, that is still in tai but converts to unix utc timestamp when accessing `.unix`.

For convenience and clarity, I would suggest to compare times to `INVALID_TIME` as now defined in the module.


MaxNoe:
@morcuended it deals with nan like it would be zero.

morcuended:
> @morcuended it deals with nan like it would be zero.
> 
> So your choice now, what you want `timestamp` to be if it is missing, the conversion functin now supports both

not strong preference, I would leave it as it is now with `nan` provided it works

rlopezcoto:
what about the interpolation, is it dealing with it properly?

MaxNoe:
Can you point me to the interpolation code?

morcuended:
> Can you point me to the interpolation code?

functions `linear_imputer` and `impute_pointing` in `lstchain/reco/utils.py` which interpolate missing alt/az coordinates directly. Thus missing time values should not be a problem.

morcuended:
May you update the branch with the last merge fixing the muon output filename so I can test it?

MaxNoe:
> Thus missing time values should not be a problem.

These are not missing anymore but `1970-01-01`, so maybe you need to explicitly filter with` time != INVALID_TIME`

rlopezcoto:
@morcuended did you test this? can we move ahead with the merge?

morcuended:
> @morcuended did you test this? can we move ahead with the merge?

Yes, I tested that the script runs well without problems and files are produced normally. However, I did not have the time to look into the time values themselves.

PR #396, opened 2020-05-04
Fix interleaved events code

FrancaCassol:
Two corrections:
- correct the configuration given to the calibration calculator, at present the flat-field and pedestal Component do not get the correct cards
- Check that some statistic is collected (at least one event) before storing the interleaved results

### DIFF
--- a/lstchain/calib/camera/pedestals.py
+++ b/lstchain/calib/camera/pedestals.py
@@ -164,7 +164,8 @@ class PedestalIntegrator(PedestalCalculator):
         """
 
         if self.num_events_seen == 0:
-            raise ValueError("No pedestal events in statistics, zero results")
+            self.log.info("No pedestal events in statistics, zero results")
+            return False
 
         container = event.mon.tel[self.tel_id].pedestal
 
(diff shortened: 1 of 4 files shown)


lstchain/calib/camera/pedestals.py
```diff
@@ -163,6 +164,7 @@ def store_results(self, event):
          event : general event container
         """
 
+        # something wrong if you are here and no statistic is there
         if self.num_events_seen == 0:
             raise ValueError("No pedestal events in statistics, zero results")
```
moralejo, line 169: This will exit the execution of the script, simply if there are no interleaved pedestals? 
If so, please transform it into a warning. I am pretty sure we will have otherwise valuable data without interleaved pedestals, without interleaved flatfield events, with none of those, and even with both of them, in some lucky cases.

FrancaCassol: You should not call this function if you do not have collected statistics.  The check must be done before.

moralejo: > You should not call this function if you do not have collected statistics. The check must be done before.

Ok, good, but with your latest fix, is that the case when one runts r0_to_dl1 script?

FrancaCassol:
This was not a good idea. Things had still to be fixed. I will do an other PR.

rlopezcoto:
I can still revert it, shall I do it?

FrancaCassol:
Don't worry, let go for an other one.

MaxNoe:
@FrancaCassol If something is not ready yet, you should open it as "Draft PR". The Open PR button has a dropdown for that option.

FrancaCassol:
Thanks @MaxNoe, interesting to know.

MaxNoe:
There is also the "WIP" label for "Work in Progress"

PR #397, opened 2020-05-05
Solve dl2 and dl1_src_dependent writing

rlopezcoto:
With #384, and discussing what was the best way to write the DL1 info and which one to write in the DL2 file, we actually forgot to write the DL2 information (!)
Also now, if src_dependent parameters are present in the DL1 file, they are removed in the DL2 one (regardless of whether they are used or not in the dl1_to_dl2 passage)

### DIFF
--- a/lstchain/scripts/lstchain_dl1_to_dl2.py
+++ b/lstchain/scripts/lstchain_dl1_to_dl2.py
@@ -144,7 +144,7 @@ def main():
 
                 h5in.copy_node(k, g, overwrite=True)
 
-    write_dl2_dataframe(dl2.astype(float), output_file)
+    write_dl2_dataframe(dl2, output_file)
 
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -142,6 +144,7 @@ def main():
 
                 h5in.copy_node(k, g, overwrite=True)
 
+    write_dl2_dataframe(dl2.astype(float), output_file)
```
vuillaut, line 147: Why convert dl2 as float?

MaxNoe: Yes, this does not make sense. There are integer columns in there and they should stay integers.

rlopezcoto: sorry, just copied the line from the old version of the file

PR #400, opened 2020-05-05
Add generic path parsing function, factor out match conversion

MaxNoe:


rlopezcoto:
Just seen this PR, maybe [this](https://github.com/cta-observatory/cta-lstchain/pull/382#issuecomment-624319863) could have been brought up here.

PR #402, opened 2020-05-11
Solve matplotlib style problem that was making labels non-legible

rlopezcoto:
Solution to work out the problem with the matplotlib style/backend that was described in #401. While this is worked out, these lines should force the matplotlib style to the 'default' one, that was working before for all the plotting.

### DIFF
--- a/lstchain/__init__.py
+++ b/lstchain/__init__.py
@@ -11,5 +11,3 @@ from .io import standard_config
 from .version import get_version
 __version__ = get_version(pep440=False)
 
-import matplotlib as mpl
-mpl.style.use('default')
--- a/lstchain/__init__.py
+++ b/lstchain/__init__.py
@@ -12,4 +12,4 @@ from .version import get_version
 __version__ = get_version(pep440=False)
 
 import matplotlib as mpl
-mpl.use.style('default')
+mpl.style.use('default')
--- a/lstchain/__init__.py
+++ b/lstchain/__init__.py
@@ -10,3 +10,6 @@ from .io import standard_config
 
 from .version import get_version
 __version__ = get_version(pep440=False)
+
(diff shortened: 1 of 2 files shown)


vuillaut:
Hi
- I just shipped ctaplot 0.5.2 fixing the styling issue - one now can actively set ctaplot style with `ctaplot.set_style()` if desired
- We can still force default style in lstchain to avoid problems in the future.

MaxNoe:
Both these solution are not ideal.

matplotlib supports context manager for setting styles. Do not set them globally! And don't override user defined styles by setting it back to defaults.

This breaks any matplotlib mechanism to set styles outside of code (environment variables, local matplotlibrc, global matplotlibrc).

vuillaut:
> This breaks any mechanism to set matplotlib options outside of code.
> 
> Don't mess with these things, please just fix the underlying issue.

Well, as said, if the problem was indeed ctaplot, I think it is fixed.

rlopezcoto:
Agreed that this may not be ideal, but at least we need a temporary solution for all the datacheck/calibration plots to be displayed. 
We can discuss a better, more general solution to remove those lines in the future, but we really need to move on with the release right now, and without this fix, the next set of reprocessed data will have very bad style, not legible plots...

rlopezcoto:
ok, I tested `ctaplot=0.5.2` and the style is again ok, so I removed the forcing of the matplotlib "default" style. 
Still, we may want to think on a way of avoiding these problems in the future...

vuillaut:
> ok, I tested `ctaplot=0.5.2` and the style is again ok, so I removed the forcing of the matplotlib "default" style.
> Still, we may want to think on a way of avoiding these problems in the future...

Glad this is fixed and sorry for the mess.

PR #403, opened 2020-05-11
Fix use of UCTS trigger in interleaved pedestals events

FrancaCassol:
Remove the check of UCTS trigger in interleaved pedestal events, because often wrong.

### DIFF
--- a/lstchain/calib/camera/calibration_calculator.py
+++ b/lstchain/calib/camera/calibration_calculator.py
@@ -206,11 +206,8 @@ class LSTCalibrationCalculator(CalibrationCalculator):
         new_ped = False
         new_ff = False
 
-        # if pedestal event (check also the ucts trigger for the moment)
-        if (
-                LSTEventType.is_pedestal(event.r1.tel[self.tel_id].trigger_type) or
-                LSTEventType.is_pedestal(event.lst.tel[self.tel_id].evt.ucts_trigger_type)
-        ) :
+        # if pedestal event
+        if LSTEventType.is_pedestal(event.r1.tel[self.tel_id].trigger_type): 
 
             new_ped = self.pedestal.calculate_pedestals(event)
 


lstchain/calib/camera/calibration_calculator.py
```diff
@@ -206,11 +206,8 @@ def process_interleaved(self, event):
         new_ped = False
         new_ff = False
 
-        # if pedestal event (check also the ucts trigger for the moment)
-        if (
-                LSTEventType.is_pedestal(event.r1.tel[self.tel_id].trigger_type) or
-                LSTEventType.is_pedestal(event.lst.tel[self.tel_id].evt.ucts_trigger_type)
```
MaxNoe, line 212: To understand this: the ucts_trigger_type wrongly identifies an event as a pedestal although it is not?

How does that happen?

DirkHoffmann:
> Remove the check of UCTS trigger

Is this because the TTYPE is wrong, or is the TIB information correct in these casees?

lstchain/calib/camera/calibration_calculator.py
rlopezcoto: there is a hardware problem that is being worked out. Until then, we cannot trust on this flag

MaxNoe: @DirkHoffman seems to disagree that this is a hardware problem

rlopezcoto: There have been several problems of UCTS/TIB bit disappearance at EVB and/or event class misidentification and there are many open discussion in camera/analysis calls.
The bottomline is that we cannot trust this flag for the time being.

FrancaCassol:
> > Remove the check of UCTS trigger
> 
> Is this because the TTYPE is wrong, or is the TIB information correct in these casees?

For the cases I could see the TIB information is either correct or absent.
But, we will have a more global  view of that after the next data processing.

Issue #406, opened 2020-05-11
error converting MC files where IGNORE_TELESCOPEs option from simtel was used


yrenier:
I have corsika files with 7 telescopes and used IGNORE_TELESCOPEs=1,2,7 to keep only the LSTs (telescopes 3, 4, 5 and 6) during the sim_telarray simulations.
When I try to get dl1 datafiles using lstchain_mc_r0_to_dl1, I get an error:

```
/usr/bin/python3 /home/yves/ctasoft/cta-lstchain/lstchain/scripts/lstchain_mc_r0_to_dl1.py --infile /home/yves/Downloads/proton_run170_South.simtel.gz --outdir /home/yves/Downloads --config=lstchain/data/lstchain_own_prod_config_PMT.json
/home/yves/.local/lib/python3.7/site-packages/ctapipe/io/simteleventsource.py:164: UserWarning: Backseeking to start of file.
  warnings.warn('Backseeking to start of file.')
USING FILTERS:  Filters(complevel=0, shuffle=False, bitshuffle=False, fletcher32=False, least_significant_digit=None)
0
Traceback (most recent call last):
  File "/home/yves/ctasoft/cta-lstchain/lstchain/scripts/lstchain_mc_r0_to_dl1.py", line 48, in <module>
    main()
  File "/home/yves/ctasoft/cta-lstchain/lstchain/scripts/lstchain_mc_r0_to_dl1.py", line 44, in main
    r0_to_dl1.r0_to_dl1(args.infile, output_filename=output_filename, custom_config=config)
  File "/home/yves/ctasoft/cta-lstchain/lstchain/reco/r0_to_dl1.py", line 308, in r0_to_dl1
    write_subarray_tables(writer, event, metadata)
  File "/home/yves/ctasoft/cta-lstchain/lstchain/io/io.py", line 570, in write_subarray_tables
    writer.write(table_name="subarray/trigger", containers=[event.dl0, event.trig])
  File "/home/yves/.local/lib/python3.7/site-packages/ctapipe/io/hdf5tableio.py", line 279, in write
    self._append_row(table_name, containers)
  File "/home/yves/.local/lib/python3.7/site-packages/ctapipe/io/hdf5tableio.py", line 253, in _append_row
    value = self._apply_col_transform(table_name, colname, value)
  File "/home/yves/.local/lib/python3.7/site-packages/ctapipe/io/tableio.py", line 115, in _apply_col_transform
    value = tr(value)
[...]

PR #408, opened 2020-05-11
Enable the usage of simtel files produced with IGNORE_TELESCOPEs option

yrenier:
As explained in #406 , this PR allows the treatment of MC files where IGNORE_TELESCOPEs option from simtel was used.


rlopezcoto:
Great Issue+solution @yrenier !

PR #338, opened 2020-04-09
Improve calibration

yrenier:
PR #331 has been cancelled, but there has been several proposed and implemented improvements that I belive should not be lost, so this new PR is to include them.

PR #409, opened 2020-05-14
adding if condition to remove keys

vuillaut:
DL1 to DL2 fails on merged DL1 files where images have been removed.

Fixed this and added unit tests for merged DL1 files.


MaxNoe:
I would suggest a bug fix release, since no other PRs have been merged yet.

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -103,6 +104,27 @@ def test_lstchain_mc_rfperformance():
     assert os.path.exists(file_model_energy)
 
 
+@pytest.mark.run(after='test_lstchain_mc_r0_to_dl1')
+def test_lstchain_merge_dl1_hdf5_files():
+    run_program('lstchain_merge_hdf5_files',
```
MaxNoe, line 109: Where does it get it's input files from?

MaxNoe:
Apperently the next version of sklearn has been released with the removal of joblib. You just need to replace the imports of joblib with `import joblib` and add it to the dependencies.

lstchain/scripts/tests/test_lstchain_scripts.py
vuillaut: merge all hdf5 files in the dir passed with `-d`.
here `output dir`

MaxNoe: Ah, was confused by the `output_dir`, so here it's actually the input

vuillaut:
> Apperently the next version of sklearn has been released with the removal of joblib. You just need to replace the imports of joblib with `import joblib` and add it to the dependencies.

shall we fix the version of joblib to ensure compatibility of the models?

MaxNoe:
@vuillaut I don't think joblib affects compatibility. I tested this for the aict tools and there were no problems with loading models stored with older versions of joblib with newer versions of joblib.

The story is different for sklearn, so you should always load a model with the same version of sklearn that was used to train it.

rlopezcoto:
> Apperently the next version of sklearn has been released with the removal of joblib. You just need to replace the imports of joblib with import joblib and add it to the dependencies.

The `joblib` issue is addressed here:
https://github.com/cta-observatory/cta-lstchain/pull/338/files
as soon as [this](https://github.com/cta-observatory/cta-lstchain/pull/338/files#r425006523) is addressed, it can be merged.

MaxNoe:
As the other PR also does other changes, I would rather merge this here fast, that be simpler, right?

rlopezcoto:
> As the other PR also does other changes, I would rather merge this here fast, that be simpler, right?

but without doing any joblib modification, right? Then I agree (as soon as it passes travis checks)

MaxNoe:
@vuillaut with pickle protocol 5 (python 3.8) and numpy 1.18, joblib becomes obsolete and we can just use the stdlib pickle module

MaxNoe:
> but without doing any joblib modification, right? Then I agree (as soon as it passes travis checks)

Why? no need to revert this. The changes are even exactly the same, so it won't even create a merge conflict.

morcuended:
> damn, that one was missed... it means that for this production, we either have to merge at a dl2 level or release a bugfix version for the processing... @morcuended were you thinking on performing the merge before producing the `DL2`?

No, we also produce DL2 files subrun-wise

morcuended:
Is the problem with missing joblib module fixed in this PR?

rlopezcoto:
> Why? no need to revert this. The changes are even exactly the same, so it won't even create a merge conflict.

I did not see that it was already [included](https://github.com/cta-observatory/cta-lstchain/pull/409/commits/e35c3d65ead280d89cf8770adc22fa959cf8cce9). So fine then

vuillaut:
- Concerning joblib, I made the changes already don't worry
- Concerning merge, it's a problem for the IRF production at least (and needs to be fixed anyway)

rlopezcoto:
> Is the problem with missing joblib module fixed in this PR?

should be

> No, we also produce DL2 files subrun-wise

it seems that @vuillaut needs this, so I'll do the v0.5.1 release and install it in the IT cluster

vuillaut:
> @vuillaut with pickle protocol 5 (python 3.8) and numpy 1.18, joblib becomes obsolete and we can just use the stdlib pickle module

good to know.
To keep in mind when we support only python >= 3.8 ;-)

vuillaut:
Ready for review.

vuillaut:
And we are celebrating 100th unit test 🥳

PR #410, opened 2020-05-15
add benchmark charge script

vuillaut:
Benchmark script to check calibration.

Make the following plots:
threshold20

### DIFF
--- a/lstchain/scripts/benchmarks/charge_benchmark.py
+++ b/lstchain/scripts/benchmarks/charge_benchmark.py
@@ -1,0 +1,105 @@
+
+
+import argparse
+import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


lstchain/scripts/benchmarks/charge_benchmark.py
```diff
@@ -0,0 +1,105 @@
...
+                    default=None
+                    )
+
+
+args = parser.parse_args()
+
+
+def main():
+
+    ctaplot.set_style()
```
rlopezcoto, line 49: I hope this does not break anything 😄

MaxNoe: Matplotlib supports context managers for style changes.

Issue #416, opened 2020-05-22
Multiprocessing error in check_dl1


morcuended:
Here is another error I've been getting during the reprocessing. In this case, it seems to be related to the multiprocessing part in the `check_dl1` function included in the r0_to_dl1 step. When this error raises, datacheck h5 file is not produced whereas the DL1 and muon files are normally generated.

```python
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/numpy/lib/histograms.py", line 1018, in histogramdd
    N, D = sample.shape
AttributeError: 'list' object has no attribute 'shape'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/multiprocessing/pool.py", line 47, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/lstchain/datachecks/dl1_checker.py", line 280, in process_dl1_file
    geom, bins)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/lstchain/datachecks/containers.py", line 205, in fill_event_wise_info
    bins=histogram_binnings.
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/matplotlib/pyplot.py", line 2621, in hist2d
    data is not None else {}), **kwargs)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/matplotlib/__init__.py", line 1565, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/matplotlib/cbook/deprecation.py", line 296, in wrapper
    return func(*args, **kwargs)
[...]

Issue #414, opened 2020-05-21
Error when merging datacheck h5 files due to mistyped muon info


morcuended:
For some of the runs I found out that an error related to muon info raises when running `lstchain_check_dl1`

For example:
```
$ lstchain_check_dl1 --input-file="/fefs/aswg/data/real/DL1/20200125/v0.5.1_v03/datacheck_dl1_LST-1.Run01863.*.h5"
```

```python
Traceback (most recent call last):
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/astropy/table/operations.py", line 858, in _vstack
    out[out_name] = col_cls.info.new_like(cols, n_rows, metadata_conflicts, out_name)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/astropy/table/column.py", line 174, in new_like
    ('meta', 'unit', 'format', 'description'))
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/astropy/utils/data_info.py", line 642, in merge_cols_attributes
    out['dtype'] = metadata.common_dtype(cols)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/astropy/utils/metadata.py", line 64, in common_dtype
    raise tme
astropy.utils.metadata.MergeConflictError: Arrays have incompatible types ['float64', 'bool']

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/bin/lstchain_check_dl1", line 8, in <module>
    sys.exit(main())
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/lstchain/scripts/lstchain_check_dl1.py", line 102, in main
    plot_datacheck(filenames, args.output_dir, args.muons_dir)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/lstchain/datachecks/dl1_checker.py", line 819, in plot_datacheck
    muons_table = vstack([muons_table, t])
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/astropy/table/operations.py", line 292, in vstack
    out = _vstack(tables, join_type, col_name_map, metadata_conflicts)
[...]

PR #417, opened 2020-05-22
Safeguard in DL1 data check against empty muon files + solved pyplot related issue

moralejo:
Solves issues occurring when muon files have no entries, as well as a strange pyplot.hist2d error, which failed very few times (I don't know why) when inputs were pandas data series (instead of numpy arrays).

This hopefully addresses issues #414 and #416. Tested it on some of the failing files, I think it should work for all of them.

### DIFF
--- a/lstchain/datachecks/containers.py
+++ b/lstchain/datachecks/containers.py
@@ -159,7 +159,7 @@ class DL1DataCheckContainer(Container):
                                  bins=histogram_binnings.hist_nislands)
         self.hist_nislands = counts
 
-        intensity = table['intensity'][mask]
+        intensity = np.array(table['intensity'][mask])
         counts, _, _ = plt.hist(intensity,
                                 bins=histogram_binnings.hist_intensity)
         self.hist_intensity = counts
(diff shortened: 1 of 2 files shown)


lstchain/datachecks/containers.py
```diff
@@ -173,35 +173,38 @@ def fill_event_wise_info(self, subrun_index, table, mask, geom,
...
-                                     table['skewness'][mask],
+        counts, _, _, _ = plt.hist2d(intensity,
+                                     np.array(table['skewness'][mask]),
                                      bins=histogram_binnings.hist_skewness)
         self.hist_skewness = counts
 
-        intercept = table['intercept'][mask]
         counts, _, _, _ = \
-            plt.hist2d(intensity, intercept,
+            plt.hist2d(intensity, np.array(table['intercept'][mask]),
```
MaxNoe, line 190: is tables a pandas series? if yes, `table.loc[mask, 'intercept'].to_numpy()` should be more efficient.

lstchain/datachecks/containers.py
```diff
@@ -173,35 +173,38 @@ def fill_event_wise_info(self, subrun_index, table, mask, geom,
...
                        bins=histogram_binnings.hist_tgrad_vs_length)
         self.hist_tgrad_vs_length = counts
+
+        # We noticed an occasional pyplot error that seems to be fixed by
+        # making sure that the coordinates passed to hist2d are ndarrays
+        # (instead of Pandas data series)
+
         counts, _, _, _ = \
-            plt.hist2d(length[intensity > 200], tgrad[intensity > 200],
+            plt.hist2d(np.array(length[intensity > 200]),
```
MaxNoe, line 206: also here, `.to_numpy()` instead of np.array

moralejo: Ok, done!

morcuended:
Tested on real data. With these changes, those errors/warnings reported on #414 and #416 do not raise any more.

PR #419, opened 2020-05-25
Fix src dependent

rlopezcoto:
There was a small bug on the `lstchain_dl1_to_dl2.py` script that did not allow to process data using src_dependent analysis. I also got problems with `data_src_dep.set_index('index', drop=True)` why was this line there @SeiyaNozaki?

I also updated the src_dependent input card, that still had some old variables defined.

### DIFF
--- a/lstchain/data/lstchain_src_dep_config.json
+++ b/lstchain/data/lstchain_src_dep_config.json
@@ -5,7 +5,7 @@
     "length": [0, Infinity],
     "wl": [0, 1],
     "r": [0, 1],
-    "leakage": [0, 1]
+    "leakage2_intensity": [0, 1]
   },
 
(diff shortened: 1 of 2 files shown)


SeiyaNozaki:
`index` column was there after adding source-dependent parameters,  so I dropped this column.
I think #381 solved this issue, so now no need to add this line.
Thanks!

rlopezcoto:
>I think #381 solved this issue, so now no need to add this line.

thanks @SeiyaNozaki, I removed the line completely

rlopezcoto:
@SeiyaNozaki then, do you agree with the changes? if yes, could you approve the PR?

SeiyaNozaki:
yes, please! :)

Issue #420, opened 2020-05-25
Transformation of camera to Sky coordinates works only for MC


JouvinLea:
Hola,
Today, we realize that the utils method camera_to_sky in utils (https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/reco/utils.py) works only for MC since the obstime is hard coded.
I propose to do a small PR to fix the bug adding a time for the reference frame.
 Is it ok to replace the existing method ?

We could also add directly in the DL2 generation two columns ra and dec for the events!

MaxNoe:
`camera_to_sky` transforms from `CameraFrame` to `HorizonFrame`, for which no obstime is needed.

It's actually badly named, as "sky" is not really clear, it should be `horizon` or `altaz`.

No obstime should be needed for processing simulations. Why was this hard coded in the first place? You don't need it for simulations and for observations it's wrong.

rlopezcoto:
when you define `altaaz` you need to give an `obstime`, no?
https://github.com/cta-observatory/cta-lstchain/blob/6f98711c8bb7464ac3781abc0b9de054477eefad/lstchain/reco/utils.py#L47

this way you can directly transform `AltAz` to sky coordinates, or am I misunderstanding something?

MaxNoe:
This works perfectly fine for mc:

```
pointing = SkyCoord(alt=alt, az=az, frame=AltAz())
camera_coord = SkyCoord(x=x, y=y, frame=CameraFrame(telescope_pointing=pointing, focal_length=focal_length)
altaz_coord = camera_coord.transform_to(AltAz())
```

Transformations from `CameraFrame` up to `AltAz`, including `TelescopeFrame` and `NominalFrame` do not neither require `location` nor `obstime`. 

For our current simulations, there is no well-defined way to go from `AltAz` to sky fixed positions, e.g. Equatorial Coordinates (ICRS) or Galactic Coordinates.
This might change if you produce gamma simulations based on a trajectory, but then you'd also have a well-defined location and obstime for simulations or if you simulate a star field for noise, in which case you could also do it by including a proper set of obstime / location in the mc or just provide pointing positions both in horizontal and in equatorial/galactic frame.

rlopezcoto:
[...]

PR #421, opened 2020-05-26
Fix method camera_to_sky for real events

JouvinLea:
This PR fix the problem in the method computing camera_to_sky for event to take properly into account the time of the observation in the computation.

For the moment, just use time if provided and the default obs time in utils if not printing a warning.

I think the solution to use HA/DEC for MC event instead of RA/DEC should be done in another method when you compute ra/dec for those SkyCoords. For example, if I read correctly to compute the HA from RA, we need to extract the sideral time from the obs time and the location.
```obstime = sky_coords.obstime.value
scale_time = sky_coords.obstime.scale
longitude = sky_coords.location.longitude
latitude = sky_coords.location.latitude
time=Time(obstime, scale, location=(longitude, latitude)
HA = sky_coords.icrs.ra.to("hourangle") -time.sidereal_time("apparent")
dec = sky_coords.icrs.dec
```

Also I need to improve the test to add a check on a ra/dec value when we add the time.

### DIFF
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -203,16 +203,24 @@ def reco_source_position_sky(cog_x, cog_y, disp_dx, disp_dy, focal_length, point
     return camera_to_sky(src_x, src_y, focal_length, pointing_alt, pointing_az)
 
 
-def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
+def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az, time = None):
     """
+    Compute camera to SkyCoord. For MC assume the default ObsTime.
 
     Parameters
     ----------
-    pos_x: X coordinate in camera (distance)
-    pos_y: Y coordinate in camera (distance)
-    focal: telescope focal (distance)
-    pointing_alt: pointing altitude in angle unit
-    pointing_az: pointing altitude in angle unit
+    pos_x: `~astropy.units.Quantity`
+        X coordinate in camera (distance)
+    pos_y: `~astropy.units.Quantity`
+        Y coordinate in camera (distance)
+    focal: `~astropy.units.Quantity`
+        telescope focal (distance)
+    pointing_alt: `~astropy.units.Quantity`
+        pointing altitude in angle unit
+    pointing_az: `~astropy.units.Quantity`
+        pointing altitude in angle unit
+    time: `~astropy.time.Time`
+
 
     Returns
     -------
@@ -230,6 +238,13 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
     sky_coords = utils.camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az)
 
     """
+
+    if time:
+        horizon_frame = AltAz(location=location, obstime=time)
+    else:
+        horizon_frame = AltAz(location=location, obstime=obstime)
+        print("No time given, use default observation time. To be use only for MC data.")
+
     pointing_direction = SkyCoord(alt=clip_alt(pointing_alt), az=pointing_az, frame=horizon_frame)
 
     camera_frame = CameraFrame(focal_length=focal, telescope_pointing=pointing_direction)
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -236,7 +236,7 @@ def camera_to_altaz(pos_x, pos_y, focal, pointing_alt, pointing_az, obstime = No
     focal = 28*u.m
     pointing_alt = np.array([1.0, 1.0]) * u.rad
     pointing_az = np.array([0.2, 0.5]) * u.rad
-    sky_coords = utils.camera_to_altaz_frame(pos_x, pos_y, focal, pointing_alt, pointing_az)
+    sky_coords = utils.camera_to_altaz(pos_x, pos_y, focal, pointing_alt, pointing_az)
 
     """
     if not obstime:
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -240,7 +240,7 @@ def camera_to_altaz(pos_x, pos_y, focal, pointing_alt, pointing_az, obstime = No
 
     """
     if not obstime:
-        logging.warning("No time given. To be use only for MC data.")
+        logging.info("No time given. To be use only for MC data.")
     horizon_frame = AltAz(location=location, obstime=obstime)
 
     pointing_direction = SkyCoord(alt=clip_alt(pointing_alt), az=pointing_az, frame=horizon_frame)
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -26,7 +26,7 @@ __all__ = [
     'cal_cam_source_pos',
     'get_event_pos_in_camera',
     'reco_source_position_sky',
-    'camera_to_sky',
+    'camera_to_altaz_frame',
     'sky_to_camera',
     'source_side',
     'source_dx_dy',
@@ -200,12 +200,12 @@ def reco_source_position_sky(cog_x, cog_y, disp_dx, disp_dy, focal_length, point
     sky frame: `astropy.coordinates.sky_coordinate.SkyCoord`
     """
     src_x, src_y = disp.disp_to_pos(disp_dx, disp_dy, cog_x, cog_y)
-    return camera_to_sky(src_x, src_y, focal_length, pointing_alt, pointing_az)
+    return camera_to_altaz_frame(src_x, src_y, focal_length, pointing_alt, pointing_az)
 
 
-def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az, time = None):
+def camera_to_altaz_frame(pos_x, pos_y, focal, pointing_alt, pointing_az, time = None):
     """
-    Compute camera to SkyCoord. For MC assume the default ObsTime.
+    Compute camera to Horizontal frame (Altitude-Azimuth system). For MC assume the default ObsTime.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


lstchain/reco/utils.py
```diff
@@ -203,16 +203,24 @@ def reco_source_position_sky(cog_x, cog_y, disp_dx, disp_dy, focal_length, point
     return camera_to_sky(src_x, src_y, focal_length, pointing_alt, pointing_az)
 
 
-def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
+def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az, time = None):
```
rlopezcoto, line 206: maybe we should change the name of the function. As @MaxNoe pointed out in #420 , this method is converting to a `HorizonFrame`

lstchain/reco/utils.py
```diff
@@ -200,12 +200,12 @@ def reco_source_position_sky(cog_x, cog_y, disp_dx, disp_dy, focal_length, point
     sky frame: `astropy.coordinates.sky_coordinate.SkyCoord`
     """
     src_x, src_y = disp.disp_to_pos(disp_dx, disp_dy, cog_x, cog_y)
-    return camera_to_sky(src_x, src_y, focal_length, pointing_alt, pointing_az)
+    return camera_to_altaz_frame(src_x, src_y, focal_length, pointing_alt, pointing_az)
 
 
-def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az, time = None):
+def camera_to_altaz_frame(pos_x, pos_y, focal, pointing_alt, pointing_az, time = None):
```
MaxNoe, line 206: The frame in the end is a little strange (Either its `camera_frame_to_altaz_frame` or `camera_to_altaz`).
I would just leave out the frame.

`time` should be `obstime` to be consistent with the astropy API, and then you also don't need the `if` statement below.

lstchain/reco/utils.py
```diff
@@ -227,9 +235,16 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
     focal = 28*u.m
     pointing_alt = np.array([1.0, 1.0]) * u.rad
     pointing_az = np.array([0.2, 0.5]) * u.rad
-    sky_coords = utils.camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az)
+    sky_coords = utils.camera_to_altaz_frame(pos_x, pos_y, focal, pointing_alt, pointing_az)
 
     """
+
+    if time:
```
MaxNoe, line 242: This if is not needed. You use `obstime=obstime` and it work fine also in the case where `obstime is None`

lstchain/reco/utils.py
```diff
@@ -200,24 +200,32 @@ def reco_source_position_sky(cog_x, cog_y, disp_dx, disp_dy, focal_length, point
...
+    pointing_az: `~astropy.units.Quantity`
+        pointing altitude in angle unit
+    time: `~astropy.time.Time`
+
 
     Returns
     -------
-    sky frame: `astropy.coordinates.sky_coordinate.SkyCoord`
-
+    sky frame: `astropy.coordinates.AltAz`
```
MaxNoe, line 227: Actually the return value is a `astropy.coordinates.SkyCoordinate` that has a frame of `AltAz`.

So maybe write ``` `astropy.coordinates.SkyCoord` in AltAz frame` ```

lstchain/reco/utils.py
```diff
@@ -200,24 +200,32 @@ def reco_source_position_sky(cog_x, cog_y, disp_dx, disp_dy, focal_length, point
...
+        X coordinate in camera (distance)
+    pos_y: `~astropy.units.Quantity`
+        Y coordinate in camera (distance)
+    focal: `~astropy.units.Quantity`
+        telescope focal (distance)
+    pointing_alt: `~astropy.units.Quantity`
+        pointing altitude in angle unit
+    pointing_az: `~astropy.units.Quantity`
+        pointing altitude in angle unit
+    time: `~astropy.time.Time`
```
MaxNoe, line 222: Call it obstime

lstchain/reco/utils.py
JouvinLea: Thanks @MaxNoe for your usefull input.
Here I disagree with you but may I misunderstood. I think for MC events we need to use the obstime_mc in order to be able to create a posteriori a ra/dec for MC event in HA/dec as proposed @moralejo . To perform that, t I understood we wiil need:
```scale_time = sky_coords.obstime.scale
longitude = sky_coords.location.longitude
latitude = sky_coords.location.latitude
time=Time(obstime, scale, location=(longitude, latitude)
HA = sky_coords.icrs.ra.to("hourangle") -time.sidereal_time("apparent")
dec = sky_coords.icrs.dec
```

So we need to have a false obstime to be able to then transform to HA.

lstchain/reco/utils.py
```diff
@@ -43,8 +44,8 @@
 
 # position of the LST1
 location = EarthLocation.from_geodetic(-17.89139 * u.deg, 28.76139 * u.deg, 2184 * u.m)
-obstime = Time('2018-11-01T02:00')
-horizon_frame = AltAz(location=location, obstime=obstime)
+obstime_mc = Time('2018-11-01T02:00')
```
MaxNoe, line 47: Just remove this.

Why was this added anyway? What is so special about 2018-11-01T02:00 that it shoud be used as time fo mc?

lstchain/reco/utils.py
```diff
@@ -227,9 +236,16 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
...
     pointing_az = np.array([0.2, 0.5]) * u.rad
-    sky_coords = utils.camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az)
+    sky_coords = utils.camera_to_altaz_frame(pos_x, pos_y, focal, pointing_alt, pointing_az)
 
     """
+
+    if obstime:
+        horizon_frame = AltAz(location=location, obstime=obstime)
+    else:
+        horizon_frame = AltAz(location=location, obstime=obstime_mc)
```
MaxNoe, line 246: Again: just don't use an obstime for mc please. It has no meaning.

lstchain/reco/utils.py
MaxNoe: I don't think @moralejo was proposing to actually calculate HA, just mentioning the possibility.

At the moment, there is no usecase I know of for having simulations in another frame than horizon.

A fake time however would never be the solution!

lstchain/reco/utils.py
```diff
@@ -239,12 +239,9 @@ def camera_to_altaz(pos_x, pos_y, focal, pointing_alt, pointing_az, obstime = No
     sky_coords = utils.camera_to_altaz_frame(pos_x, pos_y, focal, pointing_alt, pointing_az)
 
     """
-
-    if obstime:
-        horizon_frame = AltAz(location=location, obstime=obstime)
-    else:
-        horizon_frame = AltAz(location=location, obstime=obstime_mc)
+    if not obstime:
         logging.warning("No time given, use default observation time. To be use only for MC data.")
```
MaxNoe: Now, you don't use any obstime (obstime is None, you are not using the global)

lstchain/reco/utils.py
MaxNoe: I think that the monte carlo processing should be fully agnostic of any obstime. Introducing a fake obstime is a hack (that should be unnecessary).

If it at some position is necessary to introduce a fake time, I would be glad to know, because that means we nead it fixed.

MaxNoe:
Btw. the frame to get local hourangle and declination would be the Celestial Intermediate Reference System:
https://docs.astropy.org/en/stable/api/astropy.coordinates.CIRS.html

lstchain/reco/utils.py
JouvinLea: yes this is what you ask no?

MaxNoe: But the logging message says something about a default obstime, you not using any obstime

MaxNoe: And I would say, as this is the expected situation for every simulated event, this should not be a "warning".

Warnings should be for unexpected conditions that require the user to do something. Meaningless warnings are dangerous because then users start ignoring them and potentially start ignoring also the important warnings.

JouvinLea: This is because you are too fast, I already modify the print in the time you did your comment

JouvinLea: I agree for the warning

MaxNoe:
There is an open issue for requesting the local equatorial frame in astropy here:
https://github.com/astropy/astropy/issues/10239

lstchain/reco/utils.py
```diff
@@ -227,9 +236,13 @@ def camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az):
     focal = 28*u.m
     pointing_alt = np.array([1.0, 1.0]) * u.rad
     pointing_az = np.array([0.2, 0.5]) * u.rad
-    sky_coords = utils.camera_to_sky(pos_x, pos_y, focal, pointing_alt, pointing_az)
+    sky_coords = utils.camera_to_altaz_frame(pos_x, pos_y, focal, pointing_alt, pointing_az)
```
rlopezcoto, line 239: `camera_to_altaz_frame` -> `camera_to_altaz`

MaxNoe:
Is the obstime defined globally in the `utils.py` file used somewhere else? It should be removed.

JouvinLea:
if you think it can be merge, I am personally done.
otherwise tell me what I should change.

Obstime at the beginning of the files is used to define the horizon_frame used in other methods . And I will not do this clean up in this PR. maybe one PR can be done to clean the utils.py in the future.

Once this is merge, I will do another PR for implementing ra/dec in the DL2 dataframe but other PR is better because this is another subject.

Issue #422, opened 2020-05-27
image extractor config


yrenier:
Looking at the default config ( in lstchain/data/lstchain_standard_config.json) we have:
```
  "image_extractor": "LocalPeakWindowSum",
  "image_extractor_for_muons": "GlobalPeakWindowSum",
  "image_extractor_config": {},
  "LocalPeakWindowSum":{
     "window_shift": 4,
     "window_width":8
    },
  "GlobalPeakWindowSum":{
     "window_shift": 4,
     "window_width":8
    },
```
In load_image_extractor_from_config() we have:
```
def load_image_extractor_from_config(custom_config):
    config = replace_config(get_standard_config(), custom_config)
    conf = Config({config['image_extractor']: config['image_extractor_config']})
    image_extractor = getattr(extractor, config['image_extractor'])
    return image_extractor(conf)
```
so the image extractor is loading the extractor settings from the `image_extractor_config` parameter in the config. Also changing `window_shift` or `window_width` in `LocalPeakWindowSum` has no effect to the calibrator obtained by load_calibrator_from_config() which I find very confusing.

What is the expected behavior:
- put the desired `window_shift` and/or `window_width` in `image_extractor_config` ? In that case ` LocalPeakWindowSum` and `GlobalPeakWindowSum`should be removed to avoid confusion.
- the settings of the extractor selected by `image_extractor` should be used ? In that case there is a bug and load_image_extractor_from_config() should be:
```
def load_image_extractor_from_config(custom_config):
    config = replace_config(get_standard_config(), custom_config)
    image_extractor_sel = config['image_extractor']:
    conf = Config({image_extractor_sel: config[image_extractor_sel]})
    image_extractor = getattr(extractor, config['image_extractor'])
    return image_extractor(conf)
```

vuillaut:
Hi.
I agree, the config looks wrong.
I would expect (that's what was intended originally I think at least):
```
 "image_extractor": "LocalPeakWindowSum",
  "image_extractor_config": {
     "window_shift": 4,
     "window_width":8   
},
```

Plus, looking at the code, I don't see where the parameters of the `image_extractor_for_muons` can be adjusted. So the standard parameters are always used.
Actually, I think that the extractor itself is not read from the config.
```
        r1_dl1_calibrator_for_muon_rings = LSTCameraCalibrator(calibration_path = calibration_path,
                                                               time_calibration_path = time_calibration_path,
[...]

PR #423, opened 2020-05-28
Config

vuillaut:
Is the config clearer this way?
- The parameters that do not differ from the standard config are not repeated in derived ones.
- the MC image extractor parameters are clearly specified and separated from the muons ones

Solves #422


lstchain/data/lstchain_standard_config.json
```diff
@@ -90,70 +85,95 @@
...
+    3,
+    4
+  ],
   "max_events": null,
   "custom_calibration": false,
   "write_pe_image": false,
   "image_extractor": "LocalPeakWindowSum",
-  "image_extractor_for_muons": "GlobalPeakWindowSum",
-  "image_extractor_config": {},
+  "image_extractor_config": {
```
MaxNoe, line 98: Why are both configs given here if only `LocalPeakWindowSum` is used?

lstchain/data/lstchain_standard_config.json
```diff
@@ -90,70 +85,95 @@
...
-
-  "LocalPeakWindowSum":{
-     "window_shift": 4,
-     "window_width":8
-    },
-  "GlobalPeakWindowSum":{
-     "window_shift": 4,
-     "window_width":8
+  "image_extractor_for_muons": "GlobalPeakWindowSum",
+  "image_extractor_for_muons_config": {
```
MaxNoe, line 113: Same as above. Why not only give the config for the `GlobalPeakWindowSum`?

lstchain/data/lstchain_standard_config.json
vuillaut: Not need, may (or not) be more clear to be exhaustive on the possible configs, for example if the user wants to switch the `image_extractor`, the parameters are set.
Feedback welcome.

MaxNoe: `--help-all` should show all possible settings for tools.

vuillaut: > `--help-all` should show all possible settings for tools.

Nice.
For that we would need to make Tools out of all scripts.

vuillaut: I'd be happy to limit the standard config to only the necessary one if you think it's clearer (I do) - it wasn't, so I kept it this way.

MaxNoe: @vuillaut that is needed anyways to get to a consistent api and to be able to properly use the configuration system. Currently, this all fells really hacked together around some of the limitation the ctapipe config system has (or had, a lot are fixed in 0.8)

vuillaut:
I think this should be considered review, we still are unable to configure the image extractor parameters for MC in the present config.
If there is something that needs to be changed, please let me know.

lstchain/data/lstchain_standard_config.json
vuillaut: For this PR I'd like to simply fix the config bug.
The whole config will need to be changed and improved with the use of tools and ctapipe0.8.

vuillaut:
After lstchain 0.6 release, it was easier to open a new PR #496

PR #428, opened 2020-06-01
Add constant charge scaling and pulse shape scaling correction 

FrancaCassol:
In this PR I add trialets to the LSTCameraCalibrator  and to the LSTCalibrationCalculator in order to give the possibility to:

1.  Scale the integrated charges with constant values different for HG and LG

2. Scale the integrated charge for the window width on the base of the present reference pulse shape and the ctapipe code used for MC data. This is  for testing different integration widths (and eventually pulse shapes). I would like to notice that the present window (8 ns width, 4 ns shift) corresponds to a scaling correction of [1.2032, 0.9486] for [HG, LG]. The LG value is less than one due to the negative undershoot of the present pulse shape (see attached figure), I admit i am not totally sure this is the correct value to be used.
3. Finally, I add also the possibility to correct the flat-field signal for the integration window before calculating the gain, in this case, due to the large window used (12 ns), the scaling correction is  close to 1 : [1.0552, 0.9978] for [HG, LG]

For the moment, the default values of the trailers are such  that no scaling is applied.
reference_pulse_shape_MC


moralejo:
Thanks for this much needed feature. How is the cross-calibration of high and low gain done in case this correction is activated? Is it done on the already corrected values (for both gains)? Or is the correction done afterwards?

FrancaCassol:
> Thanks for this much needed feature. How is the cross-calibration of high and low gain done in case this correction is activated? Is it done on the already corrected values (for both gains)? Or is the correction done afterwards?

Actually I did not touch the cross-calibration. Could you remember me what did you expect to change?

moralejo:
> > Thanks for this much needed feature. How is the cross-calibration of high and low gain done in case this correction is activated? Is it done on the already corrected values (for both gains)? Or is the correction done afterwards?
> 
> Actually I did not touch the cross-calibration. Could you remember me what did you expect to change?

Ok, so you are applying the ENF method for HG and LG independently, and obtain conversion factors for both, but do not "enforce" the consistency of them, i.e. that for the flatfield events they come up with the same number of p.e.  One may argue that in case there is a discrepancy (as was the case before), it is better to have a smooth HG to LG transition, which can be achieved precisely by enforcing this consistency, rather than having a "step" in the change from one to the other. One might either change the LG conv. factor, or the HG factor, or both, to do this cross-calibration.

FrancaCassol:
> Ok, so you are applying the ENF method for HG and LG independently, and obtain conversion factors for both, but do not "enforce" the consistency of them, i.e. that for the flatfield events they come up with the same number of p.e. 

Actually at present (after last corrections) HG and LG come out with the same number of photon electrons with the F-factor method. But what is indeed be consolidated is the gain value (which depends on the signal charge and its correction)

>  One may argue that in case there is a discrepancy (as was the case before), it is better to have a smooth HG to LG transition, which can be achieved precisely by enforcing this consistency, rather than having a "step" in the change from one to the other. One might either change the LG conv. factor, or the HG factor, or both, to do this cross-calibration.

Yes, I agree. Let me verify which step we have now. One could also think to linearly move from one calibration to the other.

rlopezcoto:
@FrancaCassol @moralejo this implementation will change when moving to ctapipe v0.8, so I propose to accept it as it is now and maybe open an issue about the cross-calibration in the HG/LG transition for the future implementation, how does that sound?

FrancaCassol:
I agree @rlopezcoto, let's keep it as it is now, we will adapt it to ctapipe 0.8 next week

moralejo:
> 2. Scale the integrated charge for the window width on the base of the present reference pulse shape and the ctapipe code used for MC data. This is  for testing different integration widths (and eventually pulse shapes). I would like to notice that the present window (8 ns width, 4 ns shift) corresponds to a scaling correction of [1.2032, 0.9486] for [HG, LG]. The LG value is less than one due to the negative undershoot of the present pulse shape (see attached figure), I admit i am not totally sure this is the correct value to be used.

I am indeed afraid it will not be correct for the real data. If you look at slide 7 of Yukiho's presentation linked below you can see that the undershoot is nearly non-existent. Perhaps Yukiho can provide a better curve, can you contact him?
I think 

https://indico.cta-observatory.org/event/2853/contributions/24597/attachments/17750/23825/ChargeRecoYukiho20200629.pdf

rlopezcoto:
@FrancaCassol could you please open an issue to keep this in mind, also with the relevant links/plots? I'll merge this anyways, so we can move forward.

PR #429, opened 2020-06-02
filtering also inf params

vuillaut:
Parameters with `inf` values should also be removed by default.


garciagenrique:
Thanks for the patch @vuillaut

PR #431, opened 2020-06-12
New plots to understand when pedestal tagging becomes unreliable

moralejo:
The new plots show the average rate of occurrence of pixel signals above 10 and 30 p.e., vs. subrun. This is done both for pedestal-tagged and cosmics-tagged events. In the latter one can also identify shots from the MAGIC LIDAR, below an example from two nights ago:

MAGIC_LIDAR_in_LST

The PR also includes some code for the future bokeh display, but the output is disabled. Please feel free to ignore that for now. I think these new plots will be very useful in the datacheck to test the coming fix of the UCTS firmware. So I'd like to have this merged and released as soon as possible.


lstchain/datachecks/dl1_checker.py
```diff
@@ -1193,3 +1262,190 @@ def merge_dl1datacheck_files(file_list):
...
+            if line.strip()[0] != '#':
+                pixel_hardware_info.append([])
+        file.seek(0)
+        for line in file:
+            if line[0] == '#':
+                continue
+            data = [int(s) for s in line.split()]
+            pixel_hardware_info[data[0]] = [data[1], data[2], data[3]]
+
+    return pixel_hardware_info[pix_id]
```
morcuended, line 1451: looking at the `LST_pixid_to_cluster.txt`, wouldn't it be better to use something like:
```
data = np.genfromtxt(infilename, comments='#')
```

morcuended:
@moralejo, it's not clear to me when looking at the code, do you make these plots by using both UCTS and TIB trigger tagging?

lstchain/datachecks/dl1_checker.py
MaxNoe: Or pandas read_csv or read_fwf

moralejo: > data = np.genfromtxt(infilename, comments='#')

Thanks, very useful!

MaxNoe: If you do not comment the line with the column names, `pd.read_fwf` will directly give you a dataframe with the correct column names

lstchain/datachecks/dl1_checker.py
```diff
@@ -1436,16 +1436,10 @@ def get_pixel_location(pix_id):
     # The first time we read in the data stored in the io directory:
     infilename = Path(Path(os.path.dirname(__file__)).parent,'io',
```
MaxNoe, line 1437: Better use
`from pkg_resources import resource_filename` and `resource_filename('lstchain', 'resources/LST_pixid_to_cluster.txt').`

This is also where the file should go, otherwise you will need to adapt the setup py to also include this file in the distributed package explicitly.

MaxNoe: Ah, sorry. resources is not setup. So you have to include this file in package_data in the `setup.py` file

lstchain/datachecks/dl1_checker.py
```diff
@@ -1436,16 +1436,10 @@ def get_pixel_location(pix_id):
...
-        for line in file:
-            if line.strip()[0] != '#':
-                pixel_hardware_info.append([])
-        file.seek(0)
-        for line in file:
-            if line[0] == '#':
-                continue
-            data = [int(s) for s in line.split()]
-            pixel_hardware_info[data[0]] = [data[1], data[2], data[3]]
+    pixel_hardware_info.extend([None]*(1 + data[:,0].max()))
```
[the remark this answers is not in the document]
moralejo: It is a global list, yes. Probably can be done with something like a static variable? (just my ignorance)
The function does what explained in its description: return the "hardware coordinates" of a pixel from its software id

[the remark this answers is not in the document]
MaxNoe: Okay, I see. This should be split into two functions. One for reading this information from the file resources into an appropriate in-memory data structure (e.g. a mapping dict or a pandas DataFrame).

And a second function performing the lookup. 
The second function might not be needed if the data structure used directly offers what you need.

lstchain/datachecks/dl1_checker.py
moralejo: Done, thanks!

moralejo:
> @moralejo, it's not clear to me when looking at the code, do you make these plots by using both UCTS and TIB trigger tagging?

No, only the TIB trigger tag, @FrancaCassol suggested that one is more often correct. But today they're updating the UCTS firmware, which I think may solve the problem with ucts_trigger_type. I have changed the DL1 data check to use ucts_trigger_type by default, and we will see as soon as we take data and process them with this update if the problem is gone or not.

lstchain/datachecks/dl1_checker.py
moralejo: > Ah, sorry. resources is not setup. So you have to include this file in package_data in the `setup.py` file

I had not seen this. Done now.

lstchain/datachecks/dl1_checker.py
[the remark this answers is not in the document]
moralejo: > Okay, I see. This should be split into two functions. One for reading this information from the file resources into an appropriate in-memory data structure (e.g. a mapping dict or a pandas DataFrame).

But isn't the simple python list fine?  I also do not see the advantage of splitting into two functions.

> 
> And a second function performing the lookup.
> The second function might not be needed if the data structure used directly offers what you need.

I think it does, because the index of the list is directly pixel id.

rlopezcoto:
>  I have changed the DL1 data check to use ucts_trigger_type by default, and we will see as soon as we take data and process them with this update if the problem is gone or not.

this may break the backwards compatibility for some of the datacheck plots whenever we reprocess the data and re-run it.

morcuended:
> > I have changed the DL1 data check to use ucts_trigger_type by default, and we will see as soon as we take data and process them with this update if the problem is gone or not.
> 
> this may break the backwards compatibility for some of the datacheck plots whenever we reprocess the data and re-run it.

Does it makes sense to use both at the same time, i.e. add the same plots but using ucts_trigger_type?

lstchain/datachecks/dl1_checker.py
moralejo: > If you do not comment the line with the column names, `pd.read_fwf` will directly give you a dataframe with the correct column names

Ok, thx for the info, but not needed here I'd say.

moralejo:
> > > I have changed the DL1 data check to use ucts_trigger_type by default, and we will see as soon as we take data and process them with this update if the problem is gone or not.
> > 
> > 
> > this may break the backwards compatibility for some of the datacheck plots whenever we reprocess the data and re-run it.

> 
> Does it makes sense to use both at the same time, i.e. add the same plots but using ucts_trigger_type in addition to those existing ones using (tib) trigger_type?

That would be a lot more plots, just to display not one, but two _wrong_ event tags... :)  Let's see how things go with the newer data.

moralejo:
> > I have changed the DL1 data check to use ucts_trigger_type by default, and we will see as soon as we take data and process them with this update if the problem is gone or not.
> 
> this may break the backwards compatibility for some of the datacheck plots whenever we reprocess the data and re-run it.

I do not understand why. Some plots would change, but that is all.

rlopezcoto:
> I do not understand why. Some plots would change, but that is all.

yes, this is what I meant, that there will be the same plots produced using a different variable, and having different results may be a bit confusing, but ok

moralejo:
> > I do not understand why. Some plots would change, but that is all.
> 
> yes, this is what I meant, that there will be the same plots produced using a different variable, and having different results may be a bit confusing, but ok

I agree, but we may do this later. We need to do a release now...

FrancaCassol:
@moralejo, if you want to keep a backward compatibility you can put an if statement on the run number (unfortunately the firmware version is not propagated to the data)

PR #435, opened 2020-06-18
DL1 scaling

### LINKED ISSUE #433: scaling MC intensity
Discussed during LST analysis call, June 15th.

We need to scale the MC to match the real data (seems to be easier than the other way around).
Is it necessary to do it at R1/DL0 level or can it be done at DL1 level (on the charge images, before cleaning and parameters extraction)?
The former will be much more complicated.

Note: it might get even more complicated when going for ctapipe v0.8 and the DL1 writer tool which is not meant for these kinds of tweaks but we can leave that point for later.

tagging @moralejo @FrancaCassol for discussion

vuillaut:
resolves #433

### DIFF
--- a/lstchain/data/lstchain_src_dep_config.json
+++ b/lstchain/data/lstchain_src_dep_config.json
@@ -85,7 +85,7 @@
   "max_events": null,
   "custom_calibration": false,
   "write_pe_image": false,
-  "mc_dl1_scaling_factor": 1,
+  "mc_image_scaling_factor": 1,
   "image_extractor": "LocalPeakWindowSum",
   "image_extractor_for_muons": "GlobalPeakWindowSum",
   "image_extractor_config": {},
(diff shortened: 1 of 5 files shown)


moralejo:
Thanks!
Just for clarity,  I would rename  mc_dl1_scaling_factor  to  "mc_dl1a_scaling_factor"  or "mc_light_scaling_factor"

MaxNoe:
> Just for clarity, I would rename mc_dl1_scaling_factor to "mc_dl1a_scaling_factor" or "mc_light_scaling_factor"

It's applied only to `dl1.image`, so why not `mc_image_scaling`?

vuillaut:
`image` seemed more appropriate, I made the change.

vuillaut:
Note: with `ctapipe v0.8` this correction can be done with the `relative_factor`  in the DL1 Calibration Container.

PR #436, opened 2020-06-18
Tailcut cleaning with additional threshold

pawel21:
Hi,

I have developed new cleaning method using interleaved pedestal events.
This method modified `tailcuts_clean` method from `ctapipe` adding additional threshold calculated using interleaved pedestal events.
@vuillaut , @rlopezcoto , @FrancaCassol , @moralejo  Let me know, what do you think?

Ref.:

- [image cleaning optimzation lst-analysis call]( https://indico.cta-observatory.org/event/2799/contributions/24184/attachments/17569/23424/Image_cleaning_optimization.pdf)
- [image_cleanig_optim_lst_meeting](https://indico.cta-observatory.org/event/2551/contributions/24353/attachments/17657/23609/Optimization_of_the_image_cleaning_PG_LST_GM.pdf)


MaxNoe:
@pawel21 This is a very general method that is probably of interest for anyone. I would recommend to add this (also) to ctapipe itself.

We should contribute general methods directly to ctapipe, not to lstchain.

If it concerns urgent fixes, adding it to both might be needed, but general methods should go to ctapipe.

pawel21:
@MaxNoe , thank you for comment. I agree, I will add this also to `ctapipe`
But, as I understand now we are using `ctapipe` 0.7 in `cta-lstchain`, then we want move to `ctapipe` 0.8, so if I add this method to `ctapipe` this method will go to new relase 'ctapipe` 0.9, correct? 
So, first let's test this method here and then I will add this method to `ctapipe`, OK? What do you think?

rlopezcoto:
Hi @pawel21, thank you very much for this nice addition. I agree with @MaxNoe that these general methods should be included in `ctapipe`, but since we need it to be used in `lstchain` now, what he was suggesting is that you can implement it in both libraries. 
In any case, we can stick to this implementation in lstchain to comment about it and improve it, and then later you can think about the best place for its implementation in ctapipe, does that sound ok?

pawel21:
hi @rlopezcoto , sounds OK. 
Probably, will be possible to find some cleaver way to join standard 'tailcuts_clean' from `ctapipe` with new method, which I have implemented here. Because, I have only changed one line in `tailcuts_clean` method:
`pixels_above_picture = image >= picture_thresh` ->  
```pixels_above_picture = np.logical_and(image>= picture_thresh,
                                          image >= interleaved_pedestal_thresh) ```

`interleaved_pedestal_thresh` is array stored threshold from interleaved events for each pixel. The way to obtain those values will be more tricked and will be depend of telescope.

**Comment by vuillaut** [comment]
Is this still live?

**Comment by pawel21** [comment]
> Is this still live?

Yes, still alive. It will be useful to merge into master

File: lstchain/image/cleaning/cleaning.py [hunk 1]
```diff
@@ -0,0 +1,136 @@
...
+import numpy as np
+
+from lstchain.io.io import dl1_params_tel_mon_ped_key, dl1_params_tel_mon_cal_key
+
+
+def tailcuts_clean_with_pedestal_threshold(
+    geom,
+    image,
+    interleaved_pedestal_thresh,
+    picture_thresh=7,
```
rlopezcoto, line 11: maybe include here the standard
`picture_thresh=6`
`boundary_thresh=3`
?

lstchain/image/cleaning/cleaning.py
```diff
@@ -0,0 +1,136 @@
...
+        cleaning level
+    Returns
+    -------
+    picture_thresh: np.ndarray
+        picture threshold calculated using interleaved pedestal events
+    """
+    high_gain = 0
+    interleaved_events_id = 1
+    ped_mean_pe, ped_rms_pe = get_bias_and_rms(dl1_path)
+    threshold_clean_pe = ped_mean_pe + sigma_clean*ped_rms_pe
```
rlopezcoto, line 128: `sigma_clean * ped_rms_pe`

rlopezcoto (commented):
General: we just had a call in which it was underlined the importance of writing test units, maybe this code could benefit from some.

lstchain/image/cleaning/cleaning.py
moralejo: > maybe include here the standard
> `picture_thresh=6`
> `boundary_thresh=3`
> ?

Actually they are a bit too low for real data, after the "re-calibration" (correction for pulse tails) that we introduced in the latest lstchain release. So I would put there whatever Julian and Pawel think is reasonable according to their tests.

lstchain/image/cleaning/cleaning.py
```diff
@@ -0,0 +1,136 @@
...
+        path to dl1 file
+    Returns
+    -------
+    bias, rms: np.ndarray, np.ndarray
+        bias and rms in p.e.
+    """
+    f = tables.open_file(dl1_file)
+    ped = f.root[dl1_params_tel_mon_ped_key]
+    ped_charge_mean = np.array(ped.cols.charge_mean)
+    ped_charge_rms = np.array(ped.cols.charge_std)
```
moralejo, line 103: For clarity, better use always std or std dev, not rms

moralejo: I don't understand this, ped.cols.charge_mean and charge_std do not contain just the interleaved event info, right @FrancaCassol ? They contain also (in row 0) the values from the dedicated calibration file, potentially taken way before the current subrun, and hence with different star positions.

pawel21: Yes, in row 0 we have pedestal for calibration run, in row 1 we have pedestal from interleaved events (at least that's how I understand it )

lstchain/image/cleaning/cleaning.py
pawel21: I will add script to reprocess data, where we will be able give threshold values in config file.

lstchain/image/cleaning/cleaning.py
moralejo: Ok, but where do you select row 1?

jsitarek: Hi @moralejo 
I think Paweł selects those here:
https://github.com/cta-observatory/cta-lstchain/blob/601684f5390210b7c6de49e4bfbf3ef60f29e269/lstchain/image/cleaning/cleaning.py#L136

as far as I remember what @FrancaCassol is doing in calculation of those pedestal updates is to use all the interleaved events of a given subrun to compute a single interleaved update of pedestals, so then you should have always exactly two values: original from dedicated calibration/pedestal run, and second (only one) from interleaveds. If we ever change this and could have 2 values of interleaveds in one subrun this code would need to be updated, but I think there is no point in doing it now. There is another problem, if the subrun is very short and have no interleaved pedestals for the update. I'm not sure if e.g. if you have 20 pedestal events if the values are computed at all (and have huge uncertainties) or not. In both cases it would be probably simplest to use the ped/cal run values then (it would either way not make much difference since those would be just small fractions of the whole dataset

moralejo: Ok, thanks a lot, it is clear now.
About short subruns, I agree it is not a big deal because they are a small fraction of data. But I would not be too strict... already with 10 pedestal events (collected in just 0.1 s) one should be able to tell a pixel containing a bright star from one with no star. And in any case, if the std dev has a large error, we may again argue that anyway it is few events that will be affected.
Eventually one can use also (properly treated) cosmics to improve the precision of pedestal std dev's.

jsitarek: I was more thinking about regular, non-star-affected pixels that by a fluctuation will get a higher pedestal RMS and thus we would start increasing their threshold without any need.
I made a quick check with toy MC, Assuming that you get RMS of 1 phe and that you have 20 pedestal measurements the accuracy on pedestal bias is 0.22 phe (1 phe / sqrt(20) ), which if you have an actual values of bias of ~ 2phe would end up with ~10% accuracy. In the case of RMS of 1 phe , those 20 events would produce uncertainty of it of 0.16 phe (i.e. 16%, there was also a formula on the accuracy of sigma parameter from N measurements, but I do not remember it :-) ), so well, not that bad, but it is also enough to start slightly increasing thresholds of some pixels, so very short runs can be indeed affected, but probably only slightly.

moralejo: Probably better than completely omitting the anti-noisy pixels procedure. 

I don't know what the Travis CI problem is, that should also be addressed. And once we have a working dl1b script we can also arrange with @morcuended for testing it in La Palma with some full pet runs (2011 & 2012 for instance)

morcuended: Regarding Travis:
```
ERROR: Could not build wheels for ctapipe which use PEP 517 and cannot be installed directly
The command "pip install \
  https://github.com/cta-observatory/ctapipe/archive/$CTAPIPE_VERSION.tar.gz \
  https://github.com/cta-observatory/ctapipe_io_lst/archive/$CTAPIPE_IO_LST_VERSION.tar.gz \
  pytest-cov codecov
  " failed and exited with 1 during .
```
Since GitHub actions are already in place why don't we get rid of Travis already?

> And once we have a working dl1b script we can also arrange with @morcuended for testing it in La Palma with some full pet runs (2011 & 2012 for instance)

Sure, let me know the details whenever is done. I remind you that PR #543 could be merged already. The new cleaning feature using pedestal rms can be implemented in a separate PR (?).

morcuended: > Since GitHub actions are already in place why don't we get rid of Travis already?

I guess it'd be just merging #578

kosack:
Just to note, this is not really a "new" cleaning method - it is exactly the method used by the Whipple 10m for years (thresholds defined as a multiplier * pedestal_dispersion). 

One question: do you really need a new function for this? In the original function, both `picture_threshold` and `boundary_threshold` can be arrays.  E.g if you have an array x of the pedestal dispersion for each pixel, this still works:

```py3
ped_disp = get_pedestal_dispersions()  # array of pedestal dispersions
mask = tailcuts_clean(geom, image, picture_thresh=3.0*ped_disp, boundary_thresh=1.5*ped_disp)
```

jsitarek:
Hi @kosack 

I think Whipple uses only this condition, i.e. each pixel has different threshold. This optimizes the threshold, but makes a mess with reproducing it fully in MCs, since you have different thresholds in each pixel. 

The idea here is that for most of the pixels you use fixed thresholds (like in standard CTA cleaning), however a small fraction of pixels with higher thresholds (where the noise is significantly larger then the rest of the pixels).

About using the table of thresholds in the existing functon, it might work, however the pedestal RMS  changes with time, for the moment we want to use the same values throughout the subrun (following how the interleaved pedestal updates are done), but this might change in the future, therefore it is probably cleaner to have already a separate function for this.

lstchain/image/cleaning/cleaning.py
```diff
@@ -0,0 +1,140 @@
...
+        of picture neighbors. This has no effect in case keep_isolated_pixels is True
+    Returns
+    -------
+    A boolean mask of *clean* pixels.  To get a zero-suppressed image and pixel
+    list, use `image[mask], geom.pix_id[mask]`, or to keep the same
+    image size and just set unclean pixels to 0 or similar, use
+    `image[~mask] = 0`
+    """
+
+    pixels_above_picture = np.logical_and(image>= picture_thresh,
```
maxnoe, line 56: You do not need your own function to do this.

You can just use `picture_thresh = np.clip(pedestal_threshhold, picture_threshold, None)` and use that as `picture_threshold` for the normal image cleaning

pawel21: Good idea, thank you

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    return ped_charge_mean_pe, ped_charge_std_pe
+
+def get_threshold_from_dl1_file(dl1_path, sigma_clean):
+    """
+    Function to get picture threshold from dl1 from interleaved pedestal events.
+    Parameters
+    ----------
+    input_filename: str
+        path to dl1 file
+    sigma_clean: float
```
moralejo, line 37: More info, either here or above, of what  the returned threshold is?

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    picture_thresh: np.ndarray
+        picture threshold calculated using interleaved pedestal events
+    """
+    high_gain = 0
+    ped_mean_pe, ped_std_pe = get_bias_and_std(dl1_path)
+
+    # If problem with interleaved pedestal std values occur, take pedestal
+    # std values from calibration run.
+    # Correct interleaved pedestal std array should have shape (2,2,1855)
+    if ped_std_pe.shape[0] == 2:
```
moralejo, line 50: Why testing the first index?

moralejo: A bit more explanation would help.

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        interleaved_events_id = 1
+    else:
+        interleaved_events_id = 0
+    threshold_clean_pe = ped_mean_pe + sigma_clean*ped_std_pe
+    # find pixels with std = 0 and mean = 0 <=> dead pixels in interleaved
+    # pedestal event likely due to stars
+    dead_pixel_ids = np.where(threshold_clean_pe[interleaved_events_id, high_gain, :] == 0)[0]
+    # for dead pixels set max value of threshold
```
moralejo, line 58: Why the max in any pixel? Seems a bit arbitrary - but irrelevant if pixel is really switched off.

lstchain/scripts/lstchain_dl1ab_pedestal_cleaning.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python3
```
moralejo, line 1: Why a different script? It should rather be an option of the existing dl1ab script, right?

lstchain/scripts/lstchain_dl1ab_pedestal_cleaning.py
```diff
@@ -0,0 +1,186 @@
...
+                                    keep_isolated_pixels,
+                                    min_number_picture_neighbors
+                                    )
+                n_pixels = np.count_nonzero(signal_pixels)
+                if n_pixels > 0:
+                    num_islands, island_labels = number_of_islands(camera_geom, signal_pixels)
+                    n_pixels_on_island = np.bincount(island_labels.astype(np.int))
+                    n_pixels_on_island[0] = 0  # first island is no-island and should not be considered
+                    max_island_label = np.argmax(n_pixels_on_island)
+                    signal_pixels[island_labels != max_island_label] = False
```
moralejo, line 146: This should be configurable with a command-line option or through config file. With this star-aware cleaning we probably will do away with the "main island" parameters, and use all cleaning-surviving pixels instead.

moralejo:
>    I found one problem, where `ped_charge_std` array shape is not (2,2,1855), but e.g (7,2,1855) something is wrong with interleaved events (for instance in `dl1_LST-1.Run03281.0000.h5`), in this case pedestal std values is much higher, so if this happen I took pedestal value from calibration run to have reasonable picture threshold values.

Can you comment on this, @FrancaCassol ?  The 7 means there are 6 "bunches" of processed pedestals, correct? This could happen if the cosmics rate is low, and hence many interleaved pedestals go into a subrun. But that run seems to have 4 kHz of cosmics.
It may make sense to use the average of all the interleaved pedestals inside the monitoring, rather than the pedestal values from the dedicated calibrated file. The latter may have been recorded way before the data, so stars may be in very different locations.

lstchain/scripts/lstchain_dl1ab_pedestal_cleaning.py
pawel21: This solution will be better. So after PR #543 I can add this option, right?

moralejo: > This solution will be better. So after PR #543 I can add this option, right?

ah, ok, right, better wait for #543 which should be nearly ready I think.

lstchain/calib/camera/utils.py
pawel21: If we have strong star in camera FoV some pixels are switched off (std of signal in this pixels is 0) 
image , so for those pixels, I set max value from another pixels.

moralejo: Yes, it just seemed a bit arbitrary to set the max of all pixels, but no problem. I don't have a better suggestion.

jsitarek: It might have been a better solution to simply use 0 for those pixels. This means that they would have the standard image cleaning applied to them, which as Abelardo says does not matter since they have no signal.

pawel21: > It might have been a better solution to simply use 0 for those pixels. This means that they would have the standard image cleaning applied to them, which as Abelardo says does not matter since they have no signal.

OK, let me test this idea.

pawel21: In this case (use 0 for those pixels), number of surviving pedestal images increased a bit from 48 to 56 (with 224 pedestal images in total)

moralejo: > In this case (use 0 for those pixels), number of surviving pedestal images increased a bit from 48 to 56 (with 224 pedestal images in total)

So despite they have 0 mean pedestal and std dev they are not dead, right? Perhaps the mean and std are 0 because they did not reach a minimum number of valid events for the calculation. It may be safer then to keep a high threshold.  
By the way the fraction of pedestal survivors is pretty high, are you testing with low thresholds?

pawel21: I was testing Run 3004 with (picture threshold = 8, boundary th. = 4 and sigma=3), this run is noisy. I will try find better value of threshold.

pawel21:
I have modified script `lstchain_dl1ab.py`, if configuration contains "tailcuts_clean_with_pedestal_threshold" cleaning with pedestal information will be applied. 
Example config file:
``` 
"tailcuts_clean_with_pedestal_threshold": {
    "picture_thresh":8,
    "boundary_thresh":4,
    "sigma":3, 
    "keep_isolated_pixels":false,
    "min_number_picture_neighbors":1
  }
```

pawel21:
@moralejo , @jsitarek any feedback about "new" script will be useful :)

lstchain/io/io.py
```diff
@@ -1138,3 +1140,22 @@ def merge_dl2_runs(data_tag, runs, columns_to_read=None, n_process=4):
     observation_time = sum([t.total_seconds() for t in observation_times])
     df = pd.concat(df_list)
     return observation_time, df
+
+def get_cleaninig_params(config, clean_method_name):
```
moralejo, line 1144: Typo: "cleaninig"

moralejo:
> > I found one problem, where `ped_charge_std` array shape is not (2,2,1855), but e.g (7,2,1855) something is wrong with interleaved events (for instance in `dl1_LST-1.Run03281.0000.h5`), in this case pedestal std values is much higher, so if this happen I took pedestal value from calibration run to have reasonable picture threshold values.
> 
> Can you comment on this, @FrancaCassol ? The 7 means there are 6 "bunches" of processed pedestals, correct? This could happen if the cosmics rate is low, and hence many interleaved pedestals go into a subrun. But that run seems to have 4 kHz of cosmics.
> It may make sense to use the average of all the interleaved pedestals inside the monitoring, rather than the pedestal values from the dedicated calibrated file. The latter may have been recorded way before the data, so stars may be in very different locations.

@FrancaCassol please have a look at this issue above when possible.

moralejo:
> @moralejo , @jsitarek any feedback about "new" script will be useful :)

Hi @pawel21, seems good to me. Have you tested all this under ctapipe 0.10.1? We are soon moving to it, see #587

jsitarek:
> > @moralejo , @jsitarek any feedback about "new" script will be useful :)
> 
> Hi @pawel21, seems good to me. Have you tested all this under ctapipe 0.10.1? We are soon moving to it, see #587

also to me it looks fine.
one could add an explicit test for this method, but this is a bit tricky if you would like to do it on a prepared example rather then a real file, so maybe it can go to another PR

morcuended:
> > > @moralejo , @jsitarek any feedback about "new" script will be useful :)
> > 
> > 
> > Hi @pawel21, seems good to me. Have you tested all this under ctapipe 0.10.1? We are soon moving to it, see #587
> 
> also to me it looks fine.
> one could add an explicit test for this method, but this is a bit tricky if you would like to do it on a prepared example rather then a real file, so maybe it can go to another PR

Let's merge #586 so we can test with real files

lstchain/io/io.py
pawel21: fixed

lstchain/scripts/lstchain_dl1ab_pedestal_cleaning.py
rlopezcoto: now that #543 is merged, can we implement this as an option of the  `lstchain_dl1ab.py` script?

lstchain/calib/camera/utils.py
rlopezcoto: what about using a value in an invalid range so those pixels are ruled out? We do not want them surviving any cleaning.

pawel21:
Let me wait with my change until PR #587 will be merge to avoid a mess.

rlopezcoto:
> Let me wait with my change until PR #587 will be merge to avoid a mess.

Thanks @pawel21, but let's separate this PR from that of the new script to produce drs4 files. Now with this one, remember that you will have to push a branch directly to lstchain, not to your fork, see here the quick fix for the PR:
https://github.com/cta-observatory/cta-lstchain/pull/553#issuecomment-782148862

rlopezcoto:
@pawel21 could you please move this to a new branch so we can have a second look and merge it before next release? Thanks

pawel21:
New PR #612

PR #443, opened 2020-07-03
Set up environment of the lstchain_ctapipe0.8 branch

rlopezcoto:
Setting up new environment and updating travis and environment variables with the new software versions.
Notice that this PR has as a target the branch `lstchain_ctapipe0.8`, the development branch that will contain the changes to update to ctapipe v0.8

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     packages=find_packages(),
     install_requires=[
         'astropy',
-        'ctapipe~=0.7.0',
+        'ctapipe~=0.8.0',
         'ctaplot~=0.5.3',
         'eventio~=0.20.3',
         'gammapy>=0.17',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -42,10 +42,10 @@ def find_scripts(script_dir, prefix):
     version=version,
     packages=find_packages(),
     install_requires=[
-        'astropy',
-        'ctapipe~=0.7.0',
+        "astropy>=3,<5",
```
moralejo, line 45: Why not astropy 4?

rlopezcoto: I copied it from the ctapipe setup, I'm not sure if astropy 4 was _required_ by ctapipe or it was just _compatible_. I would prefer to leave it as it is in the ctapipe installation. In any case, I just checked and with this condition, the version I got is `4.0.1.post1`

PR #451, opened 2020-07-06
Changes to adapt muon ring code to ctapipe 0.8

moralejo:


moralejo:
Not yet tested to process muon rings properly, but I think all API changes are done

lstchain/image/muon/muon_analysis.py
```diff
@@ -56,7 +49,7 @@ def pixel_coords_to_telescope(geom, equivalent_focal_length):
     )
     tel_coord = camera_coord.transform_to(TelescopeFrame())
 
-    return tel_coord.delta_az, tel_coord.delta_alt
```
rlopezcoto, line 59: `delta_alt` and `delta_az` should also be modified in the function docstring

moralejo: Done

moralejo:
> For consistency, we should also change the names of the (`delta_alt`/`delta_az`) variables in l.271/272

Done

lstchain/image/muon/muon_analysis.py
```diff
@@ -181,30 +176,31 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
+        # Use same hard-coded value for pedestal fluctuations as the previous
+        # version of ctapipe:
+        pedestal_stddev = 1.1*np.ones(len(image))
+
+        muonintensityoutput = \
+            intensity_fitter(1,
+                             muonringparam.ring_center_x,
+                             muonringparam.ring_center_y,
+                             muonringparam.ring_radius,
+                             image*dist_mask,
```
moralejo, line 192: This is not going to work well... in this way there will be many pixels with 0 charge but they will be used in the muon ring likelihood fit (sure, they should be pixels with ~0 predicted light, but we want to ignore them completely). It seems this new ctapipe implementation does not allow to set a mask, or pass an image with only some pixels, since it always uses the full camera geometry, as far as I understand. Something to fix on the ctapipe side, do you agree, @MaxNoe ?

rlopezcoto: but what about performing a cleaning and passing the `clean_mask`?

moralejo: > but what about performing a cleaning and passing the `clean_mask`?

You mean masking the image array (it is an ndarray) before passing it? Would not work, because you need to pass the mask as well so that it knows which pixels' coordinates to use. Or am I missing anything?

rlopezcoto: maybe I'm missing the problem here: what you want is that the pixels set to 0 in the `image * dist_mask` are not used by the likelihood whatsover?

moralejo:
I made some more changes needed for the changes of variable names in ctapipe

PR #458, opened 2020-07-07
Add subarray to calibration class

FrancaCassol:
This is a first fix for the camera calibration, that permit to apply the calibration
(other fixes related to the calibration containers and calibration calculation can come in a second PR)

### DIFF
--- a/lstchain/calib/camera/calibrator.py
+++ b/lstchain/calib/camera/calibrator.py
@@ -61,7 +61,8 @@ class LSTCameraCalibrator(CameraCalibrator):
 
     ).tag(config=True)
 
-    def __init__(self, **kwargs):
+
+    def __init__(self, subarray, **kwargs):
         """
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


rlopezcoto:
@FrancaCassol now that you are at it working with the `calibrator.py`, the `self` included in the print in:
https://github.com/cta-observatory/cta-lstchain/blob/cba047818cdec88122aebcee242d3966d3aff300/lstchain/calib/camera/calibrator.py#L281

will not work because this function is outside of the `LSTCameraCalibrator` class, right?

FrancaCassol:
You are right @rlopezcoto,  actually all this can be eliminated with the ctapipe 8.0 charge extractor,
because the corrections are already automatically done at that level (by the way, can one switch the off in ctapipe?).  
Do I remove them here?

rlopezcoto:
Is this at the moment used anywhere? If yes, then you could change the function to that of `ctapipe`, otherwise you could just removed. 
This is not really a change absolutely needed for this step because the code can run anyways without it, so do it only if it is something quick (I would also suggest to separate this in a different PR), otherwise we can wait to the second stage of lstchain adaptation to ctapipe

rlopezcoto:
@FrancaCassol shall we merge this here and take care of that bug/deletion in a separate PR? I can open an Issue to keep track of it if you cannot do it immediately, but it would be great to have the calibration fixes implemented

FrancaCassol:
Hi Ruben,
yes accept this, I am now working on new PR where I put all the remaining changes for the having the calibration working (actually there are several others)

Issue #357, opened 2020-04-21
Switching to ctapipe 0.8


MaxNoe:
ctapipe 0.8 is not yet released, but we can collect necessary changes here we already know about.

Some of the points are more general ctapipe adaption and not related to 0.8.
These could be done already, so that the amount of work is less when ctapipe 0.8 comes out.

This is what I came up with, feel free to add stuff I forgot.

- [ ] Use ctapipe `Tool` and configuration system for executables. (Could be done now already)
- [ ] Adapt the changes to the calibration code
- [ ] Use the new cleaning components
- [ ] Clean up containers (e.g. no dl1-mother-of-all-containers)
- [ ] CameraGeometry → CameraDescription.geometry
- [ ] Largest island 
- [ ] Use `Path` traitlet instead of `Unicode` where appropriate
- [ ] Pointing container

rlopezcoto:
What do you mean by `Largest island`?

MaxNoe:
There is now a function in ctapipe that gives you the largest island after cleaning., I think this got implemented here as well and this could be dropped for the ctapipe version.

moralejo:
I would say it is better to start addressing this after the next lstchain release, right?
Otherwise the release will be delayed. We would like to have soon a new version of calibrated data to make progress e.g. with cleaning studies.

rlopezcoto:
> There is now a function in ctapipe that gives you the largest island after cleaning., I think this got implemented here as well and this could be dropped for the ctapipe version.

ok, right, this was implemented and we are currently using the main island for the analysis. In any case and after discussing with @jsitarek, for the lowest energies this may not be the best solution (using only the largest island), but that can be studied using Data/MC in the future

rlopezcoto:
@moralejo I think this is just a proposal to keep track of these points, as discussed yesterday, as soon as we have a code that writes the information of interleaved events and the first datacheck scripts we can proceed with the 0.5 release.

moralejo:
> @moralejo I think this is just a proposal to keep track of these points, as discussed yesterday, as soon as we have a code that writes the information of interleaved events and the first datacheck scripts we can proceed with the 0.5 release.

Ok, I got nervous with the "could be done _already_".  Fine if already means right after 0.5.0

MaxNoe:
I just wanted to note which changes are possible with the current release and which changes are releated to the new ctapipe release.

moralejo:
[...]

PR #461, opened 2020-07-08
WIP: Add paths

rlopezcoto:
First steps to transform path string into `Path` from `pathlib` (not addressing yet the usage of `Path` ctapipe traitlet as in the #357 list).

Some questions popped up and I wanted to discuss them before moving on:
- Shall functions like `get_dataset_path` return a `Path` instead of a `str` (it makes sense from the name)?
- Shall we move to a complete usage of `Path` in definitions like those in `test_lstchain_scripts.py` and if yes, do we implement it directly in the test files or do we want to use the definitions in `paths.py`?


vuillaut:
Is this still alive?

PR #462, opened 2020-07-08
API changes to adapt lstchain to ctapipe v0.8

vuillaut:


lstchain/reco/r0_to_dl1.py
```diff
@@ -221,7 +224,10 @@ def r0_to_dl1(
     metadata = global_metadata(source)
     write_metadata(metadata, output_filename)
 
-    cal_mc = load_calibrator_from_config(config)
+    # cal_mc = load_calibrator_from_config(config)
+    from ctapipe.calib import CameraCalibrator
+    # from ctapipe.image.extractor import LocalPeakWindowSum
+    cal_mc = CameraCalibrator(subarray=source.subarray)
```
moralejo, line 230: Where is the config for the calibrator read now?

vuillaut: Working on it ;-)

vuillaut:
Note: `width` and `length` were computed in degrees.
I am keeping it that way for now  (and attaching the proper unit in the DL1 container)

lstchain/calib/camera/calib.py
```diff
@@ -115,22 +113,22 @@ def lst_calibration(event, telescope_id, high_gain_threshold):
...
+    image, peak_time = gain_selection(data, signals, peak_time, high_gain_threshold)
 
     event.dl1.tel[telescope_id].image = image
-    event.dl1.tel[telescope_id].pulse_time = pulse_time
+    event.dl1.tel[telescope_id].peak_time = peak_time
 
 
 @deprecated('28/06/2019', message='gain selection is now performed at <= R1 calibration level')
-def gain_selection(waveform, charges, pulse_time, threshold):
+def gain_selection(waveform, charges, peak_time, threshold):
```
rlopezcoto, line 131: shall we get rid of the deprecated functions already?

lstchain/io/io.py
```diff
@@ -396,7 +395,7 @@ def write_array_info(event, output_filename):
                         )
                         continue
 
-            camera.to_table().write(
+            camera.geometry.to_table().write(
                 output_filename,
                 path=f'/instrument/telescope/camera/{camera_name}',
```
rlopezcoto, line 400: at the moment, what is the `camera_name` we are writing to the df: `LSTCam` or `LSTCam-004`? I think that regardless of the version, we are writing `LSTCam`, right? If so, we should consider changing that

lstchain/reco/r0_to_dl1.py
```diff
@@ -555,12 +562,12 @@ def r0_to_dl1(
...
                     # FIXME: no need to read telescope characteristics like foclen for every event!
-                    foclen = event.inst.subarray.tel[telescope_id].optics.equivalent_focal_length
-                    mirror_area = u.Quantity(event.inst.subarray.tel[telescope_id].optics.mirror_area, u.m ** 2)
+                    foclen = subarray.tel[telescope_id].optics.equivalent_focal_length
+                    mirror_area = u.Quantity(subarray.tel[telescope_id].optics.mirror_area, u.m ** 2)
                     width = np.rad2deg(np.arctan2(dl1_container.width, foclen))
                     length = np.rad2deg(np.arctan2(dl1_container.length, foclen))
-                    dl1_container.width = width.value
-                    dl1_container.length = length.value
+                    dl1_container.width = width
```
rlopezcoto, line 569: this has `deg` units, can we include that in the container?

vuillaut: already the case

lstchain/io/io.py
vuillaut: We were writing LSTCam regardless of the version, exact.
Let me check the status with this PR.

lstchain/reco/r0_to_dl1.py
rlopezcoto: ok so we were writing them before in `deg`, but the unit in the containers was wrong?

lstchain/io/io.py
vuillaut: Still is `LSTCam`

lstchain/calib/camera/calib.py
vuillaut: Can I do it in a subsequent PR?

rlopezcoto: sure, probably better

lstchain/io/io.py
rlopezcoto: I do not think we should write the same name, specially with the big difference that we have between `LSTCam` and `LSTCam-00x`, but not a showstopper here

PR #463, opened 2020-07-08
Calibration fixes ctapipe 0.8

FrancaCassol:
This is the PR for the calibration fixes. 
There is still some small thing to  be fixed with calibration calculator, I will work on that in the early afternoon


vuillaut:
Thanks @FrancaCassol !
I think I can review this afternoon and at the end day we might have stage 1 done.

lstchain/calib/camera/calibrator.py
```diff
@@ -207,14 +184,14 @@ def _calibrate_dl1(self, event, telid):
         if self._check_dl0_empty(waveforms):
             return
 
-        if self.image_extractor.requires_neighbors():
-            camera = event.inst.subarray.tel[telid].camera
-            self.image_extractor.neighbors = camera.neighbor_matrix_where
+        # for the moment we do the gain selection afterwards
+        # use dummy gain mask
+        no_gain_selection= np.zeros((waveforms.shape[0],waveforms.shape[1]), dtype=np.int)
```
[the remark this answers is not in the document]
FrancaCassol: Done, let me know if it is clear enough

FrancaCassol:
I just added the fixes mentioned above. Things should work for the calibration now (except than for the time calibration that is fixed by Pawel in un other PR)
Be careful, the old calibration files will not work anymore, due to some changes in the calibration containers in ctapipe 0.8. I tested the creation of a drs4 and of a calibration file and it seems all fine.

lstchain/calib/camera/calibrator.py
```diff
@@ -207,82 +184,39 @@ def _calibrate_dl1(self, event, telid):
...
+        # correct charge for global scale
+        corrected_charge = charge * np.array(self.charge_scale)[:,np.newaxis]
 
         # correct time with drs4 correction if available
         if self.time_corrector:
             pulse_time = self.time_corrector.get_corr_pulse(event, pulse_time)
 
         # add flat-fielding time correction
-        pulse_time_ff_corrected = pulse_time + self.mon_data.tel[telid].calibration.time_correction
+        pulse_time_ff_corrected = pulse_time + self.mon_data.tel[telid].calibration.time_correction.value
```
vuillaut, line 206: how can we insure `pulse_time` and the `time_correction` are in the same unit?

lstchain/visualization/plot_drs4.py
```diff
@@ -79,8 +80,8 @@ def plot_pedestals(data_file, pedestal_file, run=0 , plot_file="none", tel_id=1,
             ped_data = event.mon.tel[tel_id].pedestal
             break
 
-    camera = CameraGeometry.from_name("LSTCam", 2)
-
+    camera = reader.subarray.tels[tel_id].camera.geometry
```
vuillaut, line 83: `camera_geometry` would be more explicit,  especially now that they are different objects in ctapipe

lstchain/calib/camera/calibrator.py
FrancaCassol: The are both calculated by the charge extractor using the waveform. But, you are right, as it is done now we can not be sure. The problem is that the  time do not have units in the R1Container and DL1 container

lstchain/visualization/plot_drs4.py
rlopezcoto: solved

PR #465, opened 2020-07-09
More changes to muon analysis

moralejo:


lstchain/image/muon/muon_analysis.py
```diff
@@ -114,9 +113,17 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
 
     Returns
     ---------
-    muonintensityoutput ``
-    muonringparam       ``
-    good_ring           `bool` it determines whether the ring can be used for analysis or not
+
+    muonintensityoutput MuonEfficiencyContainer
+    ring_size           float, in p.e. total intensity in ring
```
rlopezcoto, line 118: to be compliant with the container names in ctapipe, I would suggest to move:
`ring_size` -> `ring_intensity`
`size_outside_ring` -> `intensity_outside_ring`
...
and so on

lstchain/io/io.py
```diff
@@ -443,6 +445,239 @@ def read_array_info(filename):
     return array_info
 
 
+def read_optic(filename, telescope_name):
```
rlopezcoto, line 448: read_optics

lstchain/io/io.py
```diff
@@ -443,6 +445,239 @@ def read_array_info(filename):
...
+    Returns
+    -------
+    `ctapipe.instrument.camera.geometry.CameraGeometry`
+    """
+    camera_geometry_path = f"/configuration/instrument/telescope/camera/geometry_{camera_name}"
+    camera_geometry = CameraGeometry.from_table(Table.read(filename, camera_geometry_path))
+    return camera_geometry
+
+
+def read_camera_geometries(filename):
```
rlopezcoto, line 513: can we use something more intuitive like `read_camera_geometry_table` whenever it is more than one?
Same for the other functions

rlopezcoto:
I did some comparisons using the same integrators for lstchain v0.5.1 (lstchain 5) and the current development branch (lstchain 6). The size/intensity is very similar:
Screen Shot 2020-07-10 at 11 37 30
the width is slightly larger, probably because right now we cannot pass a mask to apply the likelihood:
Screen Shot 2020-07-10 at 11 37 35

@moralejo also reported that the absolute efficiency is lower because of the same reason as the width:
image

moralejo:
Good, thanks for this test!  On second thought, the difference in efficiency is too large, I suspect some other error in the new code. Will check when I have some time.

rlopezcoto:
we should pin this conversation for the future...
in any case, is there any other muon modification or can we move ahead with this PR?

lstchain/image/muon/muon_analysis.py
moralejo: We will soon move to output format common with ctapipe, we can discuss it then.

lstchain/io/io.py
moralejo: Probably this got in here by a mistake of mine? This was introduced in a different PR which I probably merged for tests, and then pushed by mistake.

lstchain/io/io.py
moralejo: Ditto

lstchain/image/muon/muon_analysis.py
```diff
@@ -177,49 +185,55 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
...
+
+        # No longer existing in ctapipe 0.8:
+        # muonintensityoutput.ring_pix_completeness =  \
+        #     (pix_ringwidth_im > tailcuts[0]).sum() / len(pix_ringwidth_im)
 
     else:
-        # just to have the default values:
+        # just to have the default values with units:
         muonintensityoutput = MuonEfficiencyContainer()
+        muonintensityoutput.width = u.Quantity(np.nan, u.deg)
```
MaxNoe, line 230: Why didn't you open an issue at ctapipe? The defaults should have the correct units

rlopezcoto: already fixed in the next ctapipe version?
https://github.com/cta-observatory/ctapipe/pull/1380

lstchain/image/muon/muon_analysis.py
```diff
@@ -394,14 +409,16 @@ def create_muon_table():
             'muon_efficiency': [],
             'ring_containment': [],
             'ring_completeness': [],
-            'ring_pixel_completeness': [],
+            #  missing in ctapipe 0.8:
+            # 'ring_pixel_completeness': [],
```
rlopezcoto, line 413: were we using the `ring_pixel_completeness` anywhere anymore? I think this was some remnant from the initial muon code... it may be ok to leave it out, otherwise we should go back to its implementation in `ctapipe`

lstchain/image/muon/muon_analysis.py
```diff
@@ -410,29 +427,38 @@ def create_muon_table():
...
+    # output_parameters['ring_pixel_completeness'].append(muonintensityparam.ring_pix_completeness)
+    output_parameters['impact_parameter'].append(muonintensityparam.impact.value)
+    output_parameters['impact_x_array'].append(muonintensityparam.impact_x.value)
+    output_parameters['impact_y_array'].append(muonintensityparam.impact_y.value)
     output_parameters['radial_stdev'].append(radial_distribution['standard_dev'].value)
     output_parameters['radial_skewness'].append(radial_distribution['skewness'])
     output_parameters['radial_excess_kurtosis'].append(radial_distribution['excess_kurtosis'])
-    output_parameters['num_pixels_in_ring'].append(np.sum(muonintensityparam.mask))
+    #  missing in ctapipe 0.8:
+    #output_parameters['num_pixels_in_ring'].append(np.sum(muonintensityparam.mask))
```
rlopezcoto, line 461: and this one? was it used?

lstchain/image/muon/muon_analysis.py
```diff
@@ -410,29 +427,38 @@ def create_muon_table():
...
-                    size_outside_ring, mean_pixel_charge_around_ring,
-                    hg_peak_sample=np.nan, lg_peak_sample=np.nan):
+                    size, size_outside_ring, mean_pixel_charge_around_ring,
+                    muonparameters, hg_peak_sample=np.nan,
+                    lg_peak_sample=np.nan):
 
     output_parameters['event_id'].append(event_id)
     output_parameters['event_time'].append(event_time)
-    output_parameters['ring_size'].append(muonintensityparam.ring_size)
+    output_parameters['ring_size'].append(size)
```
rlopezcoto, line 436: let's keep in mind that the ring `size` (or better `intensity`) is missing from the ctapipe containers list. I'll open an issue

lstchain/image/muon/muon_analysis.py
moralejo: We are not using it as far as I know.

lstchain/image/muon/muon_analysis.py
moralejo: No, and anyway we cannot use the mask now to do the efficiency fit, so this is something to be modified when the needed fix in ctapipe is made (or we do it provisionally in lstchain by copying the corresponding class).

moralejo:
> Good, thanks for this test! On second thought, the difference in efficiency is too large, I suspect some other error in the new code. Will check when I have some time.

Solved in ctapipe, see https://github.com/cta-observatory/ctapipe/pull/1379
For now we also include a patch in lstchain so it works also under ctapipe 0.8

moralejo:
For the record: in the new ctapipe 0.8 muon code, that we copied into lstchain (intensity_fitter.py), the pixel "width" used in the calculations is different than in the old lstchain code. In the latter (muon_analysis.py) we used a hard-coded, approximate value of 0.1 deg.  In the new ctapipe code, the width is determined from the pixel area stored in the geometry, i.e. 2*sqrt(A/pi). This new value (which makes more sense) is 0.1053 deg.   The effective pixel area used in the telescope efficiency  estimate is hence 10.8% larger, and the telescope efficiency as determined from muons is hence smaller by the same fraction. A comparison is shown below.
It also seems that the new code has more successful ring intensity fits, since a population of events with 0 efficiency (about 10%) is now gone.
New_vs_old_efficiency
The rest of muon parameters are very close in the two versions, so I'd say that we can move forward with this.

vuillaut:
Consider accepting PR https://github.com/moralejo/cta-lstchain/pull/3 before ;-)

moralejo:
> Consider accepting PR [moralejo#3](https://github.com/moralejo/cta-lstchain/pull/3) before ;-)

Sure! Done, testing now, back to draft for a while...

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+    radius: float
+        radius of circle
+    rho: float
+        fractional distance of impact point from circle center
+    phi: ndarray in radians
+        rotation angles to calculate length
+
+    Returns
+    -------
+    ndarray: chord length
```
vuillaut, line 44: incomplete docstring here.
phi can also be a float and in that case, a float is returned

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+    return mirror_length - hole_length
+
+
+def pixels_on_ring(radius, pixel_diameter):
+    """Calculate number of pixels of diameter ``pixel_diameter`` on the circumference
+    of a circle with radius ``radius``
+    """
+    circumference = 2 * np.pi * radius
+    n_pixels = u.Quantity(circumference / pixel_diameter)
+    return int(n_pixels.to_value(u.dimensionless_unscaled))
```
vuillaut, line 95:
```
n_pixels = circumference / pixel_diameter
return n_pixels
```
should do, no?

+ incomplete  docstring (do you expect astropy quantities here? - it will work in any case but better be explicit)

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+    phi: float
+        Rotation angle of muon image
+
+    Returns
+    -------
+    ndarray
+        Chord length for each angle
+    """
+    circumference = 2 * np.pi * radius
+    pixels_on_circle = int(circumference / pixel_diameter)
```
vuillaut, line 130: isn't this what `pixels_on_ring` function was made for?

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+        fractional distance of impact point from circle center
+    phi: ndarray in radians
+        rotation angles to calculate length
+
+    Returns
+    -------
+    ndarray: chord length
+    """
+    scalar = np.isscalar(phi)
+    phi = np.array(phi, ndmin=1, copy=False)
```
vuillaut, line 47: I don't understand the need for the `ndmin` arg
If `phi` was already a multi-dim array, it won't flatten it.

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+    of a circle with radius ``radius``
+    """
+    circumference = 2 * np.pi * radius
+    n_pixels = u.Quantity(circumference / pixel_diameter)
+    return int(n_pixels.to_value(u.dimensionless_unscaled))
+
+
+@lru_cache(maxsize=1000)
+def linspace_two_pi(n_points):
+    return np.linspace(-np.pi, np.pi, n_points)
```
vuillaut, line 100: Is this function intended to be used more than once?
If not I don't think it is necessary.

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+    ang: ndarray
+        Angles over which to integrate
+    phi: float
+        Rotation angle of muon image
+
+    Returns
+    -------
+    ndarray
+        Chord length for each angle
+    """
```
vuillaut, line 128: incomplete docstring

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+        pixel_x.to_value(u.rad),
+        pixel_y.to_value(u.rad),
+        pixel_diameter.to_value(u.rad),
+        oversampling=oversampling,
+        min_lambda_m=min_lambda.to_value(u.m),
+        max_lambda_m=max_lambda.to_value(u.m),
+    )
+
+
+def image_prediction_no_units(
```
vuillaut, line 198: what was the motivation for not using astropy units right away? perf?

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+    gauss = cdfs[0] - cdfs[1]
+
+    # interpolate profile to find prediction for each pixel
+    pred = np.interp(ang, ang_prof, profile)
+
+    # Multiply by integrated emissivity between 300 and 600 nm, and rest of factors to
+    # get total number of photons per pixel
+    # ^ would be per radian, but no need to put it here, would anyway cancel out below
+
+    pred *= alpha * (min_lambda_m ** -1 - max_lambda_m ** -1)
```
vuillaut, line 254: `1/min_lambda` would be ~twice faster

lstchain/image/muon/intensity_fitter.py
```diff
@@ -0,0 +1,562 @@
...
+    # diameter. In any case, since in the end we do a data-MC comparison of the muon
+    # ring analysis outputs, it is not critical that this value is exact.
+    pred *= CIRCLE_SQUARE_AREA_RATIO
+
+    return pred
+
+
+def calc_likelihood(image, pred, spe_width, ped):
+    """Calculate likelihood of prediction given the measured signal,
+    gaussian approx from [denaurois2009]_
```
vuillaut, line 280: A more specific biblio would be appreciated. I suppose it's  10.1016/j.astropartphys.2009.09.001 ?

lstchain/image/muon/muon_analysis.py
```diff
@@ -142,30 +158,29 @@ def analyze_muon_event(event_id, image, geom, equivalent_focal_length,
                                                             tailcuts)
 
     mirror_radius = np.sqrt(mirror_area / np.pi)
```
vuillaut, line 160: ?
what is the unit of `mirror_area`?

lstchain/image/muon/intensity_fitter.py
MaxNoe: This function works both with and without. It is actually not used I think in the current code.

lstchain/image/muon/intensity_fitter.py
MaxNoe: yes

lstchain/image/muon/intensity_fitter.py
MaxNoe: It enables the other case: passing a single number.

lstchain/image/muon/intensity_fitter.py
MaxNoe: It is used.

MaxNoe:
@vuillaut This code is a verbatim copy of the code in ctapipe. Please don't review here.

Any fixes should be applied in ctapipe.

lstchain/image/muon/intensity_fitter.py
vuillaut: yes, once.

vuillaut:
> @vuillaut This code is a verbatim copy of the code in ctapipe. Please don't review here.

what part?

lstchain/image/muon/intensity_fitter.py
MaxNoe: Yes, this function is called repeatedly by the minuit minimizer and astropy units were a dominant factor

MaxNoe:
Everything you commented on, the `intensity_fitter.py`

lstchain/image/muon/intensity_fitter.py
vuillaut: Thanks.

moralejo:
> Everything you commented on, the `intensity_fitter.py`

Yes, sorry for that "importing", but I had to do it in order to keep the code working (because we want to apply masks before the muon intensity fit)

MaxNoe:
> Yes, sorry for that "importing", but I had to do it in order to keep the code working (because we want to apply masks before the muon intensity fit)

How about instead opening a PR adding that feature in ctapipe?

moralejo:
> > Yes, sorry for that "importing", but I had to do it in order to keep the code working (because we want to apply masks before the muon intensity fit)
> 
> How about instead opening a PR adding that feature in ctapipe?

I opened an issue, apologies for no PR but it really takes me quite a while to make sure I don't mess anything up...

vuillaut:
> Everything you commented on, the `intensity_fitter.py`

Ok, don't mind my comments on this part then.
I'll have a look into ctapipe.

moralejo:
> > > Yes, sorry for that "importing", but I had to do it in order to keep the code working (because we want to apply masks before the muon intensity fit)
> > 
> > 
> > How about instead opening a PR adding that feature in ctapipe?
> 
> I opened an issue, apologies for no PR but it really takes me quite a while to make sure I don't mess anything up...

Ok, I opened https://github.com/cta-observatory/ctapipe/pull/1389

lstchain/image/muon/muon_analysis.py
moralejo: > ?
> what is the unit of `mirror_area`?

square meters - specified now.

PR #466, opened 2020-07-09
write array info a la ctapipe v0.8

vuillaut:
Double the info (small overhead) for now to not change the data format.
This is temporary for lstchain v0.6


vuillaut:
image

PR #467, opened 2020-07-09
Fix lstchain_rfperformance tests

rlopezcoto:
There were two problems that were making the tests from the `scripts` folder not to pass:
- If the `fake_proton_file` was created from that folder, since we relied on absolute paths wrt the `lstchain/tests` folder.
- If the number of `selected_gammas` in `lstchain_mc_rfperformance` was 0, the code was breaking.

Now tests pass also from inside the `scripts` folder and regardless of the input files used.

### DIFF
--- a/lstchain/scripts/lstchain_mc_rfperformance.py
+++ b/lstchain/scripts/lstchain_mc_rfperformance.py
@@ -14,20 +14,18 @@ $>python lstchain_mc_rfperformance.py
 
 """
 
+import argparse
 import logging
 import sys
-import numpy as np
-import pandas as pd
-import argparse
+
 import matplotlib.pyplot as plt
-from distutils.util import strtobool
+import pandas as pd
+
+from lstchain.io import standard_config, replace_config, read_configuration_file
+from lstchain.io.io import dl1_params_lstcam_key
 from lstchain.reco import dl1_to_dl2
 from lstchain.reco.utils import filter_events
 from lstchain.visualization import plot_dl2
-from lstchain.reco import utils
-import astropy.units as u
-from lstchain.io import standard_config, replace_config, read_configuration_file
-from lstchain.io.io import dl1_params_lstcam_key
 
 try:
     import ctaplot
--- a/lstchain/scripts/lstchain_mc_rfperformance.py
+++ b/lstchain/scripts/lstchain_mc_rfperformance.py
@@ -14,6 +14,8 @@ $>python lstchain_mc_rfperformance.py
 
 """
 
+import logging
+import sys
 import numpy as np
 import pandas as pd
 import argparse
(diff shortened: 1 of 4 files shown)


lstchain/scripts/lstchain_mc_rfperformance.py
```diff
@@ -111,6 +113,10 @@ def main():
 
     selected_gammas = dl2.query('reco_type==0 & mc_type==0')
 
+    if(len(selected_gammas) == 0):
+        log.warning('No gammas selected, I will not plot any output') 
+        sys.exit()
```
garciagenrique, line 118: default sys.exit() is a 0 exit code - meaning that the code will exit without prompting any error. If you want that the code exits with an error, you should add `sys.exit(1)`.
However, I'm not sure whether a unitary test will fail or not if the exit error is 0.

rlopezcoto: yeah, that was exactly the problem, if the script exits with an error, then the tests do not pass. Although this is a very particular case, I precisely found it while trying to solve the first problem. 
Maybe I could raise here a warning, but what is really needed is that the code does not continue executing. If you have a suggestion, it would be great

garciagenrique: Ok, nevermind. Of course if the code exits with and error, the test will not pas... 😅

garciagenrique: (sorry for the noise, I'm mixing things between the exit kind of the script and the unit test)

rlopezcoto: no problem, reviews are always very welcome! ;-)

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -88,9 +89,13 @@ def test_lstchain_mc_trainpipe():
 @pytest.mark.run(after='test_lstchain_mc_r0_to_dl1')
 def test_lstchain_mc_rfperformance():
     gamma_file = dl1_file
-    produce_fake_dl1_proton_file()
+    produce_fake_dl1_proton_file(dl1_file)
+    #produce_fake_dl1_proton_file()
```
vuillaut, line 93: can  be removed

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -88,9 +89,13 @@ def test_lstchain_mc_trainpipe():
...
     proton_file = fake_dl1_proton_file
 
+    print('proton_file', proton_file)
+    print('gamma_file', gamma_file)
```
vuillaut, line 97: can be removed

rlopezcoto: sorry, remnants of the debugging... done

lstchain/scripts/tests/test_lstchain_scripts.py
rlopezcoto: done

PR #468, opened 2020-07-09
add functions to read instrument info from DL1 file

vuillaut:
Necessary as long there is no DL1 event source

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -656,8 +656,9 @@ def read_telescopes_positions(filename):
     """
     subarray_table = read_subarray_table(filename)
     pos_dict = {}
-    for row in subarray_table['tel_id', 'pos_x', 'pos_y', 'pos_z'].iterrows():
-        pos_dict[row[0]] = row[1], row[2], row[3]
+    pos_unit = subarray_table['pos_x'].unit
+    for row in subarray_table:
+        pos_dict[row['tel_id']] = np.array([row['pos_x'], row['pos_y'], row['pos_z']]) * pos_unit
     return pos_dict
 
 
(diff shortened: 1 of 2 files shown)


lstchain/io/io.py
```diff
@@ -443,6 +445,239 @@ def read_array_info(filename):
     return array_info
 
 
+def read_optic(filename, telescope_name):
```
moralejo, line 448: Function name is not so nice, and a bit too close to the other one. Any better option?
e.g. read_single_optics    (and keep read_optics as it is)

moralejo: And thanks a lot for this! Will try it.

vuillaut: Good point, I will rename this one and similar with `_single_`

vuillaut:
I have absolutely no idea why the test is failing on something completely unrelated and that is working fine otherwise: `CameraDisplay(geom, image)`
`astropy` version ?

rlopezcoto:
> I have absolutely no idea why the test is failing on something completely unrelated and that is working fine otherwise: CameraDisplay(geom, image) astropy version ?

I'm getting those tests pass in my local version of this PR, are they also passing in your computer?

vuillaut:
> > I have absolutely no idea why the test is failing on something completely unrelated and that is working fine otherwise: CameraDisplay(geom, image) astropy version ?
> 
> I'm getting those tests pass in my local version of this PR, are they also passing in your computer?

yes they are.
It must be an install problem but it does not seem to replicate on other PRs... 🙄

rlopezcoto:
then, let's move on...

PR #473, opened 2020-07-10
allow to use subruns to create time calib file

pawel21:
After discuss with @moralejo I have created PR for new version of script to create time calibration file with multi subruns. 
To do it: `python lstchain_data_create_time_calibration_file.py --input-file 'LST-1.1.Run01625.000*.fits.fz' --output-file time_calib_1625_multi.h5 --max-events 60000`

### DIFF
--- a/lstchain/scripts/lstchain_data_create_time_calibration_file.py
+++ b/lstchain/scripts/lstchain_data_create_time_calibration_file.py
@@ -8,12 +8,13 @@ Script to create drs4 time correction coefficients
 
 Usage:
 $> python lstchain_data_create_time_calibration_file.py
---input-file LST-1.1.Run01625.0000.fits.fz
+--input-file LST-1.1.Run01625.0000.fits.fz or input-file LST-1.1.Run01625.000*.fits.fz
 --output-file time_calibration.Run1625.0000.hdf5
 
 """
 
 import argparse
+import glob
 import numpy as np
 from traitlets.config.loader import Config
 from ctapipe.io import event_source
@@ -28,7 +29,8 @@ parser = argparse.ArgumentParser()
 # Required arguments
 parser.add_argument("--input-file", action='store', type=str,
                     dest='input_file',
-                    help="Path to fits.fz file used to create the time calibration file.",
+                    help="Path to fits.fz file used to create the time calibration file. \
+                    Allowed to use regular expression in given path to process subruns",
                     default=None, required=True)
 
 parser.add_argument("--output-file", action='store', type=str,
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_data_create_time_calibration_file.py
```diff
@@ -77,24 +78,29 @@ def main():
     # declare the pedestal calibrator
     lst_r0 = LSTR0Corrections(pedestal_path=args.pedestal_file, config=config)
 
+    reader = event_source(input_url=path_list[0], max_events=args.max_events)
```
rlopezcoto, line 81: shouldn't this `event_source` be substituted by `LSTEventSource`?
At least should be considered for the future:
https://github.com/cta-observatory/ctapipe/issues/1177#issuecomment-555429621

lstchain/scripts/lstchain_data_create_time_calibration_file.py
```diff
@@ -77,24 +78,29 @@ def main():
...
-    for i, event in enumerate(reader):
-        if event.index.event_id % 5000 == 0:
-            print(event.index.event_id)
+    for path in path_list:
+        print("--> Processing: {}".format(path))
+        reader = event_source(input_url=path, max_events=args.max_events)
+        print("--> Number of files", reader.multi_file.num_inputs())
+        for i, event in enumerate(reader):
+            if event.index.event_id % 5000 == 0:
+                print(event.index.event_id)
```
rlopezcoto, line 95: maybe `logging` instead of `print`?
also it would be nice to have something like:
`log.info(f'File {j} out of {len(path_list)}')` 
(for that you need to change l.89 to `for i, path in enumerate(path_list)`)

lstchain/scripts/lstchain_data_create_time_calibration_file.py
pawel21: Yes. I changed to `LSTEventSource`

lstchain/scripts/lstchain_data_create_time_calibration_file.py
pawel21: Yes, will be much better. 
Done.

lstchain/scripts/lstchain_data_create_time_calibration_file.py
```diff
@@ -77,24 +83,30 @@ def main():
...
     tel_id = timeCorr.tel_id
 
-    for i, event in enumerate(reader):
-        if event.index.event_id % 5000 == 0:
-            print(event.index.event_id)
+    for i, path in enumerate(path_list):
+        log.info(f'File {i+1} out of {len(path_list)}')
+        log.info(f'Processing: {path}')
+        reader = LSTEventSource(input_url=path, max_events=args.max_events)
+        log.info(f'Number of files in this subrun: {reader.multi_file.num_inputs()}')
```
rlopezcoto, line 98: It this the number of `streams` in the subrun? Shouldn't this be always 4?

pawel21: Yes, should be always 4

rlopezcoto: then we probably do not want to print it every loop, right? I would even suggest not to print it at all

pawel21: Yes, you are right. I removed.

rlopezcoto: thanks!

PR #475, opened 2020-07-10
Image Extractor

vuillaut:
Include things  from #474 - should come after  it.
To be discussed, there must be a better way to handle this.
Resolve issue #482

### DIFF
--- a/lstchain/calib/camera/calibrator.py
+++ b/lstchain/calib/camera/calibrator.py
@@ -192,11 +192,12 @@ class LSTCameraCalibrator(CameraCalibrator):
 
         no_gain_selection = np.zeros((waveforms.shape[0], waveforms.shape[1]), dtype=np.int)
 
-        # the image extractor work per channel
-        charge0, peak_time0 = self.image_extractor(waveforms[0], telid, no_gain_selection[0])
-        charge1, peak_time1 = self.image_extractor(waveforms[1], telid, no_gain_selection[1])
-        charge = np.array([charge0, charge1])
-        peak_time = np.array([peak_time0, peak_time1])
+        charge = np.zeros(waveforms.shape[0])
+        peak_time = np.zeros(waveforms.shape[0])
+        # image extraction for each channel:
+        for i in range(waveforms.shape[0]):
+            charge[i], peak_time[i] = self.image_extractor(waveforms[i], telid, no_gain_selection[i])
+
 
         # correct charge for global scale
         corrected_charge = charge * np.array(self.charge_scale, dtype=np.float32)[:, np.newaxis]
--- a/lstchain/calib/camera/calibrator.py
+++ b/lstchain/calib/camera/calibrator.py
@@ -192,7 +192,11 @@ class LSTCameraCalibrator(CameraCalibrator):
 
         no_gain_selection = np.zeros((waveforms.shape[0], waveforms.shape[1]), dtype=np.int)
 
-        charge, peak_time = self.image_extractor(waveforms, telid, no_gain_selection)
+        # the image extractor work per channel
+        charge0, peak_time0 = self.image_extractor(waveforms[0], telid, no_gain_selection[0])
+        charge1, peak_time1 = self.image_extractor(waveforms[1], telid, no_gain_selection[1])
+        charge = np.array([charge0, charge1])
+        peak_time = np.array([peak_time0, peak_time1])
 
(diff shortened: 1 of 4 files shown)


lstchain/calib/camera/calibrator.py
```diff
@@ -191,12 +190,16 @@ def _calibrate_dl1(self, event, telid):
         #      - gain selection before charge integration
 
 
-        no_gain_selection= np.zeros((waveforms.shape[0],waveforms.shape[1]), dtype=np.int)
+        no_gain_selection = np.zeros((waveforms.shape[0], waveforms.shape[1]), dtype=np.int)
 
-        charge, peak_time = self.image_extractor(waveforms, telid, no_gain_selection)
+        # the image extractor work per channel
+        charge0, peak_time0 = self.image_extractor(waveforms[0], telid, no_gain_selection[0])
```
rlopezcoto, line 196: maybe clearer if you define `charge` and `peak_time` before:
```
charge = np.zeros(2)
peak_time = np.zeros(2)

charge[0], peak_time[0] = self.image_extractor(waveforms[0], telid, no_gain_selection[0])
charge[1], peak_time[1] = self.image_extractor(waveforms[1], telid, no_gain_selection[1])
```

FrancaCassol: Hi @vuillaut, 
why do you propose to make the charge extractor to work per channel?

MaxNoe: I guess because this is how it was in ctapipe 0.7? ctapipe 0.8 only calculates one image, as it expects gain selection to be done before image extraction: https://github.com/cta-observatory/ctapipe/pull/1167

This is actually a change that should now (when completely switching to 0.8) be done in the event source, as ctapipe expects to start at the DL0 level now, where gain selection has already happened.

vuillaut: Yes, as Max explains, the image extractor no longer accepts data cubes of dimensions > 2.
So... maybe there is no better solution if we want to keep analysing both channels for real data.

vuillaut:
If no one sees a better way to handle this, I undraft...

lstchain/calib/camera/calibrator.py
FrancaCassol: @vuillaut, yes it does.  You do not need to separate HG and LG. Try for example:

```
from ctapipe.image.extractor import LocalPeakWindowSum
extractor = LocalPeakWindowSum(subarray=reader.subarray, window_width=11, window_shift=4)
selected_gain_channel = np.zeros((2,1855), dtype=np.int)
waveform=np.ones((2,1855,40))
waveform[0]=3
charge, pulse_time = extractor(waveform, 1, selected_gain_channel)

print(charge)
```

Which type of problems do to you have?

vuillaut:
Hi @FrancaCassol 
Thank you for the comment. At some point last week, the tests were not passing with an error on the waveform shape at this line. I must have messed up with the all the PRs or something.
Indeed it works fine, sorry for the mess.

vuillaut:
Reopening after #482

lstchain/calib/camera/calibrator.py
```diff
@@ -191,12 +190,17 @@ def _calibrate_dl1(self, event, telid):
         #      - gain selection before charge integration
 
 
-        no_gain_selection= np.zeros((waveforms.shape[0],waveforms.shape[1]), dtype=np.int)
+        no_gain_selection = np.zeros((waveforms.shape[0], waveforms.shape[1]), dtype=np.int)
+
+        charge = np.zeros(waveforms.shape[0])
+        peak_time = np.zeros(waveforms.shape[0])
```
moralejo, line 196: I am getting an error and I suspect it is here... charge[i] and peak_time[i] will later contain each an array with one entry per pixel. Why this initialization??

moralejo: This seems to work:
        charge = np.zeros(waveforms.shape[0]*waveforms.shape[1], dtype='float32')                                                                              
        peak_time = np.zeros(waveforms.shape[0]*waveforms.shape[1], dtype='float32')                                                                           

Did it work for anyone with the other initialization?

MaxNoe: It shouldn't have. (n_gains, n_pixels) is the correct shape here, here it's just a two element array

PR #478, opened 2020-07-11
Removed the duplicated masks for  DL1 parameters & images

moralejo:
Removed the duplicated masks for the DL1 parameters and DL1 images tables,
since they have one-to-one corresponding rows. Besides, code did not work
because images table now lacks event_id

### DIFF
--- a/lstchain/datachecks/containers.py
+++ b/lstchain/datachecks/containers.py
@@ -129,9 +129,9 @@ class DL1DataCheckContainer(Container):
         n_jump = 1+int(self.num_events/n_samples)
         # keep some info every n-jump-th event:
         sampled_event_ids = np.array(table['event_id'][mask][0::n_jump])
-        tib_time = np.array(table['tib_time'][mask][0::n_jump])
-        ucts_time = np.array(table['ucts_time'][mask][0::n_jump])
-        dragon_time = np.array(table['dragon_time'][mask][0::n_jump])
+        tib_time = np.array(table['tib_time'][mask][0::n_jump])*u.s
+        ucts_time = np.array(table['ucts_time'][mask][0::n_jump])*u.s
+        dragon_time = np.array(table['dragon_time'][mask][0::n_jump])*u.s
         # in case the resulting number of entries is <n_samples, we have to pad
         # the arrays, because hdf vector columns must have the same number of
         # elements in each row. We repeat the last value in the array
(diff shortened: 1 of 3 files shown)


MaxNoe:
>  images table now lacks event_id

That should not be, everything needs to be identifiable

moralejo:
> > images table now lacks event_id
> 
> That should not be, everything needs to be identifiable

I thought it was by design, since the rows in the two tables correspond one to one, so the event_id in both would be redundant. As long as things work properly, of course! If we want absolute safety we could put back that event_id in ctapipe...

MaxNoe:
>  If we want absolute safety we could put back that event_id in ctapipe...

This is clearly a bug, since the tool has the option to only write the image table

moralejo:
> 
> But wait: you are talking about the lstchain dl1 processing, not the ctapipe tool? Then this has nothing to do with ctapipe, it needs to be fixed here!

This happened as a result of a change in ctapipe, but indeed it makes no sense to change anything in ctapipe (or even in lstchain) since we will move as soon as possible to the new DL1. So no need to do anything about this.

lstchain/datachecks/containers.py
```diff
@@ -129,9 +129,9 @@ def fill_event_wise_info(self, subrun_index, table, mask, geom,
         n_jump = 1+int(self.num_events/n_samples)
         # keep some info every n-jump-th event:
         sampled_event_ids = np.array(table['event_id'][mask][0::n_jump])
-        tib_time = np.array(table['tib_time'][mask][0::n_jump])
-        ucts_time = np.array(table['ucts_time'][mask][0::n_jump])
-        dragon_time = np.array(table['dragon_time'][mask][0::n_jump])
+        tib_time = np.array(table['tib_time'][mask][0::n_jump])*u.s
```
MaxNoe, line 132: Do not use `array * u.<unit>`, it is much slower than `u.Quantity(somearray, u.unit, copy=False)`

MaxNoe:
> This happened as a result of a change in ctapipe,

The containers don't duplicate this information anymore, this is right. For writing it out, we use the `EventIndex` and `TelescopeEventIndex` containers.

lstchain/datachecks/containers.py
moralejo: ok!

PR #479, opened 2020-07-13
Update to ctapipe 0.8

MaxNoe:
Please don't push directly to this branch but open single PRs with this branch as target.


rlopezcoto:
After #488 this should be ready to be merged so we can proceed with the `v0.6` release

PR #483, opened 2020-07-17
Fixed initialization of charge and peak_time arrays

moralejo:
Current version does not work for new (r0 ro dl1 stage in real data). This seems to fix it.


vuillaut:
Thanks for the quick fix. I don't understand how tests were passing before...

PR #486, opened 2020-07-21
Update lstchain.ipynb notebook

lfoffano:
Updates and corrects the main functions of the notebook


vuillaut:
Thank you for the update @lfoffano !

PR #487, opened 2020-07-21
Avoid partial filling of tables (of pedestals, flatfield or cosmics)

moralejo:
Don't fill anything on tables (of pedestals, flatfield or cosmics) unless everything can be filled. This avoids problems in the later merging of tables, when some of them were partially filled.


moralejo:
This is a very minor change, which solves a pending issue with data check (occurring only when last subrun has very little statistics). So I proceed to merge.

PR #490, opened 2020-07-24
Update lstchain_mc_sensitivity.py

lfoffano:
Improved readibility, plots saved as figures, extracts sensitivity instead of percentage of sensitivity wrt Crab from sensitivity.py

updated ctaplot function: plot_angular_resolution_cta_requirement   instead of       plot_angular_resolution_cta_requirementS
updated ctaplot function: pctaplot.plot_angular_resolution_per_energy     instead of    ctaplot.plot_angular_res_per_energy
updated default parameter: noff = 3    instead of  5 

Please experts check if the plots are correct or not. I did not change much and I followed the documentation of ctaplot (leaving old lines as comments), but for example the sensitivity seems not correct (was not correct also before).

Note: In my opinion we should change the name and/or description of the input files. They are used in a different way in the program and may create confusion.

Thank you


lstchain/scripts/lstchain_mc_sensitivity.py
```diff
@@ -40,31 +43,32 @@
...
-    n_bins_gammaness = 11  #  Number of gammaness bins
+    ntelescopes_gamma = 1
+    ntelescopes_protons = 1
+    n_bins_energy = 10  #  Number of energy bins
+    n_bins_gammaness = 10  #  Number of gammaness bins
     n_bins_theta2 = 10  #  Number of theta2 bins
     obstime = 50 * 3600 * u.s
-    noff = 5
-
+    noff = 3
```
rlopezcoto, line 69: `noff` should be 5 to calculate the sensitivity using the CTA-style

lstchain/scripts/lstchain_mc_sensitivity.py
```diff
@@ -40,31 +43,32 @@
...
 
 args = parser.parse_args()
 
 
 def main():
-    ntelescopes_gamma = 4
-    ntelescopes_protons = 4
-    n_bins_energy = 20  #  Number of energy bins
-    n_bins_gammaness = 11  #  Number of gammaness bins
+    ntelescopes_gamma = 1
```
rlopezcoto, line 63: `ntelescopes_gamma` and `ntelescopes_protons` should be 4 because we are using the standard MC, in which 4 telescopes are simulated

lstchain/scripts/lstchain_mc_sensitivity.py
```diff
@@ -40,31 +43,32 @@
...
+    ntelescopes_protons = 1
+    n_bins_energy = 10  #  Number of energy bins
```
rlopezcoto, line 65: I would prefer to leave the standard binning as it was because of the plots produced, how do the new plots look like with the binning that you are proposing?

lstchain/scripts/lstchain_mc_sensitivity.py
```diff
@@ -97,78 +107,122 @@ def main():
...
+    plt.savefig("theta2.png")
+    
     plt.figure(figsize=(12, 8))
-    ctaplot.plot_angular_res_per_energy(gammas_mc.reco_alt, gammas_mc.reco_az, gammas_mc.mc_alt, gammas_mc.mc_az,
-                                        10 ** (gammas_mc.reco_energy - 3),
-                                        )
-
-    ctaplot.plot_angular_res_cta_requirements('north', color='black')
-
+    #ctaplot.plot_angular_resolution_per_energy(gammas_mc.reco_alt, gammas_mc.reco_az, gammas_mc.mc_alt, gammas_mc.mc_az, 10 ** (gammas_mc.reco_energy - 3)  )
```
rlopezcoto, line 156: remove

lstchain/scripts/lstchain_mc_sensitivity.py
```diff
@@ -97,78 +107,122 @@ def main():
...
+    plt.figure(figsize=(12, 8))
     gamma_ps_simu_info = read_simu_info_merged_hdf5(args.dl1file_gammas)
     emin = gamma_ps_simu_info.energy_range_min.value
     emax = gamma_ps_simu_info.energy_range_max.value
     total_number_of_events = gamma_ps_simu_info.num_showers * gamma_ps_simu_info.shower_reuse
     spectral_index = gamma_ps_simu_info.spectral_index
     area = (gamma_ps_simu_info.max_scatter_range.value - gamma_ps_simu_info.min_scatter_range.value) ** 2 * np.pi
     ctaplot.plot_effective_area_per_energy_power_law(emin, emax, total_number_of_events, spectral_index,
-                                                     10 ** (gammas_mc.reco_energy - 3)[gammas_mc.tel_id == 1],
+                                                     #10 ** (gammas_mc.reco_energy - 3)[gammas_mc.tel_id == 1],
```
rlopezcoto, line 187: remove

lstchain/scripts/lstchain_mc_sensitivity.py
MaxNoe: Where do you have this info from? Would be good to link a source if it is written somewhere that CTA expects 5 off positions.

lstchain/scripts/lstchain_mc_sensitivity.py
MaxNoe: I think cta wants to have 10 bins per decade of energy, so something like `10**np.arange(log_e_min, log_e_max, 0.1)`

rlopezcoto: The standard is 5 bins per decade:
https://www.cta-observatory.org/science/cta-performance/#1472563157332-1ef9e83d-426c

lstchain/scripts/lstchain_mc_sensitivity.py
rlopezcoto: I could not find it in the IRF webpage (https://www.cta-observatory.org/science/cta-performance/), although I may be missing it.
In any case, one can quote the [Monte Carlo performance studies for the site selection of the Cherenkov Telescope Array](https://www.sciencedirect.com/science/article/abs/pii/S0927650517300087)paper:
"_A ratio of the off-source to on-source exposure of 5 is considered_"

lstchain/scripts/lstchain_mc_sensitivity.py
MaxNoe: ah, right, sorry, so 0.1 → 0.2

lstchain/scripts/lstchain_mc_sensitivity.py
lfoffano: OK good

lstchain/scripts/lstchain_mc_sensitivity.py
lfoffano: OK good thanks

lstchain/scripts/lstchain_mc_sensitivity.py
lfoffano: OK so 20 bins from 0.01 to 100 TeV

lstchain/scripts/lstchain_mc_sensitivity.py
lfoffano: ah yes, thanks

lstchain/scripts/lstchain_mc_sensitivity.py
lfoffano: yes, thanks

rlopezcoto (approved):

PR #491, opened 2020-07-24
Update sensitivity.py

lfoffano:
I was reporting an error due to the wrong shape of alpha in the calculation of the Li&Ma significance (shape (20,) instead of (20,11,10), with n_energy_bins, gammaness bins, and theta2_bins in order).
Problem solved, together with improved output and readibility (prints)    

Details:
Fixed function >  calculate_sensitivity_lima
alpha was given only as an array of lenght equal to len(n_on), but it was not of the same shape (20,11,10).
Correcting the input  np.ones_like(final_gamma)) instead of np.ones(len(final_gamma)) solves the problem
                                                                        
Please experts check if the plots are correct or not. I did not change much, but for example the sensitivity seems not correct (was not correct also before).

Thank you
Luca


lstchain/mc/sensitivity.py
```diff
@@ -179,10 +179,17 @@ def calculate_sensitivity_lima(n_excesses, n_background, alpha, n_bins_energy, n
...
+    if any(len(a) != n_bins_energy for a in (n_excesses, n_background, alpha)):
+        raise ValueError(
+            'Excess, background and alpha arrays must have the same length')
+           
+    
     stat = WStatCountsStatistic(
-        n_on=np.ones_like(n_background),
-        n_off=n_background,
-        alpha=alpha)
+        n_on = n_excesses,
```
rlopezcoto, line 188: here, if you want to calculate the sensitivity using this quantity, you would need to include not the excesses but the number of events in the on region (`n_excesses` + `n_background`). It is anyways not necessary, because we only use it to calculate the excess matching a 5 sigma significance, that is not dependent on `n_excesses`

lstchain/mc/sensitivity.py
```diff
@@ -179,10 +179,17 @@ def calculate_sensitivity_lima(n_excesses, n_background, alpha, n_bins_energy, n
...
+            'Excess, background and alpha arrays must have the same length')
+           
+    
     stat = WStatCountsStatistic(
-        n_on=np.ones_like(n_background),
-        n_off=n_background,
-        alpha=alpha)
+        n_on = n_excesses,  
+        n_off = n_background,
+        alpha = alpha*np.ones_like(n_background)
```
rlopezcoto, line 190: here `alpha` should be an array, so no need to multiply it by `np.ones_like(n_background)`

Also: if I'm not wrong, according to pep8, there should be no spaces between the `=` in function definitions/calls. There should, however, be spaces between operation symbols (`*`)

lstchain/mc/sensitivity.py
```diff
@@ -227,9 +234,11 @@ def calculate_sensitivity_lima_ebin(n_excesses, n_background, alpha, n_bins_ener
             'Excess, background and alpha arrays must have the same length')
 
     stat = WStatCountsStatistic(
-        n_on=np.ones_like(n_background),
-        n_off=n_background,
-        alpha=alpha)
+        n_on = n_excesses,
+        n_off = n_background,
+        alpha = alpha*np.ones_like(n_background))
```
rlopezcoto, line 239: same as above

lstchain/mc/sensitivity.py
```diff
@@ -352,8 +361,8 @@ def find_best_cuts_sensitivity(simtelfile_gammas, simtelfile_protons,
     mc_par_p['emin'] = mc_par_p['emin'].to(u.TeV)
     mc_par_p['emax'] = mc_par_p['emax'].to(u.TeV)
 
-    mc_par_g['area_sim'] = mc_par_g['area_sim'].to(u.cm ** 2)
-    mc_par_p['area_sim'] = mc_par_p['area_sim'].to(u.cm ** 2)
+    mc_par_g['area_sim'] = mc_par_g['area_sim'].to(u.m ** 2)
+    mc_par_p['area_sim'] = mc_par_p['area_sim'].to(u.m ** 2)
```
rlopezcoto, line 365: why this change of units here?

lstchain/mc/sensitivity.py
```diff
@@ -420,21 +432,20 @@ def find_best_cuts_sensitivity(simtelfile_gammas, simtelfile_protons,
 
     total_rate_proton = np.sum(rate_weighted_p)
     total_rate_gamma = np.sum(rate_weighted_g)
-    print("Total rate triggered proton [Hz]", total_rate_proton)
-    print("Total rate triggered gamma [Hz]", total_rate_gamma)
+    print("Total rate triggered proton {:.3f} Hz".format(total_rate_proton))
+    print("Total rate triggered gamma  {:.3f} Hz".format(total_rate_gamma))
```
rlopezcoto, line 436: all these printings should be substituted by loggings, but let's make a full refactor of the code in a different PR...

lstchain/mc/sensitivity.py
```diff
@@ -768,10 +780,11 @@ def sensitivity(simtelfile_gammas, simtelfile_protons,
...
     dl2 = pd.DataFrame(columns=events.keys())
+    
     for i in range(0, n_bins_energy):
         df_bin = events[(events.mc_energy < energy[i+1]) & (events.mc_energy > energy[i]) \
                                & (events.gammaness > gcut[i]) & (events.theta2 < tcut[i])]
 
         dl2 = pd.concat((dl2, df_bin))
 
-    return energy, sensitivity, result, units, dl2
+    return energy, sensitivity_flux, result, units, dl2
```
rlopezcoto, line 790: the output being written in the `result` tuple is actually `sensitivity`, instead of `sensitivity_flux`. We can discuss about changing it, but I would not change it now because that will break some compatibility of people using this code

lstchain/mc/sensitivity.py
lfoffano: I agree. Actually I did not modify this part, but I think someone wrote "n_excess" but it is actually the number of ON events. In the following parts of the program where it is used, the number of ON events is provided.

I am changing the variables name in the function, for better readibility: from n_excess to n_on_events

Thank you

lstchain/mc/sensitivity.py
lfoffano: Ah yes, thanks for the alpha suggestion.

And also for the spaces.

lstchain/mc/sensitivity.py
lfoffano: ah ok thanks, I will keep it in mind

lstchain/mc/sensitivity.py
lfoffano: OK I will keep sensitivity, but then I will add some lines in the lstchain_mc_sensitivity in order to plot the sensitivity_flux

lstchain/mc/sensitivity.py
lfoffano: There was a reason but I forget why I changed them. I will go back to the previous version with cm.

rlopezcoto (dismissed):

rlopezcoto:
@lfoffano travis is now complaining about some changes, could you please have a look?

lfoffano:
> @lfoffano travis is now complaining about some changes, could you please have a look?

yes, sure, let me take a look

lfoffano:
@rlopezcoto yes, basically it was looking for "events.leakage2_intensity" while filtering the events. I changed the feature in "leakage_intensity_width_2".

lstchain/mc/sensitivity.py
```diff
@@ -179,10 +177,17 @@ def calculate_sensitivity_lima(n_excesses, n_background, alpha, n_bins_energy, n
 
     """
 
+    if any(len(a) != n_bins_energy for a in (n_on_events, n_background, alpha)):
```
rlopezcoto, line 180: Hi @lfoffano, Travis is still complaining about this line. If I'm not wrong, `calculate_sensitivity_lima` is a function that is expecting `int/float` because it calculates the LiMa significance for a single energy bin, so this `if` statement will fail as reported in:
https://travis-ci.org/github/cta-observatory/cta-lstchain/jobs/727298230

rlopezcoto (approved):

PR #493, opened 2020-07-27
remove gps_time

vuillaut:
gps time is not used anywhere and is now written as nan value in dl1, causing errors down the line.

Same error can get caused by `mc_core_distance` filled by nan


tysaito2008:
thank you for removing it. It was just confusing.

vuillaut:
> I have tested the changes running lstchain_dl1_to_dl2.py and it is running well.

Thank you for the test and review

PR #496, opened 2020-07-30
Integration config

### LINKED ISSUE #422: image extractor config

vuillaut:
- correct the way to load the config for the image extractor
- There were 2 integration set of parameters in the standard config. I suppose the latest one prevails, so I kept only this one: (5,12)
- add unit test to check everything is working

fix #422


rlopezcoto:
regarding the two sets of parameters, if I'm not wrong, one of them was used for the flatfield events and the other one for the integration of cosmic events, right? (@FrancaCassol could comment on that).
In any case, the signal extraction that we would like to maintain by default for cosmic events is still (4, 8), now that we can calculate the correction for the amount of signal that has not been integrated

FrancaCassol:
Hi @rlopezcoto,

> regarding the two sets of parameters, if I'm not wrong, one of them was used for the flatfield events and the other one for the integration of cosmic events, right?

 yes, I confirm, for the moment this is the use of the two settings

rlopezcoto:
thanks @FrancaCassol. Then we agree that we should keep (8, 4) for cosmic events and (5, 12) for flatfield events

moralejo:
Related to this: @FrancaCassol, can you confirm that we can now use the pulse-integration correction on the real data without problems? 
Even if that is the case, it is probably better to keep a wider integration window for the calibration (so that most of the light is included in it).

yukihok:
Thank you for this helpful PR!
Concerning the integration correction for cosmic events, I think I can provide a typical cosmic pulse shape obtained from real data if it’s needed.
For the integration of calibration pulses, may we try to have a wider window to extract more charge from the pulses?
As far as I studied, the window (5, 12) will still lose ~8-9% of charge in both HG and LG.
(Though using wider window also increases the risk of going over a readout region…)

lstchain/data/lstchain_standard_config.json
```diff
@@ -148,14 +151,6 @@
        "charge_median_cut_outliers": [-0.5,0.5],
        "charge_std_cut_outliers": [-10,10],
        "time_cut_outliers": [2,38]
-     },
```
rlopezcoto, line 151: this is the configuration for the calibrator and should be left as it was

lstchain/data/lstchain_standard_config.json
```diff
@@ -98,20 +98,23 @@
...
-  "image_extractor_config": {},
   "gain_selector": "ThresholdGainSelector",
   "gain_selector_config": {
     "threshold":  3500
   },
   "charge_scale": [1.142,1.104],
   "LocalPeakWindowSum":{
-     "window_shift": 4,
-     "window_width":8
+     "window_shift": 5,
```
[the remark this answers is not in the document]
vuillaut: updated

lstchain/data/lstchain_standard_config.json
vuillaut: ha yes I think I misread the config, reverted that

rlopezcoto (approved):

lstchain/calib/camera/tests/test_calib.py
```diff
@@ -58,11 +59,18 @@ def test_load_calibrator_from_config_GlobalPeakWindowSum():
 def test_load_image_extractor_from_config():
     from ctapipe.image import LocalPeakWindowSum
 
+    config = get_standard_config()
+    image_extractor = load_image_extractor_from_config(config, subarray)
+
+    assert isinstance(image_extractor, LocalPeakWindowSum)
+    assert image_extractor.window_shift[0][2] == 5
```
rlopezcoto, line 66: this should be a 4 now

lstchain/calib/camera/tests/test_calib.py
```diff
@@ -58,11 +59,18 @@ def test_load_calibrator_from_config_GlobalPeakWindowSum():
...
+    assert image_extractor.window_width[0][2] == 12
```
rlopezcoto, line 67: and this an 8

rlopezcoto (approved):

PR #497, opened 2020-07-31
Ucts fix to account for wrongly tagged pedestal events due to skipping events

moralejo:
This introduces and offline fix for the shifting of the UCTS info by 1-event jumps.

It also makes that the DL1 output contains now all events, not just those with valid image parameters. I think this is much better to allow study e.g. of interleaved events in the standard DL1 files.


moralejo:
By the way, in order to work, the fix needs to have the proper dragon timestamps, so the r0 to dl1 step has to be launched with the command-line options to build them.

contrera:
Is this correction compatible with running the subruns in parallel ?

DirkHoffmann:
> Is this correction compatible with running the subruns in parallel ?

No, the comparisons and corrections need strictly incremented event order N → N+1 (and later N → N+2 etc., if the skip happends multiple times in the same run).

Unless someone has found a brilliant solution for that.

moralejo:
The correction is indeed compatible with running the subruns in parallel. But one needs the same that is needed to build a proper dragon time stamp: this is done now automatically in the onsite analysis, so no problem there.

DirkHoffmann:
Sorry, first of all I must admit that I mis-read "subruns" and understood streams (sub-files 1-4). 

But splitting a run into several parts still seems not obvious to me:

> The correction is indeed compatible with running the subruns in parallel.

How do you decide to shift the event fragments of later subruns (and by how much), if you have not yet processed all previous subruns of the same run, in order to know if there was a shift (or more)?

moralejo:
> Sorry, first of all I must admit that I mis-read "subruns" and understood streams (sub-files 1-4).
> 
> But splitting a run into several parts still seems not obvious to me:
> 
> > The correction is indeed compatible with running the subruns in parallel.
> 
> How do you decide to shift the event fragments of later subruns (and by how much), if you have not yet processed all previous subruns of the same run, in order to know if there was a shift (or more)?

The trick is that the process which launches each subrun gets information from the daily check which has already gone through the whole run. It somehow identifies in each subrun how to build the absolute dragon timestamp for the first event in the subrun (or the first with valid ucts info to be precise), and this is passed to the r0_to_dl1 process through two command-line arguments. With valid dragon time stamps we can compare the times with ucts and search for the jumps.

moralejo:
Ok, tests fail because I now write out all events to the DL1 output, and it is comparing one so-produced dl1 file, with a fixed test one which contains only survivors of the image parametrization...

contrera:
OK. I think right now the onsite analysis is not yet ready for this, we only consider one correction per run. Adapting the code to it may take time.

moralejo:
> OK. I think right now the onsite analysis is not yet ready for this, we only consider one correction per run. Adapting the code to it may take time.

No, I think it is ready. The dragon time stamps are built with one correction per subrun, I believe, at least Isidro provides them like that. And I think we could not have built at all the dragon timestamp (and hence detect the Crab pulsar) without that.

moralejo:
> 
> No, I think it is ready. The dragon time stamps are built with one correction per subrun, I believe, at least Isidro provides them like that. And I think we could not have built at all the dragon timestamp (and hence detect the Crab pulsar) without that.

After discussion with @contrera: it is not correct that it is 'one correction per subrun';  since the Dragon counter is reliable, we only need one fixed point in the whole run which connects Dragon counter and absolute time. And so the same fixed point (from subrun 0) is used for all subruns.

This however does not mean the UCTS event shift correction introduced in this PR will not work on a subrun-wise basis: only the first N events of a subrun cannot be corrected, with N being the number of jumps in the UCTS info that have happened in previous subruns. Since at least in current data these jumps are rare, I think this is a very minor limitation.

moralejo:
> Ok, tests fail because I now write out all events to the DL1 output, and it is comparing one so-produced dl1 file, with a fixed test one which contains only survivors of the image parametrization...

The test that fails is actually not against a fixed file, but against a file produced with lstchain_mc_dl1ab (i.e. opening the DL1 file and recalculating image parameters). However the differences between the files are not because of different number of events. It seems a different cleaning is used. I do not understand how the check was working before...

contrera:
> > No, I think it is ready. The dragon time stamps are built with one correction per subrun, I believe, at least Isidro provides them like that. And I think we could not have built at all the dragon timestamp (and hence detect the Crab pulsar) without that.
> 
> After discussion with @contrera: it is not correct that it is 'one correction per subrun'; since the Dragon counter is reliable, we only need one fixed point in the whole run which connects Dragon counter and absolute time. And so the same fixed point (from subrun 0) is used for all subruns.
> 
> This however does not mean the UCTS event shift correction introduced in this PR will not work on a subrun-wise basis: only the first N events of a subrun cannot be corrected, with N being the number of jumps in the UCTS info that have happened in previous subruns. Since at least in current data these jumps are rare, I think this is a very minor limitation.

You are right. Good. Then it can run in parallel and modifications are minimal.  :-)

lstchain/reco/r0_to_dl1.py
```diff
@@ -535,9 +545,79 @@ def r0_to_dl1(
...
+                        # list. We have another similar array for the trigger
+                        # types, previous_ucts_trigger_type
+                        #
+                        if len(previous_ucts_time_unix) > 0:
+                            # keep the time & trigger type read for this
+                            # event (which really correspond to a later event):
+                            current_ucts_time = dl1_container.ucts_time
+                            current_ucts_trigger_type = dl1_container.ucts_trigger_type
+                            # put in dl1_container the proper time for this
+                            # event:
```
MaxNoe, line 580: This whole section is unnecessarily complex and could be improved a lot by making use of the methods python lists provide, i.e. pop and appe d. 

I think here, only a call `l.pop(0)` is needed.

It looks like collections.deque (double ended queue) would even be a better solution, as one only needs to popleft and append which is much more efficient and easier to read than these shifting loops.

moralejo: > This whole section is unnecessarily complex and could be improved a lot by making use of the methods python lists provide, i.e. pop and appe d.
> 
> I think here, only a call `l.pop(0)` is needed.
> 
> It looks like collections.deque (double ended queue) would even be a better solution, as one only needs to popleft and append which is much more efficient and easier to read than these shifting loops.

Well I guess it depends on one's command of python. It is explained, and not hard to read.

moralejo:
Turns out that writing out the events which do not survive cleaning and parametrization to the DL1 file is pretty difficult, because the (deprecated) DL1ParametersContainer is a mess. Basically, a reset() leaves it in a state (initialized with None's, no units) which is incompatible with writing it later (that seems to me an ugly feature - anyone can think of an easy fix, wirth applying to a deprecated container?). 
Now tests are passing, this was the underlying reason.

moralejo:
**Note:** as of now, the correction of the UCTS info is done quite late in the loop, after calibration and parametrization, just because the needed dragon time is calculated there. Obviously this is inconvenient if one wants to use the info from interleaved pedestals e.g. in the cleaning. Eventually we have to move both the dragon time calculation and the ucts info shifting to a function called right after an event is read from the event source. Or even putting it directly into the event source, if it is possible...  For now it is still very useful like it is, because we will have all calibrated images properly tagged in the DL1 output, so one can re-do the cleaning starting from DL1 files, for optimizing cleaning strategies.

lstchain/reco/r0_to_dl1.py
MaxNoe: You will be able to substitute this whole sections with just two small lines. And replace a very inefficient loop with idiomatic code that will run much faster:

This should do it (I did not test it though)
```
dl1_container.ucts_time = previous_ucts_times.pop(0)
previous_ucts_times.append(current_ucts_time)
```

Even more efficient would be, as I said, the use of a double ended queue:

```
from collections import deque

previous_ucts_times = deque()

...
dl1_container.ucts_time = previous_ucts_times.popleft()
previous_ucts_times.append(current_ucts_time)
```

Much cleaner, don't you agree?

moralejo: Thanks, I like the first version. Will give it a try.

lstchain/reco/r0_to_dl1.py
```diff
@@ -154,12 +153,28 @@ def get_dl1(
...
+        # integer parameters.
+        #
+        for key in dl1_container.keys():
+            dl1_container[key] = u.Quantity(0, dl1_container.fields[key].unit)
+
+        # Fields width and length do not have in their declaration the units
+        # that are actually expected later in the program, so we set them here.
+        # We now use nans since these are floats, and will be later used to
+        # calculate W/L...
+        dl1_container.width = u.Quantity(np.nan, u.m)
```
rlopezcoto, line 170: shouldn't this unit declaration be included in the container definition instead of here?

lstchain/reco/r0_to_dl1.py
```diff
@@ -535,9 +560,75 @@ def r0_to_dl1(
...
+                            previous_ucts_time_unix.\
+                                insert( 0, dl1_container.ucts_time)
+                            previous_ucts_trigger_type.\
+                                insert(0, dl1_container.ucts_trigger_type)
+                            dl1_container.ucts_trigger_type = -1
+
                         # Select the timestamps to be used for pointing interpolation
                         if config['timestamps_pointing'] == "ucts":
-                            event_timestamps = ucts_time_utc.unix
+                            event_timestamps = dl1_container.ucts_time
```
rlopezcoto, line 631: are we sure this time is in the right format? I lost track after so many changes...

lstchain/scripts/lstchain_mc_dl1ab.py
```diff
@@ -143,11 +143,18 @@ def main():
...
-                        params[ii][p] = Quantity(dl1_container[p]).value
                 else:
-                    for p in parameters_to_update:
-                        params[ii][p] = 0
+                    # for consistency with r0_to_dl1.py:
+                    for key in dl1_container.keys():
+                        dl1_container[key] = \
+                            u.Quantity(0, dl1_container.fields[key].unit)
+
+                    dl1_container.width = u.Quantity(np.nan, u.m)
```
rlopezcoto, line 152: same comment as above

lstchain/scripts/lstchain_mc_dl1ab.py
```diff
@@ -143,11 +143,18 @@ def main():
...
+                    dl1_container.length = u.Quantity(np.nan, u.m)
+                    dl1_container.wl  = u.Quantity(np.nan, u.m)
+
+            for p in parameters_to_update:
```
rlopezcoto, line 156: shall we define `parameters_to_update` as a list external to the  `lstchain_mc_dl1ab.py` script?

lstchain/reco/r0_to_dl1.py
morcuended: I would say yes, `dl1_container.ucts_time` is defined above as `ucts_time_utc.unix` (unix format and UTC scale) and whenever it has to be corrected due to these jumps it'd always refer to `dl1_container.ucts_time` container earlier filled. One note: in OSA we use `"timestamps_pointing":"dragon"` in order to make the pointing interpolation using the corrected dragon timestamps as we agreed time ago.

lstchain/reco/r0_to_dl1.py
moralejo: The lstchain DL1 container is deprecated, as you know. The default unit in the definition is degrees, but in the filling (of successfully reconstructed events) meters are used. Hence that line. Of course I could fix all of that from scratch, but it is not the purpose of this PR. The purpose is to keep exactly the same behaviour as before, but assigning the right UCTS info to each event, when possible.

lstchain/scripts/lstchain_mc_dl1ab.py
moralejo: Ditto

lstchain/scripts/lstchain_mc_dl1ab.py
moralejo: No, they're always the same. And again, we want to move to ctapipe standard DL1 container as soon as possible, no need to improve this now.

lstchain/reco/r0_to_dl1.py
```diff
@@ -154,12 +153,28 @@ def get_dl1(
...
     else:
-        return None
+        # No image was parametrized, so we put zeros (instead of the default
+        # Nones) in all parameters: a container reset() is not an option because
+        # the default None values prevent the container to be written out. We
+        # cannot use np.nan either, because upon writing it complains for the
+        # integer parameters.
+        #
+        for key in dl1_container.keys():
+            dl1_container[key] = u.Quantity(0, dl1_container.fields[key].unit)
```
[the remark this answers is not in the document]
moralejo: I agree with you from a theoretical point of view. But here none of the parameters will be used for anything if intensity is 0. And we are anyway moving to the ctapipe DL1 containers as soon as we can, so why bother about something in the deprecated container, which even now has no impact?  What we badly need now is proper event tagging, and hopefully this provide it for most data.

rlopezcoto (approved):

PR #499, opened 2020-08-10
Fix mc_info writing for real data

SeiyaNozaki:
It seems `mc_type` was written after DL1 processing with real data.
I moved two lines for `mc_type` and `mc_core_distance` to be written only when MC data is used.


SeiyaNozaki:
I agree with Ruben, so I moved the lines for those mc parameters in `fill_mc`.

rlopezcoto (dismissed):

SeiyaNozaki:
I forgot this PR completely...
I'm happy if anyone has a look codes again!

rlopezcoto (approved):

vuillaut (approved):

PR #502, opened 2020-08-27
Add badges and modify README

rlopezcoto:
Add PyPI, codacy, travis and codecov badges to the README, and transform it to an `rst` one instead of the old `md`


MaxNoe (commented):
If you want to use rest, you should check it's validity, otherwise the pypi upload will fail.

I think markdown is a lot easier and probably sufficient for this README.

To lint the document, do:

```
$ pip install restructuredtext-lint
$ rst-lint README.rst
```

I see some issues, e.g. headings not having the correct length

vuillaut:
What is the reason to go for `rst` instead of `md`?

vuillaut:
Going through old PR.
Shall we close or update?
The badges can be added to the markdown README.

PR #503, opened 2020-08-27
Add sphinx docs

rlopezcoto:
First stab at adding documentation on lstchain.
Since the repo is growing up, and there are a growing number of users willing to analyze data, this should be helpful to initiate them. It will be complicated to keep it up to date because of the rapid evolution of the software, but we should try to include at least some basic documentation.
At the moment I just included the structure and enabled it in my personal fork:

https://rlopezcoto.github.io/cta-lstchain

if people agree, I'll move to include it in the upstream repo.


mexanick:
Hi Ruben, thanks for taking care of this. I just have one question: do we need the `index.html` tracked by the repository? I believe it is reproducible by the build from `index.rst` and thus should not be included

rlopezcoto:
@mexanick yes, you are right, I'll correct that and work this PR out a little bit more, so I'll move it to Draft.

vuillaut:
Is this still alive?
Quite needed in my opinion.

.github/workflows/ci.yml
```diff
@@ -69,3 +69,36 @@ jobs:
...
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install doc dependencies
+        run: |
+          sudo apt update --yes && sudo apt install --yes git build-essential pandoc graphviz
+          pip install -U pip setuptools wheel
+          pip install -e .[docs]
```
morcuended, line 89: I think this is currently not working since there is no `docs` requirement in `setup.py`

Issue #438, opened 2020-06-19
Store energy in the muon MC analysis


rlopezcoto:
At the moment we are not storing the simulated energy in the output file of the muon analysis and it would be necessary to produce spectra. This may not be necessary once we move to ctapipe v0.8, but it may be a reasonable addition in the meantime

rlopezcoto:
Solved in #506

PR #506, opened 2020-08-28
Add energy to muon tables

rlopezcoto:
Useful feature until we move to a full ctapipe-based muon analysis (see #438)


moralejo (approved):

moralejo:
Would be better to solve the codacy issues, but it is just trailing whitespaces.

rlopezcoto:
> Would be better to solve the codacy issues, but it is just trailing whitespaces.

thanks, I solved them and got a few new ones. There is still one warning about an unused variable that is the output of a function, but I'd rather leave it as for the time being

moralejo (approved):

PR #510, opened 2020-09-10
Update to astropy v4

rlopezcoto:
Following the discussion on slack:
https://ctapipe.slack.com/archives/C0160M3K362/p1596115158284400

```py3
astropy>=3,<5
```
may install astropy v3.2.3, which is not compatible with lstchain >= v0.6, so this should solve the problem

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
     version=version,
     packages=find_packages(),
     install_requires=[
-        "astropy>=3,<5",
+        "astropy>3,<5",
         'ctapipe~=0.8.0',
         'ctaplot~=0.5.3',
         "eventio>=1.1.1,<2.0.0a0",  # at least 1.1.1, but not 2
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -42,7 +42,7 @@ def find_scripts(script_dir, prefix):
     version=version,
     packages=find_packages(),
     install_requires=[
-        "astropy>=3,<5",
+        "astropy>3,<5",
```
MaxNoe, line 45: using `~=4.0` would be more idiomatic

MaxNoe (approved):

setup.py
MaxNoe: It shows explicitly which version is wanted instead of just excluding which are not

rlopezcoto: and what when we get 4.1?
wouldn't `astropy>3,<5` install the latest one?

MaxNoe: The tilde notation does the same, it allows any version `4.x` greater than the one given.

So `~=4.0` is equivalent to `>=4.0,<5`

MaxNoe: https://www.python.org/dev/peps/pep-0440/

rlopezcoto: ok, thanks, I updated it!

MaxNoe (approved):

PR #511, opened 2020-09-10
Generating event lists and 2 IRFs, to create DL3 fits file 

chaimain:
Proceeding with creating event lists of LST DL2 files as per the standard https://gamma-astro-data-formats.readthedocs.io/en/latest/ and using pyIRF to create simple point-like IRFs - Energy Dispersion and Effective Area.

Usage of pyIRF is without using any cuts optimization for now, and have made a copy of the config file used by pyIRF.
pyIRF's master branch still has old dependencies (gammapy and ctapipe) which were not compatible for the current script. I used a separate [branch](https://github.com/chaimain/pyirf/tree/gammapy017) with gammapy v0.17 dependency to produce the IRFs.

I have made a separate file for cleaning the data in a separate new path lstchain/clean/. This is most likely unnecessary, but I tried to have a separate code instead of the filter_events in lstchain/reco/utils.py, to separate it with user-defined cuts.

I have also made a separate file hdu_table.py in a separate new path lstchain/hdu/ to include the main functions required in creating the event lists and HDU and Obs index files.

The event list script works on a single file right now, but can easily be changed to work for a list of runs in the same path. Accordingly, I separated the script for creating HDU and Obs index fits files from a list of DL3 fits files, as I could only write it once or overwrite the existing.


MaxNoe:
pyirf has no released version yet and will change very much before and during the coding sprint at the end of september , so I wouldn't really try to use it before that is over and a first release is made.

chaimain:
Yes, I understand that and would look into incorporating the updated version of pyIRF to create the IRFs for the DL3. I have added the option to not include the IRFs and just create DL3 files with the event lists for now. The IRFs generated are very basic and so it seemed ok to just try to include them with the event lists to generate DL3 file with LST data.
The draft PR is also intended to understand how the structure of generating DL3 files from DL2 files can be in the future.

chaimain:
I have updated the core pyIRF functions with the current developer version of pyIRF (v0.3.0dev0), changed the selection cuts config file from .yml to .json and added some tests .
There are still things to be updated, most of which shall be done in a few days.

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,464 @@
...
+    'make_aeff2d_hdu',
+    'create_obs_hdu_index',
+    'create_event_list'
+    ]
+
+### TODO:
+#once the pyIRF IRF HDUs are generalized for our use, we can import the functions
+# directly from pyIRF
+
+def make_aeff2d_hdu(
```
MaxNoe, line 22: I do not understand why you duplicated these functions. They are besides the adding of an additional dimesion to the effecive area exactly the same as in pyirf.

You should add the additional dimension and then just call the pyirf function.

Something like this
```
def add_dummy_area_bin(effective_area):
    return np.column_stack([effective_area, np.zeros_like(effective_area)])

```

And then call the function from pyirf:

```
effective_area = add_dummy_area_bin(effective_area_per_energy(...))

hdu = create_effective_area_hdu(...)
```

Duplicating the code does not make sense

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,464 @@
...
+    header["HDUCLAS2"] = "EFF_AREA"
+    header["HDUCLAS3"] = "POINT-LIKE" if point_like else "FULL-ENCLOSURE"
+    header["HDUCLAS4"] = "AEFF_2D"
+    header["DATE"] = Time.now().utc.iso
+    header["TELESCOP"] = "CTA"
+    header["INSTRUME"] = "LST-1"
+
+    return fits.BinTableHDU(aeff, header=header, name=extname)
+
+def make_edisp2d_hdu(
```
MaxNoe, line 57: Again, why did you copy this? The function of pyirf is the same.

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,464 @@
...
+    obs_table.meta["TELESCOP"] = "CTA"
+    obs_table.meta["INSTRUME"] = "LST-1"
+
+    obs_table.meta['MJDREFI'] = event_table.meta['MJDREFI']
+    obs_table.meta['MJDREFF'] = event_table.meta['MJDREFF']
+
+    filename_obs_table = 'obs-index.fits.gz'
+    obs = fits.BinTableHDU(obs_table)
+    hdu_list = fits.HDUList()
+    hdu_list.append(obs)
```
MaxNoe, line 323: You need a primary HDU as the first hdu, just do `hdu_list= fits.HDUList([fits.PrimaryHDU] + obs)`

lstchain/hdu/hdu_table.py
chaimain: Ah! Yes, I can do this. I assumed that as I am using gammapy to analyse the DL3 files, and where they still have the issue you mentioned, it will be safer to use the recreate it here, until it is resolved both in pyirf and gammapy. But I understand your point and will make the changes

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,464 @@
...
+                    pointing_alt = pointing_alt, pointing_az = pointing_az,
+                    obstime = time)
+    coord_icrs = coord.icrs
+    coord_pointing = camera_to_altaz(pos_x = 0 * u.m, pos_y=0 * u.m, focal = focal,
+                pointing_alt = pointing_alt[0], pointing_az = pointing_az[0],
+                obstime = time[0])
+
+    ##########################################################################
+    ### Event table columns
+    event_table["EVENT_ID"] = Column(
```
MaxNoe, line 379: It's much easier and much less code to just use an astropy table than declaring these low-level fits objects.

lstchain/hdu/hdu_table.py
chaimain: Understood

environment.yml
```diff
@@ -6,6 +6,7 @@ channels:
 dependencies:
   - python=3.7
   - pip
+  - pyirf>=0.2.0
```
vuillaut, line 9: I think there will be a release soon and this should be changed to the latest version

lstchain/clean/data_clean.py
```diff
@@ -0,0 +1,40 @@
...
+        data = data[data['leakage_intensity_width_2'] < data_cut["general"]["leakage_intensity_width_2"]]
+    data = data[data['intensity'] > data_cut["general"]["intensity"]]
+    data = data[data['wl'] > data_cut["general"]["wl"]]
+    data = data[data['gh_score'] > data_cut["general"]["gh_score"]]
+    data = data[data['theta'].value < data_cut["mc"]["theta_cut"]]
+    data = data[data['source_fov_offset'].value < data_cut["general"]["source_fov_offset"]]
+
+    return data
+
+def data_filter(data):
```
vuillaut, line 29: such function already exists
`lstchain.io.io.filter_events`

vuillaut: You might want a more specific function for your use case but you can reuse `filter_events`

vuillaut (changes requested):
I think it would be appropriate to create an `irf` submodule now also.

lstchain/data/data_selection_cuts.json
```diff
@@ -0,0 +1,15 @@
+{
+  "general":{
+    "gh_score": 0.8,
+    "source_fov_offset": 2.83,
+    "intensity": 200,
+    "r": 1,
+    "wl": 0.1,
+    "leakage2_intensity": 0.2,
```
vuillaut, line 8: does this parameter still exist?

lstchain/clean/data_clean.py
chaimain: I thought that using filter_events and applying cuts, might be slightly different, hence tried this way. But I will try with filter_events and check it.

vuillaut: simple suggestion, if it does not fit your needs, make your function no problem (it would just be less code)

lstchain/data/data_selection_cuts.json
chaimain: To use it for pre v0.6 format files. I have created a condition in the data_clean.py script to account for it.

environment.yml
MaxNoe: And requirung == probably makes sense. Or at least `!=0.x.0` as there will probably be breaking changes down the road until we have a more stable interface.

lstchain/clean/data_clean.py
MaxNoe: Filtering events is applying cuts. No semantic difference.

lstchain/clean/data_clean.py
```diff
@@ -0,0 +1,40 @@
...
+    if 'leakage2_intensity' in data.columns:
+        data = data[data['leakage2_intensity'] < data_cut["general"]["leakage2_intensity"]]
+    else:
+        data = data[data['leakage_intensity_width_2'] < data_cut["general"]["leakage_intensity_width_2"]]
+    data = data[data['intensity'] > data_cut["general"]["intensity"]]
+    data = data[data['wl'] > data_cut["general"]["wl"]]
+    data = data[data['gh_score'] > data_cut["general"]["gh_score"]]
+    #data = data[data['theta']] < data_cut["mc"]["theta_cut"]
+    data = data[data['source_fov_offset'].value < data_cut["general"]["source_fov_offset"]]
```
MaxNoe, line 38: Never use `.value` when unsure of units. Use `to_value(unit)` to be sure to get a numpy array with the correct unit.`

lstchain/data/data_selection_cuts.json
MaxNoe: No, it is the old name of the parameter one line below

lstchain/data/data_selection_cuts.json
```diff
@@ -0,0 +1,15 @@
...
+    "leakage_intensity_width_2": 0.2,
+    "alpha": 8
+    },
+  "mc":{
+    "theta_cut": 0.14
```
MaxNoe, line 13: I would highly recommend to also use units in the config file. This could be achieved like this:

```
"theta_cut": {"value": 0.14, "unit": "deg"}
```

And then reading with:

```
theta_cut = u.Quantity(**config["theta_cut"])
```

chaimain: Ok thanks! I will make the change.

lstchain/clean/data_clean.py
chaimain: OK, understood.

lstchain/data/data_selection_cuts.json
chaimain: Using units in this config file, would not let me use filter_events as it is. Would you then recommend to make this config file like the lstchain_standard_config to suit filter_events? As only the angular_offset parameters need units, so those cuts can be just adjusted in the main script itself.

environment.yml
maxnoe: Change to `~=0.3.0`, we made the release today.

lstchain/clean/data_clean.py
```diff
@@ -0,0 +1,40 @@
...
+### TODO: We can generalize for the user to choose which selection cuts to make
+def mc_filter(data):
+    if 'leakage2_intensity' in data.columns:
+        data = data[data['leakage2_intensity'] < data_cut["general"]["leakage2_intensity"]]
+    else:
+        data = data[data['leakage_intensity_width_2'] < data_cut["general"]["leakage_intensity_width_2"]]
+    data = data[data['intensity'] > data_cut["general"]["intensity"]]
+    data = data[data['wl'] > data_cut["general"]["wl"]]
+    data = data[data['gh_score'] > data_cut["general"]["gh_score"]]
+    data = data[data['theta'].to_value() < data_cut["mc"]["theta_cut"]]
```
maxnoe, line 24: `to_value` needs a `unit` argument: `to_value(u.deg)` for example.

maxnoe: But again, I would also put the `theta_cut` as unit, and then you don't need the `to_value` at all.

lstchain/clean/data_clean.py
```diff
@@ -0,0 +1,40 @@
...
+        'data_filter'
+]
+
+file = os.path.join(os.path.dirname(__file__),"../data/data_selection_cuts.json")
+
+with open(file) as json_file:
+    data_cut = json.load(json_file)
+
+### TODO: We can generalize for the user to choose which selection cuts to make
+def mc_filter(data):
```
maxnoe, line 16: These filtering functions are the same? And they have to be the same for data and mc.

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,350 @@
...
+
+    Parameters
+    ----------
+    filename_list : list
+        list of filenames
+    fits_dir : str
+        directory containing the fits file
+    """
+
+    # empty lists with the quantities we want
```
maxnoe, line 50: Making a list for each of the required variables is very verbose and makes it hard to follow.

I would recommend to use a single list of dictionaries instead, which makes it most of the time much easier to follow and a lot less code, which is also easier to maintain, because adding something only requires change at a single position, not at two.

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,350 @@
...
+    tstart = []
+    tstop = []
+    source = []
+    N_TELS = []
+    TELLIST = []
+
+    # loop through the files
+    for file in filename_list:
+
+        if os.path.exists(str(fits_dir)+"/"+file):
```
maxnoe, line 71: Don't use string concatenation for paths. Use `os.path.join` or `pathlib.Path` operations.

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,350 @@
...
+            try:
+                event_table = Table.read(str(fits_dir)+"/"+file, hdu="EVENTS")
```
maxnoe, line 73: same as above

lstchain/hdu/hdu_table.py
maxnoe: Why do you need to convert fits_dir to `str` anyways? If it is already `pathlib.Path`, then simply use `(fits_dir / file).is_file()`

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,350 @@
...
+                gti_table = Table.read(str(fits_dir)+"/"+file, hdu="GTI")
+                pointing_table = Table.read(str(fits_dir)+"/"+file, hdu="POINTING")
+            except Exception:
+                print(f"fits corrupted for file {file}")
```
maxnoe, line 77: Broken data files should not be ignored I think. This should raise an exception!

Also, use logging, not print statements.

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,350 @@
...
+                continue
+        else:
+            print(f"fits {file} doesn't exist")
```
maxnoe, line 80: This should also raise an error. The user wanted a file to be included, if it is not there, ignoring that silently is not good.

maxnoe: Where silently is relative, I know, but I think just printing it is not enough

lstchain/hdu/hdu_table.py
```diff
@@ -0,0 +1,350 @@
...
+        t_pnt = t_events.copy()
+        t_pnt['HDU_TYPE'] = ['pointing']
+        t_pnt['HDU_CLASS'] = ['pointing']
+        t_pnt['HDU_NAME'] = ['pointing']
+
+        hdu_tables.append(t_pnt)
+
+        ###############################################
+        #Energy Dispersion
+        if Table.read(str(fits_dir)+"/"+file, hdu="EDISP"):
```
maxnoe, line 132: same as above, don't use str operations here.

lstchain/clean/data_clean.py
chaimain: Yes, they will be. Just that the theta² cut is not applied on real data now. It will be used in gammapy analysis, when selecting the dataset's on_region. Also, I am going to remove this clean submodule and just use filter_events and separate fixed cuts of g/h score and theta cuts, directly in the main script.

lstchain/hdu/hdu_table.py
chaimain: I will change it accordingly, thanks!

lstchain/hdu/hdu_table.py
chaimain: Yes, I had forgotten about this. Will amend it.

chaimain:
I am closing this draft PR, as this was a mess with some erroneously added 'git pulls' from different branches. I will make a new draft PR, with incorporating all the comments mentioned here. I apologize again for the mess and repetition.

PR #512, opened 2020-09-14
Fixed wrong arguments in call to fill_muon_event

moralejo:
For the real data case a recently added argument (only used for MC) was missing.

### DIFF
--- a/lstchain/image/muon/muon_analysis.py
+++ b/lstchain/image/muon/muon_analysis.py
@@ -439,6 +439,8 @@ def fill_muon_event(parameters, output_parameters, good_ring, event_id,
     output_parameters['event_time'].append(event_time)
     if parameters is not None:
         output_parameters['mc_energy'].append(parameters.mc_energy)
+    else:
+        output_parameters['mc_energy'].append(-1.)
     output_parameters['ring_size'].append(size)
     output_parameters['size_outside'].append(size_outside_ring)
     output_parameters['ring_center_x'].append(muonringparam.center_x.value)
--- a/lstchain/image/muon/muon_analysis.py
+++ b/lstchain/image/muon/muon_analysis.py
@@ -437,7 +437,8 @@ def fill_muon_event(parameters, output_parameters, good_ring, event_id,
                     muonparameters, hg_peak_sample=np.nan, lg_peak_sample=np.nan):
     output_parameters['event_id'].append(event_id)
     output_parameters['event_time'].append(event_time)
-    output_parameters['mc_energy'].append(parameters.mc_energy)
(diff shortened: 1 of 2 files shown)


lstchain/reco/r0_to_dl1.py
```diff
@@ -731,7 +731,8 @@ def r0_to_dl1(
                                 lg_peak_sample = np.argmax(stacked_waveforms, axis=-1)[1]
 
                             if good_ring:
-                                fill_muon_event(muon_parameters,
+                                fill_muon_event(None,
```
rlopezcoto, line 734: although here we are inside `not is_simu`, why passing a `None` when we could pass the `parameters` data frame?

lstchain/image/muon/muon_analysis.py
```diff
@@ -437,7 +437,8 @@ def fill_muon_event(parameters, output_parameters, good_ring, event_id,
                     muonparameters, hg_peak_sample=np.nan, lg_peak_sample=np.nan):
     output_parameters['event_id'].append(event_id)
     output_parameters['event_time'].append(event_time)
-    output_parameters['mc_energy'].append(parameters.mc_energy)
+    if parameters is not None:
```
rlopezcoto, line 440: `parameters.mc_energy` should exist also for real data, just filled with `None`
https://github.com/cta-observatory/cta-lstchain/blob/93ccc5c782e89ff1f62e575d266fff085a957fa8/lstchain/io/lstcontainers.py#L75
would that give an error?

rlopezcoto (commented):
thanks for spotting @FrancaCassol and @moralejo!

lstchain/reco/r0_to_dl1.py
moralejo: ok, I was too lazy to check that. Will do.

moralejo: > ok, I was too lazy to check that. Will do.

Just added -1. for the mc_energy  for the real data case, to avoid possible issues. I see no advantage in passing a container that is not used anyway.

lstchain/image/muon/muon_analysis.py
moralejo: > `parameters.mc_energy` should exist also for real data, just filled with `None`
> https://github.com/cta-observatory/cta-lstchain/blob/93ccc5c782e89ff1f62e575d266fff085a957fa8/lstchain/io/lstcontainers.py#L75
> 
> 
> would that give an error?

Don't know, but let's move forward.  -1 is a sufficiently clear error value for an energy.

rlopezcoto (approved):

PR #513, opened 2020-09-15
Correct the gain selection mask

FrancaCassol:
- Correct the present gain mask that would bring to wrong results if the charge integration correction is activated (which is not the case now)

- Remove an obsolete card in the configuration file


rlopezcoto (commented):
Thanks @FrancaCassol, could you add a comment in the code on why for the meantime we are leaving ``` no_gain_selection[1] = 1 ``` ?

lstchain/calib/camera/calibrator.py
```diff
@@ -189,7 +189,7 @@ def _calibrate_dl1(self, event, telid):
         # TBD: - perform calibration of the R1 waveform (not DL1)
         #      - gain selection before charge integration
 
-
+        # In case of no gain selection the selected gain channels are  [0,0,..][1,1,..]
```
MaxNoe, line 192: This is not "no gain selection", right? This is "always choose high-gain"?

FrancaCassol: no, this is no gain selection, why do you say that?

rlopezcoto (approved):

PR #517, opened 2020-09-25
New script for displaying long-term evolution of DL1 data checks

moralejo:
Far from being finished, but this already produces an interactive web page in which many quantities can be checked on a run-wise and pixel-wise basis.
Example of output here: http://www.lst1.iac.es/datacheck/dl1/longterm_dl1_check.html (pass-protected, under LST1 pages)

### DIFF
--- a/lstchain/datachecks/dl1_checker.py
+++ b/lstchain/datachecks/dl1_checker.py
@@ -1349,14 +1349,15 @@ def show_camera(content, geom, pad_width, pad_height, label):
 
     allimages = []
     if np.ndim(content) == 1:
-        np.nan_to_num(content, copy=False)
+        #np.nan_to_num(content, copy=False)
         allimages.append(content)
     else:
         for i in range(1,numsets+1):
-            np.nan_to_num(content[i-1], copy=False)
+            #np.nan_to_num(content[i-1], copy=False)
             allimages.append(content[i-1])
 
-    cam = BokehCameraDisplay(camgeom, np.min(allimages), np.max(allimages),
+    cam = BokehCameraDisplay(camgeom, np.nanmin(allimages),
+                             np.nanmax(allimages),
                              use_notebook=False, autoshow=False)
     cam.image = allimages[0]
 
@@ -1367,13 +1368,13 @@ def show_camera(content, geom, pad_width, pad_height, label):
         for i, x in enumerate(logcontent):
             # workaround as long as log z-scale is not implemented in bokeh camera:
             if x <= 0:
-                logcontent[i] = -3.
+                logcontent[i] = np.nan
             else:
                 logcontent[i] = np.log10(image[i])
-            allimageslog.append(logcontent)
+        allimageslog.append(logcontent)
 
-    camlog = BokehCameraDisplay(camgeom, np.min(allimageslog),
-                                np.max(allimageslog),
+    camlog = BokehCameraDisplay(camgeom, np.nanmin(allimageslog),
+                                np.nanmax(allimageslog),
                                 use_notebook=False, autoshow=False)
     camlog.image = allimageslog[0]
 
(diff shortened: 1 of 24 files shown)


lstchain/datachecks/dl1_checker.py
```diff
@@ -29,31 +24,23 @@
...
 from ctapipe.io import HDF5TableWriter
 from ctapipe.visualization import CameraDisplay
-from ctapipe.visualization.bokeh import CameraDisplay as BokehCameraDisplay
 from datetime import datetime
 from lstchain.datachecks.containers import DL1DataCheckContainer
 from lstchain.datachecks.containers import DL1DataCheckHistogramBins
 from lstchain.io.io import dl1_params_lstcam_key
 from lstchain.paths import parse_datacheck_dl1_filename, parse_dl1_filename, \
     run_to_muon_filename, run_to_datacheck_dl1_filename
+# from lstchain.visualization.bokeh import plot_mean_and_stddev_bokeh
```
rlopezcoto, line 37: can this be simply removed?

lstchain/datachecks/dl1_checker.py
```diff
@@ -364,8 +351,11 @@ def plot_datacheck(datacheck_filename, out_path=None, muons_dir=None):
     geom = CameraGeometry.from_table(cam_description_table)
     engineering_geom = geom.transform_to(EngineeringCameraFrame())
 
+    '''
```
rlopezcoto, line 354: not sure if this is the best way to comment out large blocks of comments. `'''` is usually used for docstrings, but maybe somebody else can comment

lstchain/datachecks/dl1_checker.py
```diff
@@ -520,6 +510,8 @@ def plot_datacheck(datacheck_filename, out_path=None, muons_dir=None):
         if table_pedestals is None or len(table_pedestals) == 0:
             write_error_page('pedestals', pagesize)
         else:
+            '''
+            For future bokeh-based display, turned off for now:
```
rlopezcoto, line 514: same as above

lstchain/datachecks/dl1_checker.py
```diff
@@ -541,14 +532,16 @@ def plot_datacheck(datacheck_filename, out_path=None, muons_dir=None):
         if table_flatfield is None or len(table_flatfield) == 0:
             write_error_page('flatfield', pagesize)
         else:
+            '''
```
rlopezcoto, line 535: same as above

lstchain/datachecks/dl1_checker.py
moralejo: This is for the future bokeh version of the current run data check pdf. I prefer to keep it.

lstchain/datachecks/dl1_checker.py
moralejo: You're right, I am replacing it with hashes.

lstchain/datachecks/dl1_checker.py
moralejo: I included here a workaround to solve https://github.com/cta-observatory/cta-lstchain/issues/518. 
I know, that fix does not really belong here, apologies for that, but we have to move on and I do not have time to make a separate PR and test it (would surely make a mess with my branches).

lstchain/reco/r0_to_dl1.py
```diff
@@ -451,9 +451,17 @@ def r0_to_dl1(
                         'HillasParameterizationError in get_dl1()'
                     )
 
-                # The condition below should now be true for all events, this
-                # is a relic of previous approach in which only survivors of
-                # cleaning and parametrization were further processed.
+                # Workaround, see below.
+                if i == 0 and dl1_container.n_pixels == 0:
```
rlopezcoto, line 455: what if the second event `(i == 1)` also has `dl1_container.n_pixels == 0`?

morcuended: Could the condition be extended to the first `n` events?

moralejo: Could happen I guess, but only if normal triggers are not activated and calib fox is off. Will fix this and also display a message when an event is skipped.

moralejo: I am working on a cleaner fix, I will put this PR back to draft.

moralejo: I reverted the workaround to solve #518, will be done in a separate PR

rlopezcoto: so this error does not show up anymore when #523 is merged? I guess @moralejo you tested with one of those files found in #518, right?

moralejo: I tested it with one of the failing subruns, and with the previous (normal) subrun, and it works in both cases. Will ask @morcuended  to try it on a full night.

rlopezcoto (approved):

PR #519, opened 2020-09-28
Fix time message when no UCTS time info is available

morcuended:
There was a bug in the implementation of the logging of time information when UCTS is not available. It was always showing the message "Not valid UCTS timestamp found" even though the time information was present. This fix should prevent that message from appearing always.

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -358,8 +358,8 @@ def r0_to_dl1(
         writer._h5file.filters = filters
         print("USING FILTERS: ", writer._h5file.filters)
 
-        first_valid_ucts = None
-        first_valid_ucts_tib = None
+        first_valid_ucts = np.nan
+        first_valid_ucts_tib = np.nan
         previous_ucts_time_unix = []
         previous_ucts_trigger_type = []
 
@@ -498,7 +498,7 @@ def r0_to_dl1(
                                 # UCTS presence flag is OK
                                 ucts_time = event.lst.tel[telescope_id].evt.ucts_timestamp * 1e-9  # secs
 
-                                if first_valid_ucts is None:
+                                if math.isnan(first_valid_ucts):
                                     first_valid_ucts = ucts_time
 
                                     initial_dragon_counter = (
@@ -513,8 +513,8 @@ def r0_to_dl1(
                                         f"corresponding Dragon counter {initial_dragon_counter:.9f} s"
                                     )
 
-                                if first_event.lst.tel[1].evt.extdevices_presence & 1 \
-                                        and first_valid_ucts_tib is None:
+                                if event.lst.tel[telescope_id].evt.extdevices_presence & 1 \
+                                        and math.isnan(first_valid_ucts_tib):
                                     # Both TIB and UCTS presence flags are OK
                                     first_valid_ucts_tib = ucts_time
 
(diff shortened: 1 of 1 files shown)


lstchain/reco/r0_to_dl1.py
```diff
@@ -356,23 +356,22 @@ def r0_to_dl1(
...
+        first_valid_ucts = np.nan
+        first_valid_ucts_tib = np.nan
         previous_ucts_time_unix = []
         previous_ucts_trigger_type = []
 
         for i, event in enumerate(chain([first_event],  event_iter)):
 
             if i % 100 == 0:
-                print(i)
+                logger.info(i)
```
MaxNoe, line 369: You could just use a real progress bar like `tqdm`?

MaxNoe (commented):
Could you explain why this fixes the issue?

All I can see is that before it was None and then checked if it was None and now it is nan and then checked if it is nan.

Why does this change from `None` to `nan` fix the behaviour?

lstchain/reco/r0_to_dl1.py
morcuended: Will try. There are though some other messages from muon rings mixed with the events evolution:

```
...
700
Impact parameter=3.978 m, ring_width=0.073 deg, ring radius=1.075 deg, ring completeness=0.700
800
...
```
The message does not come from this script itself but somewhere else.

Would that be a problem with the progress bar?

MaxNoe: It will then span over muttiple lines, creating a newline each time a non-progressbar message is printed. But it will still work.

morcuended:
> Could you explain why this fixes the issue?
> 
> All I can see is that before it was None and then checked if it was None and now it is nan and then checked if it is nan.
> 
> Why does this change from `None` to `nan` fix the behaviour?

This change from `None` to `nan` does not fix the problem and it probably should be changed back to `None`. What actually solve the problem is to consider also the case in which you provide the initial timestamps/counters via command line. The redefinition of the first valid ucts timestamp found:

```
if math.isnan(first_valid_ucts):
    first_valid_ucts = ucts_time
```
was being done only when those initial timestamps were not given by the user

lstchain/reco/r0_to_dl1.py
```diff
@@ -545,9 +542,15 @@ def r0_to_dl1(
                                     event.lst.tel[telescope_id].evt.tib_pps_counter +
                                     event.lst.tel[telescope_id].evt.tib_tenMHz_counter * 10 ** (-7)
                             )
+
                             if event.lst.tel[telescope_id].evt.extdevices_presence & 2:
                                 # UCTS presence flag is OK
                                 ucts_time = event.lst.tel[telescope_id].evt.ucts_timestamp * 1e-9  # secs
+                                if math.isnan(first_valid_ucts):
```
MaxNoe, line 549: Ah I see now. This line is the actual fix, right?

MaxNoe: THen yes, it should be changed back to None. THen the diff also becomes more obvious.

morcuended: Yes. Actually the way I've done it is kind of confusing with too many if/else to take into account the different possibilities

morcuended: Should I change back the nan's to None's?

MaxNoe (approved):
Looks much clearer now, thanks!

lstchain/reco/r0_to_dl1.py
morcuended: OK. To be implemented in a new PR

rlopezcoto: we should get rid of these muon messages or improve them...

PR #521, opened 2020-09-28
read dl2 and sim info into pyirf internal format

vuillaut:
I will add a unit test

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -40,7 +40,8 @@ __all__ = ['read_simu_info_hdf5',
            'write_metadata',
            'write_dataframe',
            'write_dl2_dataframe',
-           'write_calibration_data'
+           'write_calibration_data',
+           'read_dl2_to_pyirf',
            ]
 
 
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -18,7 +18,9 @@ from ctapipe.tools.stage1 import Stage1ProcessorTool
 from astropy.utils import deprecated
 from ctapipe.instrument import OpticsDescription, CameraGeometry, CameraDescription, CameraReadout, \
     TelescopeDescription, SubarrayDescription
-
+from pyirf.simulations import SimulatedEventsInfo
+from astropy import table
+import pandas as pd
 
 __all__ = ['read_simu_info_hdf5',
            'read_simu_info_merged_hdf5',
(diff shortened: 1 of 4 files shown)


lstchain/io/io.py
```diff
@@ -18,7 +18,9 @@
 from astropy.utils import deprecated
 from ctapipe.instrument import OpticsDescription, CameraGeometry, CameraDescription, CameraReadout, \
     TelescopeDescription, SubarrayDescription
-
+from pyirf.simulations import SimulatedEventsInfo
+from astropy import table
+import pandas as pd
 
 __all__ = ['read_simu_info_hdf5',
            'read_simu_info_merged_hdf5',
```
chaimain, line 26: Here 'read_dl2_to_pyirf' should be included as well

chaimain: And also in the init script

vuillaut: Done

vuillaut: (in init it's only necessary to expose the function in `lstchain.io` which is useful for external users - I don't think it's necessary here)

chaimain: Oh ok! Understood, thanks!

JouvinLea (approved):

MaxNoe:
Maybe working on a pyirf branch until everything is ready might be a good idea instead of merging pieces into the master?

rlopezcoto:
I agree with @MaxNoe, specially now that we are finalizing the tests on the data before releasing v0.6.1, it would be nice not to include in the master any non-necessary changes not to break any compatibilities and have reproducible results

moralejo:
For me tests are not passing, it complains about missing pyirf.

chaimain:
Yes, the setup file also has to be updated, and as @MaxNoe suggested, we will work on a separate branch to work on our scripts, until both lstchain and pyirf are ready for the DL3 scripts.

moralejo:
Ok, please revert these changes. Amazingly I cannot even run the R0 to DL1 step - though it should not at all be related to IRFs, I guess

moralejo:
Actually it is a bit urgent, would be nice if sb can fix this today

rlopezcoto:
@moralejo I guess that you need to update the environment with:
`conda env update -n lst-dev -f environment.yml`

does that work?

moralejo:
> @moralejo I guess that you need to update the environment with:
> `conda env update -n lst-dev -f environment.yml`
> 
> does that work?

Yes, thanks! I thought pip install would do it...

PR #523, opened 2020-09-28
Add default values to containers outside of their possible ranges

moralejo:
Make a cleaner initialization of the DL1ParametersContainer, and reset it before processing every event.


lstchain/io/lstcontainers.py
```diff
@@ -51,23 +56,34 @@ class DL1ParametersContainer(Container):
...
-    alt_tel = Field(None, 'Telescope altitude pointing', unit=u.rad)
-    az_tel = Field(None, 'Telescope azimuth pointing', unit=u.rad)
-
-    obs_id = Field(None, 'Observation ID')
-    event_id = Field(None, 'Event ID')
-    calibration_id = Field(None, 'ID of the employed calibration event')
+    time_gradient = Field(np.nan, 'Time gradient in the camera')
+    intercept = Field(np.nan, 'Intercept')
+    leakage_intensity_width_1 = \
+        Field(np.float32(np.nan), 'Fraction of intensity in outermost pixels',
```
rlopezcoto, line 62: why do you use here `np.float32(np.nan)` and for `leakage_pixels_width_1` only `np.nan`?

lstchain/io/lstcontainers.py
```diff
@@ -51,23 +56,34 @@ class DL1ParametersContainer(Container):
...
+    leakage_pixels_width_2 = Field(np.nan, 'Fraction of signal pixels that are '
+                                           'in the two outermost rings of pixels')
+    n_pixels = Field(0, 'Number of pixels after cleaning')
+    concentration_cog = Field(np.nan, 'Fraction of intensity in three pixels '
+                                      'closest to the cog')
+    concentration_core = Field(np.nan, 'Fraction of intensity inside hillas '
+                                       'ellipse')
+    concentration_pixel = Field(np.nan, 'Fraction of intensity in brightest '
+                                        'pixel')
+    n_islands = Field(0, 'Number of Islands')
```
rlopezcoto, line 78: for events that do not pass analysis cuts, this may actually be 0, right? I think it would be better if we set it to -1

lstchain/io/lstcontainers.py
```diff
@@ -51,23 +56,34 @@ class DL1ParametersContainer(Container):
...
+    alt_tel = Field(None, 'Telescope altitude pointing',
```
rlopezcoto, line 79: why `alt_tel` and `az_tel` are still set to `None`? where there problems with the interpolation?

lstchain/tests/test_lstchain.py
```diff
@@ -28,7 +28,7 @@
         "length": [0, 10],
         "wl": [0, 1],
         "r": [0, 1],
-        "leakage2_intensity": [0, 1]
+        "leakage_intensity_width_2": [0, 1],
```
rlopezcoto, line 31: how was this even working?

lstchain/io/lstcontainers.py
MaxNoe: These should be `np.nan * u.deg` to have correct units.

lstchain/io/lstcontainers.py
```diff
@@ -31,18 +32,22 @@ class DL1ParametersContainer(Container):
...
+                unit=u.rad)
+    length = Field(u.Quantity(np.nan, u.deg), 'RMS spread along the major-axis',
+                   unit=u.deg)
+    width = Field(u.Quantity(np.nan, u.deg), 'RMS spread along the minor-axis',
+                             unit=u.deg)
+    psi = Field(Angle(np.nan, u.rad), 'rotation angle of ellipse', unit=u.rad)
+
+    skewness = Field(np.nan, 'measure of the asymmetry')
+    kurtosis = Field(np.nan, 'measure of the tailedness')
     disp_norm = Field(None, 'disp_norm [m]', unit=u.m)
```
MaxNoe, line 51: disp values should also be `np.nan * u.m`

lstchain/io/lstcontainers.py
```diff
@@ -51,23 +56,34 @@ class DL1ParametersContainer(Container):
     disp_miss = Field(None, 'disp_miss [m]', unit=u.m)
     src_x = Field(None, 'source x coordinate in camera frame', unit=u.m)
```
MaxNoe, line 57: `np.nan * u.m`

lstchain/io/lstcontainers.py
```diff
@@ -51,23 +56,34 @@ class DL1ParametersContainer(Container):
...
     src_y = Field(None, 'source y coordinate in camera frame', unit=u.m)
```
MaxNoe, line 58: `np.nan * u.m`

lstchain/io/lstcontainers.py
```diff
@@ -51,23 +56,34 @@ class DL1ParametersContainer(Container):
...
-    concentration_core = Field(None, 'Fraction of intensity inside hillas ellipse')
-    concentration_pixel = Field(None, 'Fraction of intensity in brightest pixel')
-    n_islands = Field(None, 'Number of Islands')
-    alt_tel = Field(None, 'Telescope altitude pointing', unit=u.rad)
-    az_tel = Field(None, 'Telescope azimuth pointing', unit=u.rad)
-
-    obs_id = Field(None, 'Observation ID')
-    event_id = Field(None, 'Event ID')
-    calibration_id = Field(None, 'ID of the employed calibration event')
+    time_gradient = Field(np.nan, 'Time gradient in the camera')
```
MaxNoe, line 59: time gradient has unit `np.nan / u.m`

lstchain/io/lstcontainers.py
```diff
@@ -51,23 +56,34 @@ class DL1ParametersContainer(Container):
...
+    n_islands = Field(0, 'Number of Islands')
+    alt_tel = Field(None, 'Telescope altitude pointing',
+                    unit=u.rad)
+    az_tel = Field(None, 'Telescope azimuth pointing',
+                   unit=u.rad)
+
+    obs_id = Field(-1, 'Observation ID')
+    event_id = Field(-1, 'Event ID')
+    calibration_id = Field(-1, 'ID of the employed calibration event')
     dragon_time = Field(None, 'Dragon time event trigger')
```
MaxNoe, line 87: should all be `np.nan` and not nan

lstchain/io/lstcontainers.py
```diff
@@ -79,17 +95,19 @@ class DL1ParametersContainer(Container):
     mc_core_x = Field(None, 'Simulated impact point x position', unit=u.m)
```
MaxNoe, line 95: `np.nan * u.m`

lstchain/io/lstcontainers.py
```diff
@@ -79,17 +95,19 @@ class DL1ParametersContainer(Container):
...
     mc_core_y = Field(None, 'Simulated impact point y position', unit=u.m)
```
MaxNoe, line 96: `np.nan * u.m`

lstchain/io/lstcontainers.py
```diff
@@ -79,17 +95,19 @@ class DL1ParametersContainer(Container):
...
     mc_h_first_int = Field(None, 'Simulated first interaction height', unit=u.m)
```
MaxNoe, line 97: `np.nan * u.m`

lstchain/io/lstcontainers.py
```diff
@@ -79,17 +95,19 @@ class DL1ParametersContainer(Container):
...
-    mc_type = Field(None, 'Simulated particle type')
+    mc_type = Field(-1, 'Simulated particle type')
     mc_az_tel = Field(None, 'Telescope MC azimuth pointing', unit=u.rad)
     mc_alt_tel = Field(None, 'Telescope MC altitude pointing', unit=u.rad)
     mc_x_max = Field(None, "MC Xmax value", unit=u.g / (u.cm ** 2))
```
MaxNoe, line 101: `np.nan * u.g /u.cm**2`

lstchain/io/lstcontainers.py
```diff
@@ -79,17 +95,19 @@ class DL1ParametersContainer(Container):
...
-    mc_core_distance = Field(None, "Distance from the impact point to the telescope", unit=u.m)
+    mc_core_distance = \
```
MaxNoe, line 102: Please don't use line continuations, rely on the parentheses.

MaxNoe:
```
mc_core_distance = Field(
    u.Quantity(np.nan, u.m), 
    "Distance from the impact point to the telescope",
    unit=u.m,
)
```

lstchain/io/lstcontainers.py
```diff
@@ -79,17 +95,19 @@ class DL1ParametersContainer(Container):
...
+        Field(u.Quantity(np.nan, u.m), "Distance from the impact point to the "
+                                       "telescope", unit=u.m)
     mc_shower_primary_id = Field(None, "MC shower primary ID 0 (gamma), 1(e-),"
                                     "2(mu-), 100*A+Z for nucleons and nuclei,"
                                     "negative for antimatter.")
 
     hadroness = Field(None, "Hadroness")
-    wl = Field(None, "width/length")
+    wl = Field(u.Quantity(np.nan), "width/length")
```
MaxNoe, line 110: just `np.nan` would be fine, but this shouldn't harm

lstchain/io/lstcontainers.py
```diff
@@ -79,17 +95,19 @@ class DL1ParametersContainer(Container):
...
 
     tel_id = Field(None, "Telescope Id")
     tel_pos_x = Field(None, "Telescope x position in the ground")
```
MaxNoe, line 113: `np.nan * u.m`

lstchain/io/lstcontainers.py
moralejo: > why do you use here `np.float32(np.nan)` and for `leakage_pixels_width_1` only `np.nan`?

I checked empirically the types of the container as filled when the image survives cleaning. The default np.nan is float64, and most parameters are like that, but for some reason (likely reducing data size) these two are float32.

lstchain/io/lstcontainers.py
moralejo: An event that has 0 pixels after cleaning has, well, 0 pixels and 0 islands.

lstchain/tests/test_lstchain.py
moralejo: > how was this even working?

It was being ignored

lstchain/io/lstcontainers.py
moralejo: Parameters that do not exist for MC must be set to None for things to work for MC. Probably there is some other solution, anyone can look for it after we have made the release and reprocessed the data.

https://github.com/cta-observatory/cta-lstchain/pull/523/commits/8a12081afd90c95e053109800d353d86a6d666e4

lstchain/io/lstcontainers.py
moralejo: Fine, but values which are not calculated in the dl1 step are better left as None and then they do not go to the output. Here I want to do minimal changes for this to work (because there are always unwanted side effects of changes). As you know this DL1 container is anyway deprecated.

lstchain/io/lstcontainers.py
moralejo: Same as above. Trying to make minimal changes...

lstchain/io/lstcontainers.py
moralejo: Are we sure this won't break anything down the line? 
I think this is an "Acceptable Deferred Defect". Will be solved when we move to standard DL1.

lstchain/io/lstcontainers.py
moralejo: You mean "nan not None", right?  Same as above,  these real-data specific things make trouble when processing the MC, if initialized to nan. Surely solvable, but better not to create the problem in the first place.

lstchain/io/lstcontainers.py
moralejo: Same as above, I just made the needed changes so that, when a pedestal event is processed, the container has the same contents as when a normal shower image is processed. This was not needed, and again I see no need to bring in potential trouble just before a release.

lstchain/io/lstcontainers.py
moralejo: ditto

lstchain/io/lstcontainers.py
moralejo: ok__

lstchain/io/lstcontainers.py
moralejo: ok

lstchain/io/lstcontainers.py
moralejo: Same as above, trying to do minimal changes.

moralejo:
Apologies for the somewhat expeditive replies, but I do not think it makes sense to really solve everything with DL1ParametersContainer. All we need now is that it works.

morcuended:
I tested the r0_to_dl1 step on the night of 2020_08_17, and now all the DL1 files are produced without noisy problems. However, the step dl1_to_dl2 failed for all the subruns with this error:

```
  File "/home/daniel.morcuende/.local/miniconda3/envs/osa-test/bin/lstchain_dl1_to_dl2", line 33, in <module>
    sys.exit(load_entry_point('lstchain', 'console_scripts', 'lstchain_dl1_to_dl2')())
  File "/fefs/home/daniel.morcuende/software/myfork/cta-lstchain/lstchain/scripts/lstchain_dl1_to_dl2.py", line 110, in main
    dl2 = dl1_to_dl2.apply_models(data, cls_gh, reg_energy, reg_disp_vector, custom_config=config)
  File "/fefs/home/daniel.morcuende/software/myfork/cta-lstchain/lstchain/reco/dl1_to_dl2.py", line 390, in apply_models
    dl2['log_reco_energy'] = reg_energy.predict(dl2[regression_features])
  File "/home/daniel.morcuende/.local/miniconda3/envs/osa-test/lib/python3.7/site-packages/sklearn/ensemble/_forest.py", line 783, in predict
    X = self._validate_X_predict(X)
  File "/home/daniel.morcuende/.local/miniconda3/envs/osa-test/lib/python3.7/site-packages/sklearn/ensemble/_forest.py", line 421, in _validate_X_predict
    return self.estimators_[0]._validate_X_predict(X, check_input=True)
  File "/home/daniel.morcuende/.local/miniconda3/envs/osa-test/lib/python3.7/site-packages/sklearn/tree/_classes.py", line 388, in _validate_X_predict
    X = check_array(X, dtype=DTYPE, accept_sparse="csr")
  File "/home/daniel.morcuende/.local/miniconda3/envs/osa-test/lib/python3.7/site-packages/sklearn/utils/validation.py", line 72, in inner_f
    return f(**kwargs)
  File "/home/daniel.morcuende/.local/miniconda3/envs/osa-test/lib/python3.7/site-packages/sklearn/utils/validation.py", line 653, in check_array
    context))
ValueError: Found array with 0 sample(s) (shape=(0, 14)) while a minimum of 1 is required.
```
I'm trying to see if there is an error with my lstchain installation or there is something wrong with the DL1 files themselves. In principle, I think it happens also when using the `cta` environment.

MaxNoe:
This means no events now survived? So somewhere a cut is applied that gets rid of all events?

morcuended:
> This means no events now survived? So somewhere a cut is applied that gets rid of all events?

Is there a way of knowing this from the DL1 files? In the dl1 parameters table, I can see all the events (53000/subrun) and everything seems normal. I've used the standard config file with usual tailcut cleaning. Checking again

MaxNoe:
Let me check the dl1_to_dl2 step.

MaxNoe:
I think the reason is that the `filter_events` function drop an event if **any** column has a `nan`, not just the columns needed for the reconstruction.

MaxNoe:
Here is the `check_valid` function, which also logs any colums that do contain nans, which might be helpful for debugging this issue:

https://github.com/fact-project/aict-tools/blob/5619074aef9a26063ca41b3c028664a0b3b8a912/aict_tools/preprocessing.py#L24

moralejo:
> I think the reason is that the `filter_events` function drops an event if **any** column has a `nan`, not just the columns needed for the reconstruction.
> 
> So you should only check the columns needed for the application of the models for `nan` values, not all columns.
> 
Yes, I had this issue yesterday, it showed up in the tests of the RF building, and this is what moved me to put back some of the nans in the initialization to None's - but there seems to be others which still make trouble.

moralejo:
@MaxNoe, what about using dropna=False in the call to filter_events? This would keep also nans in what is passed to scikitlearn, will that lead to a fatal error?

MaxNoe:
> , will that lead to a fatal error?

Yes, scikit learn will raise an Error if any element of `X` is nan

MaxNoe:
Sorry, I did not realize that filter_events is not called on data. Let me check again. I think I brought you to the wrong track (or another issue, but not the one at hand)

vuillaut:
> Sorry, I did not realize that filter_events is not called on data. Let me check again. I think I brought you to the wrong track (or another issue, but not the one at hand)

yes it is, in `lstchain_dl1_to_dl2.py`

MaxNoe:
Yes, sorry, it's getting late, I should just stop. Was about to delete my comment.

https://github.com/cta-observatory/cta-lstchain/blob/faa75c73f02af86b3f4a1c183596872cad0e30cb/lstchain/scripts/lstchain_dl1_to_dl2.py#L97

moralejo:
The filter issue is being addressed in #526

rlopezcoto (approved):

PR #525, opened 2020-09-29
Filter events dl1 taking into account regression features 

morcuended:
Trying to solve the issue in the `filter_events` function of the `dl1_to_dl2` step (see #523) when there are columns with `NaN` and not used for regression.

This will take into account only the columns used for event reconstruction (given in the lstchain config file), but it will discard the other columns, so I guess it's not the final solution. Any suggestions or comments are welcome.

The problem is that those columns filled with `NaN` (eg. `mc_core_distance`) make the current function disregard all the rows containing `NaN`

```
         intensity  log_intensity         x         y         r       phi    length     width       psi  ...  mc_core_distance        wl  tel_id  tel_pos_x  tel_pos_y  tel_pos_z  trigger_type  ucts_trigger_type  trigger_time
0      4977.034420       3.696971 -0.162801  0.042285  0.168203  2.887475  1.125409  1.026033  1.046278  ...               NaN  0.911698       1       50.0       50.0       16.0             1                 -1  1.597728e+09
1        19.918596       1.299259  0.865152 -0.116288  0.872932 -0.133613  0.063859  0.055169 -0.345889  ...               NaN  0.863921       1       50.0       50.0       16.0             1                  1  1.597728e+09
2        97.015544       1.986841  0.091031  0.812277  0.817362  1.459193  0.096502  0.066063  1.209844  ...               NaN  0.684579       1       50.0       50.0       16.0             1                  1  1.597728e+09
```

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -308,10 +308,11 @@ def build_models(filegammas, fileprotons,
         df_gamma = pd.concat([df_gamma, pd.read_hdf(filegammas, key=dl1_params_src_dep_lstcam_key)], axis=1)
         df_proton = pd.concat([df_proton, pd.read_hdf(fileprotons, key=dl1_params_src_dep_lstcam_key)], axis=1)
 
-    df_gamma = utils.filter_events(df_gamma, filters=events_filters)
-    df_proton = utils.filter_events(df_proton, filters=events_filters)
-
     regression_features = config['regression_features']
+
+    df_gamma = utils.filter_events(df_gamma, regression_features, filters=events_filters)
+    df_proton = utils.filter_events(df_proton, regression_features, filters=events_filters)
+
 
     #Train regressors for energy and disp_norm reconstruction, only with gammas
 
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -313,7 +313,6 @@ def build_models(filegammas, fileprotons,
     df_gamma = utils.filter_events(df_gamma, regression_features, filters=events_filters)
     df_proton = utils.filter_events(df_proton, regression_features, filters=events_filters)
 
-
     #Train regressors for energy and disp_norm reconstruction, only with gammas
 
     reg_energy = train_energy(df_gamma, custom_config=config)
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -46,8 +46,6 @@ __all__ = [
     'unix_tai_to_time',
 ]
 
-log = logging.getLogger(__name__)
-
 # position of the LST1
 location = EarthLocation.from_geodetic(-17.89139 * u.deg, 28.76139 * u.deg, 2184 * u.m)
 obstime = Time('2018-11-01T02:00')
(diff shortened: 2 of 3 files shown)


lstchain/reco/dl1_to_dl2.py
```diff
@@ -308,11 +308,11 @@ def build_models(filegammas, fileprotons,
         df_gamma = pd.concat([df_gamma, pd.read_hdf(filegammas, key=dl1_params_src_dep_lstcam_key)], axis=1)
         df_proton = pd.concat([df_proton, pd.read_hdf(fileprotons, key=dl1_params_src_dep_lstcam_key)], axis=1)
 
-    df_gamma = utils.filter_events(df_gamma, filters=events_filters)
-    df_proton = utils.filter_events(df_proton, filters=events_filters)
-
     regression_features = config['regression_features']
 
+    df_gamma = utils.filter_events(df_gamma, regression_features, filters=events_filters)
```
vuillaut, line 313: rather than modifying `filter_events` that could be used in another context, you may simply:
```
regression_features = config['regression_features']
classification_features = config['classification_features']
all_features = list(set(regression_features + classification_features))

df_gamma = utils.filter_events(df_gamma[all_features], filters=events_filters)
df_proton = utils.filter_events(df_proton[all_features], filters=events_filters)
```

vuillaut (changes requested):

vuillaut (changes requested):
Also, now that I look at the code, it seems that we are not filtering events when reconstructing the events!

Could you please adopt the same strategy in `apply_models` ?!

vuillaut:
> Also, now that I look at the code, it seems that we are not filtering events when reconstructing the events!
> 
> Could you please adopt the same strategy in `apply_models` ?!

It is currently done in the pipeline in the `dl1_to_dl2` script.
Since we pass the config to `apply_models`, I think it would be better to do it there.

lstchain/reco/dl1_to_dl2.py
morcuended: Yeah, much better. This will only leave those `all_features` parameters though. Is that a problem for the reconstruction or only when filtering the events for applying the models as you mention below?

morcuended: although we cannot use `classification_features` for dl1 data which include `['log_reco_energy', 'reco_disp_dx', 'reco_disp_dy']`

vuillaut: > Yeah, much better. This will only leave those `all_features` parameters though. Is that a problem for the reconstruction or only when filtering the events for applying the models as you mention below?

for training, in `build_models` it's fine since we load these events from files and don't use them any further

vuillaut: > although we cannot use `classification_features` for dl1 data which include `['log_reco_energy', 'reco_disp_dx', 'reco_disp_dy']`

ha yes you are right, so instead of 
`all_features = list(set(regression_features + classification_features))`
`all_features = [f for f in set(regression_features + classification_features) if f in df_gamma.columns and f in df_proton.columns]`

moralejo: I think what we need is that filter_events returns all columns for rows in which the filters are fulfilled, and in which none of the regression features is nan. We could simply add that condition as additional filters, i.e. "xxx is not nan" (with xxx being each of the regression features),  and implement that in filter_events instead of the "dropna" call.

morcuended:
Rebased in #526 therefore I close this PR.

PR #526, opened 2020-09-29
Filter events not droping nan on all columns anymore

vuillaut:
@morcuended I wanted to make an update on your PR #525 , made one in your branch, then finally rebased from yours considering @moralejo comments, hence the history of commits here.
Finally, your first idea was in the right direction.

### DIFF
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -421,13 +421,13 @@ def expand_tel_list(tel_list, max_tels):
 
 def filter_events(events,
                   filters=dict(intensity=[0, np.inf],
-                                 width=[0, np.inf],
-                                 length=[0, np.inf],
-                                 wl=[0, np.inf],
-                                 r=[0, np.inf],
-                                 leakage_intensity_width_2=[0, 1],
-                                 ),
-                  dropna=True,
+                               width=[0, np.inf],
+                               length=[0, np.inf],
+                               wl=[0, np.inf],
+                               r=[0, np.inf],
+                               leakage_intensity_width_2=[0, 1],
+                               ),
+                  finite_params=None,
                   ):
     """
     Apply data filtering to a pandas dataframe.
@@ -439,8 +439,8 @@ def filter_events(events,
     ----------
     events: `pandas.DataFrame`
     filters: dict containing events features names and their filtering range
-    dropna: bool
-        if True (default), `dropna()` is applied to the dataframe.
+    finite_params: optional, None or list of strings
+        extra filter to ensure finite parameters
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


moralejo (dismissed):
Looks good to me, thanks!

moralejo:
> Looks good to me, thanks!

Ouch, other than the failing CI test of course!

lstchain/reco/utils.py
```diff
@@ -452,12 +452,13 @@ def filter_events(events,
     for k in filters.keys():
         if k in events.columns:
             filter = filter & (events[k] >= filters[k][0]) & (events[k] <= filters[k][1])
+    if finite_params is not None:
+        for k in finite_params:
+            if k in events.columns:
+                filter = filter & np.isfinite(events[k])
```
MaxNoe, line 458: could be `&=` more consice and faster

lstchain/reco/utils.py
```diff
@@ -452,12 +452,13 @@ def filter_events(events,
     for k in filters.keys():
         if k in events.columns:
             filter = filter & (events[k] >= filters[k][0]) & (events[k] <= filters[k][1])
```
MaxNoe, line 454: `&=`

moralejo:
> > Looks good to me, thanks!
> 
> Ouch, other than the failing CI test of course!

Seems that regression_features is simply missing in one of the unit tests

lstchain/reco/utils.py
MaxNoe, line 458: You can do this without the loop. Also it might be helpful to log which columns contain nans and how many.

See:
https://github.com/fact-project/aict-tools/blob/5619074aef9a26063ca41b3c028664a0b3b8a912/aict_tools/preprocessing.py#L24-L35

MaxNoe: If you also want to filter out `infs` (we deal differently with those, see https://github.com/fact-project/aict-tools/blob/5619074aef9a26063ca41b3c028664a0b3b8a912/aict_tools/preprocessing.py#L14), do:

```
with pd.option_context('mode.use_inf_as_null', True):
    nans = df.isnull()
    filter &= ~nans.any(axis=1)
```

vuillaut: done, thanks for the suggestions

lstchain/reco/utils.py
```diff
@@ -451,13 +454,23 @@ def filter_events(events,
 
     for k in filters.keys():
         if k in events.columns:
-            filter = filter & (events[k] >= filters[k][0]) & (events[k] <= filters[k][1])
+            filter &= (events[k] >= filters[k][0]) & (events[k] <= filters[k][1])
 
-    if dropna:
+    _finite_params = list(set(finite_params + events.columns))
```
MaxNoe, line 459: I think you got a logic error here. You still compute `events.isnull` on all colums. and there is no `any(axis=1)` her.

Also `_finit_params` now contains all columns in events, right? I think the correct solutino would just be:

```
nans = events[finite_params].isnull()
filter &= ~nans.any(axis=1)  # not any, so all have to be finite
```
```

vuillaut: damn, I tried to be too quick, thanks.

moralejo:
Good! Would be great to merge this so that @morcuended  can test the full chain @ La Palma
 @MaxNoe, can you approve?   It "looks fine" _to me_ now, but so did it before.

MaxNoe (approved):

PR #528, opened 2020-10-01
first implementation of time-constrained image cleaning

satoshifukami0115:
Hi, I would like to share my current code of the time-constrained image cleaning.
The concept is explained [here](https://indico.cta-observatory.org/event/2995/contributions/25463/attachments/17975/24219/LSTmono_20200907_Fukami.pdf)

Currently, I put the main code under 'lstchain/image'.
I just checked the codes are working with the current v0.6.0 lstchain (and with ctapipe 0.8.0).

If you have any comments or suggestions, I would be very happy to get those.
FYI, I'm now introducing a new method referring to the one implemented in magic-cta-pipe.


MaxNoe:
Just to mention, that in ctapipe 0.8, there is a cleaning method taking time into account here: 
https://github.com/cta-observatory/ctapipe/blob/e99e39c43c5ded14ca962a0265d0be04e0d6a88f/ctapipe/image/cleaning.py#L246

This cleaning works with looking at time differences between neighboring pixels, so it should not have the problematic behaviour you experienced with your method concerning showers with a long time evolution (large energy, high impact)

satoshifukami0115:
@MaxNoe Thank you for your comment!

It seems the function 'fact~()' is a bit different from the one introduced this time (MAGIC-like).
In 'fact~()', the tailcut cleaning is first applied and then the timing cut comes later,
while in this one, the timing cut comes in each step of deciding core & boundary pixels.

I cannot imagine which one is better (of course it depends on the parameter values).
At least I agree that the high-impact problem might not happen in 'fact~()' as you said.

> Also, let me again state that methods as general as this should be added directly to ctapipe, not to lstchain.
OK, I would like to open pull request about this in ctapipe, if it is fine also for others.

satoshifukami0115:
Sorry again,
I just looked at [Powel's PR](https://github.com/cta-observatory/cta-lstchain/pull/436) and the conclusion is that it should be implemented first in lstchain, then later in ctapipe. 
Is it also the case in this PR?

MaxNoe:
@satoshifukami0115 Yes, I fully agree, the two algorithms are very different. I just wanted to note that you maybe could include it in the comparison, as both are based on time and intensity.

> Sorry again,
I just looked at Powel's PR and the conclusion is that it should be implemented first in lstchain, then later in ctapipe.
Is it also the case in this PR?

This is a question for other's not me. I would put such a general, not urgently needed algorithm for a cleaning study in ctapipe.

moralejo:
I agree with @maxnoe that this belongs more into ctapipe. For tests you can put anywhere, but I'd rather open a PR in ctapipe.

About the algorithm: the constraint of a time difference to the core of the image is dangerous for large-impact parameter events. In LST we expect way more of those at any given energy than we have in MAGIC, due both larger mirror and wider FoV (especially larger FoV of _trigger_ region). And the constraint may eliminate perfectly valid signal pixels from the image. So either it should be removed, or replaced by a constraint which is applied relative to the time 'trend' of the image (time vs. position along major axis).

satoshifukami0115:
@moralejo Thank you for your comments.
I have to change the current method anyway as it spoils large-impact events.
I would implement different methods such as ones using the time difference from neighboring core pixels or from the expected values from the trend as you suggested.
Once I check the performance, I would open a PR in ctapipe.

If you have any additional comments, please let me know. I will keep you updated frequently, anyway.

rlopezcoto:
@satoshifukami0115 did you open a PR in ctapipe to implement this functionality in the end or will you continue with the implementation in lstchain?

satoshifukami0115:
Hello @rlopezcoto I did not open PR yet in ctapipe, but I already implemented it in ctapipe in my local branch and I'm testing it. (recently I'm busy with other stuff so the progress is slow...)
It does not take time to send PR to ctapipe, but if you prefer, I can modify it to implement into lstchain.
What do you think? I would follow your suggestion.

rlopezcoto:
> I can modify it to implement into lstchain.

Hi @satoshifukami0115, no, it is ok to leave it in ctapipe since it is a general method, so should be the place to go, but then we can close this PR, right?
Thanks for the implementation!

satoshifukami0115:
@rlopezcoto I will open a new PR in ctapipe, so no problem to close this PR, thank you!

Issue #379, opened 2020-04-28
Why different formats for pedestal and calibration files?


rlopezcoto:
At the moment we write pedestal `fits` files and calibration/time_calibration `hdf5` files, is there any reason to be using different formats @FrancaCassol @pawel21?

MaxNoe:
Also: why muon results in fits and image parameters in hdf5 for the same script?

rlopezcoto:
Yeah, have to work on it...
#245

pawel21:
@rlopezcoto 
No, there is no reason.
I agree, will be better to have one format of calibration file.
I can do it, I will improve code to use hdf5 file.

DirkHoffmann:
Remember that the EVB (via ACTL library) is only able to read FITS files, no HDF5. Or you generate little-endian binary, which is much safer anyway.

vuillaut:
> Remember that the EVB (via ACTL library) is only able to read FITS files, no HDF5. Or you generate little-endian binary, which is much safer anyway.

Why EVB should need to access offline reconstruction data?

pawel21:
@DirkHoffmann Good point, I forgot.

@vuillaut EVB need this file to do online baseline correction (in fits file we store offset values for each capacitor).
We should update pedestal file in EVB, at least from time to time.
@DirkHoffmann is it possible to read HDF5 file in EVB using another library?

MaxNoe:
HDF5 is a standardized file format. It shouldn't and doesn't matter which python library you use to write an hdf5 file for it to be readable by the libhdf5.

I don't know how easy it would be to add reading of hdf to the EVB, but which python library is used is not important for that.

Actually, both h5py and tables are wrappers around the libhdf5 C code.

DirkHoffmann:
On Apr 28, Paweł Gliwny wrote:

> @DirkHoffmann is it possible to read HDF5 file in EVB using another library?

No, sorry, we do not want to introduce another library dependency in the online software.

rlopezcoto:
Hi @DirkHoffmann, is there any strong reason for that? If I'm not wrong, to read FITS files you also need the dependency of `fitsio` (C++ ?) and hdf5 is becoming pretty standard for low level data in CTA.

FrancaCassol:
Hi @pawel21,
 
> I agree, will be better to have one format of calibration file.
> I can do it, I will improve code to use hdf5 file.

[...]

PR #529, opened 2020-10-01
Tool to produce a pedestal FITS file

Bultako:
This PR adds a new Tool `lstchain_create_drs4_pedestal_file`

```
$ lstchain_create_drs4_pedestal_file --help-all
Generate a pedestal FITS file

Options
-------

Arguments that take values are actually convenience aliases to full
Configurables, whose aliases are listed on the help line. For more information
on full configurables, see '--help-all'.

--deltaT
    Flag to use deltaT correction. Default is True
--input_file=<Path> (EventSource.input_url)
    Default: None
    Path to the input file containing events.
--output_file=<Path> (PedestalFITSWriter.output_file)
    Default: 'pedestal.fits'
    Path to the generated FITS pedestal file
--log-level=<Enum> (Application.log_level)
    Default: 30
    Choices: (0, 10, 20, 30, 40, 50, 'DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL')
--config=<Path> (Tool.config_file)
    Default: None
    line parameters

Class parameters
----------------

Parameters are set from command-line arguments of the form:
`--Class.trait=value`. This line is evaluated in Python, so simple expressions
are allowed, e.g.:: `--C.a='range(3)'` For setting C.a=[0,1,2].

PedestalFITSWriter options
--------------------------
--PedestalFITSWriter.config_file=<Path>
    Default: None
    line parameters
--PedestalFITSWriter.deltaT=<Bool>
    Default: True
    Flag to use deltaT correction. Default=True
--PedestalFITSWriter.log_datefmt=<Unicode>
    Default: '%Y-%m-%d %H:%M:%S'
    The date format used by logging formatters for %(asctime)s
--PedestalFITSWriter.log_format=<Unicode>
    Default: '%(levelname)s [%(name)s] (%(module)s/%(funcName)s): %(messag...
    The Logging format template
--PedestalFITSWriter.log_level=<Enum>
    Default: 30
    Choices: (0, 10, 20, 30, 40, 50, 'DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL')
--PedestalFITSWriter.output_file=<Path>
    Default: 'pedestal.fits'
    Path to the generated FITS pedestal file

EventSource options
-------------------
--EventSource.allowed_tels=<Set>
    Default: set()
    list of allowed tel_ids, others will be ignored. If None, all telescopes in
    the input stream will be included
--EventSource.input_url=<Path>
    Default: None
    Path to the input file containing events.
--EventSource.max_events=<Int>
    Default: None

LSTR0Corrections options
------------------------
--LSTR0Corrections.offset=<Int>
    Default: 400
    Define the offset of the baseline
--LSTR0Corrections.pedestal_path=<Unicode>
    Default: ''
    Path to the LST pedestal binary file
--LSTR0Corrections.r1_sample_end=<Int>
    Default: 39
    End sample for r1 waveform
--LSTR0Corrections.r1_sample_start=<Int>
    Default: 3
    Start sample for r1 waveform
--LSTR0Corrections.tel_id=<Int>
    Default: 1
    id of the telescope to calibrate

DragonPedestal options
----------------------
--DragonPedestal.r0_sample_start=<Int>
    Default: 11
    Start sample for waveform
```

### DIFF
--- a/lstchain/tools/lstchain_create_drs4_pedestal_file.py
+++ b/lstchain/tools/lstchain_create_drs4_pedestal_file.py
@@ -34,7 +34,9 @@ class PedestalFITSWriter(Tool):
 
     aliases = {
         "input": "EventSource.input_url",
+        "i": "EventSource.input_url",
         "output": "PedestalFITSWriter.output",
+        "o": "PedestalFITSWriter.output",
     }
 
     flags = {
--- a/lstchain/tools/lstchain_create_drs4_pedestal_file.py
+++ b/lstchain/tools/lstchain_create_drs4_pedestal_file.py
@@ -64,7 +64,24 @@ class PedestalFITSWriter(Tool):
     def start(self):
 
         try:
-            pass
+            if self.deltaT:
+                self.log.info("DeltaT correction active")
+                for i, event in enumerate(self.eventsource):
+                    for tel_id in event.r0.tels_with_data:
+                        self.lst_r0.time_lapse_corr(event, tel_id)
+                        self.pedestal.fill_pedestal_event(event)
+                        if i % 500 == 0:
+                            self.log.debug("i = {}, ev id = {}".format(i, event.index.event_id))
+            else:
+                self.log.info("DeltaT correction no active")
+                for i, event in enumerate(self.eventsource):
+                    self.pedestal.fill_pedestal_event(event)
+                    if i % 500 == 0:
+                        self.log.debug("i = {}, ev id = {}".format(i, event.index.event_id))
+
+            # Finalize pedestal and write to fits file
+            self.pedestal.finalize_pedestal()
+
         except Exception as e:
             self.log.error(e)
 
--- a/lstchain/tools/lstchain_create_drs4_pedestal_file.py
+++ b/lstchain/tools/lstchain_create_drs4_pedestal_file.py
@@ -79,26 +79,21 @@ class PedestalFITSWriter(Tool):
 
         if self.deltaT:
             self.log.info("DeltaT correction active")
-            self.log.info(f"Progress bar {self.progress_bar}")
-            for event in tqdm(
-                self.eventsource,
-                desc=self.eventsource.__class__.__name__,
-                total=self.eventsource.max_events,
-                unit="ev",
-                disable=not self.progress_bar,
-            ):
+        else:
+            self.log.info("DeltaT correction not active")
+
+        for event in tqdm(
+            self.eventsource,
+            desc=self.eventsource.__class__.__name__,
+            total=self.eventsource.max_events,
+            unit="ev",
+            disable=not self.progress_bar,
+        ):
+            if self.deltaT:
                 for tel_id in event.r0.tels_with_data:
                     self.lst_r0.time_lapse_corr(event, tel_id)
                     self.pedestal.fill_pedestal_event(event)
-        else:
-            self.log.info("DeltaT correction no active")
-            for event in tqdm(
-                self.eventsource,
-                desc=self.eventsource.__class__.__name__,
-                total=self.eventsource.max_events,
-                unit="ev",
-                disable=not self.progress_bar,
-            ):
+            else:
                 self.pedestal.fill_pedestal_event(event)
 
         self.pedestal.complete_pedestal()
--- a/lstchain/tools/lstchain_create_drs4_pedestal_file.py
+++ b/lstchain/tools/lstchain_create_drs4_pedestal_file.py
@@ -1,0 +1,77 @@
+"""
+Create drs4 time pedestal fits file.
+"""
+import numpy as np
+from astropy.io import fits
+
+from ctapipe.core import Provenance, traits
+from ctapipe.core import Tool
+from ctapipe.io import EventSource
+from lstchain.calib.camera.r0 import LSTR0Corrections
+from lstchain.calib.camera.drs4 import DragonPedestal
+
+
+class PedestalFITSWriter(Tool):
+
+    name = "PedestalFITSWriter"
+    description = "Generate a pedestal FITS file"
+
+    output_file = traits.Path(
+        help="Path to the generated FITS pedestal file",
+        directory_ok=False,
+        exist=False,
+    ).tag(config=True)
+
+    deltaT = traits.Bool(
+        help="Flag to use deltaT correction. Default=True", default_value=True
+    ).tag(config=True)
+
+    aliases = {
+        "input_file": "EventSource.input_url",
+        "output_file": "PedestalFITSWriter.output_file",
+    }
+
+    flags = {
+        "deltaT": ({"PedestalFITSWriter": {"deltaT": True}}, "Flag to use deltaT correction. Default is True")
+    }
+
+    classes = [EventSource, LSTR0Corrections, DragonPedestal]
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        """
+        Tool that generates a pedestal FITS file for low level calibration.
+
+        For getting help run:
+        python lstchain_create_drs4_pedestal_file.py --help
+        """
+
+        self.eventsource = None
+
+    def setup(self):
+
+        self.log.debug(f"Open file")
+        self.eventsource = EventSource.from_config(parent=self)
+
+    def start(self):
+
+        try:
+            pass
+        except Exception as e:
+            self.log.error(e)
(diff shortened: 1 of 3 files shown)


rlopezcoto:
great, thanks @Bultako!
This is one of the points still missing from #357 and should be part of subsequent lstchain releases (currently finishing adjusting parameters for v0.6.1), so let's not merge it before the next tag.
I'd suggest that @pawel21, @FrancaCassol or anybody else who has been intensively using this script have a look to check if the products are the same with the proposed Tool

rlopezcoto:
Btw, maybe we can retake the discussion from #379 and use a common data format for all calibration files?

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,110 @@
...
+        self.lst_r0 = None
+        self.tel_id = None
+        self.ev = None
+
+    def setup(self):
+
+        self.log.debug("Opening file")
+        self.eventsource = EventSource.from_config(parent=self)
+
+        seeker = EventSeeker(self.eventsource)
```
MaxNoe, line 61: For what is the event seeker needed here? I would avoid it. There is also a breaking change coming with ctapipe 0.9 released today

Bultako: You're right.
I've refactored some code and removed `EventSeeker`
As for the breaking change with ctapipe 0.9 release, do you think we could have this Tool working with both 0.8 and 0.9 releases?

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,109 @@
...
+        directory_ok=False,
+        exists=False,
+    ).tag(config=True)
+
+    deltaT = traits.Bool(
+        help="Flag to use deltaT correction. Default=True", default_value=True
+    ).tag(config=True)
+
+    aliases = {
+        "input_file": "EventSource.input_url",
```
maxnoe, line 31: in ctapipe we now settled for `--input` and `--output` and also providing the short forms `-i` and `-o`. Maybe worth it being consistent here?

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,109 @@
...
+
+        self.eventsource = None
+        self.pixel_ids = None
+        self.pedestal = None
+        self.lst_r0 = None
+
+    def setup(self):
+
+        self.log.debug("Opening file")
+        self.eventsource = EventSource.from_config(parent=self)
```
maxnoe, line 58: Also call `add_component` here, the eventsource also has config option.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,109 @@
...
+        self.lst_r0 = self.add_component(LSTR0Corrections(parent=self))
+
+        for event in self.eventsource:
```
maxnoe, line 61: Setup is thought to be a `setup` function, I don't think you should be looping through all events already in `setup`, better move to start.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,109 @@
...
+            else:
+                self.log.info("DeltaT correction no active")
+                for i, event in enumerate(self.eventsource):
+                    self.pedestal.fill_pedestal_event(event)
+                    if i % 500 == 0:
+                        self.log.debug(f"i = {i}, ev id = {event.index.event_id}")
+
+            self.pedestal.finalize_pedestal()
+
+        except Exception as e:
```
maxnoe, line 87: This is handled in `Tool` for you, no need to add this giant try except.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,109 @@
...
+                        if i % 500 == 0:
+                            self.log.debug(f"i = {i}, ev id = {event.index.event_id}")
+            else:
+                self.log.info("DeltaT correction no active")
+                for i, event in enumerate(self.eventsource):
+                    self.pedestal.fill_pedestal_event(event)
+                    if i % 500 == 0:
+                        self.log.debug(f"i = {i}, ev id = {event.index.event_id}")
+
+            self.pedestal.finalize_pedestal()
```
maxnoe, line 85: `finalize` sounds like it should go into `finish`

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,109 @@
...
+                    for tel_id in event.r0.tels_with_data:
+                        self.lst_r0.time_lapse_corr(event, tel_id)
+                        self.pedestal.fill_pedestal_event(event)
+                        if i % 500 == 0:
+                            self.log.debug(f"i = {i}, ev id = {event.index.event_id}")
+            else:
+                self.log.info("DeltaT correction no active")
+                for i, event in enumerate(self.eventsource):
+                    self.pedestal.fill_pedestal_event(event)
+                    if i % 500 == 0:
```
maxnoe, line 82: Just use `tqdm` around the for loop, it will print a nice progress bar. See e.g. the stage 1 tool here:
https://github.com/cta-observatory/ctapipe/blob/a5ad0f181a2f1565324ce6af5c18a806c9b33dd5/ctapipe/tools/stage1.py#L125

lstchain/tools/lstchain_create_drs4_pedestal_file.py
Bultako: If I replace  `EventSource.from_config(parent=self)` by `self.add_component(EventSource(parent=self))` or by `EventSource(parent=self)` I get the following error:

```
ERROR [PedestalFITSWriter] (tool/run): Caught unexpected exception: Can't instantiate abstract class EventSource with abstract methods _generator, datalevels, is_compatible, is_simulation, obs_id, subarray
  File "/opt/miniconda3/envs/lst-dev/lib/python3.7/site-packages/ctapipe/core/tool.py", line 215, in run
    self.setup()
  File "/Users/jer/git/cta-observatory/cta-lstchain/lstchain/tools/lstchain_create_drs4_pedestal_file.py", line 65, in setup
    self.eventsource = self.add_component(EventSource(parent=self))
  File "/opt/miniconda3/envs/lst-dev/lib/python3.7/site-packages/traitlets/traitlets.py", line 955, in __new__
    inst = new_meth(cls)
TypeError: Can't instantiate abstract class EventSource with abstract methods _generator, datalevels, is_compatible, is_simulation, obs_id, subarray
```

maxnoe: Ok that's probably a bug, but `add_component` does not exist in ctapipe 0.10 anymore anyways, so just leave it out here I'd say.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,122 @@
...
+            self.log.info(f"Progress bar {self.progress_bar}")
+            for event in tqdm(
+                self.eventsource,
+                desc=self.eventsource.__class__.__name__,
+                total=self.eventsource.max_events,
+                unit="ev",
+                disable=not self.progress_bar,
+            ):
+                for tel_id in event.r0.tels_with_data:
+                    self.lst_r0.time_lapse_corr(event, tel_id)
```
maxnoe, line 91: Is this line the only difference between `if self.deltaT` and `else`? If yes, use a single loop and move the if inside, much more readable.

maxnoe: And one level more, then you get rid of the duplicated call to `fill_pedestal_event` and the `else` statement

moralejo:
Hi @Bultako, sorry for long silence on this. Would be nice to move forward with it, as well as the other three PRs for creation of Tools for lstchain's stages.

I think we have to include tests for these tools, also tests with real data. For that you'll  need to move them to a branch in the main repository (not your fork) to have access to the private data.
We plan to do a release in the next ~2 weeks latest, perhaps it is possible to include these developments in it?

rlopezcoto:
Continued in #601

PR #530, opened 2020-10-02
change HG factor in the standard config file

rlopezcoto:
Some discussions after the merge of #516 and due to the check shown in https://github.com/cta-observatory/cta-lstchain/pull/516#issuecomment-702119946, we realized that the HG factor was not fully accounting for the part of the signal that is not integrated after 8 ns in cosmic events. The factor difference seems to come from the difference of integrating:
- factor due to the different integration window in cosmic events (8 ns) and laser events (12 ns)
- factor due to the different integration window in cosmic events (8 ns) and cosmic events (12 ns)

that ratio is 0.947 as reported in https://github.com/cta-observatory/cta-lstchain/pull/516#issuecomment-702119946, so the new HG factor should be: 1.142 * 1 / 0.947 = 1.206

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -103,7 +103,7 @@
   "gain_selector_config": {
     "threshold":  3500
   },
-  "charge_scale": [1.142,1.104],
+  "charge_scale": [1.206, 1.104],
   "LocalPeakWindowSum":{
      "window_shift": 4,
      "window_width":8


FrancaCassol:
Hi @rlopezcoto,

actually I would say that 5% difference in HG is due to the 12 ns window that you used for your [comparison](https://github.com/cta-observatory/cta-lstchain/pull/516#issuecomment-702119946). In fact, as explained in this [comment](https://github.com/cta-observatory/cta-lstchain/pull/516#issuecomment-701959203), the proposed correction should permit to completely correct the charge for the integration window, so if you compare the full corrected data with data integrated in 12 ns we expect a difference of roughly [1.055, 0.998]  for [HG, LG] from the pulse shape reference (see also [here](https://github.com/cta-observatory/cta-lstchain/pull/428)), which is very coherent with what you found. To confirm this, it would be interesting to see the same comparison with a 15 ns integration window

moralejo:
I think @FrancaCassol is correct, see discussion in #516.  We should close this and move on with the factors currently in the master branch.

PR #532, opened 2020-10-05
Refactoring of sensitivity calculation code: base gammaness and theta2 cut selection in gamma efficiency

misabelber:
While knowing that the sensitivity code of lstchain will eventually be obsolete and substituted by pyirf, it is convenient to have a reliable tool to calculate the sensitivity in the meanwhile. For that reason, I've been working on simplifying and correcting the sensitivity code, and implemented the cut selection in function of the gamma efficiency, instead of searching for the best sensitivity from a grid of cuts.


lstchain/mc/plot_utils.py
```diff
@@ -150,8 +151,31 @@ def plot_MAGIC_sensitivity(ax):
...
+    --------
+    ax:    `matplotlib.pyplot.axis`  
+
+    Returns
+    --------
+    ax:    `matplotlib.pyplot.axis`  
+
+    """
+
+    s = np.loadtxt('../spectra/data/sensitivity.txt', skiprows = 1)
```
maxnoe, line 173: hard coded path to data?

lstchain/spectra/data/sensitivity.txt
```diff
@@ -0,0 +1,16 @@
+#E_med [TeV] | % Crab
```
maxnoe, line 1: Why is this data file needed?

misabelber: Ah sorry, just added it to compare in the plots the mc results with some preliminary LST sensitivity with real data. I can keep it out of the PR to avoid the dependency.

maxnoe:
A general comment: I am not really sure it makes sense to put effort into making this code better / refactor it further.

Maybe it's better to join efforts with @JouvinLea and @chaimain and get pyrif working for lstchain?

lstchain/mc/sensitivity.py
```diff
@@ -272,15 +272,11 @@ def calculate_sensitivity_lima(n_on_events, n_background, alpha):
...
-
-
     n_excesses_5sigma = stat.excess_matching_significance(5)
-    for value_excesses, value_bkg in np.nditer([n_excesses_5sigma, n_background], op_flags=['readwrite']):
-        if value_excesses < 10:
-            value_excesses[...] = 10
-        if value_excesses < 0.05 * value_bkg * alpha[0]:
-            value_excesses[...]=0.05 * value_bkg * alpha[0]
-        
+    n_excesses_5sigma[n_excesses_5sigma<10] = 10
```
maxnoe, line 276: Way do you replace smaller values with 10? That artificially introduces signal that is not there.

lstchain/mc/sensitivity.py
```diff
@@ -272,15 +272,11 @@ def calculate_sensitivity_lima(n_on_events, n_background, alpha):
...
+    bkg_5percent = 0.05*n_background*alpha
+    n_excesses_5sigma[n_excesses_5sigma<bkg_5percent] = bkg_5percent[n_excesses_5sigma<bkg_5percent]
```
maxnoe, line 278: same as above. You increase the signal amount to match the requirement. That is wrong!

misabelber: Well,  in the the way we were approaching this, the explanation is that to calculate the sensitivity, we need to know the _minimum signal needed_ in each bin to reach 5 sigma significance. This minimum signal (minimum number of excesses over background) is calculated by the Li&Ma function, but for some bins these calculated excesses are too low, so the **actual** signal needed would be higher (at least 10, or at least 5% of the background). Its not that we are increasing the amount of signal, but we are increasing the amount of signal _required_ in those bins.

moralejo: I have not checked the full code, but it seems ok to me, by increasing the number you will anyway "over-fulfill" the other conditions. Then, the larger number of gammas will translate in a higher flux (=worse sensitivity).

rlopezcoto: I think we can mark this as resolved, please consider changing this in `pyirf` if it was implemented in the way you thought was right @maxnoe

lstchain/mc/sensitivity.py
rlopezcoto: same here

rlopezcoto:
@maxnoe's suggestion is actually what makes more sense, but @misabelber has been using this code to x-check the results from `pyirf` and help her to found some problems, so now that it is working and giving proper results, let's try to merge this PR to move on.

rlopezcoto:
@misabelber since this was basically working, could you please update it in a new branch so tests can be run and we can merge it? Then for the future we can concentrate on the pyirf implementation and come back to this only in case of conflicts/discrepancies between results

misabelber:
Ok, sure!

chaimain:
@misabelber Just a reminder, this PR needs to be closed, and just like your PR#605, be pushed from a new branch of the main repo. Sorry for repeating this, if you were already planning to do it.

misabelber:
Yes @chaimain, I know, I'm working on cleaning this  branch a bit because It has become a bit of a mess, and I will open a new PR  asap. Thanks!

PR #534, opened 2020-10-06
Add pyirf in setup.py

vuillaut:


setup.py
```diff
@@ -53,6 +53,7 @@ def find_scripts(script_dir, prefix):
         'numba',
         'numpy',
         'pandas',
+        'pyirf~=0.2.0',
```
maxnoe, line 56: We just released 0.3

vuillaut: Are the things introduced previously here compatible?

maxnoe: There is already pyirf stuff in the master?

vuillaut: I mean I just wanted to solve the install issue but yeah I can bump to 0.3.0 at the same time

maxnoe: Ah, yes the reading part. I had a quick glimpse and I think nothing with respect to that changed in 0.3

environment.yml
```diff
@@ -18,4 +18,4 @@ dependencies:
       - pytest-ordering
       - git+git://github.com/cta-observatory/ctapipe_io_lst@v0.5.3
       - ctaplot
-      - pyirf~=0.3.0
+      - pyirf=0.3.*
```
maxnoe, line 21: Shouldn't anaconda just pass this to pip and it works?

vuillaut: I thought so too and applied the same semantic but it was not recognized...

vuillaut:
Tests will pass when pyirf v0.3.0 is deployed on pypi.

maxnoe:
It is deployed now, sorry we hat an error in the travis config that prevented the automatic upload

maxnoe (approved):

PR #536, opened 2020-10-12
Changed wrong exception type by generic one

moralejo:
Corrected a last-minute change introduced (and not properly tested) in longterm_dl1_check.py
Otherwise the long-term DL1 data check fails in runs in which no pedestal info is present.

Also introduced (just in case) a file existence check for muon files, though this should not be needed now as files always exist even of they contain no muons.


rlopezcoto (approved):
This change will imply a bugfix (v0.6.2) is needed for the longterm data processing. I'll do it later today if nobody opposes

Issue #514, opened 2020-09-16
Error while trying to run lstchain_mc_r0_to_dl1 script


mari-taka:
I tried to run lstchain_mc_r0_to_dl1 script to prod5 MC that includes 4 LSTs and MAGIC, but I got the following error.
I'm using lstchain 0.6.0 and ctapipe 0.8.0.
```
cp01 ~ $ lstchain_mc_r0_to_dl1 --input-file /fefs/aswg/workspace/MC_common/corsika7.7_simtelarray_20200629/prod5/4LSTs_MAGIC/gamma/zenith_20deg/south_pointing/sim_telarray/off0.4deg/data/gamma_20deg_180deg_run1___cta-prod5-lapalma_4LSTs_MAGIC_desert-2158m_4LSTs_MAGIC_mono_off0.4.simtel.gz
```
```
Traceback (most recent call last):
  File "/home/mari.takahasi/Work/miniconda3/envs/lst-dev/bin/lstchain_mc_r0_to_dl1", line 8, in <module>
    sys.exit(main())
  File "/home/mari.takahasi/Work/miniconda3/envs/lst-dev/lib/python3.7/site-packages/lstchain/scripts/lstchain_mc_r0_to_dl1.py", line 69, in main
    custom_config=config,
  File "/home/mari.takahasi/Work/miniconda3/envs/lst-dev/lib/python3.7/site-packages/lstchain/reco/r0_to_dl1.py", line 301, in r0_to_dl1
    write_array_info_08(subarray, output_filename)
  File "/home/mari.takahasi/Work/miniconda3/envs/lst-dev/lib/python3.7/site-packages/astropy/utils/decorators.py", line 124, in deprecated_func
    return func(*args, **kwargs)
  File "/home/mari.takahasi/Work/miniconda3/envs/lst-dev/lib/python3.7/site-packages/lstchain/io/io.py", line 368, in write_array_info_08
    stage1._write_instrument_configuration(subarray)
  File "/home/mari.takahasi/Work/miniconda3/envs/lst-dev/lib/python3.7/site-packages/ctapipe/tools/stage1.py", line 407, in _write_instrument_configuration
    serialize_meta=serialize_meta,
  File "/home/mari.takahasi/Work/miniconda3/envs/lst-dev/lib/python3.7/site-packages/astropy/table/connect.py", line 113, in __call__
    registry.write(instance, *args, **kwargs)
[...]

PR #539, opened 2020-10-14
Fix an error while processing prod5 MC

mari-taka:
There was a bug while trying to run lstchain_mc_r0_to_dl1 script to prod5 MC. [issue #514 ](https://github.com/cta-observatory/cta-lstchain/issues/514)
I modified `write_array_info_08` function like `write_array_info` function and became to be able to process the prod5 MC.


rlopezcoto:
thanks @mari-taka
could you please add a comment in the code saying that this is a temporary solution until we move to ctapipe v0.9.1?

mari-taka:
@rlopezcoto I add a comment that says the new function is a temporary solution. What would you think?

maxnoe:
Please also open an issue when this is merged. TODOs in code are less visible then issues, especially when issues are linked to a milestone.

rlopezcoto:
hi @mari-taka the comment is ok, but the proposed solution is not passing travis continuous integration, it seems that there is some problem with the writing of the dataframes inside the file with the modified function, could you please check?

rlopezcoto:
see #540

PR #543, opened 2020-10-23
fix a bug with the update of recalculated parameters

vuillaut:
The update of parameters was done only at the end of the loop, so only the last one was updated...


rlopezcoto (dismissed):

moralejo (approved):
As the one who introduced this bug, I thoroughly approve!

vuillaut:
Sorry I did not pay attention the tests were not passing

pawel21:
Maybe, we also should rename `lstchain_dl1ab.py`, remove "mc" from name of script, because this is confusing. What do you think?

morcuended:
The name of the script needs to be changed also for the tests.

morcuended:
I noticed that log_intensity parameter was not being updated. I fixed it and I think it is ready for approval now.
image

vuillaut:
Well spotted @morcuended , thanks.
This old PR has been ready for review/approval for a long time.

moralejo:
Hi, any idea how it is possible to get, for quite a few events, larger intensity with 10-5 than with 6-3 cleaning? I think all pixels selected by the 10-5 cleaning should be part of the 6-3 one as well. This may point to some malfunctioning of the cleaning algorithm, right?

vuillaut:
> Hi, any idea how it is possible to get, for quite a few events, larger intensity with 10-5 than with 6-3 cleaning? I think all pixels selected by the 10-5 cleaning should be part of the 6-3 one as well. This may point to some malfunctioning of the cleaning algorithm, right?

Indeed, that is not possible in theory.

morcuended:
I'm having a look at it. There are some events in the newly produced 10-5 tailcut file with `nan` as intensity. Is this expected?

vuillaut:
> Hi, any idea how it is possible to get, for quite a few events, larger intensity with 10-5 than with 6-3 cleaning? I think all pixels selected by the 10-5 cleaning should be part of the 6-3 one as well. This may point to some malfunctioning of the cleaning algorithm, right?

The mystery has been solved.
the main island is defined as “greater number of pixels”, so it may happen that an island with fewer pixels in number has greater intensity.

this PR shall be reviewed, please.

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -142,19 +155,10 @@ def main():
...
-                    for key in dl1_container.keys():
-                        dl1_container[key] = \
-                            u.Quantity(0, dl1_container.fields[key].unit)
-
-                    dl1_container.width = u.Quantity(np.nan, u.m)
-                    dl1_container.length = u.Quantity(np.nan, u.m)
-                    dl1_container.wl  = u.Quantity(np.nan, u.m)
-
-            for p in parameters_to_update:
-                params[ii][p] = u.Quantity(dl1_container[p]).value
```
moralejo, line 157: Is this patch no longer needed? The output in case the same cleaning is used is identical to the one we would obtain directly from the r0_to_dl1 script?

moralejo:
> 
> The mystery has been solved.
> the main island is defined as “greater number of pixels”, so it may happen that an island with fewer pixels in number has greater intensity.

Great that you found it. It is actually a dangerous definition of main island. With the kind of cleaning we have now, pixels surviving the cleaning around a bright star may often take precedence over the true shower image. This may help explain the differences we see in e.g. reconstructed flux when we change the cleaning levels (both in data and MC). Is this definition of main island hardcoded in ctapipe or is it configurable?

> 
> this PR shall be reviewed, please.

Will do.

maxnoe:
> Is this definition of main island hardcoded in ctapipe or is it configurable?

It is "hardcoded" (i wouldn't really call it that, since it is a very small, two line function called `largest_island`. I would rather say if you want to use something else, make a new function called `brightest_island` or similar.

https://github.com/cta-observatory/ctapipe/blob/b5c90a18bd356cd1388c771287a95c8b5858b3a1/ctapipe/image/morphology.py#L113


This could go like this:



```
def brightest_island(island_labels, image):

    island_brightness = np.zeros(island_labels.max())
    mask = island_labels > 0
    np.sum.at(island_brightness, island_labels, image)
    brightest_island = np.argmax(island_brightness)

    return island_labels == brightest_island
```

maxnoe:
PR for ctapipe here: https://github.com/cta-observatory/ctapipe/pull/1587

lstchain/scripts/lstchain_dl1ab.py
moralejo: > Is this patch no longer needed? The output in case the same cleaning is used is identical to the one we would obtain directly from the r0_to_dl1 script?

Ok, I see the test is supposed to check exactly that, so if it works, no problem.

moralejo:
The dl1ab test is not passing, don't know why. Any idea? 
Other than that, this looks good to me.

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -95,6 +101,11 @@ def main():
         'n_pixels',
         'wl',
         'r',
+        'log_intensity',
+        'x',
+        'y',
+        'disp_dx',
+        'disp_dy',
```
morcuended, line 108: I think the test is not passing because these parameters are not being defined later on, `dl1_container.disp_dx = ...`, etc

rlopezcoto:
> > Is this definition of main island hardcoded in ctapipe or is it configurable?
> 
> It is "hardcoded" (i wouldn't really call it that, since it is a very small, two line function called `largest_island`. I would rather say if you want to use something else, make a new function called `brightest_island` or similar.
> 
> https://github.com/cta-observatory/ctapipe/blob/b5c90a18bd356cd1388c771287a95c8b5858b3a1/ctapipe/image/morphology.py#L113
> 
> This could go like this:
> 
> ```
> def brightest_island(island_labels, image):
> 
>     island_brightness = np.zeros(island_labels.max())
>     mask = island_labels > 0
>     np.add.at(island_brightness, island_labels, image)
>     brightest_island = np.argmax(island_brightness)
> 
>     return island_labels == brightest_island
> ```

Hi, I see a potential problem for brightest island, and it is that if we have a very noisy pixel (or even cluster) and a very large signal is derived in only one pixel, then `brightest_island` will not correspond to the island from which we would like to calculate the image parameters.

vuillaut:
> > > Is this definition of main island hardcoded in ctapipe or is it configurable?
> > 
> > 
> > It is "hardcoded" (i wouldn't really call it that, since it is a very small, two line function called `largest_island`. I would rather say if you want to use something else, make a new function called `brightest_island` or similar.
> > https://github.com/cta-observatory/ctapipe/blob/b5c90a18bd356cd1388c771287a95c8b5858b3a1/ctapipe/image/morphology.py#L113
> > This could go like this:
> > ```
> > def brightest_island(island_labels, image):
> > 
> >     island_brightness = np.zeros(island_labels.max())
> >     mask = island_labels > 0
> >     np.add.at(island_brightness, island_labels, image)
> >     brightest_island = np.argmax(island_brightness)
> > 
> >     return island_labels == brightest_island
> > ```
> 
> Hi, I see a potential problem for brightest island, and it is that if we have a very noisy pixel (or even cluster) and a very large signal is derived in only one pixel, then `brightest_island` will not correspond to the island from which we would like to calculate the image parameters.

Please, let's have the discussion about cleaning and def of main island in a dedicated issue : #584

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -142,19 +159,21 @@ def main():
...
+                disp_dx, disp_dy, disp_norm, disp_angle, disp_sign = disp(dl1_container['x'].to_value(u.m),
+                                                                          dl1_container['y'].to_value(u.m),
+                                                                          params['src_x'][ii],
+                                                                          params['src_y'][ii],
+                                                                          )
+                dl1_container['disp_dx'] = disp_dx
+                dl1_container['disp_dy'] = disp_dy
+                dl1_container['disp_norm'] = disp_norm
+                dl1_container['disp_angle'] = disp_angle
+                dl1_container['disp_sign'] = disp_sign
```
morcuended, line 173: These definitions are not indented within `if n_pixels > 0` condition. Should they be or it does not make any difference?

vuillaut: actually, I had to put them here to recompute it even when `n_pixels=0` and in this case they are equal to `nan`

morcuended: I see, right

morcuended:
Codacy is complaining now about the arguments passed to `disp_parameters` in particular `mc_az`. Besides, I think that the `mc_` tag should be removed from the name of the parameters of that function but that's for other PR.

How can we be sure that all dl1 parameters that need to be updated are indeed updated? We have been adding parameters to the list, every time you look at it you found that there was another parameter missing.

moralejo:
> > > Is this definition of main island hardcoded in ctapipe or is it configurable?
> > 
> > 
> > It is "hardcoded" (i wouldn't really call it that, since it is a very small, two line function called `largest_island`. I would rather say if you want to use something else, make a new function called `brightest_island` or similar.
> > https://github.com/cta-observatory/ctapipe/blob/b5c90a18bd356cd1388c771287a95c8b5858b3a1/ctapipe/image/morphology.py#L113
> > This could go like this:
> > ```
> > def brightest_island(island_labels, image):
> > 
> >     island_brightness = np.zeros(island_labels.max())
> >     mask = island_labels > 0
> >     np.add.at(island_brightness, island_labels, image)
> >     brightest_island = np.argmax(island_brightness)
> > 
> >     return island_labels == brightest_island
> > ```
> 
> Hi, I see a potential problem for brightest island, and it is that if we have a very noisy pixel (or even cluster) and a very large signal is derived in only one pixel, then `brightest_island` will not correspond to the island from which we would like to calculate the image parameters.

Very noisy pixels should be taken care of by the calibration, labeling them as bad pixels, hence not to be considered any further. In any case, with proper "pedestal-rms-aware" cleaning we should no longer need to use the main island - though it is good to have it for tests.

vuillaut:
> How can we be sure that all dl1 parameters that need to be updated are indeed updated? We have been adding parameters to the list, every time you look at it you found that there was another parameter missing.

Honesty I don't see a simple way.
That's what I don't like with this scrips from the beginning, to do things well, we should have had a clean "dl1ab" function and use it in the standard pipeline as well as here. But that required too many changes and with the perspective of using the stage1 I did not want to dive into that...

moralejo (approved):

moralejo:
I think we can move on with this, right?

morcuended:
> I think we can move on with this, right?

Yes, I think so.

morcuended (approved):

vuillaut:
Thanks for all the reviews!
It helped to solve several points unseen at first.

pawel21:
Hi, I tried run script `lstchain_dl1ab.py` on real data using the following command:  
`python lstchain_dl1ab.py  -f dl1_LST-1.Run03004.0000.h5 -o dl1b_LST-1.Run03004.0000_test.h5` 
and I got error: 
` File "lstchain_dl1ab.py", line 182, in <module>
    main()
  File "lstchain_dl1ab.py", line 166, in main
    params['src_x'][ii],
ValueError: no field of name src_x`

moralejo:
> Hi, I tried run script `lstchain_dl1ab.py` on real data using the following command:
> `python lstchain_dl1ab.py -f dl1_LST-1.Run03004.0000.h5 -o dl1b_LST-1.Run03004.0000_test.h5`
> and I got error:
> ` File "lstchain_dl1ab.py", line 182, in <module> main() File "lstchain_dl1ab.py", line 166, in main params['src_x'][ii], ValueError: no field of name src_x`

I think src_x, y and others are present only if  Disp has been calculated, which should not be the case for normal DL1 files (rather, are parameters for DL2). So those should be skipped if not found. Actually, in order to recalculate them in the script you would need a RF (should be trained already with the proper cleaning...). So  I think they should be skipped even if they are found.

morcuended:
Ouch! You are right @pawel21. Sorry I did not check again the scripts after the latest changes! Indeed no `src_x`, `src_y` nor any `disp` related parameter is present in our real-data DL1 files. Are they in the MC DL1 files though?

moralejo:
Should be only in DL2.

maxnoe:
@moralejo I think `src_x` and `src_y` are the true positions and what you mean is `reco_src_x/y`. So they might be needed for the RF training?

morcuended:
Parameters to be updated: 

```
HillasParametersContainer().keys()
Out[14]: dict_keys(['intensity', 'x', 'y', 'r', 'phi', 'length', 'width', 'psi', 'skewness', 'kurtosis'])
```
plus

```
['concentration_cog',
 'concentration_core',
 'concentration_pixel',
 'intercept',
 'leakage_intensity_width_1',
 'leakage_intensity_width_2',
 'leakage_pixels_width_1',
 'leakage_pixels_width_2',
 'log_intensity',
 'time_gradient',
 'n_islands',
 'n_pixels',
 'wl']
```
and not to be updated:

```
 'mc_core_distance',
 'mc_type',
 'dragon_time',
 'event_id',
 'alt_tel',
 'az_tel',
 'calibration_id',
 'obs_id',
 'tel_id',
 'tel_pos_x',
 'tel_pos_y',
 'tel_pos_z',
 'tib_time',
 'trigger_time',
 'trigger_type',
 'ucts_time',
 'ucts_trigger_type',
```
Is that right?

morcuended:
I removed those problematic parameters and did a quick check of all the parameters to be updated:

image

It is true as @maxnoe says that `src_x`, `src_y` and `disp` related parameters are defined in `DL1ParametersContainer`. However, in one of the files I have for testing those parameters are not present.

moralejo:
> @moralejo I think `src_x` and `src_y` are the true positions and what you mean is `reco_src_x/y`. So they might be needed for the RF training?

Ah, ok! Right, so they exist in MC but not on data - hence the code just needs to check if they are there, to keep them in the new DL1 file,  but go on with no error if they're not.

maxnoe:
@morcuended They should not depend on the analysis being source dependent or not. THey are also used for training of the disp models for source independent analysis.

If `disp` is the true disp, it needs to be updated, since it depends on `cog_x` and `cog_y`.

moralejo:
> I removed those problematic parameters and did a quick check of all the parameters to be updated:
> 
> image
> 

Can you also do the test using the same cleaning as in the creation of the input DL1 file?

> It is true as @maxnoe says that `src_x`, `src_y` and `disp` related parameters are defined in `DL1ParametersContainer`. However, in one of the files I have for testing those parameters are not present. They should be in the source-dependent-analysis files though, right?

I don't think disp parameters should be in DL1 files, why would they? Probably they are defined in the container just to have them in the same table as the image parameters inside the DL2 files.
Anyway, if any higher-level reco like disp is found in the DL1 file, it should be dropped, because that would be produced from the original cleaning, and won't be "consistent" with the new cleaning anyway.

morcuended:
@moralejo, here is the check applying the same cleaning with the script dl1ab
image

vuillaut:
That's the problem not having a real data file for unit tests 🙄
`src_x` and `src_y` are *not* recalculated. However, they are needed to recalculate `disp_dx = cog_x - src_x` (and `disp_dy`) which must be recomputed to retrain the RF (`cog_x` is modified by dl1ab).

Sorry I did not think for real data again, it's hard to keep track without unit tests.

vuillaut:
So the script does what it should, but a condition for real data should be included...

morcuended:
I'm on it @vuillaut

morcuended:
BTW I have a small real data file that could be used for unit testing. Shall we include it somehow?

maxnoe:
@morcuended We already have unit test data on a password protected server approved by the EB.
 See how `ctapipe_io_lst` does it.

https://github.com/cta-observatory/ctapipe_io_lst/blob/master/download_test_data.sh

moralejo:
> BTW I have a small real data DL1 file (13MB) that could be used for unit testing. Shall we include it somehow?

It is probably better to build it from the R0 test data that Max mentioned, so that the whole chain is consistent.

morcuended:
Can we use that in lstchain as well?

maxnoe:
@morcuended see this PR: https://github.com/cta-observatory/cta-lstchain/pull/586

PR #544, opened 2020-10-25
Fix codes for adding source-dep params  when 'mc_type' of -1 is included

SeiyaNozaki:
Current codes for adding source-dependent parameters assumed all of the events have the same `mc_type`. But `mc_type` is -1 for some events if those events don't pass image cleaning. Since `mc_type` of -1 is included in data, an unexpected source position is set with the current codes. This PR will fix that issue.

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -450,17 +450,20 @@ def get_source_dependent_parameters(data, config={}):
 
     src_dep_params = pd.DataFrame(index=data.index)
 
-    is_simu = 'mc_type' in data.columns
+    is_simu = ('mc_type' in data.columns) and ~(data['mc_type'] < 0).all()
     
     if is_simu:
-        if (data['mc_type'] == 0).all():
+        if (data['mc_type'][data['mc_type']>=0] == 0).all():
             data_type = 'mc_gamma'
         else:
             data_type = 'mc_proton'
     else:
         data_type = 'real_data'
     
+    print("data type: ", data_type)
+
     expected_src_pos_x_m, expected_src_pos_y_m = get_expected_source_pos(data, data_type, config)
+    print(f"expected source position: (x, y) = ({expected_src_pos_x_m[0]} m, {expected_src_pos_y_m[0]} m)")
 
     src_dep_params['expected_src_x'] = expected_src_pos_x_m
     src_dep_params['expected_src_y'] = expected_src_pos_y_m
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -450,8 +450,9 @@ def get_source_dependent_parameters(data, config={}):
 
     src_dep_params = pd.DataFrame(index=data.index)
 
-    is_simu = ('mc_type' in data.columns) and ~(data['mc_type'] < 0).all()
-    
+    is_simu = ~(data['mc_type'] < 0).all() if ('mc_type' in data.columns) else False
+
+    # -1 means real data or events that do not pass image cleaning, 0 means gamma, and >1 means hadron
     if is_simu:
         if (data['mc_type'][data['mc_type']>=0] == 0).all():
             data_type = 'mc_gamma'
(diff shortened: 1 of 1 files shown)


rlopezcoto:
I think this comes from the fact that now `mc_type` is initialized to -1:
https://github.com/cta-observatory/cta-lstchain/blob/5ce6cd41f4d4be80620f7c92fd7503530e0180a3/lstchain/io/lstcontainers.py#L98

then if we do not fill it,

lstchain/reco/dl1_to_dl2.py
```diff
@@ -450,17 +450,20 @@ def get_source_dependent_parameters(data, config={}):
...
     if is_simu:
-        if (data['mc_type'] == 0).all():
+        if (data['mc_type'][data['mc_type']>=0] == 0).all():
             data_type = 'mc_gamma'
         else:
             data_type = 'mc_proton'
     else:
         data_type = 'real_data'
     
+    print("data type: ", data_type)
```
rlopezcoto, line 463: please remove the prints

lstchain/reco/dl1_to_dl2.py
```diff
@@ -450,17 +450,20 @@ def get_source_dependent_parameters(data, config={}):
 
     src_dep_params = pd.DataFrame(index=data.index)
 
-    is_simu = 'mc_type' in data.columns
+    is_simu = ('mc_type' in data.columns) and ~(data['mc_type'] < 0).all()
     
     if is_simu:
-        if (data['mc_type'] == 0).all():
+        if (data['mc_type'][data['mc_type']>=0] == 0).all():
```
rlopezcoto, line 456: Maybe you can include a comment here because this is becoming a bit confusing:
-1 means real data or events that do not pass image cleaning
0 means gamma
>0 hadron

rlopezcoto (approved):

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: removed

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: ok, I added the comment

lstchain/reco/dl1_to_dl2.py
```diff
@@ -450,8 +450,9 @@ def get_source_dependent_parameters(data, config={}):
 
     src_dep_params = pd.DataFrame(index=data.index)
 
-    is_simu = ('mc_type' in data.columns) and ~(data['mc_type'] < 0).all()
-    
+    is_simu = ~(data['mc_type'] < 0).all() if ('mc_type' in data.columns) else False
```
maxnoe, line 453: `~(data['mc_type'] < 0).all()`  all is very confusing to me. And I don't even think it is correct. 
Shouldn't this simply be `(data['mc_type'] > 0).all()` ?

Also you don't need parenthese around the if condition.

lstchain/reco/dl1_to_dl2.py
maxnoe: Why is the information if something passed image cleaning encoded in `mc_type`?

SeiyaNozaki: Now I understood...!
If events don't pass image cleaning, `mc_type` and `mc_core_distance` is left as default values (-1). But this should be solved by #499 because filling parameters of `mc_type` and `mc_core_distance` is included in a function of `fill_mc`.
I created this PR for the recently processed MC, but seems no need for the future. 
Is it fine to implement this change for the current MC? I guess it will not affect future MC anyway. 
If not, I will withdraw this PR.

rlopezcoto: hi @SeiyaNozaki. This is actually not only affecting MC, but also real data for which `mc_type` will be left as the default -1 as well

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: As mentioned above, the current gamma MC has -1 or 0 and for real data has -1 for all events as `mc_type`. 
So something like this?
```
is_real = (data['mc_type'] == -1).all() if 'mc_type' in data.columns else True
is_simu = not is_real
```

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: Yes, exactly. This is because I'd like to implement it. What do you think?

SeiyaNozaki:
#551 will solve instead of this PR

PR #545, opened 2020-10-25
Raise an error when keys for the filtering don't exist

SeiyaNozaki:
If the column name for the filtering does not exist, the filtering is simply not applied without warning.
I myself sometimes wasted time as I didn't notice the filtering didn't work due to the change of hillas parameters' names.
This PR is to raise an error when keys for the filtering don't exist.

### DIFF
--- a/lstchain/reco/tests/test_utils.py
+++ b/lstchain/reco/tests/test_utils.py
@@ -92,14 +92,15 @@ def test_unix_tai_to_utc():
 
 def test_filter_events():
     from lstchain.reco.utils import filter_events
-    df = pd.DataFrame({'a': [1, 2, 3, ],
+    df = pd.DataFrame({'a': [1, 2, 3],
                        'b': [np.nan, 2.2, 3.2],
                        'c': [1, 2, np.inf]}
                       )
-    np.testing.assert_array_equal(filter_events(df, finite_params=['b']),
+    np.testing.assert_array_equal(filter_events(df, filters=dict(a=[0, np.inf], b=[0, np.inf], c=[0, np.inf]), finite_params=['b']),
                                   pd.DataFrame({'a': [2, 3], 'b': [2.2, 3.2], 'c': [2, np.inf]}))
-    np.testing.assert_array_equal(filter_events(df, finite_params=['b', 'c']),
+    np.testing.assert_array_equal(filter_events(df, filters=dict(a=[0, np.inf], b=[0, np.inf], c=[0, np.inf]), finite_params=['b', 'c']),
                                   pd.DataFrame({'a': [2], 'b': [2.2], 'c': [2]}))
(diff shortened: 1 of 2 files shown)


lstchain/reco/utils.py
```diff
@@ -455,6 +454,8 @@ def filter_events(events,
     for k in filters.keys():
         if k in events.columns:
             filter &= (events[k] >= filters[k][0]) & (events[k] <= filters[k][1])
+        else:
+            raise TypeError(f'{k} does not exist in the list of keys')
```
vuillaut, line 458: - That would be a `KeyError`
- do we want an error or a warning here?

vuillaut (changes requested):

lstchain/reco/utils.py
maxnoe: error, always.

nobody reads warnings in cluster job logs.

vuillaut:
And immediately Travis reveals an Error on this point 😆

lstchain/reco/utils.py
vuillaut: In this case, there is no need for an `if - else` condition anymore.

vuillaut:
> And immediately Travis reveals an Error on this point 😆

The unit test was precisely testing that the function did not fail when passing a filter that was not in the parameters.
You will have to update the unit test accordingly.

lstchain/reco/utils.py
SeiyaNozaki: ok, you're right. I removed `if-else` condition

lstchain/reco/tests/test_utils.py
```diff
@@ -92,14 +92,15 @@ def test_unix_tai_to_utc():
...
                                   pd.DataFrame({'a': [2, 3], 'b': [2.2, 3.2], 'c': [2, np.inf]}))
-    np.testing.assert_array_equal(filter_events(df, finite_params=['b', 'c']),
+    np.testing.assert_array_equal(filter_events(df, filters=dict(a=[0, np.inf], b=[0, np.inf], c=[0, np.inf]), finite_params=['b', 'c']),
                                   pd.DataFrame({'a': [2], 'b': [2.2], 'c': [2]}))
-    np.testing.assert_array_equal(filter_events(df, finite_params=['e']), df)
-    np.testing.assert_array_equal(filter_events(df, filters={'a': [0, 1]}),
+    np.testing.assert_array_equal(filter_events(df, filters=dict(a=[0, 1])),
                                   pd.DataFrame({'a': [1], 'b': [np.nan], 'c': 1}))
+    with np.testing.assert_raises(KeyError):
+        filter_events(df, filters=dict(e=[0, np.inf]))
```
vuillaut, line 106: 👍

vuillaut (approved):

PR #546, opened 2020-10-26
adding back event_id and obs_id in dl1 tables

vuillaut:
`event_id` and `obs_id` disappeared from the dl1 image table with ctapipe v0.8

add simple test to check if these parameters are present in the file after r0 to dl1

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -659,10 +659,10 @@ def r0_to_dl1(
 
                     event.r0.prefix = ''
 
-                    writer.write(table_name = f'telescope/image/{tel_name}',
-                                 containers = [event.r0, tel, extra_im])
-                    writer.write(table_name = f'telescope/parameters/{tel_name}',
-                                 containers = [dl1_container])
+                    writer.write(table_name=f'telescope/image/{tel_name}',
+                                 containers=[event.index, tel, extra_im])
+                    writer.write(table_name=f'telescope/parameters/{tel_name}',
+                                 containers=[event.index, dl1_container])
 
                     # Muon ring analysis, for real data only (MC is done starting from DL1 files)
                     if not is_simu:
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -868,7 +868,7 @@ def write_subarray_tables(writer, event, metadata=None):
     metadata: `lstchain.io.lstcontainers.MetaData`
     """
     if metadata is not None:
-        add_global_metadata(event.dl0, metadata)
+        add_global_metadata(event.index, metadata)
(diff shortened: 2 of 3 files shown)


lstchain/reco/r0_to_dl1.py
```diff
@@ -659,10 +659,10 @@ def r0_to_dl1(
...
                     event.r0.prefix = ''
 
-                    writer.write(table_name = f'telescope/image/{tel_name}',
-                                 containers = [event.r0, tel, extra_im])
-                    writer.write(table_name = f'telescope/parameters/{tel_name}',
-                                 containers = [dl1_container])
+                    writer.write(table_name=f'telescope/image/{tel_name}',
+                                 containers=[event.index, tel, extra_im])
+                    writer.write(table_name=f'telescope/parameters/{tel_name}',
+                                 containers=[event.index, dl1_container])
```
maxnoe, line 665: If you are now using the feature to write multiple containers at once, what is the point of having this large dl1 container?

Will event_id / obs_id be able to be read back in lstchain if it is still relying on everything being in that giant DL1 container?

vuillaut: There is no good reason to use a big dl1 container anymore.
I just don't want to refactor everything if we move to stage1 soon (soon is a relative concept).

maxnoe: The second point was the more important one...

vuillaut: No, it won't.
It was not the case before, it still won't be.
Here I am merely trying to keep the level of features we already had.

vuillaut: We can discuss starting the refactoring now, it's just not the purpose of this PR.

maxnoe: Oh ok, I thought the problem was that it was there before (in the files) and it was expected to be read (e.g. in dl2) but if it wasn't read before, sure.

maxnoe (approved):

lstchain/reco/r0_to_dl1.py
vuillaut: Well, it is read in the flat table of parameters as we always do with pandas and alike.

maxnoe: Ok, that was the question ;)

vuillaut: Yes sorry there was a misunderstanding from me.
So it was in the table of parameters before and it is expected to be, e.g. to associate parameters to images (you could say the tables have a one to one correspondence but I prefer not to rely on this implicit correspondence).
However, it is not read back in containers as there is nothing doing that in lstchain at the moment.

PR #547, opened 2020-10-28
Update install instructions

vuillaut:


maxnoe (changes requested):
Let's wait for the pypi build to suceed and give the current release

maxnoe (approved):

PR #548, opened 2020-10-28
fill event info even if Hillas parameters could not be computed

vuillaut:
Some information about the events were not written in case the Hillas Parameterisation failed, although we should keep this info (otherwise there is no point keeping the event).
Interestingly enough, there was a comment in the code in this sense already:
> The condition below should now be true for all events, this dl1_container.fill_event_info(event) is a relic of the previous approach in which only survivors of cleaning and parametrization were further processed.

I also make some related cleaning.

## Changes:
- reset dl1 container before to fill it
- always fill event info (event id, obs id) and missing mc info: `mc_type` and `mc_core_distance`
- move some dl1 computation in get_dl1 or in fill_mc
- fill MC info whether or not Hillas parameters are computed
- treat events independently of the computation of the dl1 parameters

I would like to do more cleaning but there was too much (to do and to review), so more small PRs will come after this one.

## To help the review

There seems to be a lot of changes but actually, after `if dl1_filled is not None:` (line 446) I just remove an indentation from all concerned lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,19 @@ are unit tested and should be working as long as the build status is passing.
 
 - You will need to install [anaconda](https://www.anaconda.com/distribution/#download-section) first. 
 
+### As user
+
+```
+LSTCHAIN_VER=0.6.3
+wget https://raw.githubusercontent.com/cta-observatory/cta-lstchain/v$LSTCHAIN_VER/environment.yml
+conda env create -n lst -f environment.yml
+conda activate lst
+pip install https://github.com/cta-observatory/cta-lstchain/archive/v$LSTCHAIN_VER.tar.gz
+rm environment.yml
+```
+
+### As developer
+
 - Create and activate the conda environment:
 ```
 git clone https://github.com/cta-observatory/cta-lstchain.git
@@ -27,7 +40,7 @@ To update the environment (e.g. when depenencies got updated), use:
 conda env update -n lst-dev -f environment.yml
 ```
 
-- Install lstchain:
+- Install lstchain in developer mode:
 
 ```
 pip install -e .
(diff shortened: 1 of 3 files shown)


vuillaut:
This probably supersedes #544

SeiyaNozaki (dismissed):
Thank you, @vuillaut !
It seems fine, and I will withdraw #544 after this PR is merged

moralejo:
> A lot of information about the events were not written in case the Hillas Parameterisation failed, although we should keep this info (otherwise there is no point keeping the event).

Which info was missing? Is it needed for re-cleaning the data, i.e. DL1a => DL1b? (we need a real-data version of that!) That was the main driver of that change.
Sorry, it is hard for me to understand the relevant changes with the 'diff', because large blocks of code changed indentation.

There seems to be a conflict preventing merging - please test this thoroughly once the conflict is solved.

vuillaut:
> > A lot of information about the events were not written in case the Hillas Parameterisation failed, although we should keep this info (otherwise there is no point keeping the event).
> 
> Which info was missing? Is it needed for re-cleaning the data, i.e. DL1a => DL1b? (we need a real-data version of that!) That was the main driver of that change.
> Sorry, it is hard for me to understand the relevant changes with the 'diff', because large blocks of code changed indentation.
> 
> There seems to be a conflict preventing merging - please test this thoroughly once the conflict is solved.

So due to some magic, there was actually only `mc_type`, `mc_core_distance`, `event_id` and `obs_id` that were not filled.
Still need them for reprocessing.

I will go through more code cleaning after this PR to move towards using stage 1, this function has become a big plate of spaghetti.

moralejo (approved):
Seems ok to me, but should be tested in LP (if @morcuended can do it) once it is merged

SeiyaNozaki:
Let me confirm about `mc_type`.
`mc_type` will be stored as `-1` for real data, is it fine?

vuillaut:
> Let me confirm about `mc_type`.
> `mc_type` will be stored as `-1` for real data, is it fine?

Yes `mc_type` is still written in the real data as `-1`.
To remove it as the moment, it should be defined as `None` by default but could potentially affect MC... 
At the moment I'd leave it as it is.

vuillaut:
> Seems ok to me, but should be tested in LP (if @morcuended can do it) once it is merged

I have tested on a real data file, seemed to work.
@morcuended do you prefer to test before or after merging?

morcuended:
it'd be easier to test after merging

PR #552, opened 2020-11-03
Add sampling time correction to lstchain

FrancaCassol:
This PR add the possibility to switch on the time sampling correction to the integrated charge using the trailet "time_sampling_correction_path" in the json configuration files. For the moment the card is set to  "null", in order to be compatible with the present data production. In future it would be useful to apply the correction to the calibraiton events for the estimation of the dc_to_pe  factors with the F-factor method, this should reduce of a factor two the scaling of the gain due to systematics. For the moment, the code permits to apply the corrections also to cosmic events but moving the "level A" calibration code to ctapipe_io_lst, this feature will be lost (and it is indeed not necessary at this level).

The most recent version of the corrections is in the resource directory (file sampling_interval_coefficient_july_2020.hdf5), but the file is still missing some meta-data that would help for book-keeping the applied corrections (date, runs, method). Yuto should provide soon a new version.


lstchain/calib/camera/calibrator.py
```diff
@@ -38,6 +39,12 @@ class LSTCameraCalibrator(CameraCalibrator):
         help='Path to drs4 time calibration file'
     ).tag(config=True)
 
+    time_sampling_correction_path = Unicode(
```
maxnoe, line 42: Use `ctapipe.core.traits.Path` for anything that is a Path. You can use it to make sure that the path actually exista and is a file:

` = Path(exists=True, directory_ok=False)`

lstchain/calib/camera/calibrator.py
```diff
@@ -100,13 +108,31 @@ def __init__(self, subarray, **kwargs):
...
             self.time_corrector = PulseTimeCorrection(
                 calib_file_path=self.time_calibration_path
             )
         else:
             raise IOError(f"Time calibration file {self.time_calibration_path} not found!")
 
+        # declare the charge sampling corrector
+        if self.time_sampling_correction_path is not None:
+            # search the file in resources if not found
+            if not os.path.exists(self.time_sampling_correction_path):
```
maxnoe, line 123: If someone specifies a file which is then not found, an error should be raised.

Silently using a default file if a user provided config value is invalid will result in hard to debug issues.

lstchain/calib/camera/time_sampling_correction.py
```diff
@@ -0,0 +1,118 @@
...
+        ----------
+        event: general event container
+        telid: id of the telescope
+
+        Return
+        ------
+        sampling_corrections: np.array (n_gains, n_pixels, n_samples) with the correction factors
+        """
+
+        n_gains = 2
```
maxnoe, line 61: This is the `nth` location where these numbers are hard coded. These should be defined as module constants somewhere in a common place. 

e.g. `calib/camera/constants.py`

lstchain/calib/camera/time_sampling_correction.py
```diff
@@ -0,0 +1,118 @@
...
+        r1_roi_start = 3
+
+        fc_all = event.lst.tel[telid].evt.first_capacitor_id
+
+        # get the first capacitor per pixel and gain
+        fc = np.zeros((n_gains, n_pixels), dtype=np.int16)
+
+        for k in range(n_modules):
+            for n in range(n_pixel_per_module):
+                fc[0][k * 7 + n] = fc_all[8 * k + n // 2]
```
maxnoe, line 78: You define `n_pixel_per_module` above but then use hard coded magic numbers here (7 and 8). 

What is `8`? I assume the number of channels on a DRS4 chip?

lstchain/calib/camera/time_sampling_correction.py
```diff
@@ -0,0 +1,118 @@
...
+        size_drs4 = 1024
+        n_pixels = 1855
+        roi = 36
+
+        # shift the first capacitor to the value used in the r1
+        r1_roi_start = 3
+
+        fc_all = event.lst.tel[telid].evt.first_capacitor_id
+
+        # get the first capacitor per pixel and gain
```
maxnoe, line 73: This should be its own function. Also, isn't this already implemented in the `Calibrator`? If that's the same, define a function and use this function in both places.

lstchain/calib/camera/time_sampling_correction.py
```diff
@@ -0,0 +1,118 @@
...
+        fc_to_last = size_drs4 - fc_drs4[:, :]
+
+        # initialize the charge correction vector to one
+        sampling_corrections = np.ones((n_gains, n_pixels, roi))
+
+        # loop over the gains and pixels
+        for gain in range(n_gains):
+            for pix in range(n_pixels):
+
+                # if I am at the end of the 1024
```
maxnoe, line 101: This problem was solved differently and I think rather  elegently in the pedestal subtraction, where the array is just extended by `roi` values where the first values are repeated at the end.

Why not use the same approach here?

lstchain/calib/camera/calibrator.py
FrancaCassol: Done

lstchain/calib/camera/calibrator.py
FrancaCassol: Taken out

lstchain/calib/camera/time_sampling_correction.py
FrancaCassol: Yes, indeed. All these constants will be soon in ctapipe_io_lst. I wuold suggest to leave things as they and there and clean everything with the coming lstchain version crappie 10 compatible

lstchain/calib/camera/time_sampling_correction.py
FrancaCassol: Changes

lstchain/calib/camera/time_sampling_correction.py
FrancaCassol: Same comment as above. I plan to update the code using the new reader functions as soon as we move to it

lstchain/calib/camera/time_sampling_correction.py
FrancaCassol: This would mean to extend also the time_sampling_coefficients array. Why not, but is this making a big difference?

lstchain/calib/camera/pedestals.py
```diff
@@ -41,6 +43,13 @@ class PedestalIntegrator(PedestalCalculator):
         help='Interval (number of std) of accepted charge standard deviation around camera median value'
     ).tag(config=True)
 
+    time_sampling_correction_path = Unicode(
```
maxnoe, line 46: Path instead of Unicode

lstchain/calib/camera/pedestals.py
```diff
@@ -71,6 +80,21 @@ def __init__(self, subarray, **kwargs):
         self.charges = None  # charge per event in sample
         self.sample_masked_pixels = None  # pixels tp be masked per event in sample
 
+        # declare the charge sampling corrector
+        if self.time_sampling_correction_path is not None:
+            # search the file in resources if not found
+            if not os.path.exists(self.time_sampling_correction_path):
```
maxnoe, line 86: I commented on this before and you said you fixed it, but this still silently ignores a user configuration.

FrancaCassol: Hi @maxnoe,
this PR is outdated, the code is (with the corrections I promised) in the newsreader_compatible branch. I am sorry, this is surely not the right way of proceeding but things went fast and somehow overlooked that the time sampling code was still not merged. I would suggest to close this PR.

rlopezcoto:
this was implemented in a different PR, closing

Issue #132, opened 2019-05-17
Keep and handle astropy units up to final results (DL3)


vuillaut:

rlopezcoto:
we need this very badly, especially now that we are already calculating sensitivities...

vuillaut:
Yeah it's been on my todo list for too long now but it's will require many changes.
As it's becoming urgent, I'll move it higher in my list of tasks.

PR #555, opened 2020-11-05
move timing and pointing calculations in a single function

vuillaut:
step by step cleaning of the r0_to_dl1 step
In preparation for stage1: the final goal is to use directly the stage1 but a step by step approach and some code cleaning will make things easier I think and help us understand better the current status.


lstchain/reco/r0_to_dl1.py
```diff
@@ -453,189 +450,22 @@ def r0_to_dl1(
...
-                        azimuth, altitude = pointings.cal_pointingposition(event_timestamps, drive_data)
-                        event.pointing.tel[telescope_id].azimuth = azimuth
-                        event.pointing.tel[telescope_id].altitude = altitude
-                        dl1_container.az_tel = azimuth
-                        dl1_container.alt_tel = altitude
-                    else:
-                        dl1_container.az_tel = u.Quantity(np.nan, u.rad)
-                        dl1_container.alt_tel = u.Quantity(np.nan, u.rad)
+                    (first_valid_ucts, first_valid_ucts_tib, previous_ucts_time_unix, previous_ucts_trigger_type,
+                     dragon_timed) = event_timing_pointing(dl1_container,
```
morcuended, line 454: typo: `dragon_time`?

morcuended (changes requested):

lstchain/reco/r0_to_dl1.py
```diff
@@ -838,3 +669,205 @@ def rescale_dl1_charge(event, scaling_factor):
 
     for tel_id, tel in event.dl1.tel.items():
         tel.image *= scaling_factor
+
+
+def event_timing_pointing(dl1_container,
```
morcuended, line 674: It might be good to have a test unit for this function, although it is quite complex and long

lstchain/reco/r0_to_dl1.py
vuillaut: good catch, thanks !

lstchain/reco/r0_to_dl1.py
vuillaut: I agree, please be my guest :D

morcuended (approved):
Looks good to me. Maybe we could refactor the function even in two parts (timing and pointing) so it'd be easier to test them

lstchain/reco/r0_to_dl1.py
morcuended: :) maybe someday (not too far) but right now I don't have much time. Since we need to change also the tai/utc transformation by using astropy, we could implement the tests there.

lstchain/reco/r0_to_dl1.py
```diff
@@ -838,3 +669,229 @@ def rescale_dl1_charge(event, scaling_factor):
...
+
+    module_id = 82  # Get counters from the central Dragon module
+
+    if math.isnan(ucts_t0_dragon) and math.isnan(dragon_counter0) \
+            and math.isnan(ucts_t0_tib) and math.isnan(tib_counter0):
+        # Dragon/TIB timestamps not based on a valid absolute reference timestamp
+
+        dragon_time = (
+                event.lst.tel[telescope_id].svc.date +
+                event.lst.tel[telescope_id].evt.pps_counter[module_id] +
```
maxnoe, line 728: I recently learned that the LST data comes in random module order and currently I think only the `waveform` is reordered by the event source to actual pixel id order.

I suspect that here you are not really looking at module 82 but at the random module that is in position 82.

@FrancaCassol, can you confirm?

FrancaCassol: I confirm @maxnoe, 
I think I already corrected this in a mail exchange, but without too much success ;-)
There are two errors here: the module id order is not always following the array ranking order (even if sometime it does) AND the central module id is  not 82 but 132.
To make things correct one should use the
central_module_rank=np.where(event.lst.tel[tel_id].svc.module_ids==132)

moralejo: Do you know whether this has any consequence for the last reprocessing? @morcuended : is the counter value provided as input from Isidro's daily check (truly) coming from the central module?

maxnoe: In theory, it should not really matter which dragon board you take as reference, right?

morcuended: I would say it does not really matter as @maxnoe says since they are pretty well aligned. Let me check it with Isidro.

morcuended: Isidro explained to me how he was choosing these modules:

There are the `expected_modules_id` and then `expected_pixels_id` arrays. In order to get the counters corresponding to the central module (#132), he looks for 132 in the `expected_modules_id` (as Franca wrote above). That value seemed to be 82 in the LST-CAM2 for last months (i.e. `expected_modules_id[82]=132`). Hence counters are extracted according to `expected_modules_id` (`counters[82, pps_index]` and `counters[82. 10MHz_index]`). This is why I put that 82 number in the code. Checking the number of pixels corresponding to that 82 indexes, one should obtain pixels 1 to 6 (the ones of the central module).

In any case, all modules are well synchronized. According to Isidro,  PPS counters are well aligned and it may happen that some of the 10MHz counters are shifted by 100 nsec in the worst case.

maxnoe: My understanding is, that the order is random, per event. It might be that for some files, due to network latency, module 132 was often or even always at position 82. But from what Franca told me, this is nothing that should be relied on.

maxnoe: The `expected_modules_id` is given with every even for a reason.

morcuended: Yes, for sure we have to use what Franca suggested. I was just trying to understand why I put that value there.

maxnoe:
This is now done in `ctapipe_io_lst`

PR #556, opened 2020-11-05
muon analysis r0 to dl1 specific function

vuillaut:
step by step cleaning of the r0_to_dl1 step
(in preparation for stage1)

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -238,10 +238,6 @@ def r0_to_dl1(
 
     cal_mc = load_calibrator_from_config(config, subarray)
 
-    # minimum number of pe in a pixel to include it
-    # in calculation of muon ring time (peak sample):
-    min_pe_for_muon_t_calc = 10.
-
     # Dictionary to store muon ring parameters
     muon_parameters = create_muon_table()
 
@@ -640,9 +636,6 @@ def r0_to_dl1(
                 dl1_container.trigger_time = event.r0.tel[telescope_id].trigger_time
                 dl1_container.trigger_type = event.r0.tel[telescope_id].trigger_type
 
-                # FIXME: no need to read telescope characteristics like foclen for every event!
-                foclen = subarray.tel[telescope_id].optics.equivalent_focal_length
-                mirror_area = u.Quantity(subarray.tel[telescope_id].optics.mirror_area, u.m ** 2)
                 dl1_container.prefix = tel.prefix
 
                 # extra info for the image table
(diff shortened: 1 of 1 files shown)


moralejo (approved):
Look ok to me, thanks for this clean-up. 
You say "in preparation of stage 1", but if we move to using directly the stage 1 tool (with all low-level calibs in ctapipe_io_lst), the whole muon analysis will have to be inside ctapipe, I think. On the other hand, that will not be the case if we just introduce a DL1 writer "à la ctapipe" in lstchain's r0_to_dl1, which is perhaps a faster path to have ctapipe-compatible LST1 DL1 files.

maxnoe:
@moralejo I have not yet come around to doing it, because it is not needed until the rest works, but see https://github.com/cta-observatory/ctapipe/issues/1353

rlopezcoto:
Hi, thanks @vuillaut 
did you check that this works and does not modify any result in the muon analysis?
Unfortunately I think there are not many tests written on this part (partially because of the lack of r0 data to test it against), which we should correct in the future, but it would be great that this is checked to avoid future problems with modified results

vuillaut:
> Hi, thanks @vuillaut
> did you check that this works and does not modify any result in the muon analysis?
> Unfortunately I think there are not many tests written on this part (partially because of the lack of r0 data to test it against), which we should correct in the future, but it would be great that this is checked to avoid future problems with modified results

Yes, it works, I get the same muon file after processing a real data file.
Couldn't agree more about unit tests.

lstchain/reco/r0_to_dl1.py
```diff
@@ -838,3 +767,101 @@ def rescale_dl1_charge(event, scaling_factor):
...
+    # in calculation of muon ring time (peak sample):
+    min_pe_for_muon_t_calc = 10.
+
+    # FIXME: no need to read telescope characteristics like foclen for every event!
+    foclen = subarray.tel[telescope_id].optics.equivalent_focal_length
+    mirror_area = subarray.tel[telescope_id].optics.mirror_area
+
+    bad_pixels = event.mon.tel[telescope_id].calibration.unusable_pixels[0]
+    # Set to 0 unreliable pixels:
+    image = event.dl1.tel[telescope_id].image*(~bad_pixels)
```
moralejo, line 797: Is there in ctapipe already some pixel interpolator? Would be interesting to test it. Should be better in case of isolated bad pixels (perhaps a bit dangerous when a whole pixel cluster is unreliable).

moralejo (approved):

lstchain/reco/r0_to_dl1.py
maxnoe: I don't think there is, but yes, I think it is necessary.

maxnoe: It should be easy enough to implement approaches like a mean of correct neighbors using the neighbor matrix.

maxnoe:
Can we update this? I still think it would be good to separate the large imperative code in r0_to_dl1

Issue #560, opened 2020-11-09
Wrong docstring in interpolate spikes


maxnoe:
The docstring of the interpolate spike functions say:

```
        Interpolate Spike A & B.
```

but in the code, only two cases of `interpolate_spike_A` are called and there is no `interpolate_spike_B` function

jsitarek:
It used to be spikes A & B in the past, but it turned out that spikes B were caused by wrong updates of last reading time, and thus the code only has spikes A now. I updated the docstrings and made a small PR

PR #561, opened 2020-11-09
fixed description of spike interpolation

jsitarek:
fixed description (related to issue #560)


pawel21 (approved):

maxnoe (changes requested):
There is a notebook in here that looks like an accidental commit

jsitarek:
Hi @maxnoe 

the notebook was intended, however due to my mistake two separate commits (explained in the commit history) ended up in one PR. 
the first is very trivial, just a correction of a docstring following your comment
the second one is this notebook and a small feature to allow to switch off part of the time correction (this is being used in the notebook) 
can we merge both of those with this PR?

maxnoe:
@jsitarek sure, I just wanted to make sure it was intentional.

lstchain/calib/camera/calibrator.py
```diff
@@ -100,7 +102,9 @@ def __init__(self, subarray, **kwargs):
             )
 
         # declare time calibrator if correction file exist
-        if os.path.exists(self.time_calibration_path):
+        if self.time_calibration_path == '':
+            self.time_corrector = None
+        elif os.path.exists(self.time_calibration_path):
```
FrancaCassol, line 107: Hi @jsitarek ,
I think we decided we want always the time calibration file, one could indeed take out the "if self.time_corrector:" in the lines below

FrancaCassol (commented):
Concerning the notebook, could you perhaps update it to work with the new reader (ctapipe_io_lst version 0.6.4) and push it in the lstchain branch "newsreader_compatible" (where we are all push the new reader compatible code)?

lstchain/calib/camera/calibrator.py
jsitarek: Hi @FrancaCassol 

for regular processing of the data yes, it is needed and should be used
I added those lines to be able to do tests removing such a correction (I did not find any other way to do it in the current code), because Abelardo was asking how the time resolution would look like without this correction. Since there are still discussions how to correct the time in waveforms that go to ACADA, and we might need to do some more tests for this in the near future I think it makes is useful to have such an option in the code

jsitarek:
> Concerning the notebook, could you perhaps update it to work with the new reader (ctapipe_io_lst version 0.6.4) and push it in the lstchain branch "newsreader_compatible" (where we are all push the new reader compatible code)?

I removed the notebook from this branch.

lstchain/calib/camera/calibrator.py
FrancaCassol: Hi @jsitarek, 
I leave @maxnoe to review this, I thinks he asked such a security in the input trailets, in order not to risk a wrong processing of real data (which indeed demand to have the time calibration file). 
In fact, if you are testing new algorithms, you are anyway touching the code and you can locally modify this trailet.

maxnoe: Yes, please don't silently fail on a missing input file. If the user give a time calibration file and it doesn't exist, an error must be raised.

You should use `None` for `no time calibration file given` and then check `if self.time_calibration_path is None`.

Also this is already implemented in the new source. Why do we need this here at all?

jsitarek: Hi @maxnoe 

stricktly speaking, the user is *not* giving a path to a file, if the path variable is set to an empty string, but I understand 

this PR started in early November last year as a fix of 1 character in a doc string following an issue reported by @maxnoe 
then it was not merged still in the end of November when by mistake I added to it a separate PR with computing time resolution and adding this possibility to switch off the DRS4 time correction.

Since both the spikes interpolation (which was the original PR) and the DRS4 time calibration (which was the second one) now will go into general ctapipe due to PR #587 I agree with @maxnoe that it is not needed anymore, so I will close the PR

sorry for a waste of time for everybody involved

maxnoe: > stricktly speaking, the user is not giving a path to a file, if the path variable is set to an empty string, but I understand

Then test for empty string and not for `exists`.

jsitarek: Sorry, I do not understand your point. This is exactly what the change was. The old code was testing if the file exist and if so, switching on the correction, otherwise raises an error. The change that I proposed first checks if the string is empty, if so, it switches off  the correction, then it proceeds with the old behaviour (i.e. raises an error if path is wrong, or uses the correction if it is ok).

maxnoe: That behaviour you are describing is correct.

PR #562, opened 2020-11-12
Add merge muon files script

rlopezcoto:
This script merges the muon `fits` files that are usually produced per subrun to make them easier to be handled


moralejo (approved):

moralejo:
But why is Travis failing?

maxnoe:
Due to a bug in the new gammapy that produces wrong sensitivity cmvalues, see discussion slack

moralejo:
> Due to a bug in the new gammapy that produces wrong sensitivity cmvalues, see discussion slack

Thx, I was unaware!

PR #563, opened 2020-11-12
Fix gammapy error due to the new version

rlopezcoto:
Fix errors due to the name changing of function `excess_matching_significance` -> `n_sig_matching_significance` in gammapy v0.18

Also, the significance definition changed and there were some modifications on the results, so tests were also modified.

### DIFF
--- a/lstchain/mc/sensitivity.py
+++ b/lstchain/mc/sensitivity.py
@@ -10,6 +10,7 @@ from lstchain.reco.utils import reco_source_position_sky
 from astropy.coordinates.angle_utilities import angular_separation
 from lstchain.io import read_simu_info_merged_hdf5
 from lstchain.io.io import dl2_params_lstcam_key
+from pyirf.sensitivity import relative_sensitivity
 
 __all__ = [
     'read_sim_par',
@@ -23,6 +24,10 @@ __all__ = [
     'sensitivity',
     ]
 
+
+def excess_matching_significance(n_on, n_off, alpha, significance):
+    n_excess = n_on - alpha *  n_off
+    return n_excess * relative_sensitivity(n_on=n_on, n_off=n_off, alpha=alpha)
 
 def read_sim_par(dl1_file):
     """
@@ -185,8 +190,8 @@ def calculate_sensitivity_lima(n_on_events, n_background, alpha, n_bins_energy, 
         )
 
 
-    n_excesses_5sigma = stat.n_sig_matching_significance(5)
-
+    n_excesses_5sigma = excess_matching_significance(5)
+    
     for i in range(0, n_bins_energy):
         for j in range(0, n_bins_gammaness):
             for k in range(0, n_bins_theta2):
@@ -233,8 +238,7 @@ def calculate_sensitivity_lima_ebin(n_on_events, n_background, alpha, n_bins_ene
         )
         
 
-
-    n_excesses_5sigma = stat.n_sig_matching_significance(5)
+    n_excesses_5sigma = excess_matching_significance(5)
 
     for i in range(0, n_bins_energy):
         # If the excess needed to get 5 sigma is less than 10,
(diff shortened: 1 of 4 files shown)


lstchain/mc/tests/test_senstivity.py
```diff
@@ -46,25 +46,25 @@ def test_calculate_sensitivity_lima():
     
     np.testing.assert_allclose(calculate_sensitivity_lima(
             50, 10, 0.2, 1, 0, 0),
-                               ([13.48, 26.97]), rtol = 1.e-3)
+                               ([7.429626, 14.859252]), rtol = 1.e-3)
```
maxnoe, line 49: Why did these values change?

vuillaut (changes requested):
maybe fix gammapy version in env and setup?

rlopezcoto:
Following @maxnoe suggestion, I replaced the buggy `excess_matching_significance` calculation using gammapy and substituted it with `pyirf`
@vuillaut I also bumped the gammapy (and pyirf) version in the environment/setup

lstchain/mc/tests/test_senstivity.py
```diff
@@ -46,25 +46,25 @@ def test_calculate_sensitivity_lima():
...
     np.testing.assert_allclose(calculate_sensitivity_lima(
             50, 10, 0.2, 1, 0, 0),
-                               ([13.48, 26.97]), rtol = 1.e-3)
+                               ([13.49, 26.97]), rtol = 1.e-3)
     np.testing.assert_allclose(calculate_sensitivity_lima(
             200, 50, 1, 0, 1, 0),
-                               ([63.00, 31.5]), rtol = 1.e-3)
+                               ([63.01, 31.50]), rtol = 1.e-3)
     # Testing an array
     np.testing.assert_allclose(calculate_sensitivity_lima(
```
maxnoe, line 54: Why were these values changed? Shouldn't they work like before now?

lstchain/mc/sensitivity.py
```diff
@@ -178,15 +182,8 @@ def calculate_sensitivity_lima(n_on_events, n_background, alpha, n_bins_energy,
...
-    stat = WStatCountsStatistic(
-        n_on=n_on_events,  
-        n_off=n_background,
-        alpha=alpha
-        )
-
-
-    n_excesses_5sigma = stat.excess_matching_significance(5)
-
+    n_excesses_5sigma = excess_matching_significance(n_on_events, n_background, alpha, 5)
```
maxnoe, line 185: Actually, you could replace this function completely by just using

```
sensitivity = relative_sensitivity(n_on, n_off, alpha)
n_excess = sensitivity * (n_on - alpha * n_off)
return n_excess, sensitivity
```

as pyirf also takes care of the condition checks. The loops are not needed (even now they are not needed and could be replaced by simple masking).

lstchain/mc/tests/test_senstivity.py
rlopezcoto: they were essentially the same within the `rtol`, but anyways I returned them to the old values now

lstchain/mc/sensitivity.py
rlopezcoto: ok, thanks for the suggestion, we can think on improvements in the future, but I think now it is important that the `gammapy` issue, which was the original reason for this PR, gets solved

lstchain/mc/tests/test_senstivity.py
maxnoe: No, i mean the values below, which changed not just at some decimal digits

rlopezcoto: I changed some of the input values of the tests because there were some `nan`, I'm guessing because of low number of `on/off` events

vuillaut (approved):

PR #564, opened 2020-11-14
WIP Remove duplicate open PDF multipage in drs4 visualization

morcuended:
I was trying to use  `plot_pedestals` from plot_drs4.py and `plot_all` from plot_calib.py (`visualization` module) in a single script to produce drs4 and calibration data-check PDF files. The first one was produced without problems, however, the second file still contained plots from the previous file overlaid. My guess is that after producing the first file the PdfPages object was not properly closed since it was open twice. 

This problem does not happen when using the drs4 and calib visualization functions in different scripts.


lstchain/visualization/plot_drs4.py
```diff
@@ -36,9 +36,21 @@ def plot_pedestals(data_file, pedestal_file, run=0 , plot_file="none", tel_id=1,
      offset_value: baseline off_set
      """
 
-    # r0 calibrator
-    r0_calib = LSTR0Corrections(pedestal_path=pedestal_file, offset=offset_value,
-                                tel_id=tel_id )
+    # plot open pdf
+    if plot_file != "none":
+        pp = PdfPages(plot_file)
```
maxnoe: This might lead to broken pdf files as it is important to properly close a `PdfPages` object. `PdfPages` should always be used with a context manager.

morcuended:
Moved to #608

PR #553, opened 2020-11-03
Source-dependent analysis codes with wobble data

SeiyaNozaki:
This PR is to add source-dependent parameters with wobble data.
Source-dependent parameters for ON events, i.e. an expected source position is pointing direction itself, is saved in `dl1/event/telescope/parameters_src_dependent/LST_LSTCam`.
One can choose `n_off_wobble` in config file, and source-dependent parameters with each rotation angle XXX is saved in `dl1/event/telescope/parameters_src_dependentXXX/LST_LSTCam`.
The source position is updated every 1000 events for the moment to save time. Computation time will be faster as discussed in #501, then I will change this condition to compute source positions every event.

To create dl2 files, dl1 standard parameters are copied in `dl2/event/telescope/parameters/LST_LSTCam`, and random forest outputs with each roration angle is saved in `dl2/event/telescope/parameters_src_dependentXXX/LST_LSTCam` without standard parameters to avoid saving duplicate data.

Any comments are welcome.

PR #566, opened 2020-11-17
Function to calculate effective time of real data

misabelber:
After discussions in the analysis call, I have updated and optimized the function to calculate the effective time for real data events. I have included it in lstchain/data .


lstchain/data/effective_time.py
```diff
@@ -0,0 +1,32 @@
...
+    
+    deltat = np.diff(events.dragon_time)
+    deltat = deltat[(deltat > 0) & (deltat < 0.002)]
+    rate=1/np.mean(deltat)
+    dead_time = np.amin(deltat)
+    t_elapsed = events.shape[0]/rate * u.s
+    total_dead_time=events.shape[0]*dead_time
+    t_eff = t_elapsed/(1+rate*dead_time)
+
+    print("ELAPSED TIME: %.2f s\n" % t_elapsed.to_value(),
```
maxnoe, line 25: No prints here please.

maxnoe: Consider returning a dict with all these quantities if they are all of interest.

misabelber: Done! I don't think they are of much interest, just had the print around for testing purposes.

morcuended:
I suggest moving this function somewhere else, for example, to a `lstchain.utils` module (?).

There are still some trailing spaces raised by codacy. Otherwise, I think this can be merged.

lstchain/data/effective_time.py
```diff
@@ -0,0 +1,24 @@
...
+    Parameters
+    ----------
+    events: pandas DataFrame
+
+    Returns
+    -------
+    teff: float
+    """
+    
+    deltat = np.diff(events.dragon_time)
```
chaimain, line 17: You can just change `events.dragon_time` to 'events["dragon_time"].value' so that you can use the next lines of codes as is.

lstchain/data/effective_time.py
```diff
@@ -0,0 +1,24 @@
...
+    deltat = deltat[(deltat > 0) & (deltat < 0.002)]
+    rate=1/np.mean(deltat)
+    dead_time = np.amin(deltat)
+    t_elapsed = events.shape[0]/rate * u.s
```
chaimain, line 21: And coming from the earlier comment, 'events.shape[0]' -> 'len(ev)'

chaimain (commented):
Can you change these so that the events can be an astropy QTable, as we need when using pyirf functions for producing DL3 files? I checked with these changes and it works nicely!

rlopezcoto:
@misabelber can you update this PR with the latest changes of lstchain and make it ctapipe v0.10.x compatible?
It would be great if we could get it merged so it can be directly used in #593.
Remember that you will have to push a branch directly to lstchain, not to your fork, otherwise tests will not pass. See here the quick fix for the PR:
https://github.com/cta-observatory/cta-lstchain/pull/553#issuecomment-782148862

misabelber:
Ok! Will do it asap.

misabelber:
Opened new PR #605

PR #569, opened 2020-11-24
Add ctapipe-extra to setup.py

vuillaut:
ctapipe-extra is actually required when we load telescope or optics info from `ctapipe_resources` such as `OpticsDescription.from_name`

It otherwise induces a bug when lstchain is installed without the conda env.


maxnoe:
Can you not fix it by removing the `from_name` calls? Where do you need it?

rlopezcoto:
> Can you not fix it by removing the from_name calls and use EventSource.subarray instead? Where do you need it?

this is also needed for the notebooks to properly work (although they are very much outdated now...). I would include it for the time being

rlopezcoto (approved):

PR #570, opened 2020-11-24
Improve Tool of camera calibration coefficients

Bultako:
This PR polishes code in the `lstchain_create_calibration_file.py` existing Tool, that generates an HDF5 file with camera calibration coefficients, mostly to be consistent with the Tools that are being developed at this moment in different PRs. It also provides a progress bar instead of the debug logs for every 100 processed events.

### DIFF
--- a/lstchain/tools/lstchain_create_calibration_file.py
+++ b/lstchain/tools/lstchain_create_calibration_file.py
@@ -39,7 +39,9 @@ class CalibrationHDF5Writer(Tool):
 
     aliases = {
             "input": "EventSource.input_url",
+            "i": "EventSource.input_url",
             "output": "CalibrationHDF5Writer.output",
+            "o": "CalibrationHDF5Writer.output",
             "max_events": "EventSource.max_events",
             "calibration_product": "CalibrationHDF5Writer.calibration_product",
             "r0calibrator_product": "CalibrationHDF5Writer.r0calibrator_product",
--- a/lstchain/tools/lstchain_create_calibration_file.py
+++ b/lstchain/tools/lstchain_create_calibration_file.py
@@ -38,10 +38,8 @@ class CalibrationHDF5Writer(Tool):
     )
 
     aliases = {
-            "input": "EventSource.input_url",
-            "i": "EventSource.input_url",
-            "output": "CalibrationHDF5Writer.output",
-            "o": "CalibrationHDF5Writer.output",
+            "input_file": "EventSource.input_url",
+            "output_file": "CalibrationHDF5Writer.output",
             "max_events": "EventSource.max_events",
             "calibration_product": "CalibrationHDF5Writer.calibration_product",
             "r0calibrator_product": "CalibrationHDF5Writer.r0calibrator_product",
--- a/lstchain/tools/lstchain_create_calibration_file.py
+++ b/lstchain/tools/lstchain_create_calibration_file.py
@@ -2,102 +2,66 @@
 Extract flat field coefficients from flasher data files.
 """
 import numpy as np
-from traitlets import Dict, List, Unicode, Float, Bool
-
-
-from ctapipe.core import Provenance, traits
-from ctapipe.io import HDF5TableWriter
-from ctapipe.core import Tool
-from ctapipe.io import EventSource
 from ctapipe.containers import PixelStatusContainer
+from ctapipe.core import Provenance, Tool, traits
+from ctapipe.io import EventSource, HDF5TableWriter
+
 from lstchain.calib.camera.calibration_calculator import CalibrationCalculator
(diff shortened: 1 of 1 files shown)


lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -2,103 +2,77 @@
...
     ).tag(config=True)
 
-    output_file = Unicode(
-        'calibration.hdf5',
-        help='Name of the output file'
-    ).tag(config=True)
-
-    log_file = Unicode(
-        'None',
-        help='Name of the log file'
```
FrancaCassol, line 50: Hi,
This log_file trailet was originated by the wish to give the name of the file where the log output should be written, unfortunately at that time I did not menage to use it successfully (we use now a unix pipe in the calibration script), is there some news on this possibility?

maxnoe: It is builtin in ctapipe 0.10

lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -2,103 +2,77 @@
...
+        CameraR0Calibrator, default_value="NullR0Calibrator"
     )
 
-    aliases = Dict(dict(
-        input_file='EventSource.input_url',
-        output_file='CalibrationHDF5Writer.output_file',
-        log_file='CalibrationHDF5Writer.log_file',
-        max_events='EventSource.max_events',
-        pedestal_file= 'LSTR0Corrections.pedestal_path',
-        calibration_product='CalibrationHDF5Writer.calibration_product',
```
FrancaCassol, line 69: Please, keep the pedestal_file alias, this is not making big damage and can be useful for fast tests

Bultako: I can easily bring back `pedestal_file` alias, but I do not see where in the code this pedestal file could be used now. I guess you mean it could be used in a potential feature that it is still not implemented. In that case, and since it will not be used if we bring the alias back, shouldn't we just add it in the PR implementing that *now missing* feature?

FrancaCassol: You are right, in this version it is no more useful. Forget it.

lstchain/tools/lstchain_create_calibration_file.py
FrancaCassol: Perfect. Let's remember it for a next version then, for the moment the trailet is indeed not necessary.

FrancaCassol (approved):

FrancaCassol:
Hi @Bultako,
it would be good to finalise this PR. 

I wonder, the change of names in the trailets is part of a global renaming to be compatible with some protocol? Because if we introduce the "i" and "o" aliases I find more clear to keep the "input_file" and "output_file" name (in order to distinguish from "input_dir" and "output_dir" for example ). In any case, this change is affecting also the lstchain/scripts/onsite/onsite_create_calibration_file.py, so if it must be done, one should change also the names in that script

Bultako:
@FrancaCassol 

You're right. 
This PR also affects `lstchain/scripts/onsite/onsite_create_calibration_file.py`. Modifs in the names of the input and output params where suggested in another PR https://github.com/cta-observatory/cta-lstchain/pull/529#discussion_r526829121, that's why I thought better we should be consistent among the tools interface. I have reverted to the old names `input_file`and `output_file` while we take a decision, so no need to do also modifs on `onsite_create_calibration_file.py` for the moment. From my side this PR is finished and ready for review.

FrancaCassol:
> This PR also affects `lstchain/scripts/onsite/onsite_create_calibration_file.py`. Modifs in the names of the input and output params where suggested in another PR [#529 (comment)](https://github.com/cta-observatory/cta-lstchain/pull/529#discussion_r526829121), that's why I thought better we should be consistent among the tools interface. 

I see. With the finalisation of the new reader we are probably going move naturally towards ctapipe standards, so this change will probably come back later in a global conversion. 
For me this PR is fine too

FrancaCassol (dismissed):

lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -2,103 +2,77 @@
...
-    ))
-
-    classes = List([EventSource,
-                    CalibrationCalculator
-                    ]
-                   + traits.classes_with_traits(CameraR0Calibrator)
-                   + traits.classes_with_traits(CalibrationCalculator)
-                   )
+    aliases = {
+            "input_file": "EventSource.input_url",
```
maxnoe, line 41: We just realized that you can use tuples here for short option. So if you want to support e.g. `-i`, you can do:

```
("i", "input_file"): "EventSource.input_url",
```

rlopezcoto:
Continued in #598

PR #572, opened 2020-12-05
Fix width and length values in data check plots

moralejo:


morcuended:
@moralejo is the notebook `notebooks/proton_mc_vs_data.ipynb` intended to be uploaded to the repository?

moralejo:
> @moralejo is the notebook `notebooks/proton_mc_vs_data.ipynb` intended to be uploaded to the repository?

yes...  it's been there for a while. Since I use it to do some comparisons I presented and discussed a couple of time I think it should be public.

moralejo:
Ah wait... you mean why in this PR?

morcuended:
Yes, I meant both, because it seemed unrelated. But it's OK with me.

moralejo:
> Ah wait... you mean why in this PR?

no good reason, I guess!

moralejo:
I prefer to leave it here, even if it is unrelated. I am afraid I could mess something up if I try to fix this.
Can you approve it?

morcuended (approved):
Just wanted to make sure about the notebook.

moralejo:
> Just wanted to make sure about the notebook.

In the end I removed it. This made your review obsolete I think.

morcuended (approved):

Issue #571, opened 2020-11-30
spike correction should be ADC subtraction instead of the interpolation


SeiyaNozaki:
Just let me report:
If predicted spikes are located at the pulse peak position, spike correction reduces the pulse height.
![pulse_with_spike](https://user-images.githubusercontent.com/31307823/100614775-00e4cd00-335a-11eb-939f-a1f1c5f48b93.png)

As @tysaito2008 explained in the reco meeting, Tomohiko @toka2018 is studying the spike feature and implementing the new spike correction (ADC subtraction) into lstchain.

moralejo:
Ok, keep in mind that this should also go into ctapipe_io_lst, so that it works for the DL1 stage tool of ctapipe which we will eventually use as the LST standard.

PR #574, opened 2020-12-15
New correction method for the DRS4 spike

TomohikoOka:
This modification is related to #571.
I have modified the correction method of the DRS4 spike from the interpolation method.
The new method of the correction is to subtract with a fixed value of ~50 ADC counts.
Not to include a fixed value in the lst-chain script, the value of the spike ADC counts can be given from an input table by analyzers. The calculation and making a spike table can be performed with lstchain_data_create_drs4_spike_table.py in the script directory of lst-chain.
Furthermore, the correction algorithm in r0.py can be done without the input table. 
In that case, the default value described in the r0.py will be used for the spike correction.

### DIFF
--- a/lstchain/calib/camera/r0.py
+++ b/lstchain/calib/camera/r0.py
@@ -85,6 +85,11 @@ class LSTR0Corrections(CameraR0Calibrator):
                             help='Path to the LST pedestal binary file'
                             ).tag(config=True)
 
+    spike_path = Unicode('',
+                            allow_none=True,
+                            help='Path to the DRS spike text file'
+                            ).tag(config=True)
+
     def __init__(self, **kwargs):
         """
         The R0 calibrator for LST data.
@@ -115,6 +120,9 @@ class LSTR0Corrections(CameraR0Calibrator):
                                               self.size4drs+40),
                                               dtype=np.int16)
 
+        self.spike_adc = np.array([47.0, 53.0])
+
+
         self.first_cap_array = np.zeros((self.n_module,
                                          self.n_gain,
                                          self.n_pix))
@@ -137,6 +145,7 @@ class LSTR0Corrections(CameraR0Calibrator):
                                              self.n_pix))
 
         self._load_calib()
+        self._load_spike()
 
     def calibrate(self, event):
         for tel_id in event.r0.tels_with_data:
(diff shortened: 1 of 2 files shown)


moralejo:
How stable is the size of the spikes?  Can you provide some info on that?

Also: is the spike table created with a DRS4 pedestal calibration file? If so, wouldn't it be simpler to incorporate this in the script calculating the DRS4 corrections?

toka2018:
How stable is the size of the spikes? Can you provide some info on that?

Also: is the spike table created with a DRS4 pedestal calibration file? If so, wouldn't it be simpler to incorporate this in the script calculating the DRS4 corrections?
https://drive.google.com/file/d/1AO2vW1AevSu49tYo209lo1DPz8f7XQmH/view?usp=sharing

toka2018:
> How stable is the size of the spikes? Can you provide some info on that?

I summarized it. Please read my slide.
https://drive.google.com/file/d/1AO2vW1AevSu49tYo209lo1DPz8f7XQmH/view?usp=sharing


> Also: is the spike table created with a DRS4 pedestal calibration file?

Yes.

> If so, wouldn't it be simpler to incorporate this in the script calculating the DRS4 corrections?

I see. 
I separated the pedestal and spike calculation code because I avoid modifying the conventional code calculating the pedestal. 
If you think that combining the two calculation codes is better than to separate, I  will do it.

maxnoe:
> I separated the pedestal and spike calculation code because I avoid modifying the conventional code calculating the pedestal.
If you think that combining the two calculation codes is better than to separate, I will do it.

Having the same calculations done in the same script does not mean the actual calculation must be intertwined.

The calculations can easily be separated (e.g. using two components) that are then used in the same script.

toka2018:
> Having the same calculations done in the same script does not mean the actual calculation must be intertwined.
> The calculations can easily be separated (e.g. using two components) that are then used in the same script.

OK. I will incorporate the spike calculation algorithm in the pedestal calculation script (not separate the two codes).

SeiyaNozaki:
btw, should we wait for the ctapipe-compatible event source in ctapipe_io_lst and directly contribute to it after the release?

jsitarek:
About the implementation together with the DRS4 pedestal table: I think it is not as simple as substituting one entry of the table with another, because the spike position depends on current and previous first capacitor value, so simply the procedure is different (otherwise we would not have those spikes in the first place :-) ), still, things can be written in the same piece of code, but with somewhat different operations. 
Tomohiko, I do not remember from your previous studies if you tested this procedure also for randomly arriving pulses (e.g. calibration pulses intealeaved into real data)

maxnoe:
@jsitarek We were not proposing to have it in the same data structure, just compute it in the same script and store the coefficients (in a different table) in the same file.

TomohikoOka:
Hi,

Sorry for the very late reply.

I combined the above two code and tested with some data. 
https://drive.google.com/file/d/1SVkbHboI0apW6QRNnQioWNGT_HvDfmyR/view?usp=sharing

moralejo:
Hi @jsitarek, can we revive this?  What do you think of this correction?
@maxnoe made me note that the %32 spikes are currently not corrected, being relatively small. Could this method be used for them as well?

jsitarek:
Hi @moralejo @TomohikoOka 

apologies for a long delay. I reviewed the related PR on the ctapipe_io_lst. 
About this one, it has some conflicts because of refactoring the code from r0.py to ctapipe_io_lst. can you @TomohikoOka can you update the code accordingly and check that it still works with the current reading scheme, afterwards I can make a quick review of the new code

about the comment about 32-period spikes, contrary to those broad "Spike A", the period-32 spikes are not predictable, so we would have to interpolate them instead of subtracting. @SeiyaNozaki made a study that they are connected with the long time lapse times, and during regular data taking (with a few kHz rate) this happens very rarely. Adding the fact that they are also pretty small, they are not really worth to be interpolated.

TomohikoOka:
Dear @jsitarek, 

Thank you for your reply!

My code for the "cta-lstchain" was (i) to calculate the ADC value of DRS4 spikes and (ii) to correct spike events.
@maxnoe and @SeiyaNozaki currently seem to incorporate them into "ctapipe-io-lst," respectively.
(i) https://github.com/cta-observatory/cta-lstchain/pull/725 (ii) https://github.com/cta-observatory/ctapipe_io_lst/pull/96

I will help them if they want.

maxnoe:
The code this is targeting was removed in lstchain and added to `ctapipe_io_lst`. There is a new PR open regarding calculation of spike heights here and application in ctapipe_io_lst respectively. So I am closing this one.

PR #575, opened 2020-12-22
DL1 reprocess script

SeiyaNozaki:
This PR adds a DL1 reprocess script, `lstchain_reprocess_dl1.py`.
In this script, image parameters are recomputed by using DL1a data and a new configuration file, and the other parameters are copied by using DL1b data.
For now, image data of all events are saved in DL1 data as DL1a. So this procedure works if analyzers want to apply different image cleaning from default.
Any comments are welcome!

### DIFF
--- a/lstchain/io/subarray.py
+++ b/lstchain/io/subarray.py
@@ -1,0 +1,491 @@
+"""
+Description of Arrays or Subarrays of telescopes
+Those codes are from the latest version of ctapipe (ctapipe/instruments/subarray.py)
+"""
+
+__all__ = ["SubarrayDescription"]
+
+from collections import defaultdict
+from pathlib import Path
+
+import numpy as np
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.table import Table
+from astropy.utils import lazyproperty
+import tables
+from copy import copy
+
+import ctapipe
+
+from ctapipe.coordinates import GroundFrame, CameraFrame
+from ctapipe.instrument.telescope import TelescopeDescription
+from ctapipe.instrument.camera import CameraDescription, CameraReadout, CameraGeometry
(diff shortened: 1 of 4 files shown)


pawel21:
Very useful script :) Thank you. 

> So this procedure works if analyzers want to apply different image cleaning from default.
By "apply different image cleaning" you mean different image parameters (picture and boundary threshold)?
We still can use only one cleaning method 'tailcuts cleaning' with this script? Do I understand it correctly?

vuillaut:
Hi @SeiyaNozaki 

Unless I am missing something, this script already exists in lstchain: `lstchain_mc_dl1ab.py`
It's been tested and used in the past.
Though it is called `_mc_` for historical reasons, it should be working fine on real data since this part of parameters recalculation is identical.

SeiyaNozaki:
@vuillaut Thanks for pointing it out!  I didn't notice that... and I confirmed it is working actually with real data.
Maybe `_mc_` should be removed...?

@pawel21 your understanding is correct and the existing script also assumes `tailcut` cleaning. I think it would be better to have an option to choose a cleaning method in a config file as other calibration methods (e.g. image_extractor). It can make it easy to study the image cleaning method!

vuillaut:
> @vuillaut Thanks for pointing it out! I didn't notice that... and I confirmed it is working actually with real data.
> Maybe `_mc_` should be removed...?

So I assume this PR can be closed.
You may propose changes to the said script if needed.

> @pawel21 your understanding is correct and the existing script also assumes `tailcut` cleaning. I think it would be better to have an option to choose a cleaning method in a config file as other calibration methods (e.g. image_extractor). It can make it easy to study the image cleaning method!

That would.
If not urgent, I would suggest waiting for changes using the stage1 but if you need that, please propose a PR.

PR #578, opened 2020-12-23
make github ci pass and remove travis

vuillaut:
Fix the issue with the unit test concerning the version.


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
             "git+https://github.com/cta-observatory/ctapipe_io_lst@$CTAPIPE_IO_LST_VERSION" \
             pytest-cov
 
-          pip install -e .
```
morcuended, line 47: I noticed that Travis and GitHub actions were running tests in one of the pull requests and found this one still not merged.

Just for curiosity, why is the `-e` option preventing from passing the tests?

maxnoe: I don't think it is.

maxnoe: But -e is also not needed for travis, no one will edit code there :wink:

maxnoe (approved):

PR #579, opened 2021-01-07
Fix version of iminuit to < 2

maxnoe:


rlopezcoto (approved):
lstchain currently does not work with `iminuit` v2.0 as reported by @maxnoe on slack

PR #582, opened 2021-01-20
Remove n_islands from config files

rlopezcoto:
As discussed some time ago, with the changes in Data/MC regarding integration windows and correction factors, the number of islands between data and MC was not matched anymore when using the standard cleaning settings, so there was a mismatch in background rates after classification noticed by @moralejo 
The temporary solution is to remove the `n_islands` from the Random Forest training

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -68,7 +68,6 @@
     "r",
     "time_gradient",
     "leakage_intensity_width_2",
-    "n_islands"
   ],
 
   "classification_features": [
@@ -85,7 +84,6 @@
     "r",
     "time_gradient",
     "leakage_intensity_width_2",
-    "n_islands",
     "log_reco_energy",
     "reco_disp_dx",
     "reco_disp_dy"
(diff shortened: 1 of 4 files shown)


lstchain/data/lstchain_standard_config.json
```diff
@@ -68,7 +68,6 @@
     "r",
     "time_gradient",
     "leakage_intensity_width_2",
```
vuillaut, line 70: last line must not have a `,`

lstchain/data/soft_cut_config.json
```diff
@@ -82,7 +82,6 @@
     "r",
     "time_gradient",
     "leakage_intensity_width_2",
```
vuillaut, line 84: same here

rlopezcoto: done

lstchain/data/lstchain_standard_config.json
rlopezcoto: done

vuillaut (approved):

garciagenrique:
Just for the record (sorry if it seems obvious), I will keep tagging the MC productions with the last released version of lstchain to be consistent. 
Meaning that until we don't have >=v0.6.4, all the prods will contain the `wo_n_islands` tag (until this version is not implemented `n_islands` would still be in the standard config file).

rlopezcoto:
that sounds good to me @garciagenrique

PR #585, opened 2021-02-05
Update only disp params if already present in the file

morcuended:
Trying to fix #543 for real data since there are no disp related parameters in those files. They are only in MC files

I removed the re-calculation of `r` because it is already done by the `hillas_parameters` function.

### DIFF
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -101,10 +101,7 @@ def main():
         'time_gradient',
         'n_pixels',
         'wl',
-        'r',
-        'log_intensity',
-        'x',
-        'y',
+        'log_intensity'
     ])
 
     nodes_keys = get_dataset_keys(args.input_file)
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -105,11 +105,6 @@ def main():
         'log_intensity',
         'x',
         'y',
-        'disp_dx',
-        'disp_dy',
-        'disp_norm',
-        'disp_angle',
-        'disp_sign',
     ])
 
     nodes_keys = get_dataset_keys(args.input_file)
@@ -120,6 +115,8 @@ def main():
 
     with tables.open_file(args.input_file, mode='r') as input:
         image_table = input.root[dl1_images_lstcam_key]
+        dl1_params_input = input.root[dl1_params_lstcam_key].colnames
+
         with tables.open_file(args.output_file, mode='a') as output:
 
             params = output.root[dl1_params_lstcam_key].read()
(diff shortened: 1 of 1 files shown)


rlopezcoto (approved):
thanks @morcuended 
maybe let's get this merged and solve the problem generated by #543 and later decide to go and write some unit tests for real data

morcuended:
For an MC file:
image

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -158,19 +152,29 @@ def main():
...
-                                                                          dl1_container['y'].to_value(u.m),
-                                                                          params['src_x'][ii],
-                                                                          params['src_y'][ii],
-                                                                          )
-                dl1_container['disp_dx'] = disp_dx
-                dl1_container['disp_dy'] = disp_dy
-                dl1_container['disp_norm'] = disp_norm
-                dl1_container['disp_angle'] = disp_angle
-                dl1_container['disp_sign'] = disp_sign
+                if ['disp_dx', 'disp_dy', 'disp_norm', 'disp_angle', 'disp_sign'] in dl1_params_input:
```
morcuended, line 157: This is not working as expected, either I check that only one of those `disp_` parameters are present or do something like `if disp_params.issubset(dl1_params_input)` which slows a lot the runtime. Any suggestion?

maxnoe: This is a check that has to run once per file. How can that be slowing it down by a lot?

```
if set(dl1_params.intersection({'disp_dx', 'disp_dy', 'disp_norm', 'disp_angle', 'disp_sign'}):
```

maxnoe: Also:

```

In [5]: %timeit set(dl1_columns).intersection({'disp_dx', 'disp_dy', 'disp_norm', 'disp_angle', 'disp_sign'})
777 ns ± 2.24 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

how can this be a problem?

maxnoe: Dependening on if you want `all` behaviour or `any` behaviour (i would go for `any`), use `intersection` or `issubset`

morcuended: I was doing that inside the loop over all the events `for ii, row in enumerate(image_table)`

maxnoe: Even then. It takes less than *a microsecond*`. But why do you do that in the loop? The `parameters_to_update` will be fixed for the whole file, or not?

morcuended: Yes, you are right. I will refactor that.

I got the impression in my first tests that it was much slower but then tested with the another `if 'disp_dx'...` condition and it was pretty much the same.

morcuended: Done

rlopezcoto (approved):

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -158,19 +153,21 @@ def main():
                     length = np.rad2deg(np.arctan2(dl1_container.length, foclen))
                     dl1_container.width = width
                     dl1_container.length = length
-                    dl1_container.r = np.sqrt(dl1_container.x ** 2 + dl1_container.y ** 2)
```
vuillaut, line 161: Sorry, why did you remove that line?

vuillaut: nevermind, got it

vuillaut (approved):

lstchain/scripts/lstchain_dl1ab.py
morcuended: It is already calculated by `hillas_parameters` function some lines above if I'm not wrong.

```
return HillasParametersContainer(
        x=u.Quantity(cog_x, unit),
        y=u.Quantity(cog_y, unit),
        r=u.Quantity(cog_r, unit),
        phi=Angle(cog_phi, unit=u.rad),
        intensity=size,
        length=u.Quantity(length, unit),
        length_uncertainty=u.Quantity(length_uncertainty, unit),
        width=u.Quantity(width, unit),
        width_uncertainty=u.Quantity(width_uncertainty, unit),
        psi=Angle(psi, unit=u.rad),
        skewness=skewness_long,
        kurtosis=kurtosis_long,
    )
```

PR #586, opened 2021-02-05
Download the private test data

maxnoe:
I added the necessary credentials to the repository secrets and also added a small example test showing how to use it.

### DIFF
--- a/download_test_data.sh
+++ b/download_test_data.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+set -eu
+
+TEST_DATA_URL=${TEST_DATA_URL:-https://big-tank.app.tu-dortmund.de/lst-testdata/}
+TEST_DATA_USER=${TEST_DATA_USER:-""}
+TEST_DATA_PASSWORD=${TEST_DATA_PASSWORD:-""}
+
+
+if [ -z "$TEST_DATA_USER" ]; then
+	echo -n "Username: "
+	read TEST_DATA_USER
+	echo
+fi
+
+if [ -z "$TEST_DATA_PASSWORD" ]; then
+	echo -n "Password: "
+	read -s TEST_DATA_PASSWORD
+	echo
+fi
+
+
+wget \
+	-R "*.html*,*.gif" \
+	--no-host-directories --cut-dirs=1 \
+	--no-parent \
+	--user="$TEST_DATA_USER" \
+	--password="$TEST_DATA_PASSWORD" \
+	--no-verbose \
+	--recursive \
+	--directory-prefix=test_data \
+	"$TEST_DATA_URL"
--- a/lstchain/tests/test_lstchain.py
+++ b/lstchain/tests/test_lstchain.py
@@ -8,6 +8,7 @@ from lstchain.io import standard_config
 from lstchain.reco.utils import filter_events
 import astropy.units as u
 import tables
+from pathlib import Path
 
 test_dir = 'testfiles'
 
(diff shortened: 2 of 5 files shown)


lstchain/tests/test_lstchain.py
```diff
@@ -21,6 +22,7 @@
 file_model_energy = os.path.join(test_dir, 'reg_energy.sav')
 file_model_disp = os.path.join(test_dir, 'reg_disp_vector.sav')
 file_model_gh_sep = os.path.join(test_dir, 'cls_gh.sav')
+r0_file = Path('test_data/real/R0/20200218/LST-1.1.Run02008.0000_first50.fits.fz')
```
morcuended, line 25: With this filename, r0_to_dl1 is complaining:
```
Filename dl1_LST-1.1.Run02008.0000_first50.h5 does not match pattern re.compile('dl1_LST-(\\d+)(?:.(\\d+))?.Run(\\d+).(\\d+)(?:.fits)?.(?:h5|hdf5|hdf)')
```
Shall we modify it or just copy it as `test_data/real/R0/20200218/LST-1.1.Run02008.0000.fits.fz`?

morcuended (changes requested):

lstchain/tests/test_lstchain.py
maxnoe: I would just update the regex to also allow additional suffixes

morcuended: I agree. The PR looks good to me then.

morcuended (approved):

vuillaut (approved):
Awesome, thanks !

download_test_data.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+if [ -z "$TEST_DATA_PASSWORD" ]; then
+	echo -n "Password: "
+	read -s TEST_DATA_PASSWORD
+	echo
+fi
+
+
+wget \
```
vuillaut, line 23: add `-nc` to avoid human users re-downloading data when running `pytest` on their laptop?

morcuended: don't you download the data once first and that's it? Does `pytest` try to download the data everytime you run it locally?

maxnoe: Yes, good option

maxnoe: No, pytest does not do any downloading. But we should make sure that people can simply rerun the command when we update data on the server.

maxnoe: But I think better to go with `-N` instead of `-nc`

morcuended: ok, I see

vuillaut: but maybe new data should be checked for everytime pytest is run?
Imagine
- a dev download the test data once
- there is a new test requiring new data on master
- the dev runs pytest, everything passes because he is lacking new data (or new version of the data)
- tests fail in GitHub CI because the new tests is not actually passing

Note this is not critical, just more convenient for devs.
So it is just a small suggestion for improvement, feel free to merge as it is already.

maxnoe: We can do this using functions from ctapipe 0.10 as soon as we updated using `get_dataset_path`.

maxnoe: The download script was more a temporal solution until that is there.

vuillaut: ha great. then let's merge and we can update later to improve it !

morcuended:
I think this can be merge already.

In order to test the creation of drs4 and calibration files we could think of add DRS4 and ped-cal R0 files as well.

vuillaut:
Although the tests were passing here, the CI is now broken on other PRs.
See there for an example:
https://github.com/cta-observatory/cta-lstchain/pull/589/checks?check_run_id=1879616377

morcuended:
Are there any additional settings that need to be adapted to have the PRs passing the tests?

PR #587, opened 2021-02-08
Update to use ctapipe 0.10 and ctapipe_io_lst 0.6

FrancaCassol:
This is draft PR to  migrate lstchain to the new version of ctapipe_io_lst (v0.6.0) and ctapipe 0.10.1. 
The status of the code is : work progress

For the moment the main changes are on the calibration code. 
In particular, the calibration part of the r0_to_dl1 script has been updated in order to be compatible with the new reader, but  to make it working again, other changes in  reconstruction part must be done. So I could not test interleaved calibration results yet.

I encourage all lstchain experts to verify their part of code and eventually push changes in this branch


lstchain/reco/r0_to_dl1.py
```diff
@@ -453,192 +448,19 @@ def r0_to_dl1(
...
-                                first_valid_ucts = ucts_time
-                            if first_valid_ucts_tib is None \
-                                    and event.lst.tel[telescope_id].evt.extdevices_presence & 1:
-                                first_valid_ucts_tib = ucts_time
-                        else:
-                            ucts_time = math.nan
+                    # !!! to be filled  from the source !!!
+                    dl1_container.ucts_time = 0
+                    dl1_container.dragon_time = 0
+                    dl1_container.tib_time = 0
```
moralejo, line 454: I would keep only dragon_time which is the most reliable one now, and the one calculated by the event source. In any case, once we move to using the standard DL1 format, only one timestamp will be kept. 
The cross-checks involving comparisons of different times etc should be made if needed at an earlier stage (and I think they are already done in UCM-GAE's daily check)

environment.yml
```diff
@@ -6,7 +6,7 @@ channels:
 dependencies:
   - python=3.7
   - pip
-  - ctapipe=0.8.0
+  - ctapipe>=0.10.1
   - gammapy>=0.18
   - h5py
   - ipython
```
moralejo, line 12: @FrancaCassol : ctapipe_io_lst 0.6.0 is needed, right?

lstchain/reco/r0_to_dl1.py
moralejo: I did some comparisons of the r0_to_dl1 output (the one in this PR vs. the one in the latest lstchain release) and it seems to me that we are using the "waveform-shifting" option of the new LST R1 event source. I would rather switch that off for now, since it will introduce some differences in reconstructed pixel charges and times,  which will make it harder to detect other possible differences between the new and the old DL1.

lstchain/reco/r0_to_dl1.py
```diff
@@ -221,10 +219,30 @@ def r0_to_dl1(
...
+            "PointingSource":{
+                "drive_report_path": pointing_file_path
+            },
+            "LSTR0Corrections":{
+                "drs4_pedestal_path": pedestal_path,
+                "calibration_path": calibration_path,
+                "drs4_time_calibration_path": time_calibration_path,
+            }
+        }
+    }
```
morcuended, line 241: Why is this configuration here? Shouldn't it be placed in a config file?

morcuended:
Maybe it's a silly question but, why do we need the r0_to_dl1 script in lstchain whenever dl1 files are produced by `ctapipe-stage1` script?

maxnoe:
I think the idea here was to do an intermediate step of using r0_to_dl1 with the new source instead of directly using stage1 since there are still some unresolved issues for the `R1 → Dl1` analysis.

* muon analysis
* pointing information handling
* probably something I forgot here (let's make an issue)

So I think it is a good idea to 

* Move to ctapipe 0.10 as soon as possible
* Remove the now duplicated and slower code from lstchain by using the source
* Keep r0_to_dl1 for now

Unfortunately, this PR mixes doing that with introducing a new feature (new calibration code).

I would propose closing this or cleaning it up so it only does what is needed to move to 0.10 and the new source without introducing something else.

environment.yml
maxnoe: We have the problem that ctapipe_io_lst cannot be on pypi, because of the rawzfitsreader dependency and you are not allowed to include dependency links in setup.py.

I am working on a solution for that. (By making protozfitsreader available on PyPI)

lstchain/reco/r0_to_dl1.py
maxnoe: For that you should just add `"apply_waveform_time_shift": false` to the `CameraCalibrator` (default?) config.

maxnoe: Reminder, this is the config I use for the stage1 calibration: 

```
  "LSTEventSource": {
    "LSTR0Corrections": {
      "drs4_pedestal_path": "../data/lst/2020-09-09/drs4_pedestal.Run2603.0000.fits",
      "drs4_time_calibration_path": "../data/lst/2020-09-09/time_calibration.Run2605.0000.hdf5",
      "calibration_path": "../data/lst/2020-09-09/calibration.Run2605.0000.hdf5",
      "gain_selection_threshold": 3500,
      "add_calibration_timeshift": true
    },
    "PointingSource": {
      "drive_report_path": "../data/lst/2020-09-09/drive_log_20_09_09.txt"
    },
    "EventTimeCalculator": {
      "ucts_t0_dragon": 1599691710450937890,
      "dragon_counter0": 7450937600,
      "ucts_t0_tib": 1599691710450937890,
      "tib_counter0": 7450940400
    }
  },
  "CameraCalibrator": {
    "apply_waveform_time_shift": false,
    "image_extractor_type": "LocalPeakWindowSum",
    "LocalPeakWindowSum": {
      "window_shift": 4,
      "window_width": 8,
      "apply_integration_correction": false
    }
  },

```

lstchain/reco/r0_to_dl1.py
FrancaCassol: Hi @morcuended,
I used this dirty solution not to change the onsite script. 
Actually the right solution is to make this script a tool, hopefully this will happen soon.

FrancaCassol:
The new calibration code  was already reviewed  in PR [https://github.com/cta-observatory/cta-lstchain/pull/552](https://github.com/cta-observatory/cta-lstchain/pull/552)

maxnoe:
More reason to not include it here.

moralejo:
> More reason to not include it here.

Although I agree this might have been split in different PRs, probably it is not worth to do it now. Let's try to make it work and review it as it is.

setup.py
```diff
@@ -45,7 +45,7 @@ def find_scripts(script_dir, prefix):
         "astropy~=4.0,>=4.0.2",
         'ctapipe>=0.10.1',
```
moralejo, line 46: Why >=0.10.1 here and 0.10.2 in ci.yml?

maxnoe: 0.10.2 is not yet on conda-forge

maxnoe: Now it is.

lstchain/reco/r0_to_dl1.py
```diff
@@ -221,10 +219,31 @@ def r0_to_dl1(
...
+                "ucts_t0_dragon": ucts_t0_dragon,
+                "dragon_counter0": dragon_counter0,
+                "ucts_t0_tib": ucts_t0_tib,
+                "tib_counter0": tib_counter0
+            },
+            "PointingSource":{
+                "drive_report_path": pointing_file_path
+            },
+            "LSTR0Corrections":{
+                "calibrate_ff_ped": False,
```
moralejo, line 236: This should be calibrate_ff_and_ped   - but I tried it and it is not recognized anyway, with ctapipe_io_lst 0.6.1

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -34,6 +34,7 @@
 optional.add_argument('--default_time_run', help="If 0 time calibration is calculated otherwise create a link to the give run time calibration",type=int, default='1625')
 optional.add_argument('--ff_calibration', help="Perform the charge calibration (yes/no)",type=str, default='yes')
 optional.add_argument('--tel_id', help="telescope id. Default = 1", type=int, default=1)
+optional.add_argument('--sub_run', help="sub-run to be processed. Default = 0", type=int, default=0)
```
morcuended, line 37: Shouldn't we try to open all the files available?

lstchain/visualization/plot_drs4.py
```diff
@@ -42,24 +42,31 @@ def plot_pedestals(data_file, pedestal_file, run=0 , plot_file="none", tel_id=1,
 
     plt.rc('font', size=15)
```
morcuended, line 43: I opened a PR some time ago cause I think the `PdfPages` object is created twice in this script, here and the same lines appear later on around L90. I think the PDF should be opened only once.

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -135,16 +149,22 @@ def main():
...
+                  f"--FlatFieldCalculator.sample_size={stat_events} --PedestalCalculator.sample_size={stat_events} " \
+                  f"--config={config_file}  >  {log_file} 2>&1"
 
             print("\n--> RUNNING...")
             os.system(cmd)
 
             # plot and save some results
-            plot_file=f"{output_dir}/log/calibration.Run{run}.0000.pedestal.Run{ped_run}.0000.pdf"
+            plot_file=f"{output_dir}/log/calibration.Run{run}.{sub_run}.pedestal.Run{ped_run}.0000.pdf"
             print(f"\n--> PRODUCING PLOTS in {plot_file} ...")
```
morcuended, line 168: logging

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -109,16 +124,15 @@ def main():
                 raise NameError()
             else:
                 time_calibration_file = file_list[0]
```
morcuended, line 126: Make sure that the corresponding default time calibration file exists.

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -79,23 +82,35 @@ def main():
             print(f">>> Error: The pedestal file {pedestal_file} do not exist.\n Exit")
```
morcuended, line 82: I think we should be able to create calibration files using drs4 pedestal files from another different day. Maybe using a glob somewhere here. This was somewhat inconvenient for us when using this script some time ago.

lstchain/reco/r0_to_dl1.py
```diff
@@ -178,10 +175,10 @@ def r0_to_dl1(
...
     time_calibration_path=None,
     pointing_file_path=None,
-    ucts_t0_dragon=math.nan,
-    dragon_counter0=math.nan,
-    ucts_t0_tib=math.nan,
-    tib_counter0=math.nan
+    ucts_t0_dragon=None,
+    dragon_counter0=None,
+    ucts_t0_tib=None,
+    tib_counter0=None,
```
morcuended, line 181: why `None`? should we use `-1` instead?

lstchain/reco/r0_to_dl1.py
```diff
@@ -453,192 +442,19 @@ def r0_to_dl1(
...
-                    ucts_time_utc = unix_tai_to_time(ucts_time)
-                    dragon_time_utc = unix_tai_to_time(dragon_time)
-                    tib_time_utc = unix_tai_to_time(tib_time)
+                    dl1_container.ucts_trigger_type = event.lst.tel[telescope_id].evt.ucts_trigger_type
 
-                    dl1_container.ucts_time = ucts_time_utc.unix
-                    dl1_container.dragon_time = dragon_time_utc.unix
-                    dl1_container.tib_time = tib_time_utc.unix
+                dl1_container.az_tel = event.pointing.tel[telescope_id].azimuth
+                dl1_container.alt_tel = event.pointing.tel[telescope_id].altitude
```
morcuended, line 453: shouldn't these lines be indented at the same level as the previous one?
```
dl1_container.ucts_trigger_type = event.lst.tel[telescope_id].evt.ucts_trigger_type
```

morcuended: still makes sense that they live inside `if not is_simu:`

morcuended (changes requested):

morcuended:
I think it would be good now to implement tests for real data scripts here. I think that PR #590 does not make sense by itself and should be adapted taking into account the changes made here.

lstchain/scripts/lstchain_data_create_time_calibration_file.py
```diff
@@ -57,7 +56,39 @@
                     help='Path to drs4 pedestal file ',
                     default=None
                     )
+parser.add_argument('--ucts-t0-dragon', action='store', type=float,
```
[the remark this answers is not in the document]
maxnoe: Yes, ints. Default should be None, not nan

lstchain/reco/r0_to_dl1.py
maxnoe: No, None is perfekt in python for "no value given" and is expected like this for by the time calculator

lstchain/scripts/lstchain_data_create_time_calibration_file.py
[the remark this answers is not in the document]
moralejo: > Yes, ints. Default should be None, not nan

I get the error below when trying to run r0_to_dl1 (which worked on this branch a few commits ago). Is it perhaps related?

```
  File "/nfs/cta-ifae/moralejo/miniconda3/envs/lst-dev/lib/python3.7/site-packages/traitlets/traitlets.py", line 2040, in validate
    self.error(obj, value)
  File "/nfs/cta-ifae/moralejo/miniconda3/envs/lst-dev/lib/python3.7/site-packages/traitlets/traitlets.py", line 681, in error
    raise TraitError(value, info or self.info(), self)
traitlets.traitlets.TraitError: (1.5953039761947725e+18, 'an int', <ctapipe_io_lst.event_time.Int object at 0x7fcc88706e50>)
```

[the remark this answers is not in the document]
maxnoe: Yes, please note that the tests are all falling right now

maxnoe:
@morcuended @moralejo I don't think reviewing makes much sense right now.  

Let's make the tests run again first. 

@FrancaCassol 
You somehow completely removed the calibrator for simulation events. 

The calibrator is only constructed when `not is_simu`

lstchain/reco/r0_to_dl1.py
```diff
@@ -220,14 +220,14 @@ def r0_to_dl1(
...
 
-    source_config={
+    source_config = {
         "LSTEventSource": {
             "allowed_tels": [1],
             "EventTimeCalculator": {
-                "ucts_t0_dragon": int(ucts_t0_dragon),
-                "dragon_counter0": int(dragon_counter0),
-                "ucts_t0_tib": int(ucts_t0_tib),
-                "tib_counter0": int(tib_counter0)
```
FrancaCassol: Hi @maxnoe,

if we do not set this as integer we get the error already mentioned by @moralejo

lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py
```diff
@@ -67,46 +65,42 @@
...
+                n_modules = event.lst.tel[tel_id].svc.num_modules
+                pedestal = DragonPedestal(tel_id=tel_id, n_module=n_modules, r0_sample_start=args.start_r0_waveform)
+
+            if args.deltaT:
+                reader.r0_r1_calibrator.update_first_capacitors(event)
+                reader.r0_r1_calibrator.time_lapse_corr(event, tel_id)
+
             pedestal.fill_pedestal_event(event)
             if i%500 == 0:
                 print("i = {}, ev id = {}".format(i, event.index.event_id))
```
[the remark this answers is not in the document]
FrancaCassol: Hi @maxnoe,
> @FrancaCassol
> You somehow completely removed the calibrator for simulation events.
> 
> The calibrator is only constructed when `not is_simu`

what do you mean for calibration? the dc_to_pe conversion or the image creation? The first is done at the  reader level (I hope also for the simulation events), the second is done by the r1_dl1_calibrator(event) that is called for all type of events in my understanding

lstchain/scripts/onsite/onsite_create_calibration_file.py
FrancaCassol: Hi @morcuended, here the idea was just to make the changes to use the new reader  (even if I actually added also some new features by mistake, so I personally was not really coherent on that).

lstchain/reco/r0_to_dl1.py
```diff
@@ -225,10 +225,10 @@ def r0_to_dl1(
...
             "calibrate_flatfields_and_pedestals": False,
             "EventTimeCalculator": {
-                "ucts_t0_dragon": ucts_t0_dragon,
-                "dragon_counter0": dragon_counter0,
-                "ucts_t0_tib": ucts_t0_tib,
-                "tib_counter0": tib_counter0
+                "ucts_t0_dragon": int(ucts_t0_dragon),
+                "dragon_counter0": int(dragon_counter0),
+                "ucts_t0_tib": int(ucts_t0_tib),
+                "tib_counter0": int(tib_counter0)
```
maxnoe, line 231: No, this is wrong. If it creates an error, it has to be fixed another way!

There cannot be an `int` call here since that will create an error for `None`. It should either be None or already be an integer at this point in the code.

moralejo: > No, this is wrong. If it creates an error, it has to be fixed another way!
> 
> There cannot be an `int` call here since that will create an error for `None`. It should either be None or already be an integer at this point in the code.
> 
> Please revert.

Can't those use -1 as error codes?

maxnoe: What error code? These are *options*.

They are the traitlets of the `EventTimeCalculator`

maxnoe: I don't see where the problems come from. These come from `argparse` command line options. Just to `type=int` and no default. Then it will be `None` when not given and an `int` when given, exactly as needed.

moralejo: > What error code? These are _options_.
> 
I meant defaults, I thought the problem was using None's as defaults. Perhaps I'm wrong, thoguh.

maxnoe: Now I see. You everywhere set non-sensical defaults for the command line options instead of using the fact that argparse can give you `None` for "no option provided".

maxnoe: > I meant defaults, I thought the problem was using None's as defaults. 

No, there is not. In fact, it is the "default default".

moralejo: Ok, let's use Nones. So what is the solution for the problem?

lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py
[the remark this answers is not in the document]
maxnoe: No. I mean the `CameraCalibrator`. There is currently no `CameraCalibrator` `if is_simu`. So r0_to_dl1 fails on simulations. See the unit test failures.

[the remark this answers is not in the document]
maxnoe: I get `Local variable r1_dl1_calibrator accessed before definition` for simulation events, because it is just not created for simulation

lstchain/reco/r0_to_dl1.py
maxnoe: This is different for container values because there, the values need to be written to a file using the correct data type, so the defaults must (or rather the missing value indicator, which could be different from the default) must be of the same type. Which is a problem with all data types but floats, since only floats have a natural missing value indicator.

lstchain/reco/r0_to_dl1.py
maxnoe: Removing the `default="NaN"` for the cli option

maxnoe:
Everyone, please run the tests locally. 

For now, let's get the r0 to dl1 script working first. So do:

```
rm -rf test_files; pytest lstchain/tests/test_lstchain.py::test_r0_to_dl1
```

maxnoe:
With the last commit, `r0_to_dl1` is running again.

maxnoe:
These are the tests that are still failing:

```
lstchain/scripts/tests/test_lstchain_scripts.py::test_all_help[lstchain_create_calibration_file] FAILED                                [ 16%]
lstchain/scripts/tests/test_lstchain_scripts.py::test_all_help[lstchain_data_r0_to_r1] FAILED                                          [ 33%]
lstchain/scripts/tests/test_lstchain_scripts.py::test_dl1ab FAILED                                                                     [ 50%]
lstchain/calib/camera/tests/test_flatfield.py::test_flasherflatfieldcalculator FAILED                                                  [ 66%]
lstchain/calib/camera/tests/test_pedestals.py::test_pedestal_calculator FAILED                                                         [ 83%]
lstchain/io/tests/test_io.py::test_read_simu_info_merged_hdf5 FAILED    
```

maxnoe:
The last failing test has me concerned a bit there might be a deeper problem:

```
        assert mcheader.simtel_version == 1462392225
>       assert mcheader.num_showers == 40000
E       assert 20000 == 40000
```

How did that change?

maxnoe:
Ah, the merge fails so it only contains one file.

maxnoe:
And the test for the `merge_check` function was missing the assert statement, so it didn't fail.

maxnoe:
Tests are all passing now. But that will probably not mean that everything works, due to missing coverage / integration checks.

So experts, please check this branch out and try using it.

If you encounter problems:

* Report here
* Propose how that problem could be detected in a unit test to help prevent breaking this in the next update.

Some points:

* Check pointing info for simulated events works as expected, as `mcheader` does not contain `run_array_direction` any more. This info is now in each event's `pointing` container.

maxnoe:
Note: lstchain now requires python 3.7 as traitlets 5 is >= 3.7 and protozfits is currently < 3.8

moralejo:
> Tests are all passing now. But that will probably not mean that everything works, due to missing coverage / integration checks.
> 

Great, thanks a lot!

> So experts, please check this branch out and try using it.
 
I managed to run the r0_to_dl1 script using preexisting calibration files. For now I checked pixel info (q's & times) and they seem nearly identical to the result with latest lstchain release. Differences are mostly for pulses late or early in the window, so as we discussed this is due to the waveform shifting option in this new version, which I think should be disabled for now.



> If you encounter problems:
> 
> * Report here
> * Propose how that problem could be detected in a unit test to help prevent breaking this in the next update.
> 
> Some points:
> 
> * Check pointing info for simulated events works as expected, as `mcheader` does not contain `run_array_direction` any more. This info is now in each event's `pointing` container.

maxnoe:
@moralejo I added the config to disable the waveform shifting in the two standard config files.

moralejo:
> 
> I managed to run the r0_to_dl1 script using preexisting calibration files. For now I checked pixel info (q's & times) and they seem nearly identical to the result with latest lstchain release. Differences are mostly for pulses late or early in the window, so as we discussed this is due to the waveform shifting option in this new version, which I think should be disabled for now.
>

Ok, was done even before I finished writing. Thanks @maxnoe!

maxnoe:
@FrancaCassol The test data now also has a file with missing module, I think it is a closed shutter pedestal file, run 3669.
I extracted only the first 50 events using the zfitsrecompressor tool.

```
test_data/real/R0/20210215/LST-1.1.Run03669.0000_first50.fits.fz 
```

maxnoe:
@FrancaCassol @moralejo The problem now is that the pixel_status is 1D since the latest change in the source.

I am thinking about where to fix this. I think we should do it here, since the original pixel status in the R1 data model is also just of shape `(n_pixels, )` and not `(n_gains, n_pixels)`

maxnoe:
On the other hand, ctapipe defines the `*_failing_pixels` to be of shape `n_gains, n_pixels` so I'd say we stick with that and might have to change ctapipe in an upcoming version to better be compatible with the data model.

lstchain/tests/test_lstchain.py
```diff
@@ -44,9 +50,28 @@ def test_r0_to_dl1():
...
+        custom_config=standard_config,
+        pedestal_path=test_drs4_pedestal_path,
+        calibration_path=test_calib_path,
+        time_calibration_path=test_time_calib_path,
+        pointing_file_path=test_drive_report,
+        ucts_t0_dragon=None,
+        dragon_counter0=None,
+        ucts_t0_tib=None,
+        tib_counter0=None,
+    )
```
morcuended: Don't you need any assertion statement here?

FrancaCassol:
Hi @maxnoe,

> On the other hand, ctapipe defines the `*_failing_pixels` to be of shape `n_gains, n_pixels` so I'd say we stick with that and might have to change ctapipe in an upcoming version to better be compatible with the data model.

this is also necessary for the calibration code.

lstchain/tests/test_lstchain.py
maxnoe: For now I just wanted to see that it runs without error.

morcuended:
I was going to say I had the same tests failing due to the mask sizes with `Run02008.0000_first50.fits.fz`

Before committing them here, do we need tests for the scripts themselves? Like the one below:

```
def test_lstchain_data_r0_to_dl1():
    run_program(
        "lstchain_data_r0_to_dl1",
        "-f",
        test_r0_path,
        "-o",
        output_dir,
        "--pedestal-file",
        test_drs4_pedestal_path,
        "--calibration-file",
        test_calib_path,
        "--time-calibration-file",
        test_time_calib_path,
        "--pointing-file",
        test_drive_report,
        "--ucts-t0-dragon",
        "1582059789516351903",
        "--dragon-counter0",
        "2516351600",
        "--ucts-t0-tib",
        "1582059789516351903",
        "--tib-counter0",
        "2516351200",
    )
    assert real_data_dl1_file.is_file()
    assert muons_file.is_file()
    assert datacheck_file.is_file()
```

Also, we can test dl1ab and dl1_to_dl2 scripts with observed data, right?

maxnoe:
@FrancaCassol yes, I know that. That was the initial reason this came up and the question I asked what should be adapted, the calibration code or the source.

maxnoe:
@morcuended yes, as said, adding more tests would be very welcome

moralejo:
> On the other hand, ctapipe defines the `*_failing_pixels` to be of shape `n_gains, n_pixels` so I'd say we stick with that and might have to change ctapipe in an upcoming version to better be compatible with the data model.

failing_pixels is just a Bool, right?  Which seems quite limiting. If it had several bits for different types of pixel errors, it could be of shape n_pixels. Then masks could be built from it as needed.
Ok to stick with (n_gains, n_pixels) for now, though.

maxnoe:
@moralejo Yes, the `pixel_status` is a bitset according to the R1 model, Section A.5 with 8 bits.

morcuended:
I'm getting timestamps = 0 with the test files `Run02008.0000_first50.fits.fz` and `Run02008.0100_first50.fits.fz`

maxnoe:
@morcuended what are you doing? I would expect timestamps = 0 only with `LSTEventSource.fill_timestamps=False`. Otherwise there should either be an error or valid timestamps.

morcuended:
The test is basically what I posted above, I'm not specifying the config file so I assume I'm using the default one. Maybe this is what's happening.

maxnoe:
When giving the counter values, we should get valid times.

Can you add the test you wrote above? And then we should start checking for valid timestamps.

morcuended:
Sure, I'm fixing the name of the produced DL1 and DL2 files. I will commit the tests afterward.

maxnoe:
@FrancaCassol One thing that confuses me is that you added code to `LSTCameraCalibrator` but that is not used anywhere now (`r0_to_dl1` uses `ctapipe.calib.CameraCalibrator`).


Can the `LSTCameraCalibrator` be removed? If no, why is it not used anywhere?

FrancaCassol:
It is not used anymore, but I preferred to remove it after that all things worked.

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -1,181 +1,303 @@
...
         if ep.module_name.startswith(package_name)
     ]
     return entrypoints
 
 
-ALL_SCRIPTS = find_entry_points('lstchain')
+ALL_SCRIPTS = find_entry_points("lstchain")
 
 
 def run_program(*args):
```
maxnoe, line 54: can we add an option `cwd` here so we can run them in the `tmp_path`?

lstchain/tests/test_lstchain.py
```diff
@@ -32,16 +34,20 @@
...
+
 
 def test_import_visualization():
-    from lstchain import visualization
+    pass
+
 
 def test_import_lstio():
-    from lstchain import io
+    pass
```
maxnoe, line 49: Why the `pass` changes? These tests now do nothing.

morcuended:
test are not passing for `lstchain_merge_hdf5_files`:

```
tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``/instrument/subarray/layout``
```
But I think I did not touch anything here apart from cosmetics

Hence the next function is also failing:
```
    @pytest.mark.run(after="test_lstchain_merge_dl1_hdf5_files")
    def test_lstchain_merged_dl1_to_dl2():
>       output_file = merged_dl1_file.replace("dl1", "dl2")
E       TypeError: replace() takes 2 positional arguments but 3 were given
```
Not sure what is happening.

Also I did not know how to better specify the name of the muons file. For the muons file the stream is stripped out, however for the others dl1, datacheck and dl2 the stream is still present. I think this should be checked and fixed.

lstchain/tests/test_lstchain.py
morcuended: It's probably an unwanted change from your previous commits. I will try to revert them.

morcuended: I think it was the automatic refactor and optimizing imports options

morcuended:
> Also I did not know how to better specify the name of the muons file. For the muons file the stream is stripped out, however for the others dl1, datacheck and dl2 the stream is still present. I think this should be checked and fixed.

I correct myself here. The output files from `` are:

```
dl1_LST-1.1.Run02008.0000_first50.h5
muons_LST-1.Run02008.0000_first50.fits
datacheck_dl1_LST-1.Run02008.0000.h5
dl2_LST-1.1.Run02008.0000_first50.h5
```
So muons and datacheck files strip out the stream number, datacheck file even removes the `_first50` suffix since I guess it is based on `lstchain.path`

morcuended:
I had to separate the directory of the real data test from the mc tests since the script for merging the h5 files was trying to merge everything within, so it failed.

Also I make the test passing expecting the names:
```
dl1_LST-1.1.Run02008.0000_first50.h5
muons_LST-1.Run02008.0000_first50.fits
datacheck_dl1_LST-1.Run02008.0000.h5
dl2_LST-1.1.Run02008.0000_first50.h5
```
but we probably should make sure that the naming of these files is coherent between them.

lstchain/reco/r0_to_dl1.py
```diff
@@ -449,8 +450,12 @@ def r0_to_dl1(
 
                 if not is_simu:
                     dl1_container.ucts_time = 0
-                    # Assumes R1 event source filled the "dragon" timestamp:
-                    dl1_container.dragon_time = event.trigger.time
+                    # Assumes R1 event source filled the "dragon" timestamp
+                    # as MJD; we convert back to unix timestamp:
```
maxnoe, line 454: Comment is wrong and code should be simpler.

Just do `dl1_container.dragon_time = event.trigger.time.unix`

lstchain/reco/r0_to_dl1.py
```diff
@@ -449,8 +450,12 @@ def r0_to_dl1(
...
+                    dl1_container.dragon_time = Time(event.trigger.time,
+                                                     format='mjd',
+                                                     scale='tai').to_value('unix')
```
morcuended, line 457: `trigger_time` keeps being in `mjd` format

lstchain/reco/r0_to_dl1.py
morcuended: Is the scale still treated properly?

lstchain/reco/r0_to_dl1.py
moralejo: > `trigger_time` keeps being in `mjd` format

No, that is the "input" of the Time() call. did you test it? For me it provides the same unix timestamps as the previous lstchain code.

lstchain/reco/r0_to_dl1.py
maxnoe: unix is UTC, unix_tai is TAI, but wrong without fixing it (like in ctapipe_io_lst) or using the not-yet releases astropy 4.2.1

lstchain/reco/r0_to_dl1.py
morcuended: I mean in the output dl1 file. Not here, sorry. It was just a side note that in the dl1 output the format is different for the two of them.

maxnoe: The input of the time call is already a Time object. It is not an mjd float. It is just convereted to an mjd float by the `HDF5TableWriter`

morcuended:
```
In [7]: dl1_df.dragon_time
Out[7]: 
0      1.582060e+09
1      1.582060e+09
2      1.582060e+09

In [8]: dl1_df.trigger_time
Out[8]: 
0      58897.877193
1      58897.877193
2      58897.877193

```

lstchain/reco/r0_to_dl1.py
moralejo: Ok, will test it

lstchain/reco/r0_to_dl1.py
maxnoe: @morcuended should be fixed now

moralejo: > ```
> In [7]: dl1_df.dragon_time
> Out[7]: 
> 0      1.582060e+09
> 1      1.582060e+09
> 2      1.582060e+09
> 
> In [8]: dl1_df.trigger_time
> Out[8]: 
> 0      58897.877193
> 1      58897.877193
> 2      58897.877193
> ```

Yes, I only modified dragon_time, not trigger_time. But can do both, is it used downstream?

maxnoe: @moralejo I already added the `.unix` also `for trigger_time`

morcuended: now it's fine. Shall test that the first event timestamps in the dl1 file is actually the timestamps in the night summary file?

maxnoe: Good test, but beware that is nanoseconds in TAI, right? And lstchain now stores UTC.

So probably the best is something like `np.isclose((Time(event_value, format='unix') - Time(summary_value / 1e9, format='unix_tai')).to_value(u.s), 0)`

maxnoe: But you probably have to use the fix in `ctapipe_io_lst`, I will just add it.

moralejo: > But you probably have to use the fix in `ctapipe_io_lst`, I will just add it.

What fix? I don't see any recent one in ctapipe_io_lst

moralejo: I checked on one run (2291.0024) that the dragon_time in DL1 files is the same for this new code and for lstchain's previous one.

moralejo: > I checked on one run (2291.0024) that the dragon_time in DL1 files is the same for this new code and for lstchain's previous one.

However trigger_time is different by 6 s....  Because in old files trigger_time is 6 s behind dragon_time. I do not think we were using that trigger_time at all, while dragon_time was used, and checked. So I think that having the same values in both fields after this PR is ok.

morcuended: we were not using `trigger_time` that I remember. If I remember correctly it was not calculated in the same place that `dragon_time`, `tib_time` and `ucts_time` but in the old `_io_lst` reader (?). Not sure right now where those 6 secs come from.

maxnoe: > What fix? I don't see any recent one in ctapipe_io_lst

This fix for astropy to use the correct epoch for `unix_tai`:

https://github.com/cta-observatory/ctapipe_io_lst/blob/82228781c91a697c92ace3cba024fea55123ec99/ctapipe_io_lst/event_time.py#L13-L16

maxnoe: @moralejo that's the same story again with using the run start timestamp as reference for the times:

https://github.com/cta-observatory/ctapipe_io_lst/blob/dc087286fd061d1d2093022b1d6a2e6abd2b9e60/ctapipe_io_lst/__init__.py#L418-L421

moralejo: > > However trigger_time is different by 6 s.... Because in old files trigger_time is 6 s behind dragon_time (with this new code they are identical).
> 
> @moralejo that's the same story again with using the run start timestamp as reference for the times:
> 
> https://github.com/cta-observatory/ctapipe_io_lst/blob/dc087286fd061d1d2093022b1d6a2e6abd2b9e60/ctapipe_io_lst/__init__.py#L418-L421

Ok, thanks! So we stick to what we now have in this PR. Bot times the same.

moralejo: Do all of you manage to locally run pytest successfully? I have everything up to date from this PR, and consistently get a couple of errors. Below is the first one.
I interrupted it once after the first error, and it seems the directory testfiles/scripts/real_data is not created at all.

```
 def test_dl1ab_realdata():
        output_real_data_dl1ab = output_dir_realdata / “dl1ab_realdata.h5"
run_program(“lstchain_dl1ab”, “-f”, real_data_dl1_file, “-o”, output_real_data_dl1ab)
lstchain/scripts/tests/test_lstchain_scripts.py:274:
args = (‘lstchain_dl1ab’, ‘-f’, PosixPath(‘testfiles/scripts/real_data/dl1_LST-1.1.Run02008.0000_first50.h5’), ‘-o’, PosixPath(‘testfiles/scripts/real_data/dl1ab_realdata.h5’))
result = CompletedProcess(args=(‘lstchain_dl1ab’, ‘-f’, PosixPath(’testfiles/scripts/real_data/dl1_LST-1.1.Run02008.0000_first5...undError: [Errno 2] No such file or directory: \‘testfiles/scripts/real_data/dl1_LST-1.1.Run02008.0000_first50.h5\‘\n’)
```

maxnoe: This is a combination of two things, @morcuended forgot to add the `@pytest.mark.private_data` to this test to mark it as one that needs the private test data.

And by default, the tests with private data are not run, so this tests fails because it depends on another one.

And we shouldn't just make tests depend on each other. We should use pytest fixtures for things we need to produce that are needed by other tests.

To really run all tests, run `pytest -m 'private_data or not private_data'`, this is what is run on the CI

maxnoe:
@morcuended can you add the missing `@pytest.mark.private_data` for the tests that need access to the private data?


And we shouldn't rely on this ordering (or even accidental ordering).

If a test needs a dl1 file as input, use a fixture.

morcuended:
> @morcuended can you add the missing `@pytest.mark.private_data` for the tests that need access to the private data?
> 

Done 

> And we shouldn't rely on this ordering (or even accidental ordering).
> 
> If a test needs a dl1 file as input, use a fixture.

Not sure how this works. It is just adding `@pytest.fixture` in all the tests that have dependant files? is the `(scope="session")` needed?

maxnoe:
@morcuende 

No, fixtures are the inputs to those tests. So if you have a test that needs a dl1 file, you would have a fixture for producing the dl1 file and then the test gets that as input:

`scope='session'` should be appropriate here, that means the fixture is created once per call to `pytest`. 

```
@pytest.fixture(scope='session')
def observed_dl1_file(tmp_path):
      output_path = tmp_path / "dl1_....h5"
      r0_to_dl1(r0_file, output=output_path, ...)
      return output_path

@pytest.mark.private_data
def test_observed_dl1_to_dl2(observed_dl1_file):
     # do something with observed_dl1_file
```

setup.py
rlopezcoto: can we update this one with the latest ctapipe version (v0.10.3) to avoid problems?

environment.yml
```diff
@@ -5,8 +5,9 @@ channels:
   - conda-forge
 dependencies:
   - python=3.7
+  - astropy>=4.1
   - pip
-  - ctapipe=0.8.0
+  - ctapipe>=0.10.1
```
rlopezcoto, line 10: same as for setup.py

setup.py
rlopezcoto: can we update this one with the latest ctapipe version (v0.10.3) to avoid problems?

environment.yml
rlopezcoto, line 10: same as for setup.py

lstchain/data/lstchain_standard_config.json
```diff
@@ -145,18 +141,22 @@
...
-      "window_width":12
+      "window_width":12,
+      "apply_integration_correction": false
      },
     "FixedWindowSum":{
-       "window_start": 12,
-       "window_width":12
+       "window_shift": 6,
+       "window_width":12,
+       "peak_index": 18,
```
rlopezcoto, line 158: where is this `peak_index` coming from?

lstchain/data/onsite_camera_calibration_param.json
```diff
@@ -1,49 +1,65 @@
...
+	   "console"
+	 ]  
+       }              
+     } 
+     }    
+    }, 
+   "LSTEventSource": {  
+    "allowed_tels": [1],
+    "LSTR0Corrections": {
+      "select_gain": false
```
rlopezcoto, line 30: why is the gain selection set to `false` by default?

setup.py
maxnoe: Yes, sure.

lstchain/data/onsite_camera_calibration_param.json
maxnoe: This is the calibration config, you need to calculate pedestals / ff for both gain channels

lstchain/data/lstchain_standard_config.json
maxnoe: ctapipe changed the options to the extractors, instead of window start / end  you now have peak_index / shift

rlopezcoto:
> @morcuende
> 
> No, fixtures are the inputs to those tests. So if you have a test that needs a dl1 file, you would have a fixture for producing the dl1 file and then the test gets that as input:
> 
> `scope='session'` should be appropriate here, that means the fixture is created once per call to `pytest`.
> 
> ```python
> @pytest.fixture(scope='session')
> def observed_dl1_file(tmp_path):
>       output_path = tmp_path / "dl1_....h5"
>       r0_to_dl1(r0_file, output=output_path, ...)
>       return output_path
> 
> @pytest.mark.private_data
> def test_observed_dl1_to_dl2(observed_dl1_file):
>      # do something with observed_dl1_file
> ```

I think this would in general be very useful, but as you know, this is not how the lstchain tests work at the moment: e.g. we use the dl1 files produced by the `dl0_to_dl1` scripts to test the `dl1_to_dl2` scripts. I would suggest to open an issue to solve the problems generated by all these test files and do a PR that removes this way of performing tests we currently have.

maxnoe:
@rlopezcoto Since we are just adding new tests here, let's directly do it correctly.

morcuended:
I'm struggling with this @maxnoe. What is the best way of dealing with returned tuples (eg output of r0 to dl1 script: muons, dl1, datacheck files)

```
@pytest.fixture(scope='session')
def observed_dl1_file(tmp_path):
      output_path = tmp_path / "dl1_....h5"
      r0_to_dl1(r0_file, output=output_path, ...)
      return output_path1, output_path2  
```

maxnoe:
> I'm struggling with this @maxnoe. What is the best way of dealing with returned tuples (eg output of r0 to dl1 script: muons, dl1, datacheck files)

You can just unpack the tuple or use another datastructure like a dict if it gets more complicated.
For your example, I would either just unpack the tuple or use a dict and call the fixture `observed_dl1_outputs` to make clear it's not just a file.

E.g. like this:
```
@pytest.fixture(scope='session')
def observed_dl1_outputs(tmp_path):
      output_path1 = tmp_path / "dl1_....h5"
      output_path2 = tmp_path / "muons_....fits"
      r0_to_dl1(r0_file, output=output_path, ...)
      return {'dl1': output_path1, 'muon': output_path2, 'datacheck': ..., }


def test_foo(observed_dl1_outputs):
    # do something with observed_dl1_outputs["dl1"] 
    # do something with observed_dl1_outputs["muon"] 
```

moralejo:
The new DL1 output has a 'trigger_type' in the parameters table which is the one filled by the event source in event.trigger.type, i.e following CTA's convention for the meaning of the bits. This may be quite confusing (ucts_trigger_type has LST's convention)

In the former DL1 files trigger_type was simply the TIB type wit the usual LST meaning (i.e. 32=pedestal). I suggest that we fill dl1_container.trigger_type with lst.evt.tib_masked_trigger in https://github.com/cta-observatory/cta-lstchain/blob/e318da79bd07ac6ec0d50ac995c8560ba41c4893/lstchain/reco/r0_to_dl1.py#L464

Which I think would recover the old behaviour. Any objection?

morcuended:
> The new DL1 output has a 'trigger_type' in the parameters table which is the one filled by the event source in event.trigger.type, i.e following CTA's convention for the meaning of the bits. This may be quite confusing (ucts_trigger_type has LST's convention)
> 
> In the former DL1 files trigger_type was simply the TIB type wit the usual LST meaning (i.e. 32=pedestal). I suggest that we fill dl1_container.trigger_type with lst.evt.tib_masked_trigger in
> 
> https://github.com/cta-observatory/cta-lstchain/blob/e318da79bd07ac6ec0d50ac995c8560ba41c4893/lstchain/reco/r0_to_dl1.py#L464
> 
> Which I think would recover the old behaviour. Any objection?

Isn't it the same now?

image

maxnoe:
@morcuended no, `event_type` != `trigger_type`, see the R1 reference document. `event.trigger.event_type` follows the R1 event type codes, trigger_type is defined differently.

moralejo:
I am comparing image parameters (last released lstchain vs. this PR) for real data run 2291.0024. Everything looks almost identical, except time gradient and intercept which are correlated to the old ones, but have significant differences. Considering that charges and times at the single pixel level hardly ever differ, this is probably due to a change in ctapipe's calculation. Anyone knows for sure whether that is an improvement?  

@vuillaut: could you try it on MC, where we can check whether the new time gradient correlates better with impact parameter?

moralejo:
> 
> Isn't it the same now?
> 
> image

What you show is what it looks with this PR, right? It is not the same in the latest lstchain DL1 production.

maxnoe:
> Everything looks almost identical, except time gradient and intercept which are correlated to the old ones, but have significant differences. 

That will be the change of the fitting algorithm used from the siegel slopes to the LTS least squares fit.

moralejo:
> @moralejo this is not a question of LST vs. CTA definitions. Just of the (not completely clear to me why necessary) distinction between Event Type and Trigger Type.

Ok, but in this PR I think we should keep the old lstchain behaviour, I think.  When we move to stage1 we can start using the event type.

maxnoe:
@moralejo agreed!

morcuended:
> What you show is what it looks with this PR, right? It is not the same in the latest lstchain DL1 production.

Yes, right. A file produced with this PR. What do you mean with is not the same? I would expect the same outcome.

maxnoe:
@morcuended You asked "Isn't it the same now" for a plot that clearly shows it is in fact not the same. "Isn't it the same?" is normally understood as a rhethorical question implying it is the same.

morcuended:
You are right, I did not realize that both trigger_types are kind of swapped (?). Just wanted to check quickly that were trigger_types 1, 32, 42 ...

morcuended:
Actually, there is a new trigger_type = 2 that was not there before.

maxnoe:
@morcuended Again, trigger type is not the same as event type. The values differ.

Please see sections A.4.1, A.4.2 and A.4.3 of the R1 data model. You are comparing trigger type to event type values here.

morcuended:
OK I see now, sorry for the noise

moralejo:
> 
> @moralejo
> 
> I think we should at least add a column for event type, otherwise the tagging of flat field events will get lost at the dl1 stage.

Ah, the event source performs now the flat-field event recognition and stores it in event_type, right?
Let's add it, then, after the merge I can adapt the DL1 data check to use it.

moralejo:
> That will be the change of the fitting algorithm used from the siegel slopes to the LTS least squares fit.
> 
> That was introduced in ctapipe 0.9: https://github.com/cta-observatory/ctapipe/releases/tag/v0.9.0

Thanks! So the new version takes care of excluding outliers, which is good I think. No reason to suspect any problem then, as I said there is clear correlation between old and new results, and the differences are certainly due to this different approach.

maxnoe:
> Thanks! So the new version takes care of excluding outliers, which is good I think

The version before did as well, but with a different algorithms that was much, much slower.

moralejo:
I messed up with my latest commit, sorry. Fixing it...

moralejo:
FYI:  CI tests not passing my fault... will solve it soon.

lstchain/calib/camera/calibrator.py
```diff
@@ -38,6 +39,12 @@ class LSTCameraCalibrator(CameraCalibrator):
         help='Path to drs4 time calibration file'
     ).tag(config=True)
 
+    time_sampling_correction_path = Unicode(
```
maxnoe, line 42: This should be a `Path`

moralejo:
> > Thanks! So the new version takes care of excluding outliers, which is good I think
> 
> The version before did as well, but with a different algorithms that was much, much slower.

I am not convinced about the time gradient.
This is how it looks for intensities between 50 and 300:
image

And this is for intensity>300:
image


The first one worries me a bit, the correlation is quite loose...
I think we should understand it better before the release.
@maxnoe is this time fit also done with the main island alone?

maxnoe:
@moralejo I don't know what lstchain does. But which mask lstchain uses  will not have changed by changing ctapipe versions.

morcuended:
I have changed all the tests to use fixtures:
- I defined them in `conftest.py` when needed to be used across the entire repository.
- I didn't use any scope for the fixtures because it was not working for me locally.
- I use the built-in fixture `tmp_path` instead of defining by hand a test_dir. However, I would say it is convenient to have several common temporal paths so they can be shared between several tests since some scripts take a directory as input instead of the files themselves. I was not sure about which is the best way to implement this so I just returned the `tmp_path` from a given fixture and use it as input for another test when necessary.

@maxnoe, please check that everything I changed makes sense or if something can be improved.

moralejo:
I did a quick check of the time gradient: plotting its absolute value vs. impact parameter for point-like MC gammas. In the plots below the left one is for DL1 produced with the new code (this PR) and the right one is with the original code (standard MC prod, taken from DL2 files but that should not matter). The test is done with 3 MC runs.

This is for all parametrized  events:
time_gradient

And this is for intensity>200 p.e.:
time_gradient_intensity_gt_200

I'd say the correlation looks pretty similar in both cases - if anything, slightly better (narrower distribution) for the new code. I checked all other parameters, and they are practically identical to the original ones.

So I think from this point of view we can move forward with merging this PR.

moralejo:
@FrancaCassol can we set it ready for review?

morcuended:
> > > Thanks! So the new version takes care of excluding outliers, which is good I think
> > 
> > 
> > The version before did as well, but with a different algorithms that was much, much slower.
> 
> I am not convinced about the time gradient.
> This is how it looks for intensities between 50 and 300:
> image
> 
> And this is for intensity>300:
> image
> 
> The first one worries me a bit, the correlation is quite loose...
> I think we should understand it better before the release.
> @maxnoe is this time fit also done with the main island alone?

Could it be due to the fact of applying loose cleaning levels?

rlopezcoto:
> I did a quick check of the time gradient: plotting its absolute value vs. impact parameter for point-like MC gammas. In the plots below the left one is for DL1 produced with the new code (this PR) and the right one is with the original code (standard MC prod, taken from DL2 files but that should not matter). The test is done with 3 MC runs.
> 
> This is for all parametrized events:
> time_gradient
> 
> And this is for intensity>200 p.e.:
> time_gradient_intensity_gt_200
> 
> I'd say the correlation looks pretty similar in both cases - if anything, slightly better (narrower distribution) for the new code. I checked all other parameters, and they are practically identical to the original ones.
> 
> So I think from this point of view we can move forward with merging this PR.

what about for real data, did you also get similar results?

moralejo:
> > I'd say the correlation looks pretty similar in both cases - if anything, slightly better (narrower distribution) for the new code. I checked all other parameters, and they are practically identical to the original ones.
> > So I think from this point of view we can move forward with merging this PR.
> 
> what about for real data, did you also get similar results?

I wish I could know the true impact parameter in real data!  ;-)

rlopezcoto:
> I wish I could know the true impact parameter in real data! ;-)

:-D

> The "old vs. new" comparison some pages above, which made me suspicious, is for real data.

yeah, this is what I meant, if I'm not wrong in the past we had some issues with the time gradient on real data and I think they were at some point related to the cleaning as @morcuended was mentioning, although the time gradient calculation should not have changed...

moralejo:
> 
> Could it be due to the fact of applying loose cleaning levels?

They should be the same as in the previous standard production, so I don't think so. In any case, we will go for completely new cleaning. And the test on MC makes me confident that there is no fundamental problem with the new code.

moralejo:
> yeah, this is what I meant, if I'm not wrong in the past we had some issues with the time gradient on real data and I think they were at some point related to the cleaning as @morcuended was mentioning, although the time gradient calculation should not have changed...

Now it has changed (algorithm and details on how outliers are dealt with), so we cannot aim at perfect matching. With real data we cannot test it easily, let's just see how things look for Crab in the next reanalysis.

maxnoe:
> although the time gradient calculation should not have changed...

As discussed above, it did change! We implemented a much faster fitting method that still handles outliers. It is now actually the same algorithm (at least in name, didn't check the implementation) that MARS and EventDisplay use.

jsitarek:
Hi @FrancaCassol 

following the other PR #561 I'm working on updating the time_resolution macro to match the new reader, using as an example r0_to_dl1 macro, but I cannot find the place where LSTR0Corrections are being executed. 
the settings are mentioned in source_config variable, but the method is not explictely executed in this macro, and 
I do not see it explicitely run neither in ctapipe/io/eventsource.py nor in calib/camera/calibrator.py

maxnoe:
@jsitarek The code is in `ctapipe_io_lst/calibration.py` and `ctapipe_io_lst/__init__.py`:

https://github.com/cta-observatory/ctapipe_io_lst/blob/89eb21d7ba4c812676012a18faf1b910fb342ba1/ctapipe_io_lst/__init__.py#L345-L363

jsitarek:
> @jsitarek The code is in `ctapipe_io_lst/calibration.py` and `ctapipe_io_lst/__init__.py`:
> 
> https://github.com/cta-observatory/ctapipe_io_lst/blob/89eb21d7ba4c812676012a18faf1b910fb342ba1/ctapipe_io_lst/__init__.py#L345-L363
> 
> It is part of the `LSTEventSource`.

Hi @maxnoe thanks for very quick reply
it seems to me that this is really a strange place to put this code. why you didn't put this large code into a separate file?

FrancaCassol:
> Hi @FrancaCassol
> 
> following the other PR #561 I'm working on updating the time_resolution macro to match the new reader, using as an example r0_to_dl1 macro, but I cannot find the place where LSTR0Corrections are being executed.
> the settings are mentioned in source_config variable, but the method is not explictely executed in this macro, and
> I do not see it explicitely run neither in ctapipe/io/eventsource.py nor in calib/camera/calibrator.py

Hi @jsitarek,
that part of the code is already working, what do you want to update?

maxnoe:
@jsitarek providing the `LSTEventSource` is the only job of `ctapipe_io_lst`, so there is in principal nothing wrong with having the class in `__init__.py`. Why add another layer of indirection?

jsitarek:
Hi

@maxnoe I was just surprised because I saw only one other case of defining class in `__init__.py`  in  ctapipe or lstchain (and it that case if was a kind of dummy class, while LSTEventSource is very bulky) 

@FrancaCassol - since the corrections are executed there is no change requested from my side, I only wanted to make sure that this is the case, since as I explained earlier I couldn't trace it in the code since I did not expect to find it in `__init__.py`

lstchain/calib/camera/flatfield.py
```diff
@@ -36,19 +35,18 @@ class FlasherFlatFieldCalculator(FlatFieldCalculator):
...
         [-3, 3],
         help='Interval (number of std) of accepted charge standard deviation around camera median value'
     ).tag(config=True)
-    time_calibration_path = Unicode(
-        None,
-        allow_none = True,
-        help = 'Path to drs4 time calibration file'
-    ).tag(config = True)
+    time_cut_outliers = List(
+        [0, 60], help="Interval (in waveform samples) of accepted time values"
```
maxnoe, line 43: Why do we except 20 later values but not previous values? Shouldn't it be symmetric around the expected range of [0, 40] ?

FrancaCassol: > @FrancaCassol - since the corrections are executed there is no change requested from my side, I only wanted to make sure that this is the case, since as I explained earlier I couldn't trace it in the code since I did not expect to find it in `__init__.py`

I see, ok

FrancaCassol: > Why do we except 20 later values but not previous values? Shouldn't it be symmetric around the expected range of [0, 40] ?

yes, you are right, this default value is still from ctapipe (chosen compatible also with nectarcam)
The value is set to [2,38] in the lstchain calibration config, which is also without real effect. Actually, the outlier cuts are still very loose for the moment

lstchain/calib/camera/pedestals.py
```diff
@@ -41,6 +42,12 @@ class PedestalIntegrator(PedestalCalculator):
         help='Interval (number of std) of accepted charge standard deviation around camera median value'
     ).tag(config=True)
 
+    time_sampling_correction_path = Path(
+        directory_ok=False,
```
maxnoe, line 46: add `exists=True` to automatically check if the given path exists.

lstchain/calib/camera/pulse_time_correction.py
```diff
@@ -91,6 +90,35 @@ def get_corr_pulse(self, event, pulse):
                                 self.n_capacitors)
         return pulse_corr
 
+    def get_pulse_time_corrections(self, event):
```
maxnoe, line 93: Isn't this the same code that is now available in `ctapipe_io_lst`? Can't we just import it here if it is needed?

maxnoe: E.g. this is the same as `ctapipe_io_lst.calibration.calc_drs4_time_correction_both_gains`, right?

lstchain/calib/camera/tests/test_pedestals.py
```diff
@@ -26,53 +27,32 @@ def test_pedestal_calculator():
...
         if ped_calculator.calculate_pedestals(data):
             assert data.mon.tel[tel_id].pedestal
             assert np.mean(data.mon.tel[tel_id].pedestal.charge_median) == (
-                ped_calculator.extractor.window_width.tel[0] * ped_level
+                ped_calculator.extractor.window_width.tel[tel_id] * ped_level
             )
             assert np.mean(data.mon.tel[tel_id].pedestal.charge_std) == 0
 
 
-def test_calc_pedestals_from_traces():
```
maxnoe, line 54: Why remove the test?

lstchain/calib/camera/time_sampling_correction.py
```diff
@@ -0,0 +1,118 @@
...
+
+        n_gains = 2
+        n_modules = 265  # number of modules in LST's camera.
+        n_pixel_per_module = 7  # number of pixels per one module.
+        size_drs4 = 1024
+        n_pixels = 1855
+        roi = 36
+
+        # shift the first capacitor to the value used in the r1
+        r1_roi_start = 3
```
maxnoe, line 69: Is this the same as `LSTR0Corrections.r1_sample_start`? If yes, what happens if someone uses a different value of three for that traitlet? Why is this hard coded here?

lstchain/conftest.py
```diff
@@ -15,11 +20,97 @@ def pytest_configure(config):
...
+    return dl1_proton_file
+
+
+@pytest.fixture(scope="session")
+def rf_models(temp_dir_simulated_files, simulated_dl1_file):
+    """Produce test random forest models."""
+    gamma_file = simulated_dl1_file
+    proton_file = simulated_dl1_file
+    models_path = temp_dir_simulated_files
+    file_model_energy = models_path / "reg_energy.sav"
```
maxnoe, line 104: These are pickled models, right? Why then not use the `.pickle` extension?

lstchain/calib/camera/pulse_time_correction.py
FrancaCassol: yes, this code is not more used and  can be removed. As said above, for the moment I did not remove the code in lstchain but if we agree we can do it.

lstchain/calib/camera/tests/test_pedestals.py
FrancaCassol: this is a test of a code that is in ctapipe, not needed here

lstchain/calib/camera/pulse_time_correction.py
maxnoe: I think we should to make sure we move to the new code everywhere where needed and be safe that nothing is still using the old code.

lstchain/calib/camera/tests/test_pedestals.py
maxnoe: ok

FrancaCassol: the test is already there

lstchain/calib/camera/time_sampling_correction.py
FrancaCassol: > Is this the same as `LSTR0Corrections.r1_sample_start`? 

yes

> If yes, what happens if someone uses a different value of three for that traitlet? 

it would be a bad choice ;-)

> Why is this hard coded here?

now we have all these hardcoded values in  the source calibrator. Are there objections to move this code there? (one could be that these corrections are more offline than online)

maxnoe: > it would be a bad choice ;-)

If this value should never be different from 3, then we shouldn't make it a traitlet.

FrancaCassol: Let say that this value must be the same that one use for the drs4 calibration

maxnoe:
@jsitarek I don't think it is good to include a large new notebook here.

We should focus on making and validating the existing code, not adding new functionality.

jsitarek:
@maxnoe @FrancaCassol 

come on, first Franca tells me to remove the notebook that was working with the old reader for a separate PR and redo it to use the new reader, and now that the thing is done, you tell me not to do it ?
so where should I put it ?
it does not interfere with the rest of the code

lstchain/conftest.py
morcuended: Do you mean just modify the filenames to `file_model_energy = models_path / "reg_energy.pickle"` an so on?

maxnoe: yes. But this is a really minor thing.

maxnoe:
@jsitarek Adapting it to the new reader is something else then including it directly in this PR.

This PR was basically finished and waiting for a final review / approval. Now you added a new notebook with 4000 lines of code that have to be reviewed again.

It would have been much easier to wait a day or two until this PR was merged and open a new one.
Or even already open a new one but marking it as draft.

And we cannot simply look at it on github because the diff is too large.

lstchain/scripts/tests/test_lstchain_scripts.py
morcuended: Is this `cwd` option still needed? Or it is just enough with the temporal directory fixtures.

maxnoe: I think it's fine like it is.

lstchain/conftest.py
morcuended: Yes, sure. I can change them. Not sure I get the points of this `.pickle` extension. It is a file that you open, dump something in it or modify it and afterward, you can recover the original content for using it in another test for example?

maxnoe: It is just about using the correct extension for the correct filetype. We also use `.h5` for the temporary hdf5 files ;)

maxnoe: See: https://stackoverflow.com/a/40433504/3838691

rlopezcoto:
> @jsitarek Adapting it to the new reader is something else then including it directly in this PR.
> 
> This PR was basically finished and waiting for a final review / approval. Now you added a new notebook with 4000 lines of code that have to be reviewed again.
> 
> And we even cannot simply look at it on github because the diff is too large.
> 
> It would have been much easier to wait a day or two until this PR was merged and open a new one.
> Or even already open a new one but marking it as draft.

@maxnoe given that this is only a notebook that will not interfere with the functionality of the code, I suggest that now that it has been included an #561 already closed, to leave it here and get it merged, is that ok with you?
@jsitarek and @FrancaCassol for the future it would be nice if we could try to limit the length of PRs and separate the inclusion of new functionality into smaller pieces that are easier to review and also easier to references in the future

jsitarek:
> @jsitarek Adapting it to the new reader is something else then including it directly in this PR.
> 

@FrancaCassol  explicitely told me to commit it to this branch that she created:

"Concerning the notebook, could you perhaps update it to work with the new reader (ctapipe_io_lst version 0.6.4) and push it in the lstchain branch "newsreader_compatible" (where we are all push the new reader compatible code)?"

> This PR was basically finished and waiting for a final review / approval. Now you added a new notebook with 4000 lines of code that have to be reviewed again.
> 
this is a notebook, so it does not affect the rest of the code, and it is also a new one that I thought could be useful to people because of the discussions about the time resolution requirement with different correction methods. 
No idea where you took "4000 lines of code", the notebook has only ~300 lines of code (there are some text outputs but this is not something that you would be reviewing. 

Nevertheless, I understand your point and I do not want to delay this PR, so I removed the notebook from it. 

> And we even cannot simply look at it on github because the diff is too large.

you could open it in git, just clicking on "view" 

> 
> It would have been much easier to wait a day or two until this PR was merged and open a new one.
> Or even already open a new one but marking it as draft.

lstchain/conftest.py
rlopezcoto: we are currently using `.sav` in several places in the repo:
https://github.com/cta-observatory/cta-lstchain/search?q=.sav

let's try to be consistent (either we change all or leave `.sav` here for the time being)

moralejo:
Sorry for the inconvenience @jsitarek, although you are right that the notebook won't interfere with anything, I think it is better to have it in a different PR, because this one became too complex already.

lstchain/conftest.py
morcuended: I was going to say that. I've always seen `.sav` extension so I did not understand at the beginning.

FrancaCassol:
> @FrancaCassol explicitely told me to commit it to this branch that she created:
> 
> "Concerning the notebook, could you perhaps update it to work with the new reader (ctapipe_io_lst version 0.6.4) and push it in the lstchain branch "newsreader_compatible" (where we are all push the new reader compatible code)?"

yes @jsitarek , sorry for that, it is my mistake. I think @maxnoe is right, we want to merge this PR as soon as possible. We need to have the notebook working with the new reader, but it can definitely wait for a new PR. I apologise.

morcuended:
After updating to ctapipe_io_lst v0.6.5 tests are not passing, the r0_to_dl1 script fails.

moralejo:
> After updating to ctapipe_io_lst v0.6.5 tests are not passing, the r0_to_dl1 script fails.

We're looking into this. Seems related to a merge of source-dependent analysis code.

rlopezcoto:
> After updating to ctapipe_io_lst v0.6.5 tests are not passing, the r0_to_dl1 script fails.

In the last `master` merging, conflicts were not properly solved and two `test_add_source_dependent_parameters` functions were left.

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -66,7 +66,7 @@ def simulated_dl1ab(temp_dir_simulated_files, simulated_dl1_file):
     )
     return output_file
 
-
+@pytest.fixture(scope="session")
 def test_add_source_dependent_parameters(simulated_dl1_file):
```
maxnoe, line 70: A test is not a fixture. With adding this, you just made that the test is not actually run.

A fixture is an input for a test. Let me check why the test is failing.

I think the test is actually good and needed and we need to figure out why it failed

rlopezcoto: sorry, the fixture is wrong, the test was running without problems

maxnoe: great!

rlopezcoto: The problem came from:
> In the last master merging, conflicts were not properly solved and two test_add_source_dependent_parameters functions were left.

maxnoe: Yes, sorry, that was me doing a not trivial merge using the web interface. Should have checked locally.

maxnoe:
If you agree that this branch is very close to being merged, I will deactivate the requirement on "python 3.7 / ctapipe 0.8" and "python 3.6 / ctapipe 0.8" and switch to requiring the build for "python 3.7 / ctapipe 0.10"

lstchain/calib/camera/calibrator.py
```diff
@@ -1,14 +1,15 @@
 import numpy as np
```
maxnoe, line 1: @FrancaCassol Shall we just remove the old calibrator here?

lstchain/calib/camera/pedestals.py
maxnoe: Why resolved but no changes made?

lstchain/calib/camera/pedestals.py
```diff
@@ -226,9 +241,9 @@ def calculate_time_results(
 ):
     """Calculate and return the sample time"""
     return {
```
maxnoe, line 243: Use of astropy units here looks off. Should probably be `to(u.s)` instead of `.value * u.s`

lstchain/calib/camera/time_sampling_correction.py
maxnoe: So it should be an argument to the function and the caller should give `LSTR0Corrections.r1_sample_start` to make sure this is always correct.

lstchain/conftest.py
maxnoe: Ah ok, then let's not change this here and leave that for another time.

lstchain/scripts/lstchain_data_create_time_calibration_file.py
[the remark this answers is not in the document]
maxnoe: The new version of `ctapipe_io_lst` will always read the night summary columns as int now and will use a masked array if there are any nans. 

So this command line tool will need to pass the correct options depending on whether or not these counters are available.

Currently the `EventTimeCalculator` will only work if ucts and tib are available. I am working on enabling to relaxing this to only requiring that UCTS is available and making tib optional. (Is there any reason to use TIB over Dragon? Dragon is always available, right?).

The problem remaining is how to handle data with missing UCTS, since no absolute timestamps are possible to my knowledge without UCTS. 

Currently the `LSTEventSource.fill_timestamp` traitlet can be set to false, which will result in all timestamps being the same invalid time marker. 

If code does not need absolute times, that would be an option.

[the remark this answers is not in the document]
maxnoe: I can work on this stuff, but I don't know what to do when.

Please give input here: https://github.com/cta-observatory/ctapipe_io_lst/issues/71

rlopezcoto:
> If you agree that this branch is very close to being merged, I will deactivate the requirement on "python 3.7 / ctapipe 0.8" and "python 3.6 / ctapipe 0.8" and switch to requiring the build for "python 3.7 / ctapipe 0.10"

sounds good @maxnoe, please go ahead and let's try to get this merge as soon as possible.

maxnoe:
@rlopezcoto done.

moralejo (approved):
I'd say, lets merge and try, if we encounter problems, or need new features, we'll adress them in new PRs of a manageable size, ok?

rlopezcoto (approved):

maxnoe:
@rlopezcoto @moralejo agreed

maxnoe (approved):

PR #589, opened 2021-02-10
Filters on mc_type for RF

vuillaut:
Issue:
- Some of the parameters are initialized with a default value = -1.
- Even in case of issues with the computation of these parameters, they can be used for training without causing issues (the values are valid predictions)
- Training and inference would not raise any error and warning but make useless predictions

I encountered the problem when using `dl1ab`, the `mc_type` would be set to -1 for some events and thus the RF would predict 3 classes and the gammaness would not be assigned to the right class.
So I also add an error raise in case of more than 2 predicted classes, which should never happen in current setup.

### DIFF
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -95,14 +95,13 @@ class DL1ParametersContainer(Container):
     mc_core_x = Field(None, 'Simulated impact point x position', unit=u.m)
     mc_core_y = Field(None, 'Simulated impact point y position', unit=u.m)
     mc_h_first_int = Field(None, 'Simulated first interaction height', unit=u.m)
-    mc_type = Field(-1, 'Simulated particle type')
+    mc_type = Field(-9999, "MC shower primary ID 0 (gamma), 1(e-),"
+                           "2(mu-), 100*A+Z for nucleons and nuclei,"
+                           "negative for antimatter.")
     mc_az_tel = Field(None, 'Telescope MC azimuth pointing', unit=u.rad)
     mc_alt_tel = Field(None, 'Telescope MC altitude pointing', unit=u.rad)
     mc_x_max = Field(None, "MC Xmax value", unit=u.g / u.cm**2)
     mc_core_distance = Field(None, "Distance from the impact point to the telescope", unit=u.m)
-    mc_shower_primary_id = Field(None, "MC shower primary ID 0 (gamma), 1(e-),"
-                                    "2(mu-), 100*A+Z for nucleons and nuclei,"
-                                    "negative for antimatter.")
 
     hadroness = Field(None, "Hadroness")
     wl = Field(u.Quantity(np.nan), "width/length")
--- a/lstchain/io/lstcontainers.py
+++ b/lstchain/io/lstcontainers.py
@@ -68,14 +68,14 @@ class DL1ParametersContainer(Container):
                                            'border pixels')
     leakage_pixels_width_2 = Field(np.nan, 'Fraction of signal pixels that are '
                                            'in the two outermost rings of pixels')
-    n_pixels = Field(np.nan, 'Number of pixels after cleaning')
+    n_pixels = Field(-1, 'Number of pixels after cleaning')
     concentration_cog = Field(np.nan, 'Fraction of intensity in three pixels '
                                       'closest to the cog')
     concentration_core = Field(np.nan, 'Fraction of intensity inside hillas '
                                        'ellipse')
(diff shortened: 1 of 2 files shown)


rlopezcoto (dismissed):

vuillaut:
I did not foresee the writing type issue... I will try and think about an elegant solution but suggestions are welcome.

moralejo:
> I did not foresee the writing type issue... (these parameters are requested to be integers and `nan` is a float)
> I will try and think about an elegant solution but suggestions are welcome.

Then the default necessarily has to be an integer outside of the physical range, right? And filter those before passing the events to the training (and to the application of the trained models)

lstchain/io/lstcontainers.py
```diff
@@ -68,14 +68,14 @@ class DL1ParametersContainer(Container):
...
-    n_pixels = Field(0, 'Number of pixels after cleaning')
+    n_pixels = Field(np.nan, 'Number of pixels after cleaning')
     concentration_cog = Field(np.nan, 'Fraction of intensity in three pixels '
                                       'closest to the cog')
     concentration_core = Field(np.nan, 'Fraction of intensity inside hillas '
                                        'ellipse')
     concentration_pixel = Field(np.nan, 'Fraction of intensity in brightest '
                                         'pixel')
-    n_islands = Field(0, 'Number of Islands')
+    n_islands = Field(np.nan, 'Number of Islands')
```
maxnoe, line 78: Use -1, since n_island is a positive integer.

lstchain/io/lstcontainers.py
```diff
@@ -95,7 +95,7 @@ class DL1ParametersContainer(Container):
     mc_core_x = Field(None, 'Simulated impact point x position', unit=u.m)
     mc_core_y = Field(None, 'Simulated impact point y position', unit=u.m)
     mc_h_first_int = Field(None, 'Simulated first interaction height', unit=u.m)
-    mc_type = Field(-1, 'Simulated particle type')
+    mc_type = Field(np.nan, 'Simulated particle type')
```
maxnoe, line 98: same here, -1 is ok, particle id is also a positive integer.

vuillaut: > > I did not foresee the writing type issue... (these parameters are requested to be integers and `nan` is a float)
> > I will try and think about an elegant solution but suggestions are welcome.
> 
> Then the default necessarily has to be an integer outside of the physical range, right? And filter those before passing the events to the training (and to the application of the trained models)

Yes that was basically how it was implemented.
`nan` has the advantage to always be outside the physical range and does not need extra care.
But yes, with the type issue, I will introduce some more precise filtering...

vuillaut:
> > I did not foresee the writing type issue... (these parameters are requested to be integers and `nan` is a float)
> > I will try and think about an elegant solution but suggestions are welcome.
> 
> Then the default necessarily has to be an integer outside of the physical range, right? And filter those before passing the events to the training (and to the application of the trained models)

Yes that was basically how it was implemented.
nan has the advantage to always be outside the physical range and does not need extra care.
But yes, with the type issue, I will introduce some more precise filtering...

lstchain/io/lstcontainers.py
vuillaut: actually I think that in the corsika nomenclatura, negative integers are possible and meant for anti-particles (-1 would be positrons then)
maybe I'll rather go for -9999 just in case ;-)

maxnoe: No, corsika uses positive numbers:

From the user's guide:
Screenshot_20210210_212528

vuillaut: :o 
thanks!
I would have sworn I had seen negative values in the past, my mistake!

maxnoe: Sorry, triple checking this, I found that simtel does not use the CORSIKA codes anymore.

From `io_hess.h`:

```
   int primary_id;      ///< Particle ID of primary. Was in CORSIKA convention
                        ///< where detector_prog_vers in MC run header was 0,
                        ///< and is now 0 (gamma), 1(e-), 2(mu-), 100*A+Z
                        ///< for nucleons and nuclei, negative for antimatter.
```

maxnoe: So this should be 0 for gamma, 101 for proton.

maxnoe: Yes, verified for our current simulations. 0 = gamma, 101 = proton, negative numbers possible.

vuillaut: Thanks for the counter check!
I'll update accordingly.

moralejo (approved):

PR #590, opened 2021-02-11
[WIP] Adding tests for real data scripts

morcuended:
I started to implement some tests for the real data scripts:

- r0_to_dl1
- dl1ab
- dl1_to_dl2

I have one doubt, should be there one test function for mc and real data or is it better to split them?

### DIFF
--- a/lstchain/scripts/tests/test_lstchain_scripts.py
+++ b/lstchain/scripts/tests/test_lstchain_scripts.py
@@ -10,6 +10,7 @@ import pytest
 from lstchain.io.io import dl1_params_lstcam_key
 from lstchain.io.io import dl1_params_src_dep_lstcam_key
 from lstchain.tests.test_lstchain import test_dir, mc_gamma_testfile, produce_fake_dl1_proton_file, fake_dl1_proton_file
+from pathlib import Path
 
 output_dir = os.path.join(test_dir, 'scripts')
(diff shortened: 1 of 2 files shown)


morcuended:
What I mean with splitting test functions into mc or real data is for example. Should we have these two function together?:

```
@pytest.mark.run(after='test_dl1ab')
def test_dl1ab_validity():
    dl1 = pd.read_hdf(os.path.join(output_dir, 'dl1_gamma_test_large.h5'), key=dl1_params_lstcam_key)
    dl1ab = pd.read_hdf(os.path.join(output_dir, 'dl1ab.h5'), key=dl1_params_lstcam_key)
    np.testing.assert_allclose(dl1, dl1ab, rtol=1e-4, equal_nan=True)


@pytest.mark.run(after='test_dl1ab_realdata')
def test_dl1ab_realdata_validity():
    output_real_data_dl1ab = real_data_dl1_path / 'dl1ab.h5'
    dl1 = pd.read_hdf(real_data_dl1_file, key=dl1_params_lstcam_key)
    dl1ab = pd.read_hdf(output_real_data_dl1ab, key=dl1_params_lstcam_key)
    np.testing.assert_allclose(dl1, dl1ab, rtol=1e-4, equal_nan=True)
```

morcuended:
Done in #587

PR #591, opened 2021-02-12
Fix ci build

maxnoe:
New pip version does not allow building wheels with "unknown" version.

This makes sure we have the git tags so version is known.

See build failure in #552


maxnoe:
I think there was a problem with having lstchain in the credentials. It seemed it now blocked lstchain anywhere in the build.
So password is still secret, but username is not.

PR #594, opened 2021-02-18
Drs4 pedestal failing pixels

pawel21:
I have modified script to create drs4 pedestal file to allow create file, when some modules are failing.
Failing module will have NaN pedestal value , and failing module id will stored in `2` field: 
`
f1 = fits.open("pedestal_3670.fits");                                 
f1[2].data  
`
cc: @morcuended , @moralejo , @jsitarek , @rlopezcoto


maxnoe:
@pawel21 The pedestal values until now were integers. NaN is only avaiable for floats. Did you change the dtype of the pedestal in the files?

lstchain/calib/camera/drs4.py
```diff
@@ -83,11 +84,14 @@ def _fill_pedestal_event_jit(waveform, expected_pixel_id, first_cap_array,
                             numped[gain, pixel, posads] += 1
 
     def finalize_pedestal(self):
-        if np.sum(self.numped==0) > 0:
-            raise RuntimeError("Not enough events to coverage all capacitor. "
-                               "Please use more events to create pedestal file.")
+        self.meanped = self.meanped / self.numped
+        pixels_with_nan_value = np.unique(np.where(np.isnan(self.meanped) == True)[1])
```
maxnoe, line 88: `== True` not needed, `np.isnan` already returns an array of booleans.

DirkHoffmann:
We are talking about the "DRS4 corrections", used as input for the baseline corrections in the EVB, correct?

> I have modified script to create drs4 pedestal file to allow create file, when some modules are failing.
> Failing module will have NaN pedestal value , and failing module id will stored in `2` field:
> `f1 = fits.open("pedestal_3670.fits"); f1[2].data `
> cc: @morcuended , @moralejo , @jsitarek , @rlopezcoto

Such a change would be incompatible with the EVB ICD. You need to define a value that is applied by the EVB for _all_ (operating) pixels. Besides the very valid remark by @maxnoe 

> @pawel21 The pedestal values until now were integers. NaN is only avaiable for floats. Did you change the dtype of the pedestal in the files?

this is another reason to reject this PR.

maxnoe:
@pawel21 I also just saw that you made very many, unrelated changes. E.g. deleting the whole `lstchain/io/io.py`.

To make the tests pass, you will also need to push a new branch to this repository (and not use your fork anymore, since it does not have access to the private data needed for observed data tests)

lstchain/calib/camera/drs4.py
maxnoe: and indices to arrays are by definition unique. I think you mean something different here.

How about checking that at least one of the gains is nan:

```
pixels_with_nan_value, = np.where(np.isnan(meanped).any(axis=0))  # any of the gain channels has a nan
```

moralejo:
> > Failing module will have NaN pedestal value , and failing module id will stored in `2` field:
> > `f1 = fits.open("pedestal_3670.fits"); f1[2].data `
> > cc: @morcuended , @moralejo , @jsitarek , @rlopezcoto
> 
> Such a change would be incompatible with the EVB ICD. You need to define a value that is applied by the EVB for _all_ (operating) pixels. Besides the very valid remark by @maxnoe

Since this is for the case in which non-operating clusters exist, clearly the produced file can only be used by the EVB for the case in which the same clusters are not working.  Does the EVB check that the corrections file it gets contains the info for all the clusters in the EVB configuration? If not, that should be added, right? 

> 
> > @pawel21 The pedestal values until now were integers. NaN is only avaiable for floats. Did you change the dtype of the pedestal in the files?
> 
> this is another reason to reject this PR.

I take it that you meant "to further discuss and polish this PR", right? Let's stay civil, please.  :)

maxnoe:
@pawel21 please don't try to repair this branch by hand. I think it would be better to just start from the current master and then making your changes again.

You anyway need to open a new PR from a branch that is pushed to this repository, not to a fork.

pawel21:
> @pawel21 I also just saw that you made very many, unrelated changes. E.g. deleting the whole `lstchain/io/io.py`.
> 
> To make the tests pass, you will also need to push a new branch to this repository (and not use your fork anymore, since it does not have access to the private data needed for observed data tests)

Sorry, I lost. How I can  push a new branch to this repository?

pawel21:
> @pawel21 please don't try to repair this branch by hand. I think it would be better to just start from the current master and then making your changes again.
> 
> You anyway need to open a new PR from a branch that is pushed to this repository, not to a fork.

Thank you.
Yes, this will be the best solution.

lstchain/calib/camera/drs4.py
```diff
@@ -83,11 +84,18 @@ def _fill_pedestal_event_jit(waveform, expected_pixel_id, first_cap_array,
...
+=======
+        pixels_with_nan_value = np.unique(np.where(np.isnan(self.meanped))[1])
+>>>>>>> try fix branch
+        if len(pixels_with_nan_value) > 0:
+            self.failing_pixels_list = pixels_with_nan_value
+            print("Failing pixels:")
+            print(self.failing_pixels_list)
         else:
-            self.meanped = self.meanped / self.numped
+            print("Everything is OK")
```
rlopezcoto, line 98: maybe something more informative like:
`print("No pixels with nan values")`

rlopezcoto (commented):
why is the `lstchain/scripts/lstchain_dl1ab.py` script removed?

jugean2:
@pawel21 @moralejo 
Hi,
The baseline file used to feed the script which will generate the file understandable by the EVB must contain exactly 1855 pixels. 
This script will select in this baseline file the modules which are contained in the configuration file (the expected modules).
So you can put any value for the missing modules but you must put something. Of course we must not use this baseline file filled 
 with 0 or any value for missing modules with a configuration file which would include this missing module.

moralejo:
> @pawel21 @moralejo
> Hi,
> The baseline file used to feed the script which will generate the file understandable by the EVB must contain exactly 1855 pixels. This script will select in this baseline file the modules which are contained in the configuration file (the expected modules). So you can put any value for the missing modules but you must put something. Of course we must not use this baseline file filled with 0 or any value for missing modules with a configuration file which would include this missing module. Hope this is clear...

Yes, that sounds good. @pawel21 can you agree with @JulienHoules and @DirkHoffmann (whoever is in charge of modifying the translator script) on what error values to put in the drs4 calibration files for the missing modules?  The script should just return an error if a module present in the evb config file contains such error values.

DirkHoffmann:
As this issue is already closed, it's better to use a [dedicated ticket](https://forge.in2p3.fr/issues/43920) for the conclusion on the calibration-EVB interface.

PR #595, opened 2021-02-22
Add batch option to dl1 datacheck scripts

morcuended:
Right now dl1 data checker tries to display some plots so I added a batch option so it is more convenient for running it in clusters without a graphical interface or simply running the tests locally.

By default batch option is set to False.

A test for these scripts is added in #587.


lstchain/datachecks/dl1_checker.py
```diff
@@ -90,15 +94,15 @@ def check_dl1(filenames, output_path, max_cores=4, create_pdf=False):
...
+    # identifying interleaved pedestals. We check which one has
     # more values == 32, which is the pedestal tag. The one called
     # "trigger_type" seems to be the TIB trigger type. The fastest way to do
     # this for the whole run seems to be using normal pytables:
     trig_tags = {'trigger_type': [], 'ucts_trigger_type': []}
     for filename in filenames:
         with tables.open_file(filename,
                               root_uep='/dl1/event/telescope/parameters') as f:
-            for name in trig_tags.keys():
+            for name in trig_tags:
```
rlopezcoto, line 105: why is the `keys()` removed?

lstchain/datachecks/dl1_checker.py
```diff
@@ -244,6 +242,12 @@ def process_dl1_file(filename, bins, trigger_source='trigger_type'):
         # trigger type tags are not reliable. We first identify flatfield events
         # by their looks.
         image = image_table.col('image')
+        # define criteria for detecting flatfield events, since as of 20200418
```
rlopezcoto, line 245: why moving the `ff_min_pixel_charge_median` and `ff_max_pixel_charge_stddev` inside the table opening?

lstchain/datachecks/dl1_checker.py
morcuended: I think it is not needed. The IDE suggested removing it.

lstchain/datachecks/dl1_checker.py
morcuended: Same here, it was suggested by the IDE: "move assignment closer to their usage".

Both changes can be reverted if necessary.

rlopezcoto (approved):

maxnoe:
SHouldn't this be solved without any additional code by just using `export MPLBACKEN=Agg` for non-interactive use of this script? 

You should have this anyway in all your job submission scripts.

morcuended:
> SHouldn't this be solved without any additional code by just using `export MPLBACKEND=Agg` for non-interactive use of this script?
> 
> You should have this anyway in all your job submission scripts.

That's what we have been doing for months now. Not a big deal. However, now that I'm running locally the tests of the datachecker these windows keep appearing all the time.

PR #598, opened 2021-02-23
Improve Tool of camera calibration coefficients.

Bultako:
This PR is the continuation of #570 developed in a forked repo and now with the code placed in a branch of `lstchain` repo.

It polishes code in the `lstchain_create_calibration_file.py` existing Tool, that generates an HDF5 file with camera calibration coefficients, mostly to be consistent with the Tools that are being developed at this moment in different PRs. It also provides a progress bar instead of the debug logs for every 100 processed events.


rlopezcoto:
outdated with the latest calibration changes

PR #573, opened 2020-12-12
Tool to generate a DL2 file from a DL1 file

Bultako:
This PR adds a new Tool `lstchain_create_dl2_file` adapting the existing code found in script `scripts/lstchain_dl1_to_dl2.py`

```
$ lstchain_create_dl2_file --help-all
Generate a HDF5 file with reconstructed energy, disp and hadroness of events

Options
-------

Arguments that take values are actually convenience aliases to full
Configurables, whose aliases are listed on the help line. For more information
on full configurables, see '--help-all'.

--input=<Path> (ReconstructionHDF5Writer.input)
    Default: None
    Path to a DL1 HDF5 file
-i <Path> (ReconstructionHDF5Writer.input)
    Default: None
    Path to a DL1 HDF5 file
--models=<Path> (ReconstructionHDF5Writer.path_models)
    Default: None
    Path where to find the trained RF models
--output=<Path> (ReconstructionHDF5Writer.output_dir)
    Default: None
    Path where to store the reconstructed DL2
-o <Path> (ReconstructionHDF5Writer.output_dir)
    Default: None
    Path where to store the reconstructed DL2
--config=<Path> (Tool.config_file)
    Default: None
    name of a configuration file with parameters to load in addition to command-
    line parameters
--log-level=<Enum> (Application.log_level)
    Default: 30
    Choices: (0, 10, 20, 30, 40, 50, 'DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL')
    Set the log level by value or name.

Class parameters
----------------

Parameters are set from command-line arguments of the form:
`--Class.trait=value`. This line is evaluated in Python, so simple expressions
are allowed, e.g.:: `--C.a='range(3)'` For setting C.a=[0,1,2].

ReconstructionHDF5Writer options
--------------------------------
--ReconstructionHDF5Writer.config_file=<Path>
    Default: None
    Path to a configuration file. If none is given, a standard configuration is
    applied
--ReconstructionHDF5Writer.input=<Path>
    Default: None
    Path to a DL1 HDF5 file
--ReconstructionHDF5Writer.log_datefmt=<Unicode>
    Default: '%Y-%m-%d %H:%M:%S'
    The date format used by logging formatters for %(asctime)s
--ReconstructionHDF5Writer.log_format=<Unicode>
    Default: '%(levelname)s [%(name)s] (%(module)s/%(funcName)s): %(messag...
    The Logging format template
--ReconstructionHDF5Writer.log_level=<Enum>
    Default: 30
[...]

PR #599, opened 2021-02-23
Tool to generate a DL2 file from a DL1 file

Bultako:
This PR is the continuation of #573 developed in a forked repo and now with the code placed in a branch of `lstchain` repo.

It adds a new Tool `lstchain_create_dl2_file` adapting the existing code found in script `scripts/lstchain_dl1_to_dl2.py`

```
Generate a HDF5 file with reconstructed energy, disp and gammaness of events

Options
=======
    <cmd> --help-all

--source-dependent
    Perform source dependent analysis
    Equivalent to: [--ReconstructionHDF5Writer.source_dependent=True]
-q, --quiet
    Disable console logging.
-i, --input=<Path>
    Path to a DL1 HDF5 file
    Default: None
    Equivalent to: [--ReconstructionHDF5Writer.input]
-o, --output=<Path>
    Path where to store the reconstructed DL2
    Default: None
    Equivalent to: [--ReconstructionHDF5Writer.output_dir]
--energy-model=<Path>
    Path where to find the Energy trained RF model file
    Default: None
    Equivalent to: [--ReconstructionHDF5Writer.path_energy_model]
--disp-model=<Path>
    Path where to find the Disp trained RF model file
    Default: None
    Equivalent to: [--ReconstructionHDF5Writer.path_disp_model]
--gh-model=<Path>
    Path where to find the Gammaness trained RF model file
    Default: None
    Equivalent to: [--ReconstructionHDF5Writer.path_gh_model]
--config=<Path>
    line parameters
    Default: None
--log-level=<Enum>
    Default: 30
-l, --log-file=<Path>
    Filename for the log
    Default: None
--log-file-level=<Enum>
    Default: 'INFO'

Class options
=============
The command-line option below sets the respective configurable class-parameter:
    --Class.parameter=value
This line is evaluated in Python, so simple expressions are allowed.
For instance, to set `C.a=[0,1,2]`, you may type this:
    --C.a='range(3)'

Application(SingletonConfigurable) options
------------------------------------------
--Application.log_datefmt=<Unicode>
    The date format used by logging formatters for %(asctime)s
    Default: '%Y-%m-%d %H:%M:%S'
--Application.log_format=<Unicode>
    The Logging format template
    Default: '[%(name)s]%(highlevel)s %(message)s'
--Application.log_level=<Enum>
    Default: 30
--Application.show_config=<Bool>
    Instead of starting the Application, dump configuration to stdout
    Default: False
--Application.show_config_json=<Bool>
    Instead of starting the Application, dump configuration to stdout (as JSON)
    Default: False

Tool(Application) options
-------------------------
--Tool.config_file=<Path>
    line parameters
    Default: None
--Tool.log_config=<key-1>=<value-1>...
    Default: {'version': 1, 'disable_existing_loggers': False, 'formatters...
--Tool.log_datefmt=<Unicode>
    The date format used by logging formatters for %(asctime)s
    Default: '%Y-%m-%d %H:%M:%S'
--Tool.log_file=<Path>
    Filename for the log
    Default: None
--Tool.log_file_level=<Enum>
    Default: 'INFO'
--Tool.log_format=<Unicode>
    The Logging format template
    Default: '[%(name)s]%(highlevel)s %(message)s'
--Tool.log_level=<Enum>
    Default: 30
--Tool.provenance_log=<Path>
    Default: None
--Tool.quiet=<Bool>
    Default: False
--Tool.show_config=<Bool>
    Instead of starting the Application, dump configuration to stdout
    Default: False
--Tool.show_config_json=<Bool>
    Instead of starting the Application, dump configuration to stdout (as JSON)
    Default: False

ReconstructionHDF5Writer(Tool) options
--------------------------------------
--ReconstructionHDF5Writer.classification_features=<list-item-1>...
    List of classification features
    Default: []
--ReconstructionHDF5Writer.config_file=<Path>
    line parameters
    Default: None
--ReconstructionHDF5Writer.events_filters=<key-1>=<value-1>...
    Dictionary with information to filter events
    Default: {}
--ReconstructionHDF5Writer.input=<Path>
    Path to a DL1 HDF5 file
    Default: None
--ReconstructionHDF5Writer.log_config=<key-1>=<value-1>...
    Default: {'version': 1, 'disable_existing_loggers': False, 'formatters...
--ReconstructionHDF5Writer.log_datefmt=<Unicode>
    The date format used by logging formatters for %(asctime)s
    Default: '%Y-%m-%d %H:%M:%S'
--ReconstructionHDF5Writer.log_file=<Path>
    Filename for the log
    Default: None
--ReconstructionHDF5Writer.log_file_level=<Enum>
    Default: 'INFO'
--ReconstructionHDF5Writer.log_format=<Unicode>
    The Logging format template
    Default: '[%(name)s]%(highlevel)s %(message)s'
--ReconstructionHDF5Writer.log_level=<Enum>
    Default: 30
--ReconstructionHDF5Writer.output_dir=<Path>
    Path where to store the reconstructed DL2
    Default: None
--ReconstructionHDF5Writer.path_disp_model=<Path>
    Path where to find the Disp trained RF model file
    Default: None
--ReconstructionHDF5Writer.path_energy_model=<Path>
    Path where to find the Energy trained RF model file
    Default: None
--ReconstructionHDF5Writer.path_gh_model=<Path>
    Path where to find the Gammaness trained RF model file
    Default: None
--ReconstructionHDF5Writer.provenance_log=<Path>
    Default: None
--ReconstructionHDF5Writer.quiet=<Bool>
    Default: False
--ReconstructionHDF5Writer.regression_features=<list-item-1>...
    List of regression features
    Default: []
--ReconstructionHDF5Writer.show_config=<Bool>
    Instead of starting the Application, dump configuration to stdout
    Default: False
--ReconstructionHDF5Writer.show_config_json=<Bool>
    Instead of starting the Application, dump configuration to stdout (as JSON)
    Default: False
--ReconstructionHDF5Writer.source_dependent=<Bool>
    Is the analysis source dependent?
    Default: False
```


lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        """
+        Tool that generates a HDF5 file with reconstructed energy, disp and hadroness of events.
+
+        For getting help run:
+        lstchain_create_dl2_file --help
+        """
```
maxnoe, line 58: It is not necessary in python to declare variables upfront in `__init__.py`. Setting them here only makes sense if you need them already set for some reason, but e.g. `data` get's immidiately reassigned in `setup`.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+        self.configuration = None
+        self.data = None
+        self.reg_energy = None
+        self.reg_disp_vector = None
+        self.cls_gh = None
+        self.dl2 = None
+
+    def setup(self):
+
+        self.log.info("Reading configuration")
```
maxnoe, line 68: Tool already supports reading configuration files. Please don't add any new configuration file logic. Adapt the tool / config to work with the existing configuration file logic.

There should absolutely be no code to deal with configurations in here.

maxnoe: We loose most of the advantages of using Tools if we again are not using it how intendend but reimplement a new configuration system on top of it.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+        self.configuration = standard_config
+        if self.config_file:
+            self.configuration = read_configuration_file(self.config_file)
+
+        self.log.info("Reading DL1 file")
+        self.data = pd.read_hdf(self.input, key=dl1_params_lstcam_key)
+        self.data = add_delta_t_key(self.data)
+        if self.configuration["source_dependent"]:
```
maxnoe, line 76: E.g. this should be solved by adding a traitlet `source_dependent = Bool().tag(config=True)`. and then just checking `if self.source_dependent` here.

maxnoe (changes requested):
Please do not reimplement the configuration loading system.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+    config_file = traits.Path(
+        help="Path to a configuration file. If none is given, a standard configuration is applied",
+        directory_ok=False,
+    ).tag(config=True)
+
+    aliases = {
+        ("i", "input"): "ReconstructionHDF5Writer.input",
+        ("o", "output"): "ReconstructionHDF5Writer.output_dir",
+        "models": "ReconstructionHDF5Writer.path_models",
+        "config": "ReconstructionHDF5Writer.config_file",
```
maxnoe, line 48: No need for this, `Tool` already comes with `config` traitlet for setting the configuration file

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+    input = traits.Path(
+        help="Path to a DL1 HDF5 file", directory_ok=False, exists=True
+    ).tag(config=True)
+    path_models = traits.Path(
+        help="Path where to find the trained RF models", file_ok=False
+    ).tag(config=True)
+    output_dir = traits.Path(
+        help="Path where to store the reconstructed DL2", file_ok=False
+    ).tag(config=True)
+    config_file = traits.Path(
```
maxnoe, line 39: This is not needed, as `Tool` already implements the configuration handling.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+
+        self.log.info("Reading DL1 file")
+        self.data = pd.read_hdf(self.input, key=dl1_params_lstcam_key)
+        self.data = add_delta_t_key(self.data)
+        if self.configuration["source_dependent"]:
+            data_src_dep = pd.read_hdf(self.input, key=dl1_params_src_dep_lstcam_key)
+            self.data = pd.concat([self.data, data_src_dep], axis=1)
+
+        self.log.info("Reading RF models")
+        self.reg_energy = joblib.load(self.path_models / "reg_energy.sav")
```
maxnoe, line 81: In general, I am not a big fan of setting an input directory and then have hard coded names for files in an input directory. I would go for three `Path` traitlets for each of the models. This is much more flexible for e.g. debugging / benchmarking different models / what not.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+        lstchain_create_dl2_file --help
+        """
+        self.configuration = None
+        self.data = None
+        self.reg_energy = None
+        self.reg_disp_vector = None
+        self.cls_gh = None
+        self.dl2 = None
+
+    def setup(self):
```
maxnoe, line 66: I think `setup` should only initialize components that need the configuration. Any actual work, like loading input data and so on should happen in `run`.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+        ):
+            # make sure there is a least one good pointing value to interpolate from
+            if np.isfinite(self.data.alt_tel).any() and np.isfinite(self.data.az_tel).any():
+                self.data = impute_pointing(self.data)
+            else:
+                self.data.alt_tel = -np.pi / 2.0
+                self.data.az_tel = -np.pi / 2.0
+        self.data = filter_events(
+            self.data,
+            filters=self.configuration["events_filters"],
```
maxnoe, line 101: Add a `List` traitlet for this and give `filters=self.filters`

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+            finite_params=self.configuration["regression_features"]
```
maxnoe, line 102: Same here, add a `List` traitlet and add `self.regression_features`.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,161 @@
...
+            + self.configuration["classification_features"],
```
maxnoe, line 103: And again. Do not use `self.configuration`!

Bultako:
Thanks @maxnoe for the code review.

I have tried to set the config file parameter as mandatory, and worked with and internal `self._dict_conf` dict instead. For event filters, regression features, classification features, I think it is better to declare that info in a config file than in traits input params. The three model files used in training I have set up as traits input params.

Otherwise, tests with data have been added and we have green light in the CI.
I'm, marking this PR as ready for review.

maxnoe:
> and worked with and internal self._dict_conf dict instead. 

This is exactly what I wanted to avoid! Still using a separate configuration system.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,188 @@
...
+        self.data_ind = None
+        self.data_src_dep = None
+        self.dl2_ind = None
+        self.dl2_src_dep_level = None
+        self.dl2_src_dep_dict = {}
+        self.reg_energy = None
+        self.reg_disp_vector = None
+        self.cls_gh = None
+
+        self._dict_conf = replace_config(standard_config, self.get_current_config())
```
maxnoe, line 65: Still using a separete config system. Please remove and use traitlets properly. Otherwise the switch to Tools is kind of pointless.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def setup(self):
+
+        if self.config_file is None:
+            raise ToolConfigurationError("Please provide configuration file with --config parameter")
```
maxnoe, line 70: This is also not be done. Check if any of the required traitlets are missing. Not that a config file has been given. Traitlets allows the user to choose what to put in the config file and what to give on the command line. Forcing a config file is not the way. Check that all data you need is set to correct values.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,188 @@
...
+
+        self.log.info("Reading DL1 file")
+        self.data_ind = pd.read_hdf(self.input, key=dl1_params_lstcam_key)
+        if self._dict_conf["source_dependent"]:
```
maxnoe, line 74: Add a traitlet `source_dependent = Bool(default_value=False, help='...').tag(config=True)` for this. Do not use a second, parallel self-implement configuration system.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,188 @@
...
+            else:
+                self.data_ind.alt_tel = -np.pi / 2.0
+                self.data_ind.az_tel = -np.pi / 2.0
+
+        # source-independent analysis
+        self.log.info("Applying models")
+        if not self._dict_conf["source_dependent"]:
+            self.data_ind = filter_events(
+                self.data_ind,
+                filters=self._dict_conf["events_filters"],
```
maxnoe, line 103: Add a `List` or `Dictionary` traitlet for the event filters.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,188 @@
...
+                finite_params=self._dict_conf["regression_features"] + self._dict_conf["classification_features"],
```
maxnoe, line 104: Add a `List` traitlet each for `regression features and `classification` features.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,188 @@
...
+            )
+            self.dl2_ind = dl1_to_dl2.apply_models(
+                self.data_ind,
+                self.cls_gh,
+                self.reg_energy,
+                self.reg_disp_vector,
+                custom_config=self._dict_conf,
```
maxnoe, line 111: If you want to be minimally invasive, construct `custom_config` from the traitlets here, but the better approach would be to implement a `Component` for `apply_models`.

Bultako:
Thanks @maxnoe 
I now get the whole point with tools configuration system 👍🏻 
I have added traits as suggested replacing the useless config vars that I used. I have only set up `source-dependent` boo trait as command tool input param in the aliases. The list of regression and classification features, as well as the filter events criteria has to be provided in the config file, and raise an error if not found.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -43,6 +56,7 @@ class ReconstructionHDF5Writer(Tool):
         "energy-model": "ReconstructionHDF5Writer.path_energy_model",
         "disp-model": "ReconstructionHDF5Writer.path_disp_model",
         "gh-model": "ReconstructionHDF5Writer.path_gh_model",
+        "source-dependent": "ReconstructionHDF5Writer.source_dependent"
```
maxnoe, line 59: This is a boolean, so using a flag might be better than an alias. WIth alias you have to do `--source-dependent true`, with a flag, that's just `--source-dependant`.

The format is `{flag: (config, help_string), ...}`

```
flags = {
    'source-dependant': (
        {'ReconstructionHDF5Writer': {'source_dependent': True}},
        'Perform source dependent analysis',
    )
}
```

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -53,25 +67,19 @@ def __init__(self, **kwargs):
...
-        self.reg_disp_vector = None
-        self.cls_gh = None
-
-        self._dict_conf = replace_config(standard_config, self.get_current_config())
 
     def setup(self):
 
-        if self.config_file is None:
-            raise ToolConfigurationError("Please provide configuration file with --config parameter")
+        if not len(self.config["events_filters"]):
```
maxnoe, line 73: Is this required? How would you perform it *without* filtering events?

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -53,25 +67,19 @@ def __init__(self, **kwargs):
...
+            raise ToolConfigurationError("Information on events filtering not found in config.")
+        if not len(self.config["classification_features"]):
```
maxnoe, line 75: `self.classification_features`

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -53,25 +67,19 @@ def __init__(self, **kwargs):
...
+            raise ToolConfigurationError("Information on classification features not found in config.")
+        if not len(self.config["regression_features"]):
```
maxnoe, line 77: `self.regression_features`

lstchain/tools/lstchain_create_dl2_file.py
maxnoe: No, that shouldn't be the case.

maxnoe: I tried accessing `self.classification_features` on your branch and it works as expected.

lstchain/tools/lstchain_create_dl2_file.py
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    flags = {
+        'source-dependent': (
+            {'ReconstructionHDF5Writer': {'source_dependent': True}},
+            'Perform source dependent analysis',
+        )
+    }
+
+    def __init__(self, **kwargs):
```
maxnoe, line 68: If `__init__` does nothing, it is not needed.

lstchain/tools/lstchain_create_dl2_file.py
maxnoe: This is how the config file has to look like:

```json
{
  "ReconstructionHDF5Writer": {
    "regression_features": ["intensity", "width", "length"],
    "classification_features": ["intensity", "width", "length"]
  }
}

```

maxnoe: Maybe you were missing the `ReconstructionHDF5Writer` level?

Bultako: I have an empty list in `self.classification_features` and the list below in `self.config["classification_features"]`

```
['log_intensity', 'width', 'length', 'x', 'y', 'psi', 'phi', 'wl', 'skewness', 'kurtosis', 'r', 
'time_gradient', 'leakage_intensity_width_2', 'log_reco_energy', 'reco_disp_dx', 'reco_disp_dy']
```

maxnoe: Then  you are missing the `ReconstructionHDF5Writer` level in the file. You need to add that.

lstchain/tools/lstchain_create_dl2_file.py
Bultako: ok, I have removed it

lstchain/tools/lstchain_create_dl2_file.py
maxnoe: Unfortunately, the config system doesn't raise errors for unrecognised stuff in the config file. Otherwise you would have gotten that.

Bultako: Would you suggest to wrap the existing `regression_features` and `classification_features` info with `ReconstructionHDF5Writer`? In that case I guess all other scripts grabbing that info would need to be modified..

maxnoe: This is how the configuration system works! Traitlets have to be in the namespace of the tool in their config file. Like the example I showed above.

Bultako: Ok, just to be sure :) 

I will modify the config files as well as all other scripts using classification/regression features and events filtering accordingly, so they can access the info from the config files.

maxnoe: This will also bring us away from the one *monster* lstchain config file to config files per tool, which is I think a good thing.

lstchain/tools/tests/test_tools.py
```diff
@@ -0,0 +1,24 @@
+from pathlib import Path
+
+from ctapipe.core import run_tool
+from lstchain.tools.lstchain_create_dl2_file import ReconstructionHDF5Writer
+
+
+def test_create_drs4_pedestal_file(temp_dir_observed_files, simulated_dl1_file, rf_models):
```
morcuended, line 7: name of the test should be `test_create_dl2_file` not create drs4

lstchain/data/lstchain_src_dep_config.json
```diff
@@ -1,84 +1,99 @@
 {
-  "events_filters": {
-    "intensity": [0, Infinity],
-    "width": [0, Infinity],
-    "length": [0, Infinity],
-    "wl": [0, 1],
-    "r": [0, 1],
-    "leakage_intensity_width_2": [0, 1]
-  },
+ "ReconstructionHDF5Writer": {
```
maxnoe, line 2: I don't think we should change the existing config files here. 

We should create new config files per tool with only the configuration options for that tool.

This is also necessary, since `Tool` options need to be in the namespace of the Tool, so you anyway need config file per tool.

lstchain/tools/lstchain_create_dl2_file.py
maxnoe: No, the code checking the traitlets and so on belongs here, but I would move the loading of the data and the models.

Any actual work should be done in `start` rather than `setup`

lstchain/data/lstchain_src_dep_config.json
maxnoe: Probably just keep the existing file like it is and add a new `lstchain/data/dl2_tool_config.json` with the correct config?

Using sensible defaults should make these config files rather small.

vuillaut:
Closing as this is stalled.

PR #533, opened 2020-10-05
Tool to produce a time calibration HDF5 file

Bultako:
This PR adds a new Tool  `lstchain_create_time_calibration_file`

```
lstchain_create_time_calibration_file --help-all
Generate a HDF5 file with time calibration coefficients

Options
-------

Arguments that take values are actually convenience aliases to full
Configurables, whose aliases are listed on the help line. For more information
on full configurables, see '--help-all'.

--input=<Path> (TimeCalibrationHDF5Writer.input)
    Default: None
    Path to the fits.fz events file or directory to glob
--glob=<Unicode> (TimeCalibrationHDF5Writer.glob)
    Default: '*'
    Filename pattern to glob files in the directory
--output=<Unicode> (TimeCalibrationHDF5Writer.output)
    Default: ''
    Path to the hdf5 time calibration file
--pedestal=<Unicode> (LSTR0Corrections.pedestal_path)
    Default: ''
    Path to the LST pedestal binary file
--max_events=<Int> (TimeCalibrationHDF5Writer.max_events)
    Default: 20000
    Maximum numbers of events to read. Default = 20000
--log-level=<Enum> (Application.log_level)
    Default: 30
    Choices: (0, 10, 20, 30, 40, 50, 'DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL')
    Set the log level by value or name.
--config=<Path> (Tool.config_file)
    Default: None
    name of a configuration file with parameters to load in addition to command-
    line parameters

Class parameters
----------------

Parameters are set from command-line arguments of the form:
`--Class.trait=value`. This line is evaluated in Python, so simple expressions
are allowed, e.g.:: `--C.a='range(3)'` For setting C.a=[0,1,2].

TimeCalibrationHDF5Writer options
---------------------------------
--TimeCalibrationHDF5Writer.config_file=<Path>
    Default: None
    name of a configuration file with parameters to load in addition to command-
    line parameters
--TimeCalibrationHDF5Writer.glob=<Unicode>
    Default: '*'
    Filename pattern to glob files in the directory
--TimeCalibrationHDF5Writer.input=<Path>
    Default: None
    Path to the fits.fz events file or directory to glob
--TimeCalibrationHDF5Writer.log_datefmt=<Unicode>
    Default: '%Y-%m-%d %H:%M:%S'
    The date format used by logging formatters for %(asctime)s
--TimeCalibrationHDF5Writer.log_format=<Unicode>
[...]

PR #600, opened 2021-02-23
Tool to produce a time calibration HDF5 file

Bultako:
This PR is the continuation of #533 developed in a forked repo and now with the code placed in a branch of `lstchain` repo.

It adds a new Tool  `lstchain_create_time_calibration_file`

```
Generate a HDF5 file with time calibration coefficients

Options
=======
    <cmd> --help-all

--debug
    Set log-level to debug, for the most verbose logging.
    Equivalent to: [--Application.log_level=10]
--show-config
    Show the application's configuration (human-readable format)
    Equivalent to: [--Application.show_config=True]
--show-config-json
    Show the application's configuration (json format)
    Equivalent to: [--Application.show_config_json=True]
-q, --quiet
    Disable console logging.
-i, --input=<Path>
    Path to the fits.fz events file or directory to glob
    Default: None
    Equivalent to: [--TimeCalibrationHDF5Writer.input]
-o, --output=<Unicode>
    Path to the time calibration file
    Default: ''
    Equivalent to: [--TimeCorrectionCalculate.calib_file_path]
--glob=<Unicode>
    Filename pattern to glob files in the directory
    Default: '*.fits.fz'
    Equivalent to: [--TimeCalibrationHDF5Writer.glob]
--max-events=<Int>
    Default: None
    Equivalent to: [--LSTEventSource.max_events]
--pedestal=<telescopeparameter-item-1>...
    Path to the LST pedestal file
    Default: [('type', '*', None)]
    Equivalent to: [--LSTR0Corrections.drs4_pedestal_path]
--run-summary-path=<telescopeparameter-item-1>...
    Path to the run summary for the correct night. If given, dragon reference
    counters are read from this file. Explicitly given values override values
    read from the file.
    Default: [('type', '*', None)]
    Equivalent to: [--EventTimeCalculator.run_summary_path]
--config=<Path>
    line parameters
    Default: None
--log-level=<Enum>
    Default: 30
-l, --log-file=<Path>
    Filename for the log
    Default: None
--log-file-level=<Enum>
    Default: 'INFO'

To see all available configurables, use `--help-all`.
```


Bultako:
I guess that for testing this PR we need bigger raw files than those already present in `test_data`. I have the exception below when testing with `LST-1.1.Run02008.0000_first50.fits.fz`  

```
Caught unexpected exception: Not enough events to coverage all capacitor. 
Please use more events to time calibration file.
```

maxnoe:
@Bultako is that an option? Could you run the test with less required events?

rlopezcoto:
> @Bultako is that an option? Could you run the test with less required events?
> 
> And maybe fix the language of the error?

For the time calibration it is complicated because you need to cover all capacitors to be able to calculate the time correction for all of them. In principle the exception is right and you should not be able to produce a time calibration file without covering all capacitors.
I cannot think on any way of performing tests without that, but maybe @FrancaCassol or @pawel21 have some idea.

maxnoe:
We can also try to look for the smallest possible file. 

Unfortunately that will be around 1000 events or so I think.

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,97 @@
...
+    name = "TimeCalibrationHDF5Writer"
+    description = "Generate a HDF5 file with time calibration coefficients"
+
+    input = traits.Path(
+        help="Path to the fits.fz events file or directory to glob"
+    ).tag(config=True)
+
+    glob = traits.Unicode(
+        help="Filename pattern to glob files in the directory",
+        default_value="*"
```
maxnoe, line 22: Maybe a better default would be `*.fits.fz`?

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,97 @@
...
+        ("i", "input"): "TimeCalibrationHDF5Writer.input",
+        ("o", "output"): "TimeCorrectionCalculate.calib_file_path",
+        "glob": "TimeCalibrationHDF5Writer.glob",
+        "max-events": "EventSource.max_events",
+        "pedestal": "LSTR0Corrections.drs4_pedestal_path",
+        "dragon-reference-time": "EventTimeCalculator.dragon_reference_time",
+        "dragon-reference-counter": "EventTimeCalculator.dragon_reference_counter",
+    }
+
+    classes = [TimeCorrectionCalculate] + traits.classes_with_traits(EventSource)
```
maxnoe, line 40: Does this script work with other files than LST observed raw data? If not, you could just explicitly use the `LSTEventSource` instead of `traits.classes_with_traits(EventSource)` to avoid showing options for event sources that are not supported by this tool

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(self, **kwargs):
```
maxnoe, line 42: Again, looks like it is not needed.

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,97 @@
...
+    def start(self):
+
+        for j, path in enumerate(self.path_list):
+            self.eventsource = EventSource(input_url=path, config=self.config)
+            self.log.info(f"File {j + 1} out of {len(self.path_list)}")
+            self.log.info(f"Processing: {path}")
+
+            self.timeCorr = TimeCorrectionCalculate(
+                subarray=self.eventsource.subarray,
+                config=self.config,
```
maxnoe, line 70: Same here, always use `parent=self` for sub components.

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,97 @@
...
+                    total=self.eventsource.max_events,
+                    unit="ev",
+                    disable=not self.progress_bar,
+            ):
+                self.timeCorr.calibrate_peak_time(event)
+
+    def finish(self):
+
+        self.timeCorr.finalize()
+        Provenance().add_output_file(
```
maxnoe, line 85: This should probably go into the time correction component, since it happens there.

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,97 @@
...
+    ).tag(config=True)
+
+    aliases = {
+        ("i", "input"): "TimeCalibrationHDF5Writer.input",
+        ("o", "output"): "TimeCorrectionCalculate.calib_file_path",
+        "glob": "TimeCalibrationHDF5Writer.glob",
+        "max-events": "EventSource.max_events",
+        "pedestal": "LSTR0Corrections.drs4_pedestal_path",
+        "dragon-reference-time": "EventTimeCalculator.dragon_reference_time",
+        "dragon-reference-counter": "EventTimeCalculator.dragon_reference_counter",
```
maxnoe, line 37: Also add `module_id` since you need all three now. And with the next release, there will also be the option to just give the path to the run summary instead of giving all the counters explicitly.

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        self.eventsource = None
+        self.path_list = None
+        self.timeCorr = None
+
+    def setup(self):
+
+        self.path_list = [str(self.input)]
+        if self.input.is_dir():
+            self.path_list = sorted(glob.glob(str(self.input/self.glob)))
```
maxnoe, line 59:
```suggestion
            self.path_list = sorted(self.input.glob(self.glob))
```

Bultako:
This PR has been rebased into master branch last commit, green CI, all requested changes in previous reviews have been addressed, and now is ready for a final review.

lstchain/tools/lstchain_create_time_calibration_file.py
```diff
@@ -0,0 +1,90 @@
...
+            self.log.info(f"File {j + 1} out of {len(self.path_list)}")
+            self.log.info(f"Processing: {path}")
+
+            self.timeCorr = TimeCorrectionCalculate(
+                subarray=self.eventsource.subarray,
+                parent=self,
+            )
+
+            for event in tqdm(
+                    self.eventsource,
```
maxnoe, line 71: Minor nitpick, but this looks indented one level too much.

maxnoe (approved):
Should we remove the corresponding scripts and update the corresponding osa scripts?

rlopezcoto:
>Should we remove the corresponding scripts and update the corresponding osa scripts?

Before removing the current scripts, I would wait to get this tool further tested, especially by experts

rlopezcoto:
this one was reviewed and approved, could you re-approve @maxnoe so we can also include it in the next release?

maxnoe:
I think we need to adapt the onsite scripts so that actually the new tool is called, right?

rlopezcoto:
I would get this tool in the release before, and then lstosa can be later updated to use it

maxnoe:
I am not talking about Isa, I am talking about the onsite scripts in this report.

rlopezcoto:
ok, let's leave this for lstchain > v0.8

vuillaut:
Should be closed?

maxnoe:
Yes, all the calibration related tools have been moved to the new lstcam-calib project, we shouldn't update the ones here

PR #601, opened 2021-02-23
Tool to produce a pedestal FITS file

Bultako:
This PR is the continuation of #529 developed in a forked repo and now with the code placed in a branch of `lstchain` repo.

It adds a new Tool `lstchain_create_drs4_pedestal_file`

```
$ lstchain_create_drs4_pedestal_file --help
Generate a pedestal FITS file

Options
=======
    <cmd> --help-all

--no-delta-t
    Switch off delta T corrections
    Equivalent to: [--PedestalFITSWriter.deltaT=False]
-q, --quiet
    Disable console logging.
-i, --input=<Path>
    Path to the input file containing events.
    Default: None
    Equivalent to: [--LSTEventSource.input_url]
-o, --output=<Path>
    Path to the generated fits pedestal file
    Default: 'pedestal.fits'
    Equivalent to: [--PedestalFITSWriter.output]
--max-events=<Int>
    Default: None
    Equivalent to: [--LSTEventSource.max_events]
--start-r0-waveform=<Int>
    Start sample for waveform
    Default: 11
    Equivalent to: [--DragonPedestal.r0_sample_start]
--config=<Path>
    line parameters
    Default: None
--log-level=<Enum>
    Default: 30
-l, --log-file=<Path>
    Filename for the log
    Default: None
--log-file-level=<Enum>
    Default: 'INFO'

To see all available configurables, use `--help-all`.
```


rlopezcoto:
it seems that tests are not passing after merging the new `master`, probably some unresolved conflict

Bultako:
yep, we have a `ModuleNotFoundError: No module named '***.calib.camera.r0'` 
I'll fix it when implementing the tests with real data
For the moment, we mark it as a draft

Bultako:
Tests added with green light in CI.
I mark this as ready for review.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,113 @@
...
+    name = "PedestalFITSWriter"
+    description = "Generate a pedestal FITS file"
+
+    output = traits.Path(
+        default_value="pedestal.fits",
+        help="Path to the generated fits pedestal file",
+        directory_ok=False,
+    ).tag(config=True)
+
+    no_deltaT = traits.Bool(
```
maxnoe, line 25: That's not what I meant :D Having the `deltaT` boolean with default True is perfect! But if a bool is by default True, than you need a flag to switch it off.  So having `deltaT = traits.Bool(default_value=True).tag(config=True)` is correct, but then the flag must be:

```
flags = {'no-delta-t': ({"PedestalFITSWriter": {"deltaT": False}}, "Switch off delta T corrections")}
```

maxnoe: The flag must always set the opposite of the default to make sense.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,112 @@
...
+    name = "PedestalFITSWriter"
+    description = "Generate a pedestal FITS file"
+
+    output = traits.Path(
+        default_value="pedestal.fits",
+        help="Path to the generated fits pedestal file",
+        directory_ok=False,
+    ).tag(config=True)
+
+    deltaT = traits.Bool(
```
maxnoe, line 25: Sorry to come at this again, but you did the opposite of what I proposed. Now the default is "no delta T correction" and one has to explicitly enable it.

Do we want that? Isn't the normal use case the one *with* delta T correction? Defaults should be as close to the normal usage as possible, so less options are required for normal use, while still enabling the not-normal use.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,112 @@
...
+    flags = {
+        "deltaT": (
+            {"PedestalFITSWriter": {"deltaT": True}},
+            "Activate delta T corrections",
+        )
+    }
+
+    classes = [EventSource, DragonPedestal]
+
+    def __init__(self, **kwargs):
```
maxnoe, line 49: This `__init__` also looks like it is not necessary

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self.log.debug("Opening file")
+        self.eventsource = EventSource.from_config(parent=self)
+
+    def start(self):
+
+        event = next(iter(self.eventsource))
+        tel_id = event.trigger.tels_with_trigger[0]
+        self.pixel_ids = event.lst.tel[tel_id].svc.pixel_ids
+        self.pedestal = DragonPedestal(tel_id=tel_id, n_module=event.lst.tel[tel_id].svc.num_modules, config=self.config)
```
maxnoe, line 72: `DragonPedestal` is a component. So please do not pass `config=self.config` but `parent=self`.

maxnoe: You should probably also already create it in `setup`. that's what setup is for.

maxnoe: And add it to `classes`

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,112 @@
...
+
+        if self.deltaT:
```
maxnoe, line 74: Less code: `self.log.info(f"Using deltaT correction: {self.deltaT}")`

However, logging the configuration is a bit strange. Also all `Tools` log out the full configuration using `DEBUG`, so if you want to know the exact configuration used, you can use that.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,112 @@
...
+            unit="ev",
+            disable=not self.progress_bar,
+        ):
+            if self.deltaT:
+                for tel_id in event.r0.tels_with_data:
+                    self.eventsource.r0_r1_calibrator.update_first_capacitors(event)
+                    self.eventsource.r0_r1_calibrator.time_lapse_corr(event, tel_id)
+            self.pedestal.fill_pedestal_event(event)
+
+        self.pedestal.complete_pedestal()
```
maxnoe, line 92: I think this conceptually should go into `finish`, right?

lstchain/tools/lstchain_create_drs4_pedestal_file.py
Bultako: I understand that this is how it works in the [lstchain_data_create_drs4_pedestal_file.py](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py#L58) script.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def finish(self):
+
+        expected_pixel_id = fits.PrimaryHDU(self.pixel_ids)
+        pedestal_array = fits.ImageHDU(np.int16(self.pedestal.meanped), name="pedestal array")
+        failing_pixels_col = fits.Column(name='failing pixels', array=self.pedestal.failing_pixels_array, format='K')
+        failing_pixels = fits.BinTableHDU.from_columns([failing_pixels_col], name="failing pixels")
```
maxnoe, line 99: Why use a `BinTableHDU` for this but `ImageHDU`s for the other information?

I think we should also add much more meta information to this file, at least but probably more:

DATE, CREATOR, VERSION, run id, ...

Bultako: agree, maybe do this in another PR which would uniform the way on how fits files are created in lstchain scripts?

lstchain/tools/lstchain_create_drs4_pedestal_file.py
maxnoe: No, the code seems to by default make the calibration and you have to pass `--deltaT=false` to deactivate it.

maxnoe: So again, for the new tool I would propose having `deltaT` with default value `True` and an option `--no-deltaT` that sets to false

DirkHoffmann:
What is the status of this ticket? The pedestal generation tool is in production (used last week after inserting new modules in the camera). Are there still more changes to come?

Bultako:
Hi @DirkHoffmann 

This PR is part of a series of PRs intending to migrate existing scripts to Tools (see also #599 and #600). I guess they all have to be rebased to master and re-adapt the code to work with last devels in `lstchain`, before passing a final code review. It's a long time since they were open.

I think there are only four Tools in the [`tools`](https://github.com/cta-observatory/cta-lstchain/tree/master/lstchain/tools) folder of `lstchain`, and none of them producing a pedestal. What do you mean exactly saying that there is a pedestal generation tool in production? I guess it is somewhere else..

Bultako:
This PR has been rebased into master branch last commit, green CI, all requested changes in previous reviews have been addressed, and now is ready for a final review.

rlopezcoto:
this one got outdated with #725

PR #604, opened 2021-02-25
Drs4 create pedestal file with failing pixel 

pawel21:
I open new PR for fix script to create pedestal file with failing pixel with lstchain repo.


morcuended:
Now it's fine!

lstchain/calib/camera/drs4.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, tel_id, n_module, **kwargs):
         self.meanped = np.zeros((n_gain, self.n_pixels, size4drs))
         self.numped = np.zeros((n_gain, self.n_pixels, size4drs))
         self.first_cap_array = np.zeros((self.n_module, n_gain, n_channel))
+        self.failing_pixels_list = None
```
rlopezcoto, line 45: will this be a list of an array?
what about initializing it as an empty list  `self.failing_pixels_list = []`?

rlopezcoto (commented):
thanks @pawel21.
What would now be the containment of the drs4_pedestal file for the pixels that are in the `failing_pixels_list`?

lstchain/calib/camera/drs4.py
pawel21: If `failing_pixels_list` is no empty fits file will contain numpy ndarray (`fits_file[2].data`), otherwise value will be empty  (`NoneType`).

rlopezcoto: then what about initializing it as an empty array, and calling it `failing_pixels_array`?

pawel21: > then what about initializing it as an empty array, and calling it `failing_pixels_array`?

return None.

rlopezcoto: no, what I mean is that since it will be an array later, it may not be the best idea to initialize it as a `None`, but as a `np.empty` array.
Also, I'm not sure if a variable length array is the best here, or maybe better to have a boolean array:

`failing_pixels_array  = numpy.full((self.n_pixels), False)` 

and set it `True` only for the failing pixels

rlopezcoto:
> What would now be the containment of the drs4_pedestal file for the pixels that are in the failing_pixels_list?

this point I think it is important for the EvB because they needed a file with the information of all 265 modules, filled with `nan`(?) for the failing pixels, so what is the containment now?

lstchain/calib/camera/drs4.py
maxnoe: > fits_file[2].data

Please use extnames to identify tables instead of indices.

lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py
```diff
@@ -100,13 +94,12 @@ def main():
     # Finalize pedestal and write to fits file
     pedestal.finalize_pedestal()
 
-    primaryhdu = fits.PrimaryHDU(event.lst.tel[tel_id].svc.pixel_ids)
-    secondhdu = fits.ImageHDU(np.int16(pedestal.meanped))
-
-    hdulist = fits.HDUList([primaryhdu, secondhdu])
+    expected_pixel_id = fits.PrimaryHDU(event.lst.tel[tel_id].svc.pixel_ids)
+    pedestal_mean_array = fits.ImageHDU(np.int16(pedestal.meanped))
+    failing_pixels = fits.ImageHDU(pedestal.failing_pixels_list)
```
maxnoe, line 99: Why do you store the failing pixels as image HDU? Is that really the best representation?

pawel21:
> this point I think it is important for the EvB because they needed a file with the information of all 265 modules, filled with `nan`(?) for the failing pixels, so what is the containment now?
Now, array where pixels are failing contains 0

moralejo:
Will this work in the same way for calibration files in which a module is missing and for one where it is present but labeled as unusable? I think we have runs of both kinds in the past weeks, are you testing this on both?

rlopezcoto:
> Now, array where pixels are failing contains 0

Does this work for EvB @DirkHoffmann, @JulienHoules?

DirkHoffmann:
> > Now, array where pixels are failing contains 0
> 
> Does this work for EvB @DirkHoffmann, @JulienHoules?

It should. Let me update the [ICD](https://edms.cern.ch/document/2411710) accordingly and discuss details there. The main open question on my side is: What does the EVB do in case a pixel is declared with 0 correction factor ("zero-pixel")? Obviously, if all zero-pixels are in a missing modul, we can run without eye twitching. If one of the zero-pixels is part of a DRAGON module which is currently active in the Camera, then the file must be rejected according to current policies.
How do you see the situation after a previously missing (broken) module is activated (repaired) again in a given run? Will you add 7 values for the "new" module by hand?

rlopezcoto:
> It should. Let me update the [ICD](https://edms.cern.ch/document/2411710) accordingly and discuss details there. The main open question on my side is: What does the EVB do in case a pixel is declared with 0 correction factor ("zero-pixel")? Obviously, if all zero-pixels are in a missing modul, we can run without eye twitching. If one of the zero-pixels is part of a DRAGON module which is currently active in the Camera, then the file must be rejected according to current policies.

would you prefer to have these elements of the array filled with `nan`?

> How do you see the situation after a previously missing (broken) module is activated (repaired) again in a given run? Will you add 7 values for the "new" module by hand?

we would need to take a new pedestal run to calculate drs4 corrections and create a new file to be given to EvB in that case.

lstchain/calib/camera/drs4.py
pawel21: > no, what I mean is that since it will be an array later, it may not be the best idea to initialize it as a `None`, but as a `np.empty` array.
> Also, I'm not sure if a variable length array is the best here, or maybe better to have a boolean array:
> 
> `failing_pixels_array = numpy.full((self.n_pixels), False)`
> 
> and set it `True` only for the failing pixels

Done.

pawel21: > > fits_file[2].data
> 
> Please use extnames to identify tables instead of indices.

Done

lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py
pawel21: I changed to `BinTableHDU`

pawel21:
According to comment I have change a bit way to create `fits` file. Now, I use extnames to identify tables.
```
f = fits.open("pedestal_3670.fits")
expected_pixel_ids = f['PRIMARY'].data
pedestal_array = f['PEDESTAL ARRAY'].data # shape (2, 1855, 4096)
failing_pixels_array = f['FAILING PIXELS'].data['failing pixels'] # 0 mean pixel is OK, 1 mean pixels is failing
```

rlopezcoto:
> According to comment I have change a bit way to create `fits` file. Now, I use extnames to identify tables.
> 
> ```
> f = fits.open("pedestal_3670.fits")
> expected_pixel_ids = f['PRIMARY'].data
> pedestal_array = f['PEDESTAL ARRAY'].data # shape (2, 1855, 4096)
> failing_pixels_array = f['FAILING PIXELS'].data['failing pixels'] # 0 mean pixel is OK, 1 mean pixels is failing
> ```

Thanks @pawel21, I have not tested the code, but if you have done it on recent data from the last few days, and also data from the first day when the module started to fail, this PR should be good to go.

rlopezcoto (approved):

FrancaCassol:
Hi,
I  have some coding comments, but better to merge this as it is now and propose improvements in a next PR

FrancaCassol (approved):

PR #605, opened 2021-02-25
Function for calculating effective time: Get obstime real

misabelber:
Re-open  #566 PR with branch directly pushed to the repository instead from fork. 
I have updated the lastest comments on the previous PR, and included the function in lstchain/reco/utils.py as suggested. 
I will close #566.


morcuended:
Thanks @misabelber. Can you please add a test in `test_utils.py` using the dl2 files produced here?

https://github.com/cta-observatory/cta-lstchain/blob/365258afc85e779a77e9d993bda64a9327afd0f7/lstchain/scripts/tests/test_lstchain_scripts.py#L306

maxnoe:
@morcuended I would say there is no need to rely on dl2 data for this. Much simpler would be to do a toy sample like this:


```
t_obs = 600
rate = 10e3
n_events =  np.random.poisson(rate * t_obs)
timestamps = np.random.uniform(0, t_obs, n_events)

events = pd.DataFrame({'dragon_time': timestamps}

assert np.isclose(calculate_time(events), t_obs)
```

lstchain/reco/utils.py
```diff
@@ -611,3 +612,25 @@ def clip_alt(alt):
...
+    """
+    
+    deltat = np.diff(events['dragon_time'])
+    deltat = deltat[(deltat > 0) & (deltat < 0.002)]
+    rate=1/np.mean(deltat)
+    dead_time = np.amin(deltat)
+    t_elapsed = len(events)/rate * u.s
+    t_eff = t_elapsed/(1+rate*dead_time)
+
+    return t_eff
```
chaimain, line 636: Can you also return t_elapsed, as it will be easier then to get the deadtime value as well for a metadata value in event lists?

chaimain (commented):
Just a minor change

lstchain/reco/utils.py
```diff
@@ -611,3 +612,25 @@ def clip_alt(alt):
...
+    -------
+    teff: float
+    """
+    
+    deltat = np.diff(events['dragon_time'])
+    deltat = deltat[(deltat > 0) & (deltat < 0.002)]
+    rate=1/np.mean(deltat)
+    dead_time = np.amin(deltat)
+    t_elapsed = len(events)/rate * u.s
+    t_eff = t_elapsed/(1+rate*dead_time)
```
maxnoe, line 634: Are you sure you need to correct for dead time when estimating the total time through the mean of the differences? Shouldn't that be directly an estimate of the effective time?

lstchain/reco/utils.py
misabelber: Sure!

morcuended:
> @morcuended I would say there is no need to rely on dl2 data for this. Much simpler would be to do a toy sample like this:
> 
> ```
> t_obs = 600
> rate = 10e3
> n_events =  np.random.poisson(rate * t_obs)
> timestamps = np.random.uniform(0, t_obs, n_events)
> 
> events = pd.DataFrame({'dragon_time': timestamps}
> 
> assert np.isclose(calculate_time(events), t_obs)
> ```

Sure

misabelber:
Ok, I will add the test right away.

lstchain/reco/utils.py
misabelber: I remember this was discussed already in the notebook where I introduced the function, but now I'm having doubts... perhaps @moralejo can help to clarify this?

maxnoe: I made a small toy experiment that I think shows pretty clearly that we already account for the dead time by taking the mean of the detected events if I didn't make a wrong assumption.

Screenshot_20210225_161500

@moralejo ?

maxnoe: Sorry, since we get the full observation time back here, we do need to take the dead time into account to get the "live time", so the code should be correct.

misabelber: Thanks for the checking Max,

chaimain:
This seems ok to merge now. @maxnoe @rlopezcoto is it ok to merge this?

lstchain/reco/utils.py
```diff
@@ -611,3 +612,26 @@ def clip_alt(alt):
...
+    t_eff: float
+    t_elapsed: float
+    """
+    
+    deltat = np.diff(events['dragon_time'])
+    deltat = deltat[(deltat > 0) & (deltat < 0.002)]
+    rate=1/np.mean(deltat)
+    dead_time = np.amin(deltat)
+    t_elapsed = len(events)/rate * u.s
+    t_eff = t_elapsed/(1+rate*dead_time)
```
moralejo, line 635: This is not really correct if applied to DL2 events, unless all triggered events are present in the file, which I think we cannot assume to be the case. Are interleaved pedestal and flatfield events still there? (event if they are now, they might eventually be filtered). Also: events for which no or too few pixels survived cleaning may not be there.

morcuended: > Are interleaved pedestal and flatfield events still there?

Yes

> events for which no or too few pixels survived cleaning may not be there.

Not all the events are present in the DL2 files. There are less than the initial 53k that are present in the DL1 files.

maxnoe: I think then we should run this immediately and store it in the metadata of the dl1 files, right?

moralejo: We can solve this if we save in the DL1 file a parameter "delta_t" which is the time difference between each event and _the previously triggered_ one.

maxnoe: Yes, storing delta_t for all events will also just work.

rlopezcoto: instead of storing an additional parameter for all events, can we apply this to DL1 data better so we do not increase the data volume as @maxnoe was proposing above? Is it foreseen that this `delta_t` is needed for anything else apart than calculating the effective time?

moralejo: Calculating livetime in DL1 is not flexible enough. You won't be able to apply e.g. zenith cuts later, except in the chunks for which livetime is precalculated

moralejo: > instead of storing an additional parameter for all events, can we apply this to DL1 data better so we do not increase the data volume as @maxnoe was proposing above? Is it foreseen that this `delta_t` is needed for anything else apart than calculating the effective time?

Not really "needed" but it is a useful diagnostic tool in some cases. If something weird appears in the data you can check if it is related with too close events (e.g. related to the DRS4 corrections)

maxnoe: > Well we need to implement the exponential fit to get the external rate (dead-time-free) instead of the observed event rate.

The mean is the maximum-likelihood estimator for the exponential and the min is the likelihood estimator for the deat time. No need to perform some kind of explicit fit, I think.

maxnoe: > @rlopezcoto : Calculating livetime in DL1 is not flexible enough. You won't be able to apply e.g. zenith cuts later, except in the chunks for which livetime is precalculated

You want to apply zenith cuts *within* a run?

moralejo: > > @rlopezcoto : Calculating livetime in DL1 is not flexible enough. You won't be able to apply e.g. zenith cuts later, except in the chunks for which livetime is precalculated
> 
> You want to apply zenith cuts _within_ a run?

We don't have a predetermined run duration yet, though typically they are short (20').  
But if you want to be able to apply a precise zenith cut to a sample, you'll be forced to cut inside runs, or apply a rough cuts on entire runs. I think the former approach is better, and then you'll want to have the effective time.

moralejo: > > Well we need to implement the exponential fit to get the external rate (dead-time-free) instead of the observed event rate.
> 
> The mean is the maximum-likelihood estimator for the exponential and the min is the likelihood estimator for the dead time. No need to perform some kind of explicit fit, I think.

You are right, it is just slightly modified in the case dead time is present:

<delta_t> = d + 1 / lambda

moralejo: Now taking into account interleaved events at a regular rate lambda_interleaved (@FrancaCassol: are flat-field & pedestal interleaved events exactly interleaved? I mean alternating ff-P-ff-P-.... with same interval between each ff andP):

- Store **delta_t** (time from previous triggered event) in DL1
- Find dead time per event: **d**  (from minimum observed delta_t)
- Make histogram of **delta_t**   **for cosmic events only**  (will anyway be truncated at the time interval between interleaved events)
- fit histogram to an exponential to obtain the true cosmic event rate: **lambda**
- calculate elapsed time (just add up the times between timestamps of **DL2 events**, excluding large values which indicate observation gaps): **t_elapsed**
- Obtain effective time as  **t_eff = t_elapsed / (1+lambda d + lambda_interleaved d)**

misabelber: Ok! After reading all the discussion, seems like there's some work to do before merging this function. We must then work on the storage of delta_t in DL1 files, shall we do that in a separate PR?

moralejo: Assuming lambda=10kHz, lambda_interleaved=200Hz, d = 7e-6 s  The error from neglecting the interleaved term above is just  0.13%  in t_eff. We can also ignore it for now and just write a note about it in the code.

moralejo: > Ok! After reading all the discussion, seems like there's some work to do before merging this function. We must then work on the storage of delta_t in DL1 files, shall we do that in a separate PR?

I don't think another PR is needed. This is precisely fot t_eff calculation, right? So why separate it?

misabelber: Yes! no problem, just asking, then let's implement it here.

rlopezcoto: @moralejo: the problem is that now this needs to be implemented in the `r0_to_dl1` and this would better be suited for `ctapipe_io_lst`, at least for the estable long term version

maxnoe: `ctapipe_io_lst` fills the ctapipe `ArrayEventContainer` which has now field for a `delta_t`, so I think the right place would be `r0_to_dl1`.

moralejo: Indeed, r0_to_dl1.

rlopezcoto: ok for the time being while we still use r0_to_dl1, but what about when we definitely move to the stage1 tool?

moralejo: I don't know if there is anything foreseen, since for now stage-1 is more MC-oriented. @maxnoe do you know if there is any discussion on this on the ctapipe side?

maxnoe: I think this has to be dealed with "the proper way" in ctapipe. There is no such info yet, but I imaging the way this is planned to be dealt with in CTA is a monitoring table where the accumulated dead time is store every couple of seconds?

How will LST provide this information? Estimating this from event distributions is probably not the best way.

maxnoe: In FACT we store the dead time fraction every 5 seconds.

moralejo: The relevant timescale seems to be 10 s: https://jama.cta-observatory.org/perspective.req#/items/26761?projectId=11 , and a monitoring table seems to be the place to keep it, yes.

moralejo: > How will LST provide this information? Estimating this from event distributions is probably not the best way.

We could just assume fixed deadtime per triggered event, but I think that would be less accurate. The delta_t distribution does not have a perfectly 'sharp rise' at the expected deadtime per event.

I'd say that when we move to stage1 we should make this in a 'ctapipe compliant' way, perhaps filling the said monitoring table in the R1 event source.

maxnoe: Wait, shouldn't be there some hardware component for LST that just writes this information?

moralejo: > Wait, shouldn't be there some hardware component for LST that just writes this information?

As said in the recent call: not that I know of, but with the trigger rate and the known (or empirically determined) dead time per event it should be straightforward to obtain that, e.g. in the R1 event source.

misabelber:
So, I have a doubtÑ According to which was discussed in the meeting, we want to include the deltat values in the step from dl1 to dl2. I was about to implement it inside the "apply_models"  function of reco/dl1_to_dl2.py, but I realized that it would not work, because in the lstchain_dl1_to_dl2.py script, we filter the events before applying the models, so the calculated deltats will not be correct. Shall I better move the deltat calculation to the lstchain_dl1_to_dl2.py script, and put it before the filtering OR else, I could leave it in apply_models but we change the order and we first apply models and afterwards filter? Any preferences about this or suggestions where to better put it?

rlopezcoto:
> So, I have a doubt: According to which was discussed in the meeting, we want to include the deltat values in the step from dl1 to dl2. I was about to implement it inside the "apply_models" function of reco/dl1_to_dl2.py, but I realized that it would not work, because in the lstchain_dl1_to_dl2.py script, we filter the events before applying the models, so the calculated deltats will not be correct. Shall I better move the deltat calculation to the lstchain_dl1_to_dl2.py script, and put it before the filtering OR else, I could leave it in apply_models but we change the order and we first apply models and afterwards filter? Any preferences about this or suggestions where to better put it?

I think it sounds better if you move the `delta_t` calculation into the `lstchain_dl1_to_dl2.py` script, but by heart I remember that this was very minimal and only callin the `dl1_to_dl2` function, right? If this function is called after the filtering of events, then the calculation should be done before.

lstchain/reco/utils.py
```diff
@@ -613,6 +614,28 @@ def clip_alt(alt):
...
+
+    Returns
+    -------
+    events: pandas DataFrame of dl2 events with delta_t
+    """
+
+    # Get delta t of real data and add it to the data frame
+    if 'dragon_time' in events.columns:
+        time= np.array(events.dragon_time)
+        delta_t = np.insert(time[1:]-time[:-1],0,0)
```
moralejo, line 635: Can't you use np.diff ?

misabelber: Yep, thanks!

lstchain/reco/utils.py
```diff
@@ -611,3 +613,48 @@ def clip_alt(alt):
...
+
+def add_delta_t_key(events):
+    """
+    Adds the time difference with the previous event to a real data
+    dataframe.
+    Should be only used only with non-filtered data frames,
+    so events are consecutive.
+    Parameters
+    ----------
+    events: pandas DataFrame of dl2 events
```
rlopezcoto, line 625: these should still be dl1 events, right? (no models applied)

rlopezcoto: thanks!

rlopezcoto (approved):

chaimain:
@misabelber the function is repeated here https://github.com/cta-observatory/cta-lstchain/blob/b7d140f49c63689e329dc74aace5f3e613df2cba/lstchain/mc/sensitivity.py#L156. I suppose it will be better to use the function in this PR there as well.

misabelber:
Ah yes, I was waiting for the other one to be approved.

lstchain/reco/utils.py
moralejo: See also #615 , I implemented a slightly simplified version of the method outlined above.

PR #607, opened 2021-02-25
Dl1 datacheck

moralejo:
Adapt the DL1 data check to the changes in the r0 to DL1 stage.


lstchain/image/muon/muon_analysis.py
```diff
@@ -193,7 +191,7 @@ def analyze_muon_event(subarray, event_id, image, geom, equivalent_focal_length,
          ])
 
     if candidate_clean_ring:
-        intensity_fitter = MuonIntensityFitter(subarray)
+        intensity_fitter = MuonIntensityFitter(subarray, hole_radius_m=0.308)
```
rlopezcoto, line 194: can't we define this `hole_radius` as a global variable? It will not change for the LST

lstchain/datachecks/dl1_checker.py
```diff
@@ -669,7 +669,7 @@ def plot_datacheck(datacheck_filename, out_path=None, batch=False, muons_dir=Non
                                  ['Flat-field mean time (ns)',
                                   'Flat-field time std dev (ns)',
                                   'FLATFIELD, pixel-wise pulse time info'],
-                                 batch, pagesize)
```
morcuended, line 672: ups! my bad

morcuended (approved):

lstchain/image/muon/muon_analysis.py
moralejo: Not worth, we will move to stage 1 in the next to next release.
I added that explicit argument just to suppress a warning (about MSTs and SSTs not being present). By default it will notice the only present telescope is an LST, and use the same value 0.308, but also throw the annoying warning with every instantiation.

morcuended:
Now I remember, what about `trigger_type` instead of `ucts_trigger_type` when no UCTS data are not available?

maxnoe:
The event source will use `tib_masked_trigger` if ucts data is not available to create the `event.trigger.event_type`. 

Maybe a better solution would thus be to convert the `EventType` back into the `TriggerType` instead of duplicating the logic again to choose which one to save

moralejo:
In the datacheck so far we were using only ucts. But for some data it may be good to use TIB indeed. 
@maxnoe I don't get what you propose, sorry, can you clarify?

maxnoe:
Sorry, I misremembered. I though you only had one field in the dl1 container for trigger type and always filled `ucts`, but you have `ucts_trigger_type` and `trigger_type` which is filled with tib.

moralejo:
@morcuended : from the DL1 data I don't know though when to switch to TIB is convenient. Any proposal?

morcuended:
Does the `event.trigger.event_type` have always the available trigger source either ucts or tib? If so we could use it. Otherwise, we could check if all `ucts_trigger_type` and `trigger_type` values are set to `-1` or there are a substantial number of `ucts_trigger_type==42` events.

image

image

maxnoe:
> Does the event.trigger.event_type have always the available trigger source either ucts or tib? 

Yes, see https://github.com/cta-observatory/ctapipe_io_lst/blob/e3d9fc29d780edbe7cbe55587f4b673682ba898d/ctapipe_io_lst/__init__.py#L500

morcuended:
> > Does the event.trigger.event_type have always the available trigger source either ucts or tib?
> 
> Yes, see https://github.com/cta-observatory/ctapipe_io_lst/blob/e3d9fc29d780edbe7cbe55587f4b673682ba898d/ctapipe_io_lst/__init__.py#L500

Then we could just use `event_type` instead.

moralejo:
We can always use event_type, yes. Actually I realized I am still not using it for recognizing flatfield events. Will change and test it later today.

moralejo:
Ok, use of event_type done and tested, but I think this revealed an issue:

image

This is a run (1881) in which tib is missing and ucts trigger type is wrong. From the plot you can see that some  flatfield events are wrongly identified as pedestals. This means that the "pedestal tag" (ucts in this case) has precedence over the empirical identification of flatfield events.  The earlier behaviour of the datacheck was: empirical identification of FF has precedence - just because it is way more reliable.

@maxnoe: I think we can for now assume that the flatfield "empirical tagging" is the safest one - any objection for us to change that in ctapipe_io_lst?

maxnoe:
> . This means that the "pedestal tag" (ucts in this case) has precedence over the empirical identification of flatfield events. The earlier behaviour of the datacheck was: empirical identification of FF has precedence - just because it is way more reliable.

On the implementation side, this "precedence" comes from the fact that the identification code is only run on "physics" events. We can change to always run it, no problem.

moralejo:
> > . This means that the "pedestal tag" (ucts in this case) has precedence over the empirical identification of flatfield events. The earlier behaviour of the datacheck was: empirical identification of FF has precedence - just because it is way more reliable.
> 
> On the implementation side, this "precedence" comes from the fact that the identification code is only run on "physics" events. We can change to always run it, no problem.

Yes, this is what I meant. Can you do it?

maxnoe:
https://github.com/cta-observatory/ctapipe_io_lst/pull/74

moralejo:
> [cta-observatory/ctapipe_io_lst#74](https://github.com/cta-observatory/ctapipe_io_lst/pull/74)

Thanks! Will try once merged, but should certainly work.

moralejo:
image
Ok, no flatfields labeled as pedestals now  (reminder: these "pedestals" are cosmics because this is run 1881 with wrong ucts info!)

rlopezcoto:
> image
> Ok, no flatfields labeled as pedestals now (reminder: these "pedestals" are cosmics because this is run 1881 with wrong ucts info!)

Great, anything else missing @morcuended, @maxnoe ?

rlopezcoto (approved):

morcuended (approved):
Good! Fine with me

PR #608, opened 2021-02-26
Use context manager for creating PDFPages in DRS4 and calib visualization functions

morcuended:
Coming from #564

**Issue:**
> I was trying to use plot_pedestals from plot_drs4.py and plot_all from plot_calib.py (visualization module) in a single script to produce drs4 and calibration data-check PDF files. The first one was produced without problems, however, the second file still contained plots from the previous file overlaid. My guess is that after producing the first file the PdfPages object was not properly closed since it was open twice.
> 
> This problem does not happen when using the drs4 and calib visualization functions in different scripts.

**Changes:**
- Use a context manager instead to create and close the PDF file.
- Close `plt` objects to avoid cluttering when using the two functions in the same script.
- Add the possibility of changing the `sample_size` in `plot_pedestals`. By default is set to 1000 events as it was before. In this way, we can use observed R0 files to test the function.
- Add tests for both functions: `plot_drs4.plot_pedestals` and `plot_calib.plot_all`.
- Black formatted.


morcuended:
I'm not able to create a test for this function with the available R0 test files. Would it be possible to add a chopped R0 DRS4 run? (eg run 2005)?

morcuended:
In order to avoid cluttering of plt objects I had to close the figures after saving them.



This is what was happening before when using both functions (drs4 and calib) in the same script:
```python
/fefs/home/daniel.morcuende/software/cta-lstchain/lstchain/visualization/plot_calib.py:89: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to each axes instance.
  plt.subplot(pad)

```
image

maxnoe:
> I'm not able to create a test for this function with the available R0 test files. Would it be possible to add a chopped R0 DRS4-run file? (eg run 2005)?

I added `LST-1.{1,2,3,4}.Run02005.0000_first50.fits.fz` to the dataserver

morcuended:
> > I'm not able to create a test for this function with the available R0 test files. Would it be possible to add a chopped R0 DRS4-run file? (eg run 2005)?
> 
> I added `LST-1.{1,2,3,4}.Run02005.0000_first50.fits.fz` to the dataserver

Thanks! In the end, I realized that the problem was the sample size that the script was using (minimum 1000 events). So I just made that flexible. In any case, those files can be in principle used to create tests for the drs4 and calibration scripts.

lstchain/visualization/plot_calib.py
```diff
@@ -49,297 +49,303 @@ def read_file(file_name, tel_id=1):
...
-        rms = np.std(n_pe[select])
-        label=f"Median {median:3.2f}, std {rms:3.2f}"
-        plt.hist(n_pe[select], bins=50,label=label)
-        plt.legend()
-        plt.subplots_adjust(top=0.92)
-        if plot_file != "none":
-            pp.savefig(plt.gcf())
-
-
-    if plot_file != "none":
```
rlopezcoto, line 344: is this condition removed?

morcuended: You only open the `PDFPages` `if plot_file is not None` and the `with` context manager closes it afterward. I just close the plt object right after saving the figure.

rlopezcoto (approved):

PR #609, opened 2021-03-01
lstchain-sensitivity tools refactoring

misabelber:
Adapted #532  to current master version, I closed the old PR and created this one following the current procedure.

### DIFF
--- a/lstchain/mc/plot_utils.py
+++ b/lstchain/mc/plot_utils.py
@@ -13,6 +13,7 @@ __all__ = [
     'format_axes_sensitivity',
     'plot_MAGIC_sensitivity',
     'plot_Crab_SED',
+    'plot_LST_preliminary_sensitivity',
     'plot_sensitivity',
     'sensitivity_minimization_plot',
     'sensitivity_plot_comparison',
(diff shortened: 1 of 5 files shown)


lstchain/mc/plot_utils.py
```diff
@@ -135,24 +136,48 @@ def format_axes_sensitivity(ax):
...
+    Parameters
+    --------
+    ax:    `matplotlib.pyplot.axis`
+
+    Returns
+    --------
+    ax:    `matplotlib.pyplot.axis`
+
+    """
+    s = np.loadtxt('../spectra/data/sensitivity.txt', skiprows = 1)
```
rlopezcoto, line 152: this file is missing

rlopezcoto (commented):
thanks @misabelber. Since these changes were used to test the sensitivity calculations and compare them with those of pyirf, I propose to merge them in case future comparisons are needed.

lstchain/mc/plot_utils.py
misabelber: Yes, I think I will remove this function, since I don't think is somenthing very general. I will take a look at the rest of the codecov issues asap.

rlopezcoto (approved):

PR #611, opened 2021-03-02
Move configuration of LST event source to json file

moralejo:
LST event source configuration is now in the json file (though command-line options, if present, override the settings in the file). So there should be no change of behavior. Now one can select the preferred trigger tag (ucts or tib) by setting it in the file.

### DIFF
--- a/lstchain/scripts/lstchain_data_r0_to_dl1.py
+++ b/lstchain/scripts/lstchain_data_r0_to_dl1.py
@@ -23,6 +23,7 @@ import logging
 import sys
 from pathlib import Path
 
+from lstchain.io import  standard_config
 from lstchain.io.config import read_configuration_file
 from lstchain.paths import parse_r0_filename, run_to_dl1_filename, r0_to_dl1_filename
 from lstchain.reco import r0_to_dl1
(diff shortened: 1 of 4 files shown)


lstchain/scripts/lstchain_data_r0_to_dl1.py
```diff
@@ -151,21 +152,35 @@ def main():
...
+    lst_event_source['EventTimeCalculator']['dragon_counter0'] = \
+        args.dragon_counter0
+    lst_event_source['EventTimeCalculator']['ucts_t0_tib'] = \
+        args.ucts_t0_tib
+    lst_event_source['EventTimeCalculator']['tib_counter0'] = \
+        args.tib_counter0
+    lst_event_source['PointingSource']['drive_report_path'] = \
+        args.pointing_file
+    lst_r0_corrections = lst_event_source['LSTR0Corrections']
+    lst_r0_corrections['drs4_pedestal_path'] = args.pedestal_file
```
rlopezcoto, line 174: I am maybe missing something, but what if we do not give any `pedestal_file` through the command line?

rlopezcoto (commented):
tests are not passing and also I'm trying to understand (below) if this is equivalent to what we have now...

moralejo:
> tests are not passing and also I'm trying to understand (below) if this is equivalent to what we have now...

Weird, they pass locally! ??

lstchain/scripts/lstchain_data_r0_to_dl1.py
moralejo: Behaviour would be the same as before, i.e. it is a required argument. But this means those settings cannot be configured through the json file as of now.

moralejo:
> > tests are not passing and also I'm trying to understand (below) if this is equivalent to what we have now...
> 
> Weird, they pass locally! ??

I correct myself: they passed locally until the last, apparently innocuous change. WIP

lstchain/scripts/lstchain_data_r0_to_dl1.py
moralejo: > Behaviour would be the same as before, i.e. it is a required argument. But this means those settings cannot be configured through the json file as of now. Will fix.

Done

rlopezcoto (approved):

PR #612, opened 2021-03-02
Tailcut cleaning with additional threshold based on interleaved pedestals

pawel21:
Hi,

I open new PR for script which recompute parameters based on dl1 file using information from interleaved pedestal.
To reprocess dl1 file with another cleaning settings:
```
python lstchain_dl1ab.py -f dl1_LST-1.Run03004.0000.h5  -o dl1b_LST-1.Run03004.0000.h5  --no-image True -c ../data/cleaning_pedestal_config.json
```


rlopezcoto:
Great, thanks @pawel21!
This PR comes from #436 in which an extended discussion was already done. Please keep that in mind when reviewing this PR.

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,73 @@
...
+        path to dl1 file
+    Returns
+    -------
+    bias, std: np.ndarray, np.ndarray
+        bias and std in p.e.
+    """
+    try:
+        f = tables.open_file(dl1_file)
+    except Exception:
+        raise RuntimeError("No found dl1 file")
```
maxnoe, line 22:
```suggestion
        raise IOError("Error opening DL1 file")
```

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,73 @@
...
+    ped = f.root[dl1_params_tel_mon_ped_key]
+    ped_charge_mean = np.array(ped.cols.charge_mean)
+    ped_charge_std = np.array(ped.cols.charge_std)
+    calib = f.root[dl1_params_tel_mon_cal_key]
+    dc_to_pe = np.array(calib.cols.dc_to_pe)
+    ped_charge_mean_pe = ped_charge_mean * dc_to_pe
+    ped_charge_std_pe = ped_charge_std * dc_to_pe
+    f.close()
```
maxnoe, line 30: please use a context manager for opening and closing files:

```
with tables.open_file(dl1_file) as f:
    ...
```

pawel21: Done

FrancaCassol:
Hi @pawel21,

before starting a more detailed review, a question: why this is not a Tool?

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,70 @@
+import tables
```
FrancaCassol, line 1: The name of this module is too general for the code that is contains, that is instead rather specific.
I would not create at all this module and rather put the functions that it contains inside the main script or, better, tool

maxnoe:
@FrancaCassol That's not the main issue here. The main issue is that it is basically a copy of the full script to add a different cleaning.

@pawel21 please change to adding an *option* to the existing script to avoid all this duplicated code.
It will be very hard to maintain both versions, any time we need to change something that affects the rest of the code.

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    input_filename: str
+        path to dl1 file
+    sigma_clean: float
+        cleaning level parameter
+    Returns
+    -------
+    picture_thresh: np.ndarray
+        picture threshold calculated using interleaved pedestal events
+    """
+    high_gain = 0
```
FrancaCassol, line 52: Better take the HG index from the ctapipe_lst_io.constants module

lstchain/calib/camera/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    # std values from calibration run.
+    # Correct interleaved pedestal std array should have shape (2,2,1855)
+    if ped_std_pe.shape[0] == 2:
+        interleaved_events_id = 1
+    else:
+        interleaved_events_id = 0
+    threshold_clean_pe = ped_mean_pe + sigma_clean*ped_std_pe
+    # find pixels with std = 0 and mean = 0 <=> dead pixels in interleaved
+    # pedestal event likely due to stars
+    dead_pixel_ids = np.where(threshold_clean_pe[interleaved_events_id, high_gain, :] == 0)[0]
```
FrancaCassol, line 65: better take the bad pixels directly from the calib.unusable_pixels

lstchain/io/config.py
```diff
@@ -60,3 +61,22 @@ def replace_config(base_config, new_config):
         config[k] = new_config[k]
 
     return config
+
+def get_cleaning_parameters(config, clean_method_name):
```
FrancaCassol, line 65: Whit a tool all this is not necessary ;-)

FrancaCassol (changes requested):

FrancaCassol:
> @FrancaCassol That's not the main issue here. The main issue is that it is basically a copy of the full script to add a different cleaning.

I see, I was indeed wondering for the reason for two scripts (I thought one had to replace the other). In any case, I think we must at certain moment no more accept the adding of new scripts if a Tool is a better option. Also it would be better to use more self-explaining names.

maxnoe:
@FrancaCassol exactly, but for this new feature I would be fine with adding a new option to the existing script instead of going the full way to convert it to a `Tool`. But we should not have to very, very similar scripts with much duplicated code doing only slightly different things.

rlopezcoto:
> In any case, I think we must at certain moment no more accept the adding of new scripts if a Tool is a better option

Regarding this point, the idea is that from next major lstchain release (v0.7) on, we will work to convert all the scripts into Tools and we will encourage people writing only Tools instead of scripts.

pawel21:
> @pawel21 please change to adding an _option_ to the existing script to avoid all this duplicated code.
> It will be very hard to maintain both versions, any time we need to change something that affects the rest of the code.

Hi Max, 
Thank you for your review. 
What do you mean by "adding an _option_ to the existing script"? Should script use standard config file and depend on this _option_ use or not cleaning with pedestal information?

maxnoe:
Yes, don't add a new script. Make the old script configurable (either via command line options or config file or both) to use or not use the new cleaning method.

lstchain/calib/camera/utils.py
pawel21: > I would not create at all this module and rather put the functions that it contains inside the main script or, better, tool

but, this module is used also in notebook to show how pedestal cleaning is working.

pawel21:
Hi,

According to comments, I have added option to `lstchain_dl1ab.py` to allow use pedestal cleaning.
I have also added parameters for pedestal cleaning in `lstchain_standard_config.json`.
Now, in order to reprocess dl1 file with pedestal cleaning settings:
```
python lstchain_dl1ab.py -f dl1_LST-1.Run03004.0000.h5 -o dl1b_LST-1.Run03004.0000.h5 
--no-image True 
-c ../data/lstchain_standard_config.json 
--pedestal-cleaning True 
```

lstchain/calib/camera/utils.py
pawel21: Done, thank you.

lstchain/calib/camera/utils.py
pawel21: Done

lstchain/io/io.py
```diff
@@ -50,12 +50,12 @@
     'write_calibration_data',
     'read_dl2_to_pyirf',
     'read_dl2_params',
-    'extract_observation_time',
-    'merge_dl2_runs'
```
rlopezcoto, line 54: `'merge_dl2_runs'` is still present in the file

rlopezcoto (commented):
in the PR, the script `lstchain_dl1ab_pedestal_cleaning.py` is still present, could you remove it by using `git rm lstchain_dl1ab_pedestal_cleaning.py`?

maxnoe:
@pawel21 Nice, thank you! See how much smaller the diff is now?

lstchain/io/io.py
pawel21: > `'merge_dl2_runs'` is still present in the file

Should I remove?

rlopezcoto: you have removed it in your PR, but the function is still included in the `io.py`, therefore you should include it again in the `__all__` definition.

rlopezcoto:
great, thanks @pawel21, this should be good to go. @maxnoe and @FrancaCassol could you please have a final look to see if everything is solved?

moralejo:
> Hi,
> 
> According to comments, I have added option to `lstchain_dl1ab.py` to allow use pedestal cleaning.
> I have also added parameters for pedestal cleaning in `lstchain_standard_config.json`.
> Now, in order to reprocess dl1 file with pedestal cleaning settings:
> 
> ```
> python lstchain_dl1ab.py -f dl1_LST-1.Run03004.0000.h5 -o dl1b_LST-1.Run03004.0000.h5 
> --no-image True 
> -c ../data/lstchain_standard_config.json 
> --pedestal-cleaning True 
> ```

The pedestal cleaning cannot be activated now through the config card, right? Not extremely urgent (no reason to delay), but would be convenient to have.

pawel21:
> The pedestal cleaning cannot be activated now through the config card, right? Not extremely urgent (no reason to delay), but would be convenient to have.
Right, parameters for pedestal cleaning are in the config card, but to use pedestal cleaning We have to set `--pedestal-cleaning True `

moralejo:
> /nfs/cta-ifae/moralejo/CTA/ctasoft/cta-lstchain/lstchain/scripts/lstchain_dl1ab.py:168: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe

Can you please suppress this warning? It appears several times in an execution.

maxnoe:
Just replace `np.int` with `np.int64`

pawel21:
> Can you please suppress this warning? It appears several times in an execution. 

Sorry, I don't know how to do it in correct way with ` logging`. Maybe @maxnoe have any idea? 

I think this is no related to my change of code.

maxnoe:
@pawel21 no need to supress it. Just fix it. replace `np.int` with `np.int64`

pawel21:
> @pawel21 no need to suppress it. Just fix it. replace `np.int` with `np.int64`

Done, thank you

maxnoe:
I did it for the rest of the occurences in lstchain here #621

lstchain/calib/camera/utils.py
FrancaCassol: We should then give a more clear status to this module.  
Is it for generic functions related to DL1 calibration data? (eg. dl1_calibration_utils.py) Or is it only for the pixel threshold estimation? (e.g. pixel_threshold_estimation.py)
According to that I would change the name of the module (and eventually of the functions) so that it is more self-explaining.

moralejo: Not urgent, right? We can leave it like this for the coming release.

moralejo:
https://github.com/cta-observatory/cta-lstchain/blob/56a2ff7bb69c4a2c7fea87c96a743b2f99bb4945/lstchain/scripts/lstchain_dl1ab.py#L165-L173

I know this might as well go in another PR, but for simplicity I suggest to do it here:

Use of main island (defined as highest # of pixels, not brightest one) is hardcoded (same as in ro_to_dl1). But I think that with this new cleaning approach we may eventually use the whole image, so I will set it as configurable.

moralejo:
Tests don't pass - sorry @pawel21 , they also failed locally, but I thought it was due to the problem of protozfitsreader on Mac. Will check.

lstchain/calib/camera/utils.py
pawel21: I renamed `utils.py` -> `pixel_threshold_estimation.py`

FrancaCassol: > Not urgent, right? We can leave it like this for the coming release.

@moralejo, why should we pospone to a coming release an easy improvement of new added code?

moralejo:
I just made a small modification so that the use of main island (or full image) for parametrization can be selected through the config file.

moralejo:
> I just made a small modification so that the use of main island (or full image) for parametrization can be selected through the config file.

Tests not passing, apologies for that, looking into it.

rlopezcoto:
@pawel21 is there anything else missing in this PR or did you already work out all suggestions and comments?

pawel21:
>  is there anything else missing in this PR or did you already work out all suggestions and comments?

I think, everything is ready for merge into master.

rlopezcoto:
@moralejo, @FrancaCassol , @maxnoe ?

lstchain/calib/camera/utils.py
moralejo: Hi @FrancaCassol, there were no other pending changes, right?

moralejo:
From my side it is ok, I have been testing it and did not find any problem - we'll have to decide on the settings, but that is a separate issue.

moralejo (approved):

FrancaCassol (approved):

PR #613, opened 2021-03-03
Use new function for effective time in sensitivity code

misabelber:
Sorry for open other PR, forgot to substitute the old function for the new effective time function.

### DIFF
--- a/lstchain/mc/sensitivity.py
+++ b/lstchain/mc/sensitivity.py
@@ -5,7 +5,7 @@ from .plot_utils import sensitivity_minimization_plot, plot_positions_survived_e
 from .mc import rate, weight
 from lstchain.spectra.crab import crab_hegra
 from lstchain.spectra.proton import proton_bess
-from lstchain.reco.utils import reco_source_position_sky
+from lstchain.reco.utils import reco_source_position_sky, get_effective_time
 from astropy.coordinates.angle_utilities import angular_separation
 from lstchain.io import read_simu_info_merged_hdf5
 from lstchain.io.io import dl2_params_lstcam_key
@@ -16,7 +16,6 @@ __all__ = [
     'read_sim_par',
     'process_mc',
     'process_real',
-    'get_obstime_real',
     'get_weights',
     'samesign',
     'diff_events_after_cut',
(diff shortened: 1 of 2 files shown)


lstchain/mc/tests/test_senstivity.py
```diff
@@ -34,22 +33,17 @@ def test_read_sim_par(simulated_dl1_file):
 
 @pytest.mark.run(after='test_apply_models')
 def test_process_mc(simulated_dl2_file):
+    #Write some delta_t in the example dl2 file
+    dl2 = pd.read_hdf(simulated_dl2_file, key=dl2_params_lstcam_key)
+    dl2["delta_t"]=0.0005
```
rlopezcoto, line 38: `simulated_dl2_file` has only one entry?

misabelber: No, but this way it writes the same value on all events which I think is ok for testing purposes.

lstchain/mc/sensitivity.py
```diff
@@ -130,7 +129,7 @@ def process_mc(dl2_file, mc_type):
 def process_real(dl2_file):
 
     events = pd.read_hdf(dl2_file, key = dl2_params_lstcam_key)
-    obstime_real = get_obstime_real(events)
+    obstime_real = get_effective_time(events)[0]
```
moralejo, line 132: Please note that with the recent changes in get_effective_time, this will now be an astropy quantity with units of seconds. Make sure this does not make trouble later!

moralejo: Currently tests are failing because get_effective_time is getting an events table with no "dragon_time"

morcuended: `simulated_dl2` does not have `dragon_time`. Is that the reason?

moralejo: Seems so. If it is simulated, it shouldn't need either to calculate the effective time.

misabelber: I updated the test to include an artifical dragon_time in the test files, just as it was donde with delta_t. Now the test are passing but  not codecov/project and I can't understand why... I'm not yet very used to this codecov. Does anybody has an idea about what can I do to pass that test?

rlopezcoto: tests are passing now, anything else missing for this PR?

misabelber: On my side it is finished.

rlopezcoto (approved):

PR #615, opened 2021-03-04
Modifications in the effective time calculation routine & test

moralejo:


lstchain/reco/tests/test_utils.py
```diff
@@ -2,6 +2,7 @@
 import astropy.units as u
 from astropy.time import Time
 from astropy.coordinates import SkyCoord
+from ctapipe.containers import EventType
```
chaimain, line 5: Where is this being used?

lstchain/reco/utils.py
```diff
@@ -21,11 +21,12 @@
 from astropy.time import Time
 from astropy.utils import deprecated
 from ctapipe.coordinates import CameraFrame
+from ctapipe.containers import EventType
```
chaimain, line 24: Same as above.

lstchain/reco/utils.py
```diff
@@ -649,12 +655,37 @@ def get_effective_time(events):
     t_eff: float
     t_elapsed: float
     """
-    
+
+    # time differences between the events in the table (which in general are
+    # NOT all triggered events):
+    time_diff = np.diff(events.dragon_time)
```
chaimain, line 661: Better to use `events["dragon_time"]` so that it is more general and can be used when we read the events as a QTable.

lstchain/reco/utils.py
```diff
@@ -649,12 +655,37 @@ def get_effective_time(events):
...
+    # might indicate the DAQ was stopped (e.g. if the table contains more
+    # than one run). We set 0.1 s as limit to decide a "break" occurred:
+    t_elapsed = np.sum(time_diff[time_diff<0.1])
+
+    # Get delta_t, the time elapsed since the previous triggered event.
+    # We exclude the first event, for which delta_t might be set to 0 if it is
+    # the first even in a file.
+    # We want this to calculate the actual trigger rate, on which the dead
+    # time depends.
     delta_t = events.delta_t[1:]
```
chaimain, line 672: As mentioned above, it should be `events["delta_t"]`

lstchain/reco/utils.py
```diff
@@ -21,11 +21,12 @@
 from astropy.time import Time
 from astropy.utils import deprecated
 from ctapipe.coordinates import CameraFrame
+from ctapipe.containers import EventType
 
 from . import disp
 
 __all__ = [
-    'add_delta_t_key'
+    'add_delta_t_key',
```
rlopezcoto, line 29: how did this past the tests without that comma??

lstchain/reco/utils.py
rlopezcoto: why is this more general for this particular case?

lstchain/reco/utils.py
rlopezcoto, line 672: do you start on the second index (`events.delta_t[1:]`) because the first one is set to 0?
if for whatever reason we calculate `delta_t` per subrun and then calculate the effective time per run, then we will have several zeros in the `delta_t` table for the first run of every subrun. Although it would be better not to have this behaviour and calculate the `delta_t` table per run, I would add some check like:
`delta_t = events.delta_t[events.delta_t > 0.]`

lstchain/reco/utils.py
chaimain: Accessing the column as an attribute is an extra feature for pandas.DataFrame, but when I use this function for my DL3 tool, where I access the DL2 file as an astropy.table object, I cannot use this method of accessing.

rlopezcoto: but there is not problem to use it here, where we are not expecting this functionality to change, right?

lstchain/reco/tests/test_utils.py
moralejo: Thanks! It was used at some point, but no longer.

lstchain/reco/utils.py
moralejo: Done, thx!

lstchain/reco/utils.py
moralejo: I have no idea! I saw it not because of any misbehaviour, but thanks to PyCharm checks!

lstchain/reco/utils.py
moralejo: > why is this more general for this particular case?
Not sure what you mean, but to obtain the elapsed time we need the intervals between the events present in the file, that's all.

moralejo: > do you start on the second index (`events.delta_t[1:]`) because the first one is set to 0?

yes

> if for whatever reason we calculate `delta_t` per subrun and then calculate the effective time per run, then we will have several zeros in the `delta_t` table for the first run of every subrun. Although it would be better not to have this behaviour and calculate the `delta_t` table per run, I would add some check like:
> `delta_t = events.delta_t[events.delta_t > 0.]`

Good point, thanks!, will do!

rlopezcoto: but if for example we add `delta_t` to dl2 files that come from dl1 subruns, the first event of every subrun will always have `delta_t=0`.
If we then merge all these dl2 subruns and calculate the effective time afterwards, the `np.amin(delta_t)` will be 0

rlopezcoto: because by skipping the first event of the `events` data frame (`events.delta_t[1:]`) you are skipping the first event of the file, but the file could have been merged after the `delta_t` have been calculated

moralejo: Thanks, all the suggestions implemented.
Besides, I realized that in the test the true livetime for comparison could be better calculated. Now the accuracy of the method is ~0.02%, so I also reduced the tolerance in the test.

lstchain/reco/utils.py
```diff
@@ -649,12 +654,38 @@ def get_effective_time(events):
...
+    # might indicate the DAQ was stopped (e.g. if the table contains more
+    # than one run). We set 0.1 s as limit to decide a "break" occurred:
+    t_elapsed = np.sum(time_diff[time_diff<0.1])
+
+    # Get delta_t, the time elapsed since the previous triggered event.
+    # We exclude the first event, for which delta_t might be set to 0 if it is
+    # the first even in a file.
+    # We want this to calculate the actual trigger rate, on which the dead
+    # time depends.
+    delta_t = events.delta_t
```
chaimain, line 671: Missed changing the access method here

lstchain/reco/utils.py
chaimain: @rlopezcoto No, on its own there is no problem here. But as we need to use it while creating DL3 file, it has to be without using this method of accessing

lstchain/reco/utils.py
chaimain: > > why is this more general for this particular case?
> 
> Not sure what you mean, but to obtain the elapsed time we need the intervals between the events present in the file, that's all.

I think this was copied for my earlier comment, by mistake

moralejo: > As mentioned above, it should be `events["delta_t"]`

ok!

lstchain/reco/utils.py
moralejo: > Missed changing the access method here

That was fast @chaimain! Done now. Please test.

misabelber (approved):
Thanks @moralejo for implementing the changes! After including the mod in the accessing method for @chaimain it looks perfect to me. Also good point @rlopezcoto  on addressing the problem of only accessing the first element, I did not thought about merged dl2 files.

moralejo:
> Thanks @moralejo for implementing the changes! After including the mod in the accessing method for @chaimain it looks perfect to me. Also good point @rlopezcoto on addressing the problem of only accessing the first element, I did not thought about merged dl2 files.

Yes, although in general it is better to calculate live time in small chunks if possible (then add up), because I think the method in principle requires stable rate.

rlopezcoto (approved):

PR #617, opened 2021-03-04
Make get_effective_time accept QTable as input

moralejo:
And always return times in seconds in any case

### DIFF
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -658,9 +658,9 @@ def get_effective_time(events):
     timestamp = np.array(events['dragon_time'])
     delta_t = np.array(events['delta_t'])
 
-    if type(timestamp) is not u.Quantity:
+    if not isinstance(timestamp, u.Quantity):
         timestamp *= u.s
-    if type(delta_t) is not u.Quantity:
+    if not isinstance(delta_t, u.Quantity):
         delta_t *= u.s
 
     # time differences between the events in the table (which in general are
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -652,13 +652,11 @@ def get_effective_time(events):
 
     Returns
     -------
-    t_eff: float
-    t_elapsed: float
+    t_eff: astropy Quantity (in seconds, if input has no units)
+    t_elapsed: astropy Quantity (ditto)
     """
     timestamp = np.array(events['dragon_time'])
     delta_t = np.array(events['delta_t'])
-
-    print('HOLA', type(timestamp))
 
     if type(timestamp) is not u.Quantity:
         timestamp *= u.s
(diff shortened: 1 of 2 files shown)


lstchain/reco/utils.py
```diff
@@ -639,36 +639,43 @@ def add_delta_t_key(events):
...
 
     Returns
     -------
     t_eff: float
     t_elapsed: float
     """
+    timestamp = np.array(events['dragon_time'])
+    delta_t = np.array(events['delta_t'])
+
+    print('HOLA', type(timestamp))
```
maxnoe, line 661: debug print?

lstchain/reco/utils.py
```diff
@@ -639,36 +639,43 @@ def add_delta_t_key(events):
...
+
+    if type(timestamp) is not u.Quantity:
```
maxnoe, line 663: this works, but the canonical way of testing this would be `if not isinstance(timestamp, u.Quantity)`.

maxnoe: However, you can just do:

```
timestamp = u.Quantity(timestamp, u.s, copy=False)
```
This will convert the series / array without units to having u.s but will also work when timestmap is already a quantity with time units.

lstchain/reco/utils.py
moralejo: Ooooook, just wanted to give the code some human warmth

lstchain/reco/utils.py
```diff
@@ -639,36 +639,41 @@ def add_delta_t_key(events):
...
-    # We exclude the first event, for which delta_t might be set to 0 if it is
-    # the first even in a file.
-    # We want this to calculate the actual trigger rate, on which the dead
-    # time depends.
-    delta_t = events['delta_t']
+    t_elapsed = np.sum(time_diff[time_diff<0.1*u.s])
+
+    # delta_t is the time elapsed since the previous triggered event.
+    # We exclude the null values that might be set for the first even in a file.
     delta_t = delta_t[delta_t>0.]
```
chaimain, line 677: Although it works, but just to be consistent, it will be good to add *u.s after 0 as well.

chaimain (approved):

lstchain/reco/utils.py
moralejo: 0 is 0 in any units... Actually I found it amusing it gets accepted, but it does!

maxnoe: See https://github.com/astropy/astropy/issues/10396 :D

moralejo: Ok! - but I think I won't reopen this to save half a microsecond in each such comparison :-D

maxnoe: Yes, agreed. But the same code was also running in `hillas_parameters` and there it made a big difference.

PR #618, opened 2021-03-04
Update ctapipe_io_lst to 0.8.0

maxnoe:


moralejo (dismissed):

README.md
```diff
@@ -12,6 +12,15 @@ are unit tested and should be working as long as the build status is passing.
 
 - You will need to install [anaconda](https://www.anaconda.com/distribution/#download-section) first. 
 
+
+### macOS requirements
+
+lstchain depends on `protozfits` to read LST raw data via `ctapipe_io_lst`.
+Currently, prebuilt binaries are only available on linux.
+On macOS, you will need `protobuf` and `zeromq` installed so that the
```
rlopezcoto, line 20: shouldn't we then include that in the environment? or at least an option for macos

maxnoe: I am working on making it work with `protobuf` and `zeromq` installed via conda and also on a conda-forge recipe, that should also work for mac. Should be done tomorrow I hope.

moralejo:
Is this finished then?

moralejo (approved):

README.md
rlopezcoto: thanks

maxnoe:
@moralejo  I think so yes. Currently working on the conda recipe, but ran into an issue there. Hope to get it out today still but the conda package will probably take a while since it has to be approved by the conda-forge guys.

moralejo:
But we can merge and it will work on linux, right? And on Mac once the conda package is there.

maxnoe:
@moralejo on mac should also already work, but it will compile from source, so installing might take 1-2 minutes.

moralejo:
Can we merge then? I'd like to test it, and that makes it easier.

maxnoe:
Fine with me

rlopezcoto (approved):

maxnoe:
I tested in my macvm and everything works fine (runing the tests, haven't tried more).

If mac support is important, we could try activating mac builds in the github CI. It was just not good on travis, but github actions has mac support I think.

moralejo:
> I tested in my macvm and everything works fine (runing the tests, haven't tried more).
> 
> If mac support is important, we could try activating mac builds in the github CI. It was just not good on travis, but github actions has mac support I think.

Would be good yes, many of us use mac. This will go in parallel and not slow the CI tests, right?

maxnoe:
Yes, should go in parallel.

rlopezcoto:
maybe not making it a `Required` test for the PRs to be accepted? Although I agree with @moralejo that many of us work with Mac and it would be pretty needed that lstchain works in this OS.

Bultako:
just for info
I had to update `cmake` compiler to version>3.15.0 to build the conda env in my mac

maxnoe:
Yes, that is required.

PR #623, opened 2021-03-05
Fix format of  run number in calibration script's input and output files

FrancaCassol:


rlopezcoto (commented):
shouldn't all these be integers?

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -79,7 +78,7 @@ def main():
 
         # search the pedestal calibration file
 
-        pedestal_file = f"{output_dir}/drs4_pedestal.Run{ped_run}.0000.fits"
+        pedestal_file = f"{output_dir}/drs4_pedestal.Run{'%05d'%ped_run}.0000.fits"
```
maxnoe, line 81: Please don't mix % formatting with f-strings, just use `f"{output_dir}/drs4_pedestal.Run{ped_run:05d}.0000.fits"`

maxnoe: Like everywhere else...

FrancaCassol: Hi @maxnoe, sorry, I do not understand the comment what do you mean?

maxnoe: Please replace
```
pedestal_file = f"{output_dir}/drs4_pedestal.Run{'%05d'%ped_run}.0000.fits"
```

with 

```
pedestal_file = f"{output_dir}/drs4_pedestal.Run{ped_run:05d}.0000.fits"
```

Like you used everywhere else in this file.

FrancaCassol: > shouldn't all these be integers?

yes, they must be integer

maxnoe: The current code is using the old `%` based formatting to create the string `0XXXX` and then uses an f-string to insert that into the file name. This doesn't make sense. Just use the f-string.

FrancaCassol: Ah, yes, that is an overlook (I made test), I am going to change it

rlopezcoto (approved):

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -162,7 +156,7 @@ def main():
             os.system(cmd)
 
             # plot and save some results
-            plot_file=f"{output_dir}/log/calibration.Run{run}.{sub_run}.pedestal.Run{ped_run}.0000.pdf"
+            plot_file=f"{output_dir}/log/calibration.Run{run:05d}.{sub_run:04d}.pedestal.Run{ped_run:04d}.0000.pdf"
```
maxnoe, line 159: Why is run sometimes formatted with 5 digits and sometimes with only 4?

rlopezcoto: this is probably a typo and should:
`Run{ped_run:04d}.0000.pdf` -> `Run{ped_run:05d}.0000.pdf`

after this I think we should be good to go with this pR as well

rlopezcoto (approved):

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -101,8 +96,7 @@ def main():
         #
         # produce drs4 time calibration file
         #
-
-        time_file = f"{output_dir}/time_calibration.Run{run}.0000.hdf5"
+        time_file = f"{output_dir}/time_calibration.Run{run:05d}.0000.hdf5"
```
morcuended, line 99: We use `.hdf5` for calibration files and `.h5` for DL1 and DL2 files. Should we unify that or it is better to keep it like this?

morcuended (dismissed):

morcuended (approved):

PR #625, opened 2021-03-05
Make sure we only use ctapipe 0.10

maxnoe:


rlopezcoto (approved):
Already thinking on the new version? :D

PR #626, opened 2021-03-08
DL1 data check, add plots on surviving pedestal fraction

moralejo:
Added plots of fraction of interleaved pedestals which survive cleaning

Removed plot of absolute number of events of each type per subrun (keep
just the rate of each)

Solved issue with precision of time in the pixwise_runsummary table

Now the r0_to_dl1 script will not automatically produce the DL1 data check file.  
@morcuended: the reason is that we want to run the data  check for the DL1 files that we will produce **with the dl1ab script**, which will use in the cleaning the info about the pixel-wise pedestal fluctuations (from stars etc). LSTOSA will have to launch  lstchain_check_dl1.py with the option --omit-pdf  for each subrun **after** running dl1ab. Then the datacheck*h5 file is created for that subrun - and the rest is the same as it was until now: running again lstchain_check_dl1.py  over all the subrun-wise datacheck files.


moralejo:
It is tested. Just replaced one plot (# of events per subrun by type)  by a more useful one, fraction of pedestal events which survive cleaning. This is to monitor the amount of spurious islands that we may have in actual cosmic events.

lstchain/datachecks/dl1_checker.py
```diff
@@ -406,21 +406,24 @@ def plot_datacheck(datacheck_filename, out_path=None, batch=False, muons_dir=Non
...
-        axes[1, 0].set_ylabel('number of events')
-        axes[1, 1].set_ylabel('rate (events/s)')
+            axes[1, 1].plot(table.col('subrun_index'),
+                            table.col('num_cleaned_events') /
+                            table.col('num_events'), fmt, label=label)
+
+        axes[1, 0].set_ylabel('rate (events/s)')
+        axes[1, 0].set_yscale('log')
+        axes[1, 1].set_ylabel('Fraction of events surviving cleaning')
+        axes[1, 0].set_yscale('linear')
```
rlopezcoto, line 424: same axis scale set twice? I guess it should be `axes[1, 1]`

rlopezcoto: also, do you need to specify it that it is linear? It should be set by default

lstchain/reco/r0_to_dl1.py
```diff
@@ -567,8 +567,8 @@ def r0_to_dl1(
         table.write(muon_output_filename, format='fits', overwrite=True)
 
         # Produce the dl1 datacheck .h5 file:
-        check_dl1(output_filename, Path(output_filename).parent,
-                  max_cores=1, create_pdf=False)
+        # check_dl1(output_filename, Path(output_filename).parent,
```
rlopezcoto, line 570: remove?

lstchain/datachecks/dl1_checker.py
moralejo: Yes, I forgot to remove the line where it is set to log, thx!

lstchain/reco/r0_to_dl1.py
moralejo: Ok

morcuended:
Test failing because the r0 to dl1 script expects to produce the datacheck files. This needs to be changed accordingly.

> @morcuended: the reason is that we want to run the data check for the DL1 files that we will produce with the dl1ab script, which will use in the cleaning the info about the pixel-wise pedestal fluctuations (from stars etc). LSTOSA will have to launch lstchain_check_dl1.py with the option --omit-pdf for each subrun after running dl1ab. Then the datacheck*h5 file is created for that subrun - and the rest is the same as it was until now: running again lstchain_check_dl1.py over all the subrun-wise datacheck files.

Noted. I'll have to work on this in the coming days.

moralejo:
One test not passing, will fix.

lstchain/scripts/longterm_dl1_check.py
```diff
@@ -745,8 +750,21 @@ def plot(filename='longterm_dl1_check.h5'):
     fig_ped_stddev.y_range = \
         Range1d(0.,1.1*np.max(runsummary['ped_charge_stddev']))
 
+    frac = runsummary['num_pedestals_after_cleaning'] / \
+           runsummary['num_pedestals']
+    err = np.sqrt(frac*(1-frac)/runsummary['num_pedestals'])
+    fig_ped_clean_fraction = \
```
maxnoe, line 756: minor style thing, but line continuations are generally discouraged. Use the implicit line continuation inside parentheses, e.g.:

```
fig_ped_clean_fraction = show_graph(
    x=...,
    y =...
)
```

moralejo: But in some cases like this one readability is not so good... line is broken at the square bracket (!)

maxnoe: At which square brackets? This is a Function call. This is how I (and I think black) would format it:

```
    fig_ped_clean_fraction = show_graph(
        x=pd.to_datetime(runsummary["time"], origin="unix", unit="s"),
        y=frac,
        xlabel="date",
        ylabel="Fraction of pedestals surviving cleaning",
        ey=err,
        xtype="datetime",
        ytype="linear",
        point_labels=run_titles,
    )

```

moralejo:
> Test failing because the r0 to dl1 script expects to produce the datacheck files. This needs to be changed accordingly.
> 

@morcuended, anyone: any idea why tests failed now? It complains that it does not find the DL1 data check script (??)

lstchain/scripts/longterm_dl1_check.py
moralejo: I meant the automatic line breaking by pycharm (probably configurable?) - but your suggestion is fine, will do.

moralejo:
> > Test failing because the r0 to dl1 script expects to produce the datacheck files. This needs to be changed accordingly.
> 
> @morcuended, anyone: any idea why tests failed now? It complains that it does not find the DL1 data check script (??)

Ok, fixed now, thx @morcuended

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -130,6 +130,14 @@ def observed_dl1_files(temp_dir_observed_files):
...
     )
 
+    run_program(
+            "lstchain_check_dl1",
+            "-b",
+            "--omit-pdf",
+            "--input-file",
+            dl1_output_path1
+    )
+
```
morcuended, line 140: Still failing. You probably need to give the output path `temp_dir_observed_files`

moralejo: Yes, this was it, thanks! should be ok now.

morcuended (approved):

PR #629, opened 2021-03-09
Update ctapipe_io_lst and ctapipe versions

rlopezcoto:
Maybe we could explicitly include ctapipe `v0.10.5` as well? No big changes from v0.10.4, right?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(
     install_requires=[
         "astropy~=4.2",
         'ctapipe>=0.10.4',
-        'ctapipe_io_lst>=0.8.2',
+        'ctapipe_io_lst~=0.8.2',
         'ctaplot~=0.5.5',
         "eventio>=1.5.1,<2.0.0a0",  # at least 1.1.1, but not 2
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -44,7 +44,7 @@ def find_scripts(script_dir, prefix):
     install_requires=[
         "astropy~=4.2",
         'ctapipe>=0.10.4',
-        'ctapipe_io_lst~=0.8.0',
+        'ctapipe_io_lst>=0.8.2',
```
maxnoe, line 47: Why `>=` here? Do we want to install 0.9 automatically when it comes out?

rlopezcoto: no, I moved it to `~=0.8.2`. Although will this always take `>v0.8.2 & <v0.9`?

maxnoe: Yes, `~=0.8.2` means the same as `>=0.8.2` and `<0.9.0a0`, just a more concise way to write it. See https://www.python.org/dev/peps/pep-0440/#compatible-release

moralejo (approved):

PR #614, opened 2021-03-03
Script for creating run summary

morcuended:
As discussed in #610 here is the script to create run summaries with reference timestamps and counters.

It writes a txt file for a given date containing:

 -   Run number
 -   Number of subruns
 -   Type of run (determined empirically)
 -   Start_of_the_run
 -   Reference source ("ucts" or "run_date")
 -   Reference timestamp
 -   Initial dragon counter
 -   Dragon module ID used to take the reference counter values

How it works:
```
lstchain_create_run_summary.py -d 20200117
```
It produces an ECSV file from an astropy Table:

```python
# %ECSV 0.9
# ---
# datatype:
# - {name: run_id, datatype: int64}
# - {name: n_subruns, datatype: int64}
# - {name: run_type, datatype: string}
# - {name: ucts_timestamp, datatype: int64}
# - {name: run_start, datatype: int64}
# - {name: dragon_reference_time, datatype: int64}
# - {name: dragon_reference_module_id, datatype: int16}
# - {name: dragon_reference_module_index, datatype: int16}
# - {name: dragon_reference_counter, datatype: uint64}
# - {name: dragon_reference_source, datatype: string}
# delimiter: ','
# meta: !!omap
# - {date: '2020-01-17'}
# - {lstchain_version: 0.6.3.post434+git32642ee}
# schema: astropy-2.0
run_id,n_subruns,run_type,ucts_timestamp,run_start,dragon_reference_time,dragon_reference_module_id,dragon_reference_module_index,dragon_reference_counter,dragon_reference_source
1803,2,DRS4,-1,1579282968000000000,1579282968000000000,90,0,6095491300,run_start
1804,6,DRS4,1579289727863850890,1579289712000000000,1579289727863850890,90,0,5863850400,ucts
1805,5,CALI,1579291426030146503,1579291413000000000,1579291426030146503,90,0,2030146000,ucts
1806,5,CALI,1579291932080485703,1579291917000000000,1579291932080485703,90,0,5080485200,ucts
1807,19,DATA,1579292477145904430,1579292461000000000,1579292477145904430,90,0,6145904000,ucts
1808,35,DATA,1579292985532016507,1579292975000000000,1579292985532016507,90,0,2532016000,ucts
1809,18,DATA,1579293682237784201,1579293667000000000,1579293682237784201,90,0,7237783700,ucts
1810,5,DRS4,1579295514220829368,1579295500000000000,1579295514220829368,90,0,7220828900,ucts
[...]

PR #632, opened 2021-03-09
Update ctapipe_io_lst to 0.9

maxnoe:
Needed to be able to give the new RunSummaryFile to the event source.


maxnoe:
We need to merge #614 first, and then switch from the old night summary to the new run summary

maxnoe:
The failing tests are fixed by the work @FrancaCassol is doing in #631, so I am closing here and pushing the version update there.

Issue #628, opened 2021-03-09
Standard cleaning settings


moralejo:
With #612 we can reduce the effect of stars by setting an additional cleaning threshold for each pixel, which depends on the standard deviation of its charge in interleaved pedestal events throughout a subrun.

We have to decide on the **standard settings** for the next reprocessing of LST data. There are 4 parameters to play with (default values used in previous processing in parentheses):

Picture threshold (6), boundary threshold (3), min_number_picture_neighbors (1), sigma (-)

sigma is the number of standard deviations above the mean pedestal charge so that a pixel can survive the cleaning (if it fulfills the rest of conditions). This is implemented in #612 and hence was not available in the previous processing.

Besides, one can set the parametrization to use all pixels surviving the cleaning, or just the main island. The previous processing used the main island as a means to reduce the effect of noisy pixels. In the next reprocessing we intend to **use all cleaning survivors**, because otherwise we may be losing parts of the image which may help in the reconstruction (e.g. by keeping just the main island we may be making all showers look more gamma-like).

To be considered:

- the "sigma" condition must be the dominant one only exceptionally, i.e. for pixels illuminated by stars, or noisy for whatever other reason. The bulk of the camera pixels at any given time must be cleaned according to the picture and boundary thresholds. If that is not the case, MC - data matching would become very complicated. The goal of the additional condition is just to avoid noisy pixels to survive too often (since that would affect the reconstruction of the actual shower images).
- Therefore,the typical values of mean_pedestal+sigma\*pedestal_rms should not be larger than the picture threshold. With our current LocalPeak integration, the mean_pedestal of camera pixels may vary between 1.7 and 2.5 p.e., with standard deviation between 1.6 and 2 p.e. With sigma=3 it would not make sense to use a picture threshold of 6, because all pixels would in practice have a higher threshold, given by sigma. We should actually have a warning for the case in which the cleaning of a large fraction of camera pixels have mean_pedestal+sigma\*pedestal_rms > picture_threshold
- we could also use timing constraints. e.g. using ctapipe.image.apply_time_delta_cleaning within lstchain's dl1ab script
[...]

PR #637, opened 2021-03-10
Update default cleaning levels 

rlopezcoto:
After discussion in #628, we finally decided for a standard cleaning defined by:
- picture threshold = 6
- boundary threshold = 3 
- sigma = 2.5
- time delta cleaning = 2
- min_number_picture_neighbors = 2

as it is included now in the configuration of `tailcuts_clean_with_pedestal_threshold`. There is still the point raised by @moralejo that `tailcut` would still be applied at the `r0_to_dl1` level, but in principle we need to compute the events using the proper cleaning by running the `lstchain_dl1ab.py`

Other places where the regular `tailcuts_clean` is used are:
- `lstchain/reco/volume_reducer.py`
- `lstchain/image/muon/muon_analysis.py` -> here a [hardcoded [10, 5] cleaning is applied](https://github.com/cta-observatory/cta-lstchain/blob/5183119f3fb732163d03bb463dd4efdac3122f48/lstchain/image/muon/muon_analysis.py#L134)

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -115,9 +115,6 @@ def get_dl1(
     if "delta_time" in cleaning_parameters:
         delta_time = cleaning_parameters["delta_time"]
 
-    print('use_main_island', use_main_island)
-    print('delta_time', delta_time)
-        
     # we use pop because ctapipe won't recognize that keyword in tailcuts
     cleaning_parameters_for_tailcuts.pop("delta_time")
     cleaning_parameters_for_tailcuts.pop("use_only_main_island")
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -104,17 +104,21 @@ def get_dl1(
 
     config = replace_config(standard_config, custom_config)
     cleaning_parameters = config["tailcut"]
+    cleaning_parameters_for_tailcuts = cleaning_parameters.copy()
     use_main_island = True
     if "use_only_main_island" in cleaning_parameters.keys():
-        # we use pop because ctapipe won't recognize that keyword in tailcuts
-        use_main_island = cleaning_parameters.pop("use_only_main_island")
+        use_main_island = cleaning_parameters["use_only_main_island"]
 
     # time constraint for image cleaning: require at least one neighbor
     # within delta_time:
     delta_time = None
     if "delta_time" in cleaning_parameters:
-        delta_time = cleaning_parameters.pop("delta_time")
+        delta_time = cleaning_parameters["delta_time"]
 
+    # we use pop because ctapipe won't recognize that keyword in tailcuts
+    cleaning_parameters_for_tailcuts.pop("delta_time")
+    cleaning_parameters_for_tailcuts.pop("use_only_main_island")
+    
     dl1_container = DL1ParametersContainer() if dl1_container is None else dl1_container
 
     dl1 = calibrated_event.dl1.tel[telescope_id]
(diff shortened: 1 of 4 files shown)


moralejo:
We need also to set the values for MC, which are the same except for the "pedestal rms" part.
The only thing is to check whether the muon identification works all the same with those settings (it should), let me have a look.

moralejo:
Ok checked, result on muons is identical regardless of cleaning, because the muon ring identification does not depend on it (nor does the ret of the muon analysis).

So for the "tailcut" entry in the standard config, set everything the same as  for tailcuts_clean_with_pedestal_threshold  (except sigma, which would anyway be ignored)

moralejo:
> Ok checked, result on muons is identical regardless of cleaning, because the muon ring identification does not depend on it (nor does the rest of the muon analysis).
> 
> So for the "tailcut" entry in the standard config, set everything the same as for tailcuts_clean_with_pedestal_threshold (except sigma, which would anyway be ignored), in this way we will have the proper cleaning also for the MC analysis directly in the r0_to_dl1 execution.

I did the changes, please check.

moralejo (approved):

rlopezcoto:
tests are no passing, let me check

rlopezcoto:
I think the difference comes from how the `main_island` and `delta_time` parameters are passed:
https://github.com/cta-observatory/cta-lstchain/blob/ecf4ee6070488ab751be4e1f630e1d41208b88fc/lstchain/scripts/lstchain_dl1ab.py#L107-L113

https://github.com/cta-observatory/cta-lstchain/blob/ecf4ee6070488ab751be4e1f630e1d41208b88fc/lstchain/reco/r0_to_dl1.py#L107-L116

Apparently `we use pop because ctapipe won't recognize that keyword in tailcuts`

rlopezcoto:
The problem was the removal of the dictionary elements (see discussion [here](https://github.com/cta-observatory/cta-lstchain/pull/630/files/43464e78a1618c1dcc074237a03b11f75b94b82d#r590330902)).

The config does not actually get reloaded for every event, so if you remove those elements from the dictionary using `.pop`, then they are not present anymore and the default values defined in `r0_to_dl1` are taken. 
This is treated in a different way in `lstchain_dl1ab.py`. It is very good that we found this, because at the moment only the first event was analyzed using the `use_only_main_island` and `delta_time` values from the config file

moralejo:
Well spotted @rlopezcoto, thanks!!

rlopezcoto:
Also, in `r0_to_dl1` the time cleaning was being applied to all pixels instead of only signal pixels

moralejo:
Still one mistake, leave it to me ;-)

moralejo:
you were fast!

moralejo (approved):

lstchain/reco/r0_to_dl1.py
```diff
@@ -104,17 +104,21 @@ def get_dl1(
 
     config = replace_config(standard_config, custom_config)
     cleaning_parameters = config["tailcut"]
+    cleaning_parameters_for_tailcuts = cleaning_parameters.copy()
```
maxnoe, line 107: I don't understand these changes, the code was good before. Doing a full copy and then later popping the values and then never using them is not an improvement.

rlopezcoto: no, the code was not good if it was not working

rlopezcoto: `cleaning_parameters` need to contain all dictionary entries (`use_main_island` and `delta_time` included) because the configuration is not reloaded every event that is analyzed.

If we pass `cleaning_parameters` to `tailcut_clean`, an error is raised because it does not identify some of the dictionary entries, that is why a new dictionary without the `use_main_island` and `delta_time` entries are included is created every time.

maxnoe: I see, then the error is that this is in the event loop in the first place. It does not change.

maxnoe (approved):

PR #593, opened 2021-02-17
Tools to create DL3 files with IRFs generated using pyirf functions.

chaimain:
Forwarded from [PR#535](https://github.com/cta-observatory/cta-lstchain/pull/535) so that the CI passes the tests with access to run the new test data file.
PSF is not added currently but can be done through another PR.

PR #640, opened 2021-03-10
Move fixtures to conftest to use them widely

morcuended:


morcuended:
Moved some fixtures to `conftest.py` so they can be used by DL3 Tool tests in #593.

maxnoe (approved):

Issue #638, opened 2021-03-10
Remove write_array_info_08 deprecated function?


morcuended:
https://github.com/cta-observatory/cta-lstchain/blob/a2b76b4f89b1a566b9fb12da65b11bd5ff79b6d0/lstchain/io/io.py#L369-L380

As specified in the deprecation message. Should we remove this function for the upcoming release?

maxnoe:
yes please

PR #641, opened 2021-03-10
Remove deprecated write array functions

morcuended:
See #638


rlopezcoto:
is this ready for review already?

morcuended:
> is this ready for review already?

not yet

lstchain/io/io.py
```diff
@@ -268,22 +267,29 @@ def merging_check(file_list):
...
-    filename0 = mergeable_list[0]
-    array_info0 = read_array_info(filename0)
-    mcheader0 = read_simu_info_hdf5(filename0)
-    thrown_events_hist0 = read_simtel_energy_histogram(filename0)
-    metadata0 = read_metadata(filename0)
+    first_file = mergeable_list[0]
+    subarray_info0 = SubarrayDescription.from_hdf(first_file)
+    metadata0 = read_metadata(first_file)
+
+    if subarray_info0.name == "MonteCarloArray":
```
morcuended, line 274: Is there a way of knowing if a file is simulated or not better than this?

morcuended:
> is this ready for review already?

ready

lstchain/datachecks/dl1_checker.py
```diff
@@ -154,10 +154,10 @@ def check_dl1(filenames, output_path, max_cores=4, create_pdf=False, batch=False
     # we assume that cam geom is the same in all files, & write the first one
     # we convert units from m to deg
     cam_description_table = \
-        Table.read(filenames[0], path='instrument/telescope/camera/LSTCam')
+        Table.read(filenames[0], path='configuration/instrument/telescope/camera/geometry_LSTCam')
```
vuillaut, line 157: good opportunity to define that path as a global variable that we can import

lstchain/io/io.py
```diff
@@ -268,22 +267,29 @@ def merging_check(file_list):
...
-        except:
+            subarray_info = SubarrayDescription.from_hdf(filename)
+
+            if subarray_info0.name == "MonteCarloArray":
+                mcheader = read_simu_info_hdf5(filename)
+                thrown_events_hist = read_simtel_energy_histogram(filename)
+                check_mcheader(mcheader0, mcheader)
+                check_thrown_events_histogram(thrown_events_hist0, thrown_events_hist)
+
+            assert subarray_info == subarray_info0
```
morcuended, line 290: Also, I'm not quite sure of this assertion. I guess it checks that all the tables are the same. Before it was done by a loop.

lstchain/io/io.py
vuillaut: good point
there is a metadata `is_simulation` in event_source
This could be added to the `global_metadata`.

vuillaut: i think this way of checking is fine for now, until we adopt ctapipe DL1

vuillaut (changes requested):

lstchain/datachecks/dl1_checker.py
vuillaut: (and that can be used everywhere else)

maxnoe: No, please just use `SubArrayDescription.to_hdf` and `SubArrayDescription.from_hdf`, no need for manually reading any tables.

vuillaut: Indeed, even better

mexanick: Can a hardcoded name (`geometry_LSTCam`) be avoided?

morcuended: naming problems should go away when using `SubArrayDescription`

morcuended: changed

lstchain/scripts/longterm_dl1_check.py
```diff
@@ -487,22 +490,17 @@ class pixwise_info(tables.IsDescription):
...
 
 
 def plot(filename='longterm_dl1_check.h5'):
 
     # First read in the camera geometry:
-    cam_description_table = \
-        Table.read(filename, path='instrument/telescope/camera/LSTCam')
-    camgeom = CameraGeometry.from_table(cam_description_table)
+    subarray_info = SubarrayDescription.from_hdf(filename)
+    camgeom = subarray_info.camera_types[0].geometry
```
morcuended, line 503: I see a problem here when having more than one camera type. Here it is just the only one present in the file, but what happens with mc arrays with more telescopes?

vuillaut: Shouldn't you read the geometry of the telescope directly (from its `tel_id`)?
Something such as
```
subarray_info.tel[tel_id].camera.geometry
```

maxnoe: Yes, `subarray.tel[tel_id].camera.geometry`

morcuended: Which is the best way of getting `tel_id`? Can it be fetched from the same subarray info, config file?

morcuended: I tried to do it this way. @vuillaut @maxnoe please check that I did it properly. I set `tel_id` by default to 1 either via command line or as an argument of the corresponding function.

moralejo (approved):

rlopezcoto (approved):

rlopezcoto:
@vuillaut are you ok with the changes performed?

vuillaut (approved):

PR #644, opened 2021-03-10
Use telescope ID together with telescope name during R0 to DL1 conversion

mexanick:
Possible fix of telescopes with the same name not being properly distinguished

Details: [Issue 633](https://github.com/cta-observatory/cta-lstchain/issues/633)

Issue #633, opened 2021-03-09
ValueError during addition of disp parameters during R0 to DL1 conversion


mexanick:
I’m trying to run the MC R0 to DL1 conversion using the most recent lstchain and have encountered a following issue. While some files pass the conversion just ok, some other issue a following error:
```shell
  File "/home/mykhailo.dalchenko/.conda/envs/lst-dev-03032021/lib/python3.7/site-packages/numpy/core/records.py", line 746, in fromrecords
    retval = sb.array(recList, dtype=descr)
ValueError: could not assign tuple of length 55 to structure with 54 fields.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mykhailo.dalchenko/.conda/envs/lst-dev-03032021/lib/python3.7/site-packages/tables/table.py", line 2221, in append
    wbufRA = numpy.rec.array(rows, dtype=self._v_dtype)
  File "/home/mykhailo.dalchenko/.conda/envs/lst-dev-03032021/lib/python3.7/site-packages/numpy/core/records.py", line 1069, in array
    return fromrecords(obj, dtype=dtype, shape=shape, **kwds)
  File "/home/mykhailo.dalchenko/.conda/envs/lst-dev-03032021/lib/python3.7/site-packages/numpy/core/records.py", line 758, in fromrecords
    _array[k] = tuple(recList[k])
ValueError: could not assign tuple of length 55 to structure with 54 fields.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mykhailo.dalchenko/.conda/envs/lst-dev-03032021/bin/convert_r0_to_dl1", line 8, in <module>
    sys.exit(main())
  File "/home/mykhailo.dalchenko/.conda/envs/lst-dev-03032021/lib/python3.7/site-packages/sim_runner/scripts/convert_r0_to_dl1.py", line 64, in main
    custom_config=config,
  File "/fefs/aswg/workspace/mykhailo.dalchenko/small_pix_cam/sw/cta-lstchain/lstchain/reco/r0_to_dl1.py", line 572, in r0_to_dl1
    add_disp_to_parameters_table(output_filename, f'dl1/event/telescope/parameters/{tel_name}', focal)
  File "/fefs/aswg/workspace/mykhailo.dalchenko/small_pix_cam/sw/cta-lstchain/lstchain/reco/r0_to_dl1.py", line 623, in add_disp_to_parameters_table
[...]

PR #646, opened 2021-03-11
Improvement/bugfixing of R0 to DL1 conversion

mexanick:
Superseeds #644

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -186,7 +186,8 @@ def get_dl1(
     return dl1_container
 
 def r0_to_dl1(
-    input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
+    input_filename=None,
+    #input_filename=get_dataset_path('gamma_test_large.simtel.gz'),
     output_filename=None,
     custom_config={},
 ):
@@ -312,6 +313,7 @@ def r0_to_dl1(
                 col_name='tels_with_trigger',
                 transform=tel_list_transform
             )
+            writer.exclude(f'subarray/trigger', 'tel')
 
         # Forcing filters for the dl1 dataset that are currently read from the pre-existing files
         # This should be fixed in ctapipe and then corrected here
@@ -395,7 +397,8 @@ def r0_to_dl1(
                 tel = event.dl1.tel[telescope_id]
                 tel.prefix = ''  # don't really need one
                 # remove the first part of the tel_name which is the type 'LST', 'MST' or 'SST'
-                tel_name = str(subarray.tel[telescope_id])[4:]
+                #tel_name = str(subarray.tel[telescope_id])[4:]
+                tel_name = f'{str(subarray.tel[telescope_id])[4:]}-{telescope_id}'
 
                 if custom_calibration:
                     lst_calibration(event, telescope_id)
@@ -435,6 +438,11 @@ def r0_to_dl1(
                     dl1_container.trigger_type = event.lst.tel[telescope_id].evt.tib_masked_trigger
                 else:
                     dl1_container.trigger_type = event.trigger.event_type
+                    writer.exclude(f'telescope/parameters/{tel_name}', 'hadroness')
+                    writer.exclude(f'telescope/parameters/{tel_name}', 'dragon_time')
+                    writer.exclude(f'telescope/parameters/{tel_name}', 'ucts_time')
+                    writer.exclude(f'telescope/parameters/{tel_name}', 'tib_time')
+                    writer.exclude(f'telescope/parameters/{tel_name}', 'ucts_trigger_type')
 
                 dl1_container.az_tel = event.pointing.tel[telescope_id].azimuth
                 dl1_container.alt_tel = event.pointing.tel[telescope_id].altitude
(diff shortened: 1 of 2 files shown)


rlopezcoto:
@mexanick thanks for this PR (see #633 for the initial reported problem). 
Some of the tests are not passing related to the expected naming of some of the tables we use for the analysis. We need to have a careful look and converge to solve this issue, but we are now on the verge of a new release, so I would say that this should go to the first bugfix release of v0.7, is that ok with you?

lstchain/reco/r0_to_dl1.py
```diff
@@ -566,9 +588,14 @@ def r0_to_dl1(
 
     if is_simu:
         # Reconstruct source position from disp for all events and write the result in the output file
-        for tel_name in ['LST_LSTCam']:
-            focal = OpticsDescription.from_name(tel_name.split('_')[0]).equivalent_focal_length
-            add_disp_to_parameters_table(output_filename, dl1_params_lstcam_key, focal)
+        for telescope_id in subarray.tels.keys():
+            if telescope_id in config['source_config']['EventSource']['allowed_tels']:
```
maxnoe, line 592: just `source.allowed_tels`?

mexanick: > @mexanick thanks for this PR (see #633 for the initial reported problem).
> Some of the tests are not passing related to the expected naming of some of the tables we use for the analysis. We need to have a careful look and converge to solve this issue, but we are now on the verge of a new release, so I would say that this should go to the first bugfix release of v0.7, is that ok with you?
> (more comments will follow as soon as we have finished with the release)

To my understanding, the tests which aren't passing now should pass (at least part of them) after merging of #641 which will (at least partially) resolve the naming. But indeed, the naming scheme has to be centrally defined. Holding it to past the 0.7 release is ok to me if you want, however I need to proceed with other steps of the simulation asap. This naming story actually makes us troubles as we are working not only on standard LST simulations, but also on SiPM-equipped camera which has different geometry and different name. So instead of hardcoding various things here and there I'd like to resolve the underlying issue.

mexanick:
> @mexanick thanks for this PR (see #633 for the initial reported problem).
> Some of the tests are not passing related to the expected naming of some of the tables we use for the analysis. We need to have a careful look and converge to solve this issue, but we are now on the verge of a new release, so I would say that this should go to the first bugfix release of v0.7, is that ok with you?
> (more comments will follow as soon as we have finished with the release)
To my understanding, the tests which aren't passing now should pass (at least part of them) after merging of #641 which will (at least partially) resolve the naming. But indeed, the naming scheme has to be centrally defined. Holding it to past the 0.7 release is ok to me if you want, however I need to proceed with other steps of the simulation asap. This naming story actually makes us troubles as we are working not only on standard LST simulations, but also on SiPM-equipped camera which has different geometry and different name. So instead of hardcoding various things here and there I'd like to resolve the underlying issue.

rlopezcoto:
@mexanick some conflicts showed up now that #641 was merged, could you please solve them to check if tests are passing now and we can move on with this PR?

morcuended:
Tried to resolve conflicts. Hope it is fine now

mexanick:
@morcuended thanks! Sorry, I somehow missed the request from @rlopezcoto . 
Considering the tests: I believe they will fail again. The reason is that I've introduced indexing of the telescopes based on the position and the parameter keys are now `.../LST_LSTCam_{tel_id}` which is not what the test cases are waiting for (they still use hardcoded dl1_params_lstcam_key).
So before moving on we should decide on the indexing. I know there's a concurrent indexing attempt in `ctapipe` and thus don't force this update, as perhaps we just want to move on with a newer ctapipe release?
Considering the warning suppression, this can be cherry-picked in a separate PR.

maxnoe:
@mexanick the output file format here should not be changed.

mexanick:
> @mexanick the output file format here should not be changed.

do you mean we must keep hardcoded `dl1_params_lstcam_key` and alike? 
I do not change the output file format, I just allow indexed param keys names. 
If we keep hardcoded table keys, this will make impossible to:
1. Process multiple telescopes 
2. Process any telescope whose name is not LST_LSTCam

If this is desired behavior, I would mandatory restrict the processing to the first telescope only (otherwise you may mix the telescopes and the current standard config is allowing it) and force to name every telescope as LST_LSTCam.

mexanick:
In aprivate discussion @maxnoe proposed to implement a temporary fix which will keep the parameter key for telescope in pos 1 intact. This is a bit ugly and requires a little difference in treatment of the telescope in pos. 1, but won't brake the rest of the lstchain. And the major format change might be propagated later through the next version of ctapipe. 
@rlopezcoto @morcuended if you agree with this, I will quickly implement it.

rlopezcoto:
@mexanick are you proposing this as a temporary solution to be able to deal with more than one telescope for the time being?
I am in favor of supporting more than one telescope analysis, but I would try to avoid a special treatment for any of the telescopes. In any case, could you please implement the fix so we can have a look?

rlopezcoto:
> @rlopezcoto This pull request changes the layout of the HDF5 file to contain keys like `LST_LSTCam_1`.

so what is the temporary fix suggested not to change this for the first telescope and do it for the rest?

> This is a change from the current layout that does not use tel ids and is also different to what the stage1 tool does. I am thus not too hot about including this change.
>
> I was under the impression, that we don't want to make changes to the dl1 data format of lstchain anymore.

right, this is also specially important now that is fixed for `v0.7.x`, so I would also delay this a little bit until it is further discussed and may enter in further lstchain versions (would that be ok @mexanick?)

maxnoe:
> so what is the temporary fix suggested not to change this for the first telescope and do it for the rest?

Basically that tel_id one is `LST_LSTCam` and we only start adding the tel_id for tel_ids > 1

mexanick:
yup, will do

PR #648, opened 2021-03-11
Fix possible KeyError in DL1 cleaning

### LINKED ISSUE #647: KeyError if certain cleaning parameters aren't in the config
https://github.com/cta-observatory/cta-lstchain/blob/8ac0048a4b4965dfaaf822de534c7562048cc6f8/lstchain/reco/r0_to_dl1.py#L119-L120
The keys which are popped may not be present in the config. I.e. standard config doesn't have them. In such case, the above lines will raise KeyError exception. 

Suggestion: use `object.pop(field, None)` instead of `object.pop(field)`

mexanick:
Fixes #647


maxnoe:
@mexanick I also just realized this issue and fixed it along with an overall simplification of the code here #649

PR #649, opened 2021-03-11
Improve cleaning config handling

maxnoe:


moralejo:
Just one comment (not really on the change, with which I agree): perhaps we should move to main_island being false by default.

maxnoe:
Another question that popped up when looking at this is that I would change the order.

I would calculate number of islands / main island *after* applying the time delta cleaning.

maxnoe:
But that can be another PR, this is uncontroversial, right? So lets's go ahead with this and I'll open more PRs for the other stuff.

mexanick (approved):

rlopezcoto:
is this again introducing the bug that was solved in #637?

maxnoe:
No

lstchain/reco/r0_to_dl1.py
```diff
@@ -103,22 +103,12 @@ def get_dl1(
...
-    if "delta_time" in cleaning_parameters:
-        delta_time = cleaning_parameters["delta_time"]
-
-    # we use pop because ctapipe won't recognize that keyword in tailcuts
-    cleaning_parameters_for_tailcuts.pop("delta_time")
-    cleaning_parameters_for_tailcuts.pop("use_only_main_island")
-    
+
+    # pop delta_time and use_main_island, so we can cleaning_parameters to tailcuts
+    cleaning_parameters = config["tailcut"].copy()
```
maxnoe, line 108: The copy is still here @rlopezcoto

rlopezcoto: ah ok, I missed that, thanks (test were passing, so it meant that changes were fine anyways)

rlopezcoto (approved):

rlopezcoto (approved):

PR #650, opened 2021-03-11
Update lstchain version in README

chaimain:


vuillaut (approved):

rlopezcoto (commented):
Thanks @chaimain. Does anybody know an automatic way to handle the version that is written into the README file?

maxnoe:
No, since you have to make a git commit to change it.

I mean everything is scriptable, but...

mexanick:
well, you can use a custom badge with the version and provide CI tag-only job, but this can serve only for reference purpose, not as a copyable text.

PR #656, opened 2021-03-17
Adds an event filter Component

chaimain:
A new ctapipe.core.Component that uses the function `filter_events` from `lstchain/reco/utils.py`. A small step to add more `Components` and `Tools` to lstchain. Can be used currently in [PR #593](https://github.com/cta-observatory/cta-lstchain/pull/593)  and [PR #599](https://github.com/cta-observatory/cta-lstchain/pull/599).


lstchain/io/event_selection.py
```diff
@@ -0,0 +1,36 @@
...
+
+class EventSelector(Component):
+    """
+    Filter values used for event filters are taken as inputs and
+    filter_events() is used on a table of events called in with the Component.
+    """
+
+    filters = Dict(
+        help="Dict of event filter parameters",
+        default_value={
```
maxnoe, line 19: These defaults look they are not cutting anything but nan values. So they should be removed and the features listed in the `finite_parameters` list.

chaimain: Ah yes! I will amend it.

lstchain/io/tests/test_eventselection.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from lstchain.io import EventSelector
+
+def test_event_selection(simulated_dl1_file):
+   from lstchain.io.io import dl1_params_lstcam_key
+   import pandas as pd
+
+   data = pd.read_hdf(simulated_dl1_file, key=dl1_params_lstcam_key)
```
maxnoe, line 8: I would propose testing no with some complicated example data but to create a custom table with some hand-picked values so you can make sure the correct events are filtered away.

chaimain: Ok

lstchain/io/event_selection.py
```diff
@@ -28,6 +25,7 @@ class EventSelector(Component):
 
     finite_params = List(
 	help="List of parameters to ensure finite values",
+	default_value={"intensity", "length", "width"},
```
maxnoe, line 28: Your default value is a set (works but is a bit strange)

chaimain: Ah! I will correct it.

lstchain/reco/utils.py
```diff
@@ -520,31 +520,40 @@ def filter_events(events,
...
 
     Returns
     -------
-    `pandas.DataFrame`
+    `pandas.DataFrame` or 'astropy.table.QTable'
     """
+    from astropy.table import QTable
 
-    filter = np.ones(len(events), dtype=bool)
+    if isinstance(events, QTable):
```
maxnoe, line 543: Maybe just use `Table`? Then it will work for both `Table` and `QTable`, since:

```

In [3]: issubclass(QTable, Table)
Out[3]: True
```

lstchain/reco/utils.py
```diff
@@ -520,31 +520,40 @@ def filter_events(events,
...
+        events_df = events.to_pandas()
+    else:
+        events_df = events
+
+    filter = np.ones(len(events_df), dtype=bool)
 
     for k in filters.keys():
```
maxnoe, line 550: Better use `for col, (lower_limit, upper_limit) in filters.items()` and then use those below. Avoids the additional lookups for nicer code.

lstchain/reco/utils.py
chaimain: Ok

lstchain/reco/utils.py
```diff
@@ -555,7 +564,7 @@ def filter_events(events,
                 if v > 0:
                     log.warning(f'{k} : {v}')
 
-    return events[filter]
+    return events[filter.to_numpy()]
```
maxnoe, line 567: This `to_numpy` should not be needed, right? Since `filter` is a numpy array? Doesn't this even fail?

chaimain: The filter is, but the mask will not be for DataFrame. It gives `pandas.Series`, but with `to_numpy` it works for both

lstchain/reco/utils.py
chaimain: Ok

lstchain/reco/utils.py
maxnoe: I don't understand, since there is only the `filter` variable here, and no `mask` variable.

maxnoe: Ah, the problem is that `array &= series` will actually result in `array` becoming a `Series`.

To prevent that, use `array[:] &= series`

chaimain: Sorry, I was incorrect earlier. for DataFrame, `filter` is `Series` and so is the corresponding `mask`. But, for Tables, `filter` and `mask` are `array`.
So, to have a common method, I used `to_numpy()` in the return

maxnoe (dismissed):

lstchain/io/tests/test_eventselection.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+from lstchain.io import EventSelector
+
+def test_event_selection():
+   import pandas as pd
+   import numpy as np
+
+   evt_fil = EventSelector()
+
+   data = pd.DataFrame({"a": [1, 2, 3], "b": [2.2, 3.2, np.nan], "c": [1, 3, np.inf]})
```
maxnoe, line 10: Does this test actually test that the cuts work? I just had a glimpse at the values and they seem all to be inside the ranges defined by the cuts, so this test is actually not testing if events are removed that are outside the cut intervals.

Also, you should probably also test with an astropy table as input, since you want to support both.

maxnoe: Maybe define the data as an astropy table with units and then run the test once for the table and once for `table.to_pandas`.

maxnoe (approved):

PR #663, opened 2021-03-29
reading focal length from files or subarray description

vuillaut:


lstchain/reco/r0_to_dl1.py
```diff
@@ -548,8 +548,8 @@ def r0_to_dl1(
     if is_simu:
         # Reconstruct source position from disp for all events and write the result in the output file
         for tel_name in ['LST_LSTCam']:
```
maxnoe, line 550: Why a loop over a single entry?

maxnoe (approved):

lstchain/reco/r0_to_dl1.py
mexanick: this is still related to the inability to work with more than 1 telescope (and it should be LST_LSTCam). I address it in #646

morcuended:
Tests are failing, you need to parse `tel_id` when needed

```python
E           ValueError: Running ***_dl1_to_dl2 failed with return code 1, output: 
E            Traceback (most recent call last):
E             File "/usr/share/miniconda/envs/ci/bin/***_dl1_to_dl2", line 8, in <module>
E               sys.exit(main())
E             File "/usr/share/miniconda/envs/ci/lib/python3.7/site-packages/***/scripts/***_dl1_to_dl2.py", line 104, in main
E               focal_length = subarray_info.tel[args.tel_id].optics.equivalent_focal_length
E           AttributeError: 'Namespace' object has no attribute 'tel_id'
```

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -63,6 +63,11 @@
                     help='Path to a configuration file. If none is given, a standard configuration is applied',
                     default=None, required=False)
 
+parser.add_argument('--tel-id', '-tid', action='store', type=str,
```
mexanick, line 66: Why this parameter is needed? The telescopes to be read out can be retrieved from the JSON config file which is already an argument here.

vuillaut: from which entry in the config?

mexanick: Though the indexing problem is actually deeper. See the conversation here: https://github.com/cta-observatory/cta-lstchain/issues/633

maxnoe:
Why is this option needed for dl1? The dl1 file contains the subarray description and that contains which telescopes are in the file.

vuillaut:
> Why is this option needed for dl1? The dl1 file contains the subarray description and that contains which telescopes are in the file.

The subarray description will contain all telescopes (with current MC, LST and MAGIC tels), here will want to apply dl1 to dl2 to a single `tel_id`.

mexanick:
but again, why don't you constrain it by setting `allowed_tels` in the json config to the telescope(s) you want to process?

vuillaut:
> but again, why don't you constrain it by setting `allowed_tels` in the json config to the telescope(s) you want to process?

We can. But it's not better.
- Event source is not dl1 to dl2.
- there are already two event source in the config
So this will induce some implicit guess on which `tel_id` to select in the config.

vuillaut:
> > but again, why don't you constrain it by setting `allowed_tels` in the json config to the telescope(s) you want to process?
> 
> We can. But it's not better.
> 
> * Event source is not dl1 to dl2.
> * there are already two event source in the config
>   So this will induce some implicit guess on which `tel_id` to select in the config.

And there was actually a third "allowed_tels" not tight to an event_source  😅  - so I'll use this one.

morcuended:
I did the same command-line parsing for `dl1ab` script:

https://github.com/cta-observatory/cta-lstchain/blob/3ae7124a8d99aa9dd09cee7d02496eec6427bb68/lstchain/scripts/lstchain_dl1ab.py#L70

Probably we can adopt the same scheme (guessing from the config) there as well. Do we contemplate having more than those allowed telescopes here?

mexanick:
> 
> And there was actually a third "allowed_tels" not tight to an event_source 😅 - so I'll use this one.

While it can be done, I think it is a bit dangerous practice... I didn't find such field the standard config (in the understanding of the lstchain_standard_config.json). Of course, it is JSON and nothing prevents anyone from adding it, but having it three times... 

finally, more pythonic way would be `tel_id = config.get('allowed_tels', 1)` and it can be a list and not a value itself

mexanick:
I personally used the EventSource one and checked whether such data is still in the file for DL1 to DL2 conversion. But I agree this might not be the optimal way

morcuended:
> `tel_id = config.get('allowed_tels', 1)` and it can be a list and not a value itself

Eventually you want to have only one value, don't you?

mexanick:
> > `tel_id = config.get('allowed_tels', 1)` and it can be a list and not a value itself
> 
> Eventually you want to have only one value, don't you?

Well, if I process N telescopes (e.g. two LSTs), I want to have N values and loop over them.

morcuended:
Right. I mean that you need to use `telescope_id` as an `int` for example in `mirror_area = subarray.tel[telescope_id].optics.mirror_area`.

maxnoe:
> here are already two event source in the config
> So this will induce some implicit guess on which tel_id to select in the config.

The whole problem is related to this. There shouldn't be a single config file that is used by all the scripts for both simulations and observations.

So I agree with the approach here and we should merge.

Everything else is a larger issue and I don't see how that should be fixed other than just switching to ctapipe.

From my point of view, it just doesn't make sense to make the lstchain dl1 to dl2 more general. Since at that point, it should just become the ctapipe dl1 to dl2 tool (for which we will have a core developer workshop soon I think).

vuillaut:
> > here are already two event source in the config
> > So this will induce some implicit guess on which tel_id to select in the config.
> 
> The whole problem is related to this. There shouldn't be a single config file that is used by all the scripts for both simulations and observations.
> 
> So I agree with the approach here and we should merge.
> 
> Everything else is a larger issue and I don't see how that should be fixed other than just switching to ctapipe.
> 
> From my point of view, it just doesn't make sense to make the lstchain dl1 to dl2 more general. Since at that point, it should just become the ctapipe dl1 to dl2 tool (for which we will have a core developer workshop soon I think).

I completely agree.

vuillaut:
So it's now ready for review.

lstchain/reco/r0_to_dl1.py
```diff
@@ -338,8 +338,8 @@ def r0_to_dl1(
                                            new_ped=True, new_ff=True)
 
                 # flat-field or pedestal:
-                if ( event.trigger.event_type == EventType.FLATFIELD or
-                     event.trigger.event_type == EventType.SKY_PEDESTAL ):
+                if (event.trigger.event_type == EventType.FLATFIELD or
+                        event.trigger.event_type == EventType.SKY_PEDESTAL):
```
mexanick, line 342: Overindent?

maxnoe (approved):

mexanick (approved):
Just one minor style comment

lstchain/reco/dl1_to_dl2.py
```diff
@@ -529,7 +527,7 @@ def calc_source_dependent_parameters(data, expected_src_pos_x_m, expected_src_po
     return src_dep_params
 
 
-def get_expected_source_pos(data, data_type, config):
+def get_expected_source_pos(data, data_type, config, focal_length=28*u.m):
```
morcuended, line 530: is this default value needed?

morcuended (changes requested):

lstchain/reco/r0_to_dl1.py
vuillaut: Is it?
I don't know all the rules, IDE auto-formatting.

lstchain/reco/dl1_to_dl2.py
vuillaut: The intention was to keep the same API.

morcuended (approved):

lstchain/reco/r0_to_dl1.py
maxnoe: black would format as

```
if (
    event.trigger.event_type == EventType.FLATFIELD
    or event.trigger.event_type == EventType.SKY_PEDESTAL
):
```

vuillaut:
Thanks

lstchain/reco/r0_to_dl1.py
vuillaut: Thanks, I'll see how to add black to PyCharm.

PR #664, opened 2021-03-29
Action to autorun black on PRs

vuillaut:
Can we auto-black PRs?


vuillaut:
I have only manually added the workflow `blacken.yml`.
The action runned black, fixed 85 files automatically, and added the fixes to the PR.

moralejo:
Is this Black (or will it be) standard in CTA? Just asking, I now tend to use single quotes (though I liked double ones) because I understood, or perhaps misunderstood, that single quotes were preferred.

maxnoe:
@moralejo Black is used for ctapipe at least. And the double vs. single quotes is my only pain point with black.

But since it is autoformatted, you don't really have to care.

vuillaut:
I don't have a strong preference for black, there are other style enforcement tools:
flake8, pylint, pyflakes, autopep8... 
Max may have suggestions, preferences?
But if ctapipe uses black, we might as well for consistency.

rlopezcoto:
> But if ctapipe uses black, we might as well for consistency.

If it is used in ctapipe and there is no change expected there, let's also use `black`

vuillaut:
@maxnoe do you know why the CI action is not reporting status here?
It seems to have run correctly:
https://github.com/cta-observatory/cta-lstchain/actions/runs/699041750
🤔

maxnoe:
@vuillaut What do you mean with reporting? This is not a checker. It added a commit that does the actual formatting:  76838f9

vuillaut:
I mean that here in the PR, I see tests are not reporting status:
image

nbiederbeck:
some questions: 

1. this seems to work on the whole repo, not only the PR if I understand the diff correctly?
2. this adds another "commiter" (autoblack), is this something you really want?
3. the [github action section in the black repo](https://github.com/psf/black#github-actions) just mentions checking, and one has to fix by themselves (e.g. with a pre-commit hook that is [used in ctapipe](https://cta-observatory.github.io/ctapipe/getting_started/index.html#developing-a-new-feature-or-code-change))

vuillaut:
> some questions:
> 
> 1. this seems to work on the whole repo, not only the PR if I understand the diff correctly?

Correct, but once the repo has been blacken, only the new or modified files in a PR will be.

> 2. this adds another "commiter" (autoblack), is this something you really want?

Why not? It's probably a good thing to know the source of these changes.

> 3. the [github action section in the black repo](https://github.com/psf/black#github-actions) just mentions checking, and one has to fix by themselves (e.g. with a pre-commit hook that is [used in ctapipe](https://cta-observatory.github.io/ctapipe/getting_started/index.html#developing-a-new-feature-or-code-change))

Sorry, I did not understand if you had a question here.
I'd rather avoid extra steps for developers if possible.

maxnoe:
> Why not? It's probably a good thing to know the source of these changes.

Agreed, this makes it for example easy to ignore the black related changes in `git blame` or `git log`

nbiederbeck:
> Why not? It's probably a good thing to know the source of these changes.

I understand and agree for whole-repo blackening. But on single PR changes I thikn I want to see the correct developer and commit message, e.g. in a `git blame`.

maxnoe:
@vuillaut Can you try pushing another commit?

maxnoe:
> But on single PR changes I thikn I want to see the correct developer and commit message, e.g. in a git blame.

I don't understand this. The tool adds another commit to the PR. The original commits from the original author are unchanged.

vuillaut:
> > The tool adds another commit to the PR.
> 
> Yes, but why? I propose that instead of two commits (changes + black) we can have one commit (already formatted changes) in the case of a pre-commit hook.

As said, I'd rather avoid relying on something on the devs side.
I don't really see the issue with having a different formatting commit.

nbiederbeck:
Alright, fine for me. I just wanted to share my thoughts, thanks for the discussion.

chaimain:
I noticed that I had not removed few extra mentions of some new IO functions (in the __init__.py) and the now removed data/data_selection_cuts.json file from setup.py. So I just removed them from here directly.

vuillaut:
Ready for review.

maxnoe:
I am a bit hesitant, since this can easily confuse developers and introduce merge conflicts even when only working alone on a branch (doing a git push, adding more commits locally, then realizing that autoblack changed the branch remotely).

vuillaut:
> I am a bit hesitant, since this can easily confuse developers and introduce merge conflicts even when only working alone on a branch (doing a git push, adding more commits locally, then realizing that autoblack changed the branch remotely).

I see your point (as it just happened in the PR). We have to weight the gain/effort balance.

The other solution would be to ask developers to use pre-commit or to run black on their side and enforce it with a check-only action blocking the PR (similar to unit tests) but I would really like to avoid more actions from the devs side for something that could be completely automatized.

mexanick:
I would also prefer a stricter code style guidelines together with proper enforcement. It is quite easy enforce automatically on the developer's side (we can provide instructions on how to run black or how to incorporate the standard in flakes or whatever...) Being required to pull to be up to date with format changes is weird and moreover may cause problems when the editor has different settings on the developer's end.

rlopezcoto:
since we did not decide to go for this (either not to), and now there are several conflicts, can we close/move this PR to draft?

vuillaut:
> since we did not decide to go for this (either not to), and now there are several conflicts, can we close/move this PR to draft?

Yes, I think we can close this one.
I would still like to propose something to ease formatting and thus avoid reviewers having to take care of that. We can use pre-commit too, I'll make another proposal soon.

PR #665, opened 2021-03-30
Fix mixing of Level A and Level B calibration

FrancaCassol:
The present code overwrites the level A calibration as soon as a Level B calibration event is ready, this is not what we want at this level


rlopezcoto:
Just a little bit more information about this PR: This solves a bug in which conversion factors were being updated in an unwanted way.

rlopezcoto (approved):

PR #668, opened 2021-03-31
Suppress Column X in Table Y is not writable warnings

mexanick:
Excludes non-existing columns from attempt to write, thus suppressing the corresponding warnings

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -764,10 +764,6 @@ def write_subarray_tables(writer, event, metadata=None):
         add_global_metadata(event.simulation, metadata)
         add_global_metadata(event.trigger, metadata)
 
-    writer.exclude('subarray/mc_shower', 'mc_shower')
-    writer.exclude('subarray/mc_shower', 'mc_tel')
-
-    writer.write(table_name="subarray/mc_shower", containers=[event.index, event.simulation])
     writer.write(table_name="subarray/trigger", containers=[event.index, event.trigger])
 
 
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -764,6 +764,9 @@ def write_subarray_tables(writer, event, metadata=None):
         add_global_metadata(event.simulation, metadata)
         add_global_metadata(event.trigger, metadata)
 
+    writer.exclude('subarray/mc_shower', 'mc_shower')
+    writer.exclude('subarray/mc_shower', 'mc_tel')
+
     writer.write(table_name="subarray/mc_shower", containers=[event.index, event.simulation])
     writer.write(table_name="subarray/trigger", containers=[event.index, event.trigger])
 
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -423,7 +423,6 @@ def r0_to_dl1(
                     dl1_container.trigger_type = event.lst.tel[telescope_id].evt.tib_masked_trigger
                 else:
                     dl1_container.trigger_type = event.trigger.event_type
-                    writer.exclude(f'telescope/parameters/{tel_name}', 'hadroness')
                     writer.exclude(f'telescope/parameters/{tel_name}', 'dragon_time')
                     writer.exclude(f'telescope/parameters/{tel_name}', 'ucts_time')
                     writer.exclude(f'telescope/parameters/{tel_name}', 'tib_time')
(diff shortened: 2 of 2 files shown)


lstchain/io/io.py
```diff
@@ -764,6 +764,9 @@ def write_subarray_tables(writer, event, metadata=None):
         add_global_metadata(event.simulation, metadata)
         add_global_metadata(event.trigger, metadata)
 
+    writer.exclude('subarray/mc_shower', 'mc_shower')
```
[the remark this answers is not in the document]
mexanick: true, the written part contains only `obs_id` and `event_id`

lstchain/reco/r0_to_dl1.py
```diff
@@ -439,6 +445,19 @@ def r0_to_dl1(
                     add_global_metadata(container, metadata)
 
                 event.r0.prefix = ''
+                if is_simu:
+                    # None values in telescope/image table
+                    writer.exclude(f'telescope/image/{tel_name}', 'image_mask')
+                    writer.exclude(f'telescope/image/{tel_name}', 'parameters')
+                    # None values in telescope/parameters table
+                    writer.exclude(f'telescope/parameters/{tel_name}', 'disp_norm')
+                    writer.exclude(f'telescope/parameters/{tel_name}', 'disp_dx')
```
morcuended, line 454: I'm not getting this. Why excluding `disp*` params for simulated data?

moralejo: > I'm not getting this. Why excluding `disp*` params for simulated data?

I think they are not yet calculated at this point in the chain.

morcuended: Ah yes

morcuended: Still not getting why excluding the parameters

moralejo: Probably not needed, in the sense that there would be no warning, right?   On the other hand, having there a parameter which is not yet calculated may be misleading.

maxnoe: So excluding them here is correct for observed data. THese are the truth values which can only be calculated on simulated data.

maxnoe: Ah wait. The if is wrong, right?

These columns should be excluded `if not is_simu`

mexanick: nope, these columns are indeed not yet reconstructed yet, the addition of disp parameters is at the very end and is done for the whole dataframe, the (in)famous `for name in ['LST_LSTCam']` loop which was fixed recently

maxnoe: But then they should be ignored for both `is_simu` and `not is_simu`, right? So the if is not needed?

moralejo: If these are true values they should not be called "disp*". All we need is true gamma direction on camera, and, if you want to store it precalculated, also "dist". The name disp as far as I know is for reconstructed quantities.  No need to change it before release, though!

mexanick: I actually realized, that I didn't test on data, so the data-related warnings were not suppressed. The corresponding fix is added

lstchain/reco/r0_to_dl1.py
```diff
@@ -440,6 +445,33 @@ def r0_to_dl1(
...
+                # None values in telescope/parameters table
+                writer.exclude(f'telescope/parameters/{tel_name}', 'hadroness')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'disp_norm')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'disp_dx')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'disp_dy')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'disp_angle')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'disp_sign')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'disp_miss')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'src_x')
+                writer.exclude(f'telescope/parameters/{tel_name}', 'src_y')
```
moralejo, line 460: I don't understand this, I thought you just wanted to exclude MC values that don't exist in real data and vice versa, to avoid warnings. But why to exclude here 'parameters' or src_x, y  both for MC and real data?

maxnoe: This is correct and also what is done in the stage1 tool. `parameters` is ignored, because it is an empty container that is not being filled by lstchain (since it creates this giant `DL1Container`, not the ctapipe container structure and the rest of the columns are values that are written later not via the container system.

moralejo: Ah, I see, thanks! It is clear now.

moralejo (approved):

PR #672, opened 2021-04-06
Convert table dtype to dtype of merge table before appending, fixes 671

maxnoe:
@morcuended Can you try this out?

### DIFF
--- a/lstchain/scripts/lstchain_merge_hdf5_files.py
+++ b/lstchain/scripts/lstchain_merge_hdf5_files.py
@@ -76,7 +76,7 @@ def main():
 
     if args.noimage:
         keys = get_dataset_keys(file_list[0])
-        keys = [k for k in keys if 'image' not in k]
+        keys = {k for k in keys if 'image' not in k}
     else:
         keys = None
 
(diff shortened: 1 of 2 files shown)


morcuended:
Sure, let me try.

lstchain/scripts/lstchain_merge_hdf5_files.py
```diff
@@ -76,7 +76,7 @@ def main():
 
     if args.noimage:
         keys = get_dataset_keys(file_list[0])
-        keys = [k for k in keys if 'image' not in k]
+        keys = {k for k in keys if 'image' not in k}
```
morcuended, line 79: Is this still needed given that it is later converted to `set()`?

maxnoe: Not strictly needed, but if keys is expected to be a set why not make it a set directly?

morcuended (approved):
Thanks @maxnoe. It worked fine with the fix. I tried with DL2/DL1, with/without `--smart` and `--no-image` option

maxnoe:
Thanks for checking

Issue #669, opened 2021-04-03
Unusable pixels error in dl1ab


morcuended:
Got this error while running the `lstchain_dl1ab` script (with pedestal cleaning and tailcut 8-4) for run 02927, sub-run 0164 (out of 245 total sub-runs) from 2020-01-18 using lstchain v0.7.1

Command:
```bash
lstchain_dl1ab \
  --input-file=/fefs/aswg/data/real/running_analysis/20201118/v0.7.1/dl1_LST-1.Run02927.0164.h5 \
  --output-file=/fefs/aswg/data/real/running_analysis/20201118/v0.7.1/tailcut84/dl1_LST-1.Run02927.0164.h5 \
  --pedestal-cleaning=True \
  --config=/fefs/aswg/lstosa/config_files_lstchain/lstchain_standard_config_tailcut84.json
```

```python
Traceback (most recent call last):
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/bin/lstchain_dl1ab", line 8, in <module>
    sys.exit(main())
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/lstchain/scripts/lstchain_dl1ab.py", line 89, in main
    pedestal_thresh = get_threshold_from_dl1_file(args.input_file, sigma)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/lstchain/calib/camera/pixel_threshold_estimation.py", line 65, in get_threshold_from_dl1_file
    unusable_pixels = get_unusable_pixels(dl1_path, interleaved_events_id)
  File "/fefs/aswg/software/virtual_env/anaconda3/envs/osa/lib/python3.7/site-packages/lstchain/calib/camera/pixel_threshold_estimation.py", line 76, in get_unusable_pixels
    HIGH_GAIN,
IndexError: index 1 is out of bounds for axis 0 with size 1
```

maxnoe:
It seems the shape of `unusable_pixels` does not match the shape of `ped_std_pe`

morcuended:
Thanks for the inspection @maxnoe. @pawel21, @rlopezcoto, @moralejo can this be fixed? From the hardware/observation point of view what changes in order to produce these different shapes? Is that sometimes you cannot identify properly interleaved pedestal events?

maxnoe:
For the file that raises the error:
```
[...]

PR #675, opened 2021-04-07
Unusable pixels in dl1ab

pawel21:
Fix Unusable pixels error in dl1ab #669.

According to @FrancaCassol and @moralejo suggestion, I took calibration_id from `f.root.dl1.event.telescope.monitoring.pedestal.col('calibration_id')` to got unusable_pixels.

This solve problem with production, but such data  as run 02927, sub-run 0164  have wrong interleaved pedestal (pedestal from calibration run instead from data run). Do you have any idea how to solve it?

### DIFF
--- a/lstchain/calib/camera/pixel_threshold_estimation.py
+++ b/lstchain/calib/camera/pixel_threshold_estimation.py
@@ -3,6 +3,7 @@ import numpy as np
 from ctapipe_io_lst.constants import HIGH_GAIN
 from lstchain.io.io import dl1_params_tel_mon_ped_key, dl1_params_tel_mon_cal_key
 
+INTERLEAVED_CALIBRATION_ID = 1
 
 def get_bias_and_std(dl1_file):
     """
@@ -56,7 +57,7 @@ def get_threshold_from_dl1_file(dl1_path, sigma_clean):
     # std values from calibration run.
     # Correct interleaved pedestal std array should have shape (2,2,1855)
     if ped_std_pe.shape[0] == 2:
-        interleaved_events_id = 1
+        interleaved_events_id = INTERLEAVED_CALIBRATION_ID
     else:
         interleaved_events_id = 0
     threshold_clean_pe = ped_mean_pe + sigma_clean*ped_std_pe
@@ -73,7 +74,7 @@ def get_unusable_pixels(dl1_path):
     with tables.open_file(dl1_path) as f:
         calibration_id = f.root.dl1.event.telescope.monitoring.pedestal.col('calibration_id')
         unusable_pixels = np.where(f.root[dl1_params_tel_mon_cal_key].col(
-                                   'unusable_pixels')[calibration_id[1],
+                                   'unusable_pixels')[calibration_id[INTERLEAVED_CALIBRATION_ID],
                                                       HIGH_GAIN,
                                                       :] == True)
     return unusable_pixels
--- a/lstchain/calib/camera/pixel_threshold_estimation.py
+++ b/lstchain/calib/camera/pixel_threshold_estimation.py
@@ -49,7 +49,7 @@ def get_threshold_from_dl1_file(dl1_path, sigma_clean):
     picture_thresh: np.ndarray
         picture threshold calculated using interleaved pedestal events
     """
-    
+
     ped_mean_pe, ped_std_pe = get_bias_and_std(dl1_path)
 
     # If problem with interleaved pedestal std values occur, take pedestal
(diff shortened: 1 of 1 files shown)


lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -62,17 +62,18 @@ def get_threshold_from_dl1_file(dl1_path, sigma_clean):
...
     # return pedestal interleaved threshold from data run for high gain
     return threshold_clean_pe[interleaved_events_id, HIGH_GAIN, :]
 
-def get_unusable_pixels(dl1_path, interleaved_events_id):
+def get_unusable_pixels(dl1_path):
     with tables.open_file(dl1_path) as f:
+        calibration_id = f.root.dl1.event.telescope.monitoring.pedestal.col('calibration_id')
         unusable_pixels = np.where(f.root[dl1_params_tel_mon_cal_key].col(
-                                   'unusable_pixels')[interleaved_events_id,
+                                   'unusable_pixels')[calibration_id[1],
```
maxnoe, line 76: Why the hardcoded 1 here?

pawel21: as I understood `1` in `calibration_id = f.root.dl1.event.telescope.monitoring.pedestal.col('calibration_id')` correspond to id for interleaved events.

maxnoe: Then please define a constant `INTERLEAVED_CALIBRATION_ID`

pawel21: Done

lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -49,30 +50,31 @@ def get_threshold_from_dl1_file(dl1_path, sigma_clean):
...
     ped_mean_pe, ped_std_pe = get_bias_and_std(dl1_path)
 
     # If problem with interleaved pedestal std values occur, take pedestal
     # std values from calibration run.
     # Correct interleaved pedestal std array should have shape (2,2,1855)
     if ped_std_pe.shape[0] == 2:
-        interleaved_events_id = 1
+        interleaved_events_id = INTERLEAVED_CALIBRATION_ID
     else:
         interleaved_events_id = 0
```
moralejo, line 62: For consistency, you should define   ORIGINAL_CALIBRATION_ID = 0   (or whatever other name)

moralejo, line 62: This is contradictory, if it is 0 it is not interleaved

lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -49,30 +50,31 @@ def get_threshold_from_dl1_file(dl1_path, sigma_clean):
...
     # return pedestal interleaved threshold from data run for high gain
     return threshold_clean_pe[interleaved_events_id, HIGH_GAIN, :]
 
-def get_unusable_pixels(dl1_path, interleaved_events_id):
+def get_unusable_pixels(dl1_path):
     with tables.open_file(dl1_path) as f:
+        calibration_id = f.root.dl1.event.telescope.monitoring.pedestal.col('calibration_id')
         unusable_pixels = np.where(f.root[dl1_params_tel_mon_cal_key].col(
-                                   'unusable_pixels')[interleaved_events_id,
+                                   'unusable_pixels')[calibration_id[INTERLEAVED_CALIBRATION_ID],
```
moralejo, line 77: Why taking out the second argument? Should be called "pedestal_events_id" or something like that (can be interleaved or not; if not, by the way, there is no guarantee that the stars are in the right position). And then that should be used inside get_unusable_pixels where you now use INTERLEAVED_CALIBRATION_ID

lstchain/calib/camera/pixel_threshold_estimation.py
pawel21: Done

pawel21: Fixed

lstchain/calib/camera/pixel_threshold_estimation.py
pawel21: Done

moralejo (approved):
Looks fine to me.

maxnoe (commented):
The ids don't actually refer to events but to a set of calibration coefficients, right?

So I'd suggest using `pedestal_id` or `pedestal_calib_id` instead of `pedestal_events_id`.

moralejo:
> The ids don't actually refer to events but to a set of calibration coefficients, right?
> 
> So I'd suggest using `pedestal_id` or `pedestal_calib_id` instead of `pedestal_events_id`.

Good point, indeed. We should keep events for camera events.

pawel21:
> The ids don't actually refer to events but to a set of calibration coefficients, right?
> 
> So I'd suggest using `pedestal_id` or `pedestal_calib_id` instead of `pedestal_events_id`.

Done

moralejo (approved):

PR #680, opened 2021-04-08
filter events before applying models

vuillaut:
The events were not filtered before applying the models, leading to an issue when `nan` values were in the data in the latest production, nicely reported by @chaimain and @garciagenrique.

It also means that the filters were not applied in case only applying the dl1_to_dl2 step (with stricter filters than when dl1 are created).

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -395,11 +395,14 @@ def apply_models(dl1, classifier, reg_energy, reg_disp_vector, focal_length=28*u
     """
 
     config = replace_config(standard_config, custom_config)
-
-    dl2 = dl1.copy()
-
     regression_features = config["regression_features"]
     classification_features = config["classification_features"]
+    events_filters = config["events_filters"]
+
+    dl2 = utils.filter_events(dl1,
+                              filters=events_filters,
+                              finite_params=config['regression_features'] + config['classification_features'],
+                              )
       
     #Reconstruction of Energy and disp_norm distance
     dl2['log_reco_energy'] = reg_energy.predict(dl2[regression_features])


rlopezcoto (approved):
Models from the standard processing are trained without any additional filters, so should not be affected by this, right?

garciagenrique:
> Models from the standard processing are trained without any additional filters, so should not be affected by this, right?

They shouldn't as the the filter_event is applied just before the training.

garciagenrique:
However it was the case for the MC prod `20210331_v0.7.0.post57+gitc741c9d_prod5_trans_80_local_tailcut_8_4`. No testing proton file was created, as the testing dl1 file was not filtered before.

vuillaut:
> Models from the standard processing are trained without any additional filters, so should not be affected by this, right?

The models are not affected indeed, since filtering is correctly applied when building them
The testing files are (necessary to produce the IRFs).

garciagenrique (approved):

garciagenrique:
@chaimain I will then create the missing dl2_proton_testing using 0.7.1 plus this merge - not ideal but I don't think that a new prod is needed.

moralejo:
> > Models from the standard processing are trained without any additional filters, so should not be affected by this, right?
> 
> The models are not affected indeed, since filtering is correctly applied when building them
> The testing files are (necessary to produce the IRFs).

Hi,

where can I find the applied filters? Are they just to exclude nans, or is there any other cut?

A.

vuillaut:
> > > Models from the standard processing are trained without any additional filters, so should not be affected by this, right?
> > 
> > 
> > The models are not affected indeed, since filtering is correctly applied when building them
> > The testing files are (necessary to produce the IRFs).
> 
> Hi,
> 
> where can I find the applied filters? Are they just to exclude nans, or is there any other cut?
> 
> A.

Here are the standard ones:
https://github.com/cta-observatory/cta-lstchain/blob/b28dcb2493599b29bc96111d53b4f554b8b8fb8c/lstchain/data/lstchain_standard_config.json#L28

moralejo:
> So basically, there is only this filter on `r` that is effective: `[0, 1]`

Ok, thanks. Event the r < 1 m (i.e. 2 degrees) seems pretty loose (which is fine).

PR #681, opened 2021-04-08
Update lstchain_standard_config.json

maxnoe:
Checking one of the runs @moralejo reported with problems for the ff tagging, I investigated a bit and came to the conclusion the we should adapt the default values for the ff tagging to min=3000, fraction=0.8.

Pixel values we are looking at for the tagging with the new min, max and the old min:

r0_sums

New resulting ratios with the new cut in ratio:

pixel_fraction

With that, the identified FF rate goes from 86 to 92 events per second.


moralejo (approved):

maxnoe:
@moralejo, sorry forgot the src dependent config.

moralejo (approved):

PR #686, opened 2021-04-09
allow more complex naming of irf files

vuillaut:
It would be great to allow more complex names for IRF files.
For example, in the production, we use tags such as `20210403_v0.7.1_prod5_trans_80_local` in the names that include the lstchain version. In this case, we would end up with `20210403_v0.fits.gz` as filename.

### DIFF
--- a/lstchain/tools/lstchain_create_irf_files.py
+++ b/lstchain/tools/lstchain_create_irf_files.py
@@ -166,12 +166,12 @@ class IRFFITSWriter(Tool):
                 )
 
         filename = self.output_irf_file.name
-        if filename.split(".")[1] != "fits":
+        if filename.split(".")[-2] != "fits" or filename.split(".")[-1] != "gz":
             self.log.warning(
                 f"{filename} is not a correct "
-                "compressed FITS file name. It will be corrected."
+                "compressed FITS file name (.fits.gz). It will be corrected."
             )
-            filename = filename.split(".")[0] + ".fits.gz"
+            filename += ".fits.gz"
             self.output_irf_file = self.output_irf_file.parent / filename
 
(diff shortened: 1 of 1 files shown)


vuillaut:
Alternatively, we can just ask for a filename ending with `fits.gz`, raising an `Error` if not.

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -166,12 +166,12 @@ def setup(self):
...
         filename = self.output_irf_file.name
-        if filename.split(".")[1] != "fits":
+        if filename.split(".")[-2] != "fits" or filename.split(".")[-1] != "gz":
             self.log.warning(
                 f"{filename} is not a correct "
-                "compressed FITS file name. It will be corrected."
+                "compressed FITS file name (.fits.gz). It will be corrected."
             )
-            filename = filename.split(".")[0] + ".fits.gz"
+            filename += ".fits.gz"
```
chaimain, line 174: Won't this just add to a wrong name, like `irf.xyz` -> `irf.xyz.fits.gz`, instead of just taking care of the extensions?

chaimain:
Maybe we can also add the `irf_type` and `mc_type`(gamma) information also in the name, to make it apparent just by looking at the filename and not just reading the relevant metadata inside the file.
Asking the user to provide a name could also be good, but maybe a default name should be worked upon.

We can also look into naming the DL3 index files as well, to add more information in the filename.

maxnoe:
> Alternatively, we can just ask for a filename ending with fits.gz, raising an Error if not.

Yes, please just check for `.fits` OR `.fits.gz`.

maxnoe (approved):

PR #255, opened 2019-12-24
add 4 scripts of prototype for applying DRS4 sampling interval correction

yutonogami:
* The create_sampling_interval.py can create the counts table of coming test-pulses for all pixels to create sampling intervals table. It should use when some LST-camera modules are changed.
* The ConvertCountsToInterval.py can merge sevelar counts tables made by create_sampling_interval.py and convert the counts to sampling intervals. It should use when some LST-camera modules are changed.
* The charge_calculation.py is example script to apply sampling intervals (using ROOT). It will be merged only the algorithm of applying sampling intervals to LSTR0Correction class. The sampling interval correction should use to calibrate raw waveform.
* The merge_hist.py can merge the result files to a root-file using ROOT. It will be delete when the algorithm of applying sampling interval correction is added to LSTR0Correction class.

PR #690, opened 2021-04-12
Tool to generate a HDF table of sampling interval coefficients

SeiyaNozaki:
This PR is a continuation of #255.
I added a new tool and class to create sampling interval coefficients tables, which is used for `TimeSamplingCorrection`.

This tool has three stages:
- count pulse peak on each capacitor
  -   we usually use 1 million events, so better to submit a subrun-wise job to save time. It seems some modules have some problems with this method (almost synchronization between test pulse and the first capacitor id), but this trouble appears randomly for all of the modules. So we take 2-3 runs for a single gain to generate sampling interval coefficients. 

- stack single fits file, convert into sampling interval coefficients, calculate charge resolution with new coefficients for the verification
  - stack multiple fits files generated by subrun-wise jobs and convert the stucked peak counts into sampling interval coefficients for each run. Then,  charge resolutions are calculated with new coefficients for the verification. If a given module has trouble in a specific run, charge resolution would be not good than expected. So the final coefficient values would be what achieved the best resolutions. The final coefficients and used run for the generation of coefficients on each module are saved for each gain.

-  merge coefficients of both gains into a single HDF table
   - merge output files of the previous stage for each gain. The generated HDF file should be compatible with `TimeSamplingCorrection` already implemented in lstchain.
  https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/calib/camera/time_sampling_correction.py#L14

### DIFF
--- a/lstchain/calib/camera/sampling_interval_coefficient_calculate.py
+++ b/lstchain/calib/camera/sampling_interval_coefficient_calculate.py
@@ -19,8 +19,8 @@ class SamplingIntervalCalculate(Component):
         self.peak_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
         self.fc_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
 
-        self.peak_count_stuck ={}
-        self.fc_count_stuck = {}
+        self.peak_count_stack ={}
+        self.fc_count_stack = {}
 
         self.sampling_interval_coefficient = {}
         self.charge_array_after_corr ={}
@@ -52,30 +52,30 @@ class SamplingIntervalCalculate(Component):
             self.fc_count[np.arange(N_PIXELS), fc % N_CAPACITORS_CHANNEL] += 1
             
 
-    def stuck_single_sampling_interval(self, file_list, gain):
+    def stack_single_sampling_interval(self, file_list, gain):
 
         for single_file in file_list:
 
             run_id, gain_in_file, peak_count, fc_count = load_single_fits_sampling_interval(single_file)
             
             if gain_in_file == gain:
-                if not run_id in self.peak_count_stuck.keys():
-                    self.peak_count_stuck[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
-                    self.fc_count_stuck[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
+                if not run_id in self.peak_count_stack.keys():
+                    self.peak_count_stack[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
+                    self.fc_count_stack[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
                 
-                self.peak_count_stuck[run_id] += peak_count
-                self.fc_count_stuck[run_id] += fc_count
+                self.peak_count_stack[run_id] += peak_count
+                self.fc_count_stack[run_id] += fc_count
 
 
     def convert_to_samp_interval_coefficient(self, gain):
         # convert peak counts to sampling interval coefficient
 
-        for run_id in self.peak_count_stuck.keys():
+        for run_id in self.peak_count_stack.keys():
             self.sampling_interval_coefficient[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL + N_SAMPLES])
 
             for pixel in range(N_PIXELS):
                 self.sampling_interval_coefficient[run_id][pixel, :N_CAPACITORS_CHANNEL] = \
-                    self.peak_count_stuck[run_id][pixel] / np.sum(self.peak_count_stuck[run_id][pixel]) * N_CAPACITORS_CHANNEL
+                    self.peak_count_stack[run_id][pixel] / np.sum(self.peak_count_stack[run_id][pixel]) * N_CAPACITORS_CHANNEL
                 
                 self.sampling_interval_coefficient[run_id][pixel, N_CAPACITORS_CHANNEL:] = \
                     self.sampling_interval_coefficient[run_id][pixel, :N_SAMPLES]
@@ -85,7 +85,7 @@ class SamplingIntervalCalculate(Component):
         self.charge_array_before_corr = np.zeros([N_PIXELS, 60000])
         self.charge_reso_array_before_corr = np.zeros(N_PIXELS)
 
-        for run_id in self.peak_count_stuck.keys():
+        for run_id in self.peak_count_stack.keys():
             self.charge_array_after_corr[run_id] = np.zeros([N_PIXELS, 60000])
             self.charge_reso_array_after_corr[run_id] = np.zeros(N_PIXELS)
             
@@ -120,7 +120,7 @@ class SamplingIntervalCalculate(Component):
 
                 self.charge_array_before_corr[pixel][count] = np.sum(waveform[gain, pixel,integ_start[pixel]:integ_last[pixel]])
 
-                for run_id in self.peak_count_stuck.keys():
+                for run_id in self.peak_count_stack.keys():
                     samp_interval_coefficient = self.sampling_interval_coefficient[run_id]
                     
                     self.charge_array_after_corr[run_id][pixel][count] = \
@@ -135,7 +135,7 @@ class SamplingIntervalCalculate(Component):
             charge_array_before_corr_final = charge_array_before_corr_final[charge_array_before_corr_final!=0]
             self.charge_reso_array_before_corr[pixel] = np.std(charge_array_before_corr_final)/np.mean(charge_array_before_corr_final)
 
-            for run_id in self.peak_count_stuck.keys():
+            for run_id in self.peak_count_stack.keys():
                 charge_array_after_corr_final = self.charge_array_after_corr[run_id][pixel]   
                 charge_array_after_corr_final = charge_array_after_corr_final[charge_array_after_corr_final!=0] 
                 self.charge_reso_array_after_corr[run_id][pixel] = np.std(charge_array_after_corr_final)/np.mean(charge_array_after_corr_final)
--- a/lstchain/calib/camera/sampling_interval_coefficient_calculate.py
+++ b/lstchain/calib/camera/sampling_interval_coefficient_calculate.py
@@ -16,8 +16,8 @@ class SamplingIntervalCalculate(Component):
         The SamplingIntervalCalculate class to create a sampling interval coefficient table for LST readout system using chip DRS4.
     """
     def __init__(self):
-        self.peak_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
-        self.fc_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
+        self.peak_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL], dtype=np.uint16)
+        self.fc_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL], dtype=np.uint16)
 
         self.peak_count_stack ={}
         self.fc_count_stack = {}
@@ -27,7 +27,7 @@ class SamplingIntervalCalculate(Component):
         self.charge_reso_array_after_corr ={}
 
         self.charge_reso_final = np.zeros(N_PIXELS)
-        self.used_run = np.zeros(N_PIXELS)
+        self.used_run = np.zeros(N_PIXELS, dtype=np.uint16)
         self.sampling_interval_coefficient_final = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
 
     def increment_peak_count(self, event, tel_id, gain, r0_r1_calibrator):
@@ -60,8 +60,8 @@ class SamplingIntervalCalculate(Component):
             
             if gain_in_file == gain:
                 if not run_id in self.peak_count_stack.keys():
-                    self.peak_count_stack[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
-                    self.fc_count_stack[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
+                    self.peak_count_stack[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL], dtype=np.uint16)
+                    self.fc_count_stack[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL], dtype=np.uint16)
                 
                 self.peak_count_stack[run_id] += peak_count
                 self.fc_count_stack[run_id] += fc_count
@@ -142,7 +142,7 @@ class SamplingIntervalCalculate(Component):
 
     def verify(self):
         n_keys=len(self.charge_reso_array_after_corr.keys())
-        run_id_array = np.zeros(n_keys)
+        run_id_array = np.zeros(n_keys, dtype=np.uint16)
         charge_reso_array_after_corr_all = np.zeros([n_keys, 1855])
         
         for i, ikey in enumerate(self.charge_reso_array_after_corr.keys()):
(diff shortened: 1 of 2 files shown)


lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+N_PIXELS = 1855
+N_CAPACITORS_CHANNEL = 1024
+N_SAMPLES = 40
+
+class SamplingIntervalCalculate(Component):
+    """
+        The SamplingIntervalCalculate class to create a sampling interval coefficient table for LST readout system using chip DRS4.
+    """
+    def __init__(self):
+        self.peak_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
```
maxnoe, line 19: `zeros` creates a float array by default, counts should be integers, right? So specify `dtype=np.int64` or another appropriate int dtype.

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+        self.fc_count = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
```
maxnoe, line 20: As above

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+            pulse_peak = np.argmax(waveform[gain], axis=1)
+            fc = first_capacitors[tel_id][gain]
+            pulse_peak_abs_pos = (fc + r1_sample_start + pulse_peak) % N_CAPACITORS_CHANNEL
+            
+            # increment peak count array
+            self.peak_count[np.arange(N_PIXELS), pulse_peak_abs_pos] += 1
+            self.fc_count[np.arange(N_PIXELS), fc % N_CAPACITORS_CHANNEL] += 1
+            
+
+    def stuck_single_sampling_interval(self, file_list, gain):
```
maxnoe, line 55: What does `stuck` mean here? Are you sure that's the right word?

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+
+        for single_file in file_list:
+
+            run_id, gain_in_file, peak_count, fc_count = load_single_fits_sampling_interval(single_file)
+            
+            if gain_in_file == gain:
+                if not run_id in self.peak_count_stuck.keys():
+                    self.peak_count_stuck[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
```
maxnoe, line 63: dtype

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+                    self.fc_count_stuck[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL])
```
maxnoe, line 64: dtype?

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+    def convert_to_samp_interval_coefficient(self, gain):
+        # convert peak counts to sampling interval coefficient
+
+        for run_id in self.peak_count_stuck.keys():
+            self.sampling_interval_coefficient[run_id] = np.zeros([N_PIXELS, N_CAPACITORS_CHANNEL + N_SAMPLES])
+
+            for pixel in range(N_PIXELS):
+                self.sampling_interval_coefficient[run_id][pixel, :N_CAPACITORS_CHANNEL] = \
```
maxnoe, line 77: Please use implicit line continuations inside parentheses instead of these line breaks with `\`.

Also, it looks like this loop is not necessary but could be avoided by using the `axis` keyword for the numpy functions.

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
SeiyaNozaki: sorry, it is a typo... it should be 'stack'

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+    def calc_charge(self, count, event, tel_id, gain, r0_r1_calibrator):
+
+        waveform = event.r1.tel[tel_id].waveform
+        
+        first_capacitors = r0_r1_calibrator.first_cap
+        r1_sample_start = r0_r1_calibrator.r1_sample_start.tel[tel_id]
+        r1_sample_end = r0_r1_calibrator.r1_sample_end.tel[tel_id]
+
+        # Check pulse
+        pulse_event_flag = np.sum(np.max(waveform[gain], axis=1) > 100) > 1800
```
maxnoe, line 102: This needs more context and probably the unnamed constants here should be transformed into either global constants or configurable traitlets.

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+            for pixel in range(N_PIXELS):
+                self.sampling_interval_coefficient[run_id][pixel, :N_CAPACITORS_CHANNEL] = \
+                    self.peak_count_stuck[run_id][pixel] / np.sum(self.peak_count_stuck[run_id][pixel]) * N_CAPACITORS_CHANNEL
+                
+                self.sampling_interval_coefficient[run_id][pixel, N_CAPACITORS_CHANNEL:] = \
+                    self.sampling_interval_coefficient[run_id][pixel, :N_SAMPLES]
+
+                
+    def set_charge_array(self, gain):
+        self.charge_array_before_corr = np.zeros([N_PIXELS, 60000])
```
maxnoe, line 85: Why 60000?

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
...
+        self.charge_reso_array_before_corr = np.zeros(N_PIXELS)
+
+        for run_id in self.peak_count_stuck.keys():
+            self.charge_array_after_corr[run_id] = np.zeros([N_PIXELS, 60000])
```
maxnoe, line 89: same as above

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,170 @@
+import numpy as np
+from astropy.io import fits
+from numba import jit, prange
+
+from ctapipe.core import Component
+
+
+__all__ = ['SamplingIntervalCalculate']
+
+N_PIXELS = 1855
```
maxnoe, line 10: you can import these from `ctapipe_io_lst.constants.`

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
```diff
@@ -0,0 +1,248 @@
...
+            self.start_stuck_verify()
+
+        if self.merge_flag:
+            self.start_merge()
+
+    def start_count(self):
+        self.log.debug('start peak counting')
+
+        for i, event in enumerate(self.eventsource):
+            if event.index.event_id % 500 == 0:
```
maxnoe, line 147: Use `tqdm` for a proper progress bar instead.

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
SeiyaNozaki: thanks, I set the proper data type (uint16 is enough).

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
SeiyaNozaki: thanks for the comment. done.
(Concerning the loop, please let me know if there is a more straightforward way...!)

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
SeiyaNozaki: added contexts, and set global constants for them

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
SeiyaNozaki: I modified related codes. If `max_events` is given, this given value will be used to generate a new array.
If not (`max_events` is `None` by default), just use a large enough number (55000) to cover all events in a single subrun.
(A single subrun usually contains ~53000 events.)
I'm sure it is not the best way... so I'm very happy if you have any good idea...

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
SeiyaNozaki: done!

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
SeiyaNozaki: ok, I didn't understand `tqdm` before:) Now `tqdm` is used during even loop.

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
maxnoe: You can actually get the number of events by doing `len(source.multi_file)`, I will also add `__len__` to the `LSTEventSource`, but that will require a new release.

A more flexible solution would be to use a python list and append to it, converting to a 2d numpy array at the end.

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
```diff
@@ -0,0 +1,262 @@
...
+from tqdm import tqdm
+
+from ctapipe.core import Provenance, traits
+from ctapipe.core import Tool
+from ctapipe_io_lst import LSTEventSource
+from ctapipe_io_lst.calibration import LSTR0Corrections
+from lstchain.calib.camera.sampling_interval_coefficient_calculate import SamplingIntervalCalculate
+from lstchain.paths import run_info_from_filename
+
+class SamplingIntervalCoefficientHDFWriter(Tool):
```
maxnoe, line 13: This seems just to be three tools in one that don't really share code and even use different traitlets. I think really doing three smaller tools would be better and less confusing (since the help will only show the options for each step, not for all steps).

SeiyaNozaki: I see. Actually, I just wanted to gather all of the procedures in a single file. Is it possible to put three smaller tools in a single file and call one of the tools when executing the command...? or one file per one tool?

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
```diff
@@ -0,0 +1,262 @@
...
+    """
+    Tool that generates a sampling interval coefficient hdf file.
+    For getting help run:
+    lstchain_create_sampling_interval_coefficient_file --help
+    """
+
+    name = "SamplingIntervalCoefficientHDFWriter"
```
FrancaCassol, line 20: Hi Seiya,
if I understand correctly you are writing fits files, instead of a HDF5 file.  
Would be it possible to write indeed a h5 file? (we decided to prefer this format)

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
```diff
@@ -0,0 +1,262 @@
...
+        help="Path to the output fits file",
+        directory_ok=False,
+    ).tag(config=True)
+
+    glob = traits.Unicode(
+        help="Filename pattern to glob files in the directory",
+        default_value="*.fits"
+    ).tag(config=True)
+
+    gain = traits.Int(
```
FrancaCassol, line 48: Is there a reason to sparate HG and LG? 
It would be better to have all the data in a single file, so better to process them together if possible

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
FrancaCassol: Forget my comment, I see that the final file is in h5 and also that you put together the gains at the ends. 
Perhaps a bit misleading the fact that you have single tool for the three steps

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
maxnoe: You can do multiple Tools per file and have different entrypoints, but I would just make three files

SeiyaNozaki:
Thank you for the comments, @maxnoe and @FrancaCassol !
I refactored the codes to make them simple and easy to understand.
Now there are two steps (`lstchain_create_peak_count_file_for_sampling_interval_calib` and `lstchain_create_sampling_interval_coefficient_file`): the first one is to count the pulse peak on each capacitor and the second is to stack, convert and verify the sampling interval coefficients.
The summary of the procedure is below:
image

I also added a function to produce a result pdf file like below:
[sampling_interval_coefficient_results_HG.pdf](https://github.com/cta-observatory/cta-lstchain/files/6409153/sampling_interval_coefficient_results_HG.pdf)

It would be very appreciated to have a look again.

SeiyaNozaki:
Hi @FrancaCassol @maxnoe,
Sorry for the delay, but came back again.
I'm happy if you can check the updated codes again when you have time!

lstchain/calib/camera/sampling_interval_coefficient_calculate.py
```diff
@@ -0,0 +1,283 @@
...
+            if criteria_flag:
+                self.sampling_interval_coefficient_final[pixel] = self.sampling_interval_coefficient[self.used_run[pixel]][pixel, :N_CAPACITORS_CHANNEL]
+            else:
+                logger.info(f'[charge resolution] {CHANNEL[self.gain]} pixel {pixel}: {self.charge_reso_final[pixel]} > {CRITERIA[self.gain]}')
+                self.sampling_interval_coefficient_final[pixel] = np.ones(N_CAPACITORS_CHANNEL)
+
+    def plot_results(self, verify_data_path, output_file):
+
+        plt.rcParams['font.size']=15
+        camera = CameraGeometry.from_name('LSTCam-002')
```
FrancaCassol, line 197: This is an outdated camera (we have now version 4).
Better to use:
```
from ctapipe_io_lst import load_camera_geometry
camera = load_camera_geometry()
camera = camera.transform_to(EngineeringCameraFrame())
```

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
```diff
@@ -20,21 +25,11 @@ class SamplingIntervalCoefficientHDFWriter(Tool):
     name = "SamplingIntervalCoefficientHDFWriter"
     description = "Generate a sampling interval coefficient table"
 
-    input_fits = traits.Path(
-        help="Path to the generated peak count fits files",
+    input_peak_count_fits = traits.Path(
+        help="Path to the generated peak count fits files by lstchain_create_peak_count_file_for_sampling_interval_calib",
```
FrancaCassol, line 29: Hi @SeiyaNozaki ,

I do not see this tool in your PR, am I wrong?

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
```diff
@@ -0,0 +1,133 @@
...
+        if self.input_peak_count_fits.is_dir():
+            self.path_list = sorted(self.input_peak_count_fits.glob(self.glob))
+
+    def start(self):
+        
+        self.sampling_interval_coefficient_calculate.stack_peak_count_fits(self.path_list)
+        self.sampling_interval_coefficient_calculate.convert_to_samp_interval_coefficient()
+        self.sampling_interval_coefficient_calculate.set_charge_array()
+
+        for i, event in enumerate(tqdm(
```
maxnoe, line 83: No need for enumarete, the event already has `event.count`

lstchain/tools/lstchain_create_sampling_interval_coefficient_file.py
maxnoe: Then this line is not needed anymore

Issue #501, opened 2020-08-27
Speeding up coordinate transformation and writing them in the DL2 files


rlopezcoto:
Discussed in #420, #421 and #437 the speeding up of the Alt/Az -> RA/Dec and to keep track of things. 
@MaxNoe has been working on https://github.com/astropy/astropy/pull/10647 (thanks!) to speed up the transformation, and we should be able to use it whenever it is accepted, likely with astropy 4.2. 
If it is urgent for people working on IRFs, we could implement a temporary, slower way of writing this information into the standard DL2 files, otherwise we could leave it to whenever this is accepted.

rlopezcoto:
2 orders of magnitude correspond to two new generation of experiments 😄 
This is great @MaxNoe, thank you very much, let us leave this Issue open until astropy 4.2 gets out and we can implement the fast coordinate transformation in lstchain

maxnoe:
Astropy 4.2rc1 is published and on pypi, so you could start testing it.

Be advised that the speed up is only possible when transforming many coordinates close in time at once (so we can fully utilize it e.g. to transform all dl2 events at once from horizotal it ICRS) but it does not help with transforming for single coordinates / timestamps like e.g. needed to transform star coordinates into camera frame for the DL0→DL1 step.

maxnoe:
Astropy 4.2 was release some time ago and is available via pip and conda.

PR #691, opened 2021-04-12
Fix pointing in DL3 Pointing HDU

chaimain:
Using the first event for pointing position, instead of a mean value to have the correct pointing position compared with the target source position.

### DIFF
--- a/lstchain/irf/hdu_table.py
+++ b/lstchain/irf/hdu_table.py
@@ -296,9 +296,6 @@ def create_event_list(
     date_end = t_stop_iso[:10]
     time_end = t_stop_iso[11:]
 
-    mean_time = Time(
-        data["dragon_time"].value.mean() * u.s, format="unix", scale="utc"
-    )
     MJDREF = Time("1970-01-01T00:00", scale="utc")
 
     # Position parameters
@@ -328,7 +325,7 @@ def create_event_list(
         # Nomenclature is to be worked out or have a separate way to mark mispointings
         mode = "UNDETERMINED"
 
-    log.error(
+    log.info(
         "Source pointing difference with camera pointing"
         f" is {source_pointing_diff:.3f}"
     )
@@ -343,8 +340,8 @@ def create_event_list(
             # Optional columns
             "GAMMANESS": data["gh_score"],
             "MULTIP": u.Quantity(np.repeat(len(tel_list), len(data)), dtype=int),
-            "GLON":src_sky_pos.galactic.l.to(u.deg),
-            "GLAT":src_sky_pos.galactic.b.to(u.deg),
+            "GLON": src_sky_pos.galactic.l.to(u.deg),
+            "GLAT": src_sky_pos.galactic.b.to(u.deg),
             "ALT": reco_alt.to(u.deg),
             "AZ": reco_az.to(u.deg),
         }
@@ -427,12 +424,18 @@ def create_event_list(
     pnt_header["MJDREFF"] = ev_header["MJDREFF"]
     pnt_header["TIMEUNIT"] = ev_header["TIMEUNIT"]
     pnt_header["TIMESYS"] = ev_header["TIMESYS"]
-    pnt_header["OBSGEO-L"] = (location.lon.to_value(),
-        'Geographic longitude of telescope (deg)')
-    pnt_header["OBSGEO-B"] = (location.lat.to_value(),
-        'Geographic latitude of telescope (deg)')
-    pnt_header["OBSGEO-H"] = (location.height.to_value(),
-        'Geographic latitude of telescope (m)')
+    pnt_header["OBSGEO-L"] = (
+        location.lon.to_value(),
+        "Geographic longitude of telescope (deg)",
+    )
+    pnt_header["OBSGEO-B"] = (
+        location.lat.to_value(),
+        "Geographic latitude of telescope (deg)",
+    )
+    pnt_header["OBSGEO-H"] = (
+        round(location.height.to_value(), 2),
+        "Geographic latitude of telescope (m)",
+    )
 
     pnt_header["TIMEREF"] = ev_header["TIMEREF"]
 
--- a/lstchain/irf/hdu_table.py
+++ b/lstchain/irf/hdu_table.py
@@ -6,6 +6,7 @@ import astropy.units as u
 from astropy.table import Table, QTable
 from astropy.io import fits
 from astropy.coordinates import SkyCoord, AltAz
+from astropy.coordinates.erfa_astrom import ErfaAstromInterpolator, erfa_astrom
 from astropy.time import Time
 
 from lstchain.__init__ import __version__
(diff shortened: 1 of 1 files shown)


lstchain/irf/hdu_table.py
```diff
@@ -427,12 +424,18 @@ def create_event_list(
...
     pnt_header["TIMEUNIT"] = ev_header["TIMEUNIT"]
     pnt_header["TIMESYS"] = ev_header["TIMESYS"]
-    pnt_header["OBSGEO-L"] = (location.lon.to_value(),
-        'Geographic longitude of telescope (deg)')
-    pnt_header["OBSGEO-B"] = (location.lat.to_value(),
-        'Geographic latitude of telescope (deg)')
-    pnt_header["OBSGEO-H"] = (location.height.to_value(), 
-        'Geographic latitude of telescope (m)')
+    pnt_header["OBSGEO-L"] = (
+        location.lon.to_value(),
```
maxnoe, line 428: Add `u.deg` here

lstchain/irf/hdu_table.py
```diff
@@ -427,12 +424,18 @@ def create_event_list(
...
+        "Geographic longitude of telescope (deg)",
+    )
+    pnt_header["OBSGEO-B"] = (
```
maxnoe, line 431: sasme

lstchain/irf/hdu_table.py
```diff
@@ -427,12 +424,18 @@ def create_event_list(
...
+        location.lat.to_value(),
+        "Geographic latitude of telescope (deg)",
+    )
+    pnt_header["OBSGEO-H"] = (
```
maxnoe, line 435: u.m

maxnoe:
What does this fix? Reading the GADF documentation, filling the *mean* makes for alt/az makes more sense than filling the first event value.

maxnoe:
> Using the first event for pointing position, instead of a mean value to have the correct pointing position compared with the target source position.

That doesn't make sense. The target source position is in ICRS, which doesn't change in time. So why should changing the time(interval) change for the alt/az position change anything with respect to the source?

maxnoe (changes requested):
I don't think this change should be made. altaz pointing should be the mean when tracking a position in ICRS.

chaimain:
> What does this fix? Reading the GADF documentation, filling the _mean_ makes for alt/az makes more sense than filling the first event value.

While using the mean value of pointing information, to calculate the offset by measuring the separation between the source position and the pointing information, I was getting large differences (compared to 0.4 deg for wobble) than when using the first events' information.

maxnoe:
Ok, looking at the code again, you don't fill `ALT_PNT` / `AZ_PNT`, but you transform the altaz pointing to icrs to fill `RA_PNT` / `DEC_PNT`. So using the first entry is fine.

But we know also have the icrs pointing position in the pointing information in `event.pointing.array_ra` and `event.pointing.array_dec`.


Please fix the unit use anyway.

lstchain/irf/hdu_table.py
```diff
@@ -309,15 +307,19 @@ def create_event_list(
...
+        frame=AltAz(obstime=time_utc[0], location=location),
+    )
+
+    with erfa_astrom.set(ErfaAstromInterpolator(30 * u.s)):
+        src_sky_pos_icrs = src_sky_pos.transform_to(frame="icrs")
+        tel_pnt_sky_pos_icrs = tel_pnt_sky_pos.transform_to(frame="icrs")
 
     # Observation modes
-    source_pointing_diff = source_pos.separation(tel_pnt_sky_pos)
+    source_pointing_diff = source_pos.separation(tel_pnt_sky_pos_icrs)
```
maxnoe, line 322: This should also be `source_pos_icrs`, otherwise astropy will make the same transform twice

chaimain: The `source_pos` is already in ICRS as I pass it from either getting the source positions from the user input or from resolving the source name

chaimain:
Updated the coordinated transformation of the reco sky positions to ICRS using astropy.erfa_astrom as per [#10647](https://github.com/astropy/astropy/pull/10647) and [Issue #501](https://github.com/cta-observatory/cta-lstchain/issues/501)

lstchain/irf/hdu_table.py
```diff
@@ -309,13 +307,16 @@ def create_event_list(
...
     tel_pnt_sky_pos = SkyCoord(
-        alt=pointing_alt.mean(),
-        az=pointing_az.mean(),
-        frame=AltAz(obstime=mean_time, location=location),
+        alt=pointing_alt[0],
+        az=pointing_az[0],
+        frame=AltAz(obstime=time_utc[0], location=location),
     ).transform_to(frame="icrs")
 
+    with erfa_astrom.set(ErfaAstromInterpolator(30 * u.s)):
```
maxnoe, line 317: I typically use five minutes now. See https://docs.astropy.org/en/stable/coordinates/index.html#improving-performance-for-arrays-of-obstime

lstchain/irf/hdu_table.py
```diff
@@ -309,13 +307,16 @@ def create_event_list(
...
+        src_sky_pos_icrs = src_sky_pos.transform_to(frame="icrs")
```
maxnoe, line 318: If src pos is already in icrs, then why transform it here again?

lstchain/irf/hdu_table.py
maxnoe: But you have a a transform_to('icrs') just above

lstchain/irf/hdu_table.py
chaimain: The positions being transformed are from the events, while, the `source_pos` I use is for metadata and to compare with the pointing positions to evaluate the observation mode or any 'mis-pointing'

lstchain/irf/hdu_table.py
chaimain: Yes, I tried to check with varying the time period, and although the time taken to transform does not differ much, the min/max separation was lower.
Although, I do not think it would be too significant for our use case. 
Is 30 seconds too low?

lstchain/irf/hdu_table.py
chaimain: That is for `src_sky_pos` which is coming from the reco position of each observed event.
Maybe, the variable name is confusing? I can change the `source_pos` to `source_std_pos` maybe?

maxnoe: No, use reco_pos or similar for the reconstructed positiom AMD Reserve source for the assumed / observed source.

maxnoe: Let me propose `reco_altaz` and `reco_icrs`

chaimain: And accordingly, for pointing position, `pnt_altaz` and `pnt_icrs`?

lstchain/irf/hdu_table.py
maxnoe: shouldn't really matter

maxnoe (approved):

PR #694, opened 2021-04-15
Fix peaktimes overwritten with nan

maxnoe:


vuillaut (approved):

moralejo (approved):

vuillaut (commented):
instead of a copy, wouldn't it be faster to pass the geometry and the peak_times with the mask ?

vuillaut (changes requested):
putting it on hold :)

vuillaut (approved):
Hotfix, to be removed with function in ctapipe v0.11

Issue #677, opened 2021-04-08
Raise an error when no `observation_mode` is provided in the configuration


rlopezcoto:
In places like:
https://github.com/cta-observatory/cta-lstchain/blob/8373d074282ce953bef0fff780e32bd6904a1aed/lstchain/reco/dl1_to_dl2.py#L563-L582

if no `observation_mode` is provided due to the usage of an old/different configuration file in which it is not defined, an error should be raised. We could set `wobble` as de default one and issue a `warning` if no `observation_mode` is provided, but this may be dangerous and is probably better if the user needs to define it

SeiyaNozaki:
Hi @rlopezcoto, thank you for pointing out this issue!
I will add lines to raise an error if `observation_mode` is not defined.

With the current codes,  analyzers should set `observation_mode` (on/wobble), and `source_name` or `source_ra,dec` for wobble data.
Now I'm thinking to change that analyzer needs to set only `source_name` or `source_ra,dec` even for on data, and no `observation_mode`. It would consume more time than setting just (0,0) for on data, but allowable I think. It also would be easier to find out any problems (pointing issue) or mistakes (wrong source). What do you think...?

rlopezcoto:
Thanks @SeiyaNozaki. I personally find `observation_mode` a good way for not-so-experienced people to explicitly set what kind of observation was performed in the analysis.

PR #698, opened 2021-04-19
Remove source-dependent parameters from a data frame of dl2_params_lstcam_key

SeiyaNozaki:
This PR is to fix a small bug in source-dependent analysis to generate dl2 data.
Source-dependent parameters should be saved in `dl2_params_src_dep_lstcam_key`, and other standard parameters should be in `dl2_params_lstcam_key`.
However, it seems some dl2 source-dependent parameters are still in `dl2_params_lstcam_key` because only dl1 source-dependent parameters are dropped...
In this PR, all of the source-dependent parameters should be dropped in a standard dl2 data frame.

I also modified the codes to raise an error if `observation_mode` is not defined mentioned as #677


rlopezcoto (approved):
thanks @SeiyaNozaki

SeiyaNozaki:
fixed via #772

PR #699, opened 2021-04-19
Sort events by timestamp before calculating effective observation time

morcuended:
I tried to calculate the effective and elapsed time from runs taken on different days. It happened that I did not sort the input events beforehand so I was getting nonsensical negative results.

The easiest solution would be to order first the events according to their timestamp then do the rest of the calculations as before.

### DIFF
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -716,6 +716,9 @@ def get_effective_time(events):
     t_eff: astropy Quantity (in seconds, if input has no units)
     t_elapsed: astropy Quantity (ditto)
     """
+    # Sorting events by timestamps so the input list of events does not have 
+    # to be ordered, for example when dealing with runs from different days. 
+    events = events.sort_values(by="dragon_time")
     timestamp = np.array(events["dragon_time"])
     delta_t = np.array(events["delta_t"])
 
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -719,11 +719,7 @@ def get_effective_time(events):
     """
     # Sorting events by timestamps so the input list of events does not have 
     # to be ordered, for example when dealing with runs from different days. 
-    if isinstance(events, pd.DataFrame):
-        events = events.sort_values("dragon_time")
-    elif isinstance(events, QTable):
-        events.sort("dragon_time")
-    timestamp = np.array(events["dragon_time"])
+    timestamp = np.sort(np.array(events["dragon_time"]))
     delta_t = np.array(events["delta_t"])
 
     if not isinstance(timestamp, u.Quantity):
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -20,6 +20,7 @@ import pandas as pd
 from astropy.coordinates import AltAz, SkyCoord, EarthLocation
 from astropy.time import Time
 from astropy.utils import deprecated
+from astropy.table import QTable
 from ctapipe.coordinates import CameraFrame
 
 from . import disp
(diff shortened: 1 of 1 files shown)


lstchain/reco/utils.py
```diff
@@ -718,7 +719,10 @@ def get_effective_time(events):
     """
     # Sorting events by timestamps so the input list of events does not have 
     # to be ordered, for example when dealing with runs from different days. 
-    events = events.sort_values(by="dragon_time")
+    if isinstance(events, pd.DataFrame):
+        events = events.sort_values("dragon_time")
+    elif isinstance(events, QTable):
+        events.sort("dragon_time")
     timestamp = np.array(events["dragon_time"])
```
moralejo, line 726: I think it is enough to do   timestamp= np.sort(np.array(events["dragon_time"]))
There is no need to sort out the delta_t's

moralejo: Related topic:

https://github.com/cta-observatory/cta-lstchain/blob/5032edb033a3d01bc6db45b930ad62d2df3a302a/lstchain/reco/utils.py#L695

Not that it will change the result significantly, but in the calculation of delta_t, if the previous event is in a different obs_id, we should set delta_t to nan instead of the calculated value (which can be negative as well if the obs_id's are not ordered)

moralejo: > Related topic:
> 
> https://github.com/cta-observatory/cta-lstchain/blob/5032edb033a3d01bc6db45b930ad62d2df3a302a/lstchain/reco/utils.py#L695
> 
> Not that it will change the result significantly, but in the calculation of delta_t, if the previous event is in a different obs_id, we should set delta_t to nan instead of the calculated value (which can be negative as well if the obs_id's are not ordered)

And if we do so, we should replace np.mean and np.min in get_effective_time by nanmean and nanmim

moralejo: And if we do so, we should replace delta_t's  mean and min (in get_effective_time) by nanmean and nanmin

maxnoe: Just compute the diff per run and concatenate.

maxnoe:
```
delta_ts = np.concatenate([
    np.diff(np.sort(run["dragon_time"]))
    for run_id, run in df.group_by("obs_id")
])
```

morcuended: > I think it is enough to do timestamp= np.sort(np.array(events["dragon_time"]))
> There is no need to sort out the delta_t's

Yes, indeed much straightforward.

morcuended: > ```
> delta_ts = np.concatenate([
>     np.diff(np.sort(run["dragon_time"].to_numpy()))
>     for run_id, run in df.group_by("obs_id")
> ])
> ```

Do you mean to calculate `delta_t` like this in https://github.com/cta-observatory/cta-lstchain/blob/5032edb033a3d01bc6db45b930ad62d2df3a302a/lstchain/reco/utils.py#L677?

maxnoe: Sorry, I misremembered. delta_t is now calculated and stored in the dl1 files right?

So there shouldn't be any time deltas between obs ids? Or is this function also run on multiple obs_ids ?

morcuended: It is not expected to run on multiple obs_ids in the lstosa pipeline. Actually, we currently do it subrun wise. But it might happen that you try to calculate the obs time from several dl2 files on a run basis for several days. Hence you end up with a data frame containing `delta_t` and `timestamps` of not sorted events.

maxnoe: yes, this is clear. But if delta t is calculated beforehand, atleaest the comment by @moralejo about diffs between obs_ids should not be a concern, right?

morcuended: Right, at least I see it like this. However, for the calculation of the `time_diff` array in `get_effective_time` I see that calculating the diffs per `obs_id` and then concatenate the arrays could be useful.

moralejo: As of now we do not calculate delta_t before DL1, we do it in the DL1 to DL2 step.

SeiyaNozaki: If there is much difference between observations (let's say 4kHz and 10kHz), the calculation of the effective time is still correct?

moralejo: @SeiyaNozaki No, I don't think so (but would be good to include that in the tests, perhaps the deviation is not large).  It is anyway very recommendable that this is applied on observations with more or less constant rate.

SeiyaNozaki: ok, thanks! I just wanted to know for my better understanding:)

PR #704, opened 2021-04-29
Save source-dependent parameters in a separate file and use an external link for original parameters

SeiyaNozaki:
The main update in this PR is to use an external link for source-independent parameters when adding source-dependent parameters to save disk space.
For now, when adding source-dependent parameters, those parameters are overwritten in the same dl1 file. So analyzer needs to copy the official dl1 data to each local space to add parameters.
This PR enables us to save a separate dl1 file with source-dependent parameters, so we don't need to copy the original data.
Furthermore, the original dl1 parameters are saved as an external link in this separate file. So we can dereference the original dl1 parameters without a copy.

Small cleanups are also included.


SeiyaNozaki:
@maxnoe 
I assigned you as a reviewer as suggested by @moralejo .
I'm happy if you have a look to find any problems or suggestions!

maxnoe:
@SeiyaNozaki I am a bit reluctant on this. I found working with liked hdf5 files quite cumbersome when playing around with it. 

Does it still work when downloading the two files to your own computer? If you change the relative paths? 

Is the datasize of the *parameters* (not the images) really that large of a factor? 

Otherwise I would recommend to just not include the images for the source dependent files.

moralejo:
@maxnoe, I suggested, on top of removing the images, not to repeat the parameter information and use instead these links. I never used them myself, so I am not aware of the possible downsides.

I was mainly thinking about the case of files living in La Palma, with links to the standard DL1 & DL2 locations, which are quite stable. As of now I think we have many terabytes of duplicated DL1/2 tables.
This use of links may be optional, in case sb is thinking about moving the file out of La Palma.

SeiyaNozaki:
Thank you for the comments,  @maxnoe and @moralejo .

> Does it still work when downloading the two files to your own computer? If you change the relative paths?

An absolute path of the original file is saved in external_link. 
So it will not work when downloading the files... but it should work when we change the path of the generated file in La Palma server.

> Is the datasize of the parameters (not the images) really that large of a factor?

I checked the data size at several conditions:

| image| dl1 srcindep | n_off | srcdep | file size |
| ------------- | ------------- | ------------- | ------------- | ------------- |
|o|o|-|-|697M (original)|
|x|o|-|-|7.5M|
|link|link|-|-|15K|
|x|o|1|o|13M|
|x|link|1|o|5.4M|
|x|o|3|o|19M|
|x|link|3|o|12M|

(There are 40 parameters in the default dl1 keys, and (7x`n_off`+6) source-dependent parameters.)
So it will save 7.5MB/subrun. What do you think of this merit?

Now I also thinking to add source-dependent parameters to the dl2 files since `None` values etc are filtered in dl1_to_dl2 process. So what I'm thinking is:
- read a dl2 table from the default dl2 file
- remove only dl2 parameters obtained with applying RF
- add source-dependent parameters
- apply RF for each source-dependent parameters and obtain dl2-source-dep parameters
- save default srcindep parameters (as an external link) and dl2-srcdep parameters

DL1 files with source-dependent parameters will be used only for applying RF, so no need to save dl1 parameters with source-dep ones, but add srcdep parameters just before applying RF. There are also some duplicates between srcdep-dl1 and srcdep-dl2 for now, my proposal will solve this duplicate.

Any comments are welcome!

rlopezcoto:
Hi all, although duplicating information, if we store the src-dep parameters for 3 OFFs, we will be increasing the file size only by 50% (12 M -> 19 M) and we would be avoiding any possible downsides of linking other files. I would then be in favor of, when running the src-dep analysis, storing by default dl1 without images (dl1b) + source-dependent parameters.

SeiyaNozaki:
Thank you for the comments, @rlopezcoto!
If @moralejo agrees, I would forget an external link solution, and just remove images when adding source-dependent parameters.

To avoid duplicates, I'd like to add an option to calculate src-dep parameters just before applying RF in `lstchain_dl1_to_dl2.py`.  Then, we don't need to save another dl1 file, but save only dl2 files with src-dep parameters. Is it fine?

rlopezcoto:
> Thank you for the comments, @rlopezcoto!
> If @moralejo agrees, I would forget an external link solution, and just remove images when adding source-dependent parameters.

Thanks to you for the implementation!

> To avoid duplicates, I'd like to add an option to calculate src-dep parameters just before applying RF in `lstchain_dl1_to_dl2.py`. Then, we don't need to save another dl1 file, but save only dl2 files with src-dep parameters. Is it fine?

you would also need to calculate src-dep parameters to train the RF

moralejo:
Hi, ok, I think given the side complications we can live with duplicating the parameters for now.
Note however that those 700 MB of images will become less than half when we move to the standard DL1 model, so in relative terms the parameters will become more relevant.

SeiyaNozaki:
thank you @moralejo . I will keep it in mind since it is directly related to my life on la palma server...!

I updated this branch, to add an option to calculate source-dependent parameters just before applying RF model.
I'm happy if you would have a look.

rlopezcoto:
Hi @SeiyaNozaki, what about this: 
> you would also need to calculate src-dep parameters to train the RF

did you also implement it in `lstchain_trainpipe.py`?

SeiyaNozaki:
Hi @rlopezcoto 
>did you also implement it in lstchain_trainpipe.py?

Not in `lstchain_trainpipe.py`, but in `build_models` in `dl1_to_dl2.py` since inputs for `build_models` are filename, not data frame.

rlopezcoto:
ok great, that sounds good, thanks Seiya!

rlopezcoto:
@SeiyaNozaki apart from conflict solving, is this ready for review?

SeiyaNozaki:
I'd like to close this PR, and just implement #888 instead of this PR.
#888 is ready for review.

rlopezcoto:
moved to #888

Issue #705, opened 2021-04-29
Profiling r0 to dl1


maxnoe:
I ran the `lstchain_data_r0_to_dl1`  on Run 2610.000 to profile the run.

Results:

![Screenshot_20210429_183755](https://user-images.githubusercontent.com/5488440/116586871-32b68b80-a91a-11eb-996d-ef26d6fafdcc.png)

Unsurprisingly, most part is spent in the low-level calibration, but one major block is the `apply_time_delta_cleaning` which was implemented very inefficiently in ctapipe 0.10.5.

Will follow up also with a profiling of the same for MCs

maxnoe:
This is with copying the `apply_time_delta_cleaning` method from the current ctapipe master:

![Screenshot_20210429_190241](https://user-images.githubusercontent.com/5488440/116590010-8b3b5800-a91d-11eb-9299-d67d6942c572.png)


140s faster, essentially it goes from a major contributor to barely visible

maxnoe:
Everything else looks basically like expected, now huge gains possible and also most time is spend in the ctapipe functions and methods, not in the lstchain code.

maxnoe:
Now on to MCs.

PR #706, opened 2021-04-29
Use apply_time_delta_cleaning from current ctapipe master

maxnoe:
See discussion in #705

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -18,7 +18,6 @@ from ctapipe.image import (
     HillasParameterizationError,
     hillas_parameters,
     tailcuts_clean,
-    apply_time_delta_cleaning,
 )
 from ctapipe.image.morphology import number_of_islands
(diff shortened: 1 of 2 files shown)


lstchain/reco/r0_to_dl1.py
```diff
@@ -173,14 +173,10 @@ def get_dl1(
...
         if delta_time is not None:
-            # makes sure only signal pixels are used in the time check:
-            cleaned_pixel_times = peak_time.copy()
-            cleaned_pixel_times[~signal_pixels] = np.nan
-
             signal_pixels = apply_time_delta_cleaning(
                 camera_geometry,
                 signal_pixels,
-                cleaned_pixel_times,
+                peak_time,
```
rlopezcoto, line 179: why changing the input from the masked pixel times after cleaning to the original ones?

maxnoe: Because that was done specifically to fix a bug in the ctapipe 0.10.5 version

rlopezcoto (approved):

PR #707, opened 2021-05-04
Add correction of the systematic term to the F-factor method

FrancaCassol:
This PR adds:
- the possibility to correct the the F-factor method for the quadratic component if the correction file is given (by default not given for the moment)
- a trailet to the calibration script to allow the selection of a configuration file different than the default one (for tests)
- two new plots in the calibration check pdf 

I am still performing some test so this is a draft PR

### DIFF
--- a/lstchain/calib/camera/calibration_calculator.py
+++ b/lstchain/calib/camera/calibration_calculator.py
@@ -3,9 +3,10 @@ Component for the estimation of the calibration coefficients  events
 """
 
 import numpy as np
+import h5py
 from ctapipe.core import Component
 from ctapipe.core import traits
-from ctapipe.core.traits import  Float, List
+from ctapipe.core.traits import  Float, List, Path
 from lstchain.calib.camera.flatfield import FlatFieldCalculator
 from lstchain.calib.camera.pedestals import PedestalCalculator
 from ctapipe.containers import EventType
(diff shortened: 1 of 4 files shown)


FrancaCassol:
Hi, 
the test shows that correcting the F-factor method for the B term we obtain a gain estimation which depends much less from the signal charge (some small systematic differences remain to be further studied):

Scan of February, before B corrections (estimated gains increases with charge):
plot_gain_vs_pe_20210218_no_corrected


Scan of February, after B corrections estimated with the scan of April (most gains are within 2% close to the gain estimated with filter 52):

plot_gain_vs_charge_20210218_B_corrected

FrancaCassol:
Hi, 
would be it possible to review this PR? 
Thanks

pawel21 (approved):

maxnoe (commented):
This code looks fine, however it does not explain where this `systematic_correction` file comes from or how one can create it.

Where is the code to construct this file? Could this be added to the help / docstrings?

maxnoe:
@FrancaCassol ok. But that also means, that there now test for this right?

Wouldn't the natural order be to first make a PR for the creation of the file and then one that depends on its existence?

If the structure is already final, I can also add the file to the test data so you can add a unit test for it.

FrancaCassol:
Ok, then I do a PR with all together

PR #708, opened 2021-05-05
Pixel noise adder & light smearer

moralejo:
This is for use in the dl1ab script when processing Monte Carlo:  optionally add noise to the pixel charges (to simulate higher NSB), and smear the light in the pixels (to simulate a worse PSF). With this one can achieve a better tuning of the MC to the data, without having to re-run sim_telarray.

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -34,6 +34,15 @@
     "leakage_intensity_width_2": [0, 1]
   },
 
+  "image_modifier": {
+    "modify_image": false,
+    "extra_noise_in_dim_pixels": 1.5,
+    "extra_bias_in_dim_pixels": 0.6,
+    "transition_charge": 8,
+    "extra_noise_in_bright_pixels": 1.44,
+    "smeared_light_fraction": 0.15
+  },
+
   "tailcut": {
     "picture_thresh":6,
     "boundary_thresh":3,
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -82,6 +82,19 @@ def main():
     else:
         config = std_config
 
+    modify_image = False
+    if "image_modifier" in config:
+        imconfig = config["image_modifier"]
+        modify_image = imconfig["modify_image"]
+        extra_noise_in_dim_pixels = imconfig["extra_noise_in_dim_pixels"]
+        extra_bias_in_dim_pixels = imconfig["extra_bias_in_dim_pixels"]
+        transition_charge = imconfig["transition_charge"]
+        extra_noise_in_bright_pixels = imconfig["extra_noise_in_bright_pixels"]
+        smeared_light_fraction = imconfig["smeared_light_fraction"]
+        if modify_image and args.noimage is False:
+            log.info("NOTE: Using the modify_image option means images will not be saved.")
+            args.noimage = True
+
     if args.pedestal_cleaning:
         log.info("Pedestal cleaning")
         clean_method_name = 'tailcuts_clean_with_pedestal_threshold'
@@ -162,23 +175,23 @@ def main():
                 image = row['image']
                 peak_time = row['peak_time']
 
-                # Add noise in pixels, to adjust MC to data noise levels.
-                # TO BE DONE: in case DL1a is saved, we have to save the modified image!
-                # TO BE DONE: in case of "pedestal cleaning" (not used now in MC) we should
-                # recalculate picture_th above!
-                qcopy = image.copy()
-                qlimit = 8
-                bias = -0.9
-                mu = 1.5
-                mu_2 = 1.44
-                image[qcopy < qlimit] += (np.random.poisson(mu, (qcopy < qlimit).sum()) + bias)
-                image[qcopy > qlimit] += (np.random.poisson(mu_2, (qcopy > qlimit).sum()) - mu_2)
+                if modify_image:
+                    # Add noise in pixels, to adjust MC to data noise levels.
+                    # TO BE DONE: in case of "pedestal cleaning" (not used now in MC) we should
+                    # recalculate picture_th above!
+                    qcopy = image.copy()
 
-                # Move part of the light in each pixel (fraction) into its neighbors, to simulate a worse PSF:
-                fraction=0.25
-                q_spread = image * camera_geom.neighbor_matrix * fraction / 6
-                q_remaining = image * (1 - fraction)
-                image = q_remaining + np.sum(q_spread, axis=1)
+                    image[qcopy < transition_charge] += (np.random.poisson(extra_noise_in_dim_pixels,
+                                                                           (qcopy < transition_charge).sum()) -
+                                                         extra_noise_in_dim_pixels + extra_bias_in_dim_pixels)
+                    image[qcopy > transition_charge] += (np.random.poisson(extra_noise_in_bright_pixels,
+                                                                           (qcopy > transition_charge).sum()) -
+                                                         extra_noise_in_bright_pixels)
+
+                    # Move part of the light in each pixel (fraction) into its neighbors, to simulate a worse PSF:
+                    q_spread = image * camera_geom.neighbor_matrix * smeared_light_fraction / 6
+                    q_remaining = image * (1 - smeared_light_fraction)
+                    image = q_remaining + np.sum(q_spread, axis=1)
 
                 signal_pixels = tailcuts_clean(camera_geom,
                                                image,
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -86,6 +86,8 @@ def main():
     if "image_modifier" in config:
         imconfig = config["image_modifier"]
         modify_image = imconfig["modify_image"]
+        if modify_image:
+            log.info('image_modifier configuration:', imconfig)
         extra_noise_in_dim_pixels = imconfig["extra_noise_in_dim_pixels"]
         extra_bias_in_dim_pixels = imconfig["extra_bias_in_dim_pixels"]
         transition_charge = imconfig["transition_charge"]
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -87,7 +87,7 @@ def main():
         imconfig = config["image_modifier"]
         modify_image = imconfig["modify_image"]
         if modify_image:
-            log.info('image_modifier configuration:', imconfig)
+            log.info(f"image_modifier configuration: {imconfig}")
         extra_noise_in_dim_pixels = imconfig["extra_noise_in_dim_pixels"]
         extra_bias_in_dim_pixels = imconfig["extra_bias_in_dim_pixels"]
         transition_charge = imconfig["transition_charge"]
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -162,6 +162,18 @@ def main():
                 image = row['image']
                 peak_time = row['peak_time']
 
+                # Add noise in pixels, to adjust MC to data noise levels.
+                # TO BE DONE: in case DL1a is saved, we have to save the modified image!
+                # TO BE DONE: in case of "pedestal cleaning" (not used now in MC) we should
+                # recalculate picture_th above!
+                qcopy = image.copy()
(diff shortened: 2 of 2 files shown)


vuillaut (commented):
Can I ask you to make a proper function to add noise in `lstchain.image` that could then be used in the script? 
```
add_charges_noise(image, extra_noise_dim_pixels=0, extra_bias_dim_pixels=0, ...)
```

maxnoe:
I was about to ask the same.

A component would be better, since that would also be usable once we switch to tools. something like this:


```
from numpy.random import default_rng


class ImageModifer(TelescopeComponent):
     additional_noise = FloatTelescopeParameter(...).tag(config=True)
     random_seed = Int(default_value=0).tag(config=True)

    def __init__(self, subarray, *args, **kwargs):
        super().__init__(self, subarray, *args, **kwargs)
       self.rng = default_rng(self.seed)


    def __call__(self, tel_id, image):
        image = image + self.rng.poisson(...)
```

moralejo:
I considered using a function, but I thought it might perhaps slow it down - if that is not the case if it of course cleaner.

maxnoe:
> but I thought it might perhaps slow it down

noooooo

moralejo:
I will make two functions now. As for the component, not sure it is really needed right now, in view of the transition to stage1 and its corresponding version of the DL1ab.

I'd like to have this released today, to see if we can test it on the full MC production, including building RFs based on the tuned MC for the Crab analysis.

moralejo:
> > but I thought it might perhaps slow it down
> 
> noooooo

oooook, so just my prejudice.

maxnoe:
At least the random state should be fixed somehow. Best would be to give the rng object to the function and create one with a specific seed (e.g. obs_id) in the r0_to_dl1 function.

So do:

```
rng = default_rng(obs_id)
add_noise_in_pixels(..., rng=rng)
```

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -172,8 +172,13 @@ def main():
         if set(dl1_params_input).intersection(disp_params):
             parameters_to_update.extend(disp_params)
 
+        if increase_nsb or increase_psf:
```
maxnoe, line 175: nitpick, but the psf does not actually use the RNG, right?

moralejo: Why should it? It is a completely deterministic spread of light :D

lstchain/image/modifier.py
```diff
@@ -0,0 +1,83 @@
...
+    camera edge and hence don't have all possible neighbors
+
+    Returns
+    -------
+    Modified (diffuminated) image
+
+    """
+
+    # number of neighbors of completely surrounded pixels:
+    max_neighbors = np.max([len(pixnbs) for pixnbs in camera_geom.neighbors])
```
maxnoe, line 75: That's gotta be pretty slow. And does not change for LST. I would go for just putting an N_PIXEL_NEIGHBORS = 6 at the top.

Faster would probably by `np.count_nonzero(camera_geometry.neighbor_matrix).sum(axis=0)`, but again, I'd just use the constant here.

lstchain/scripts/lstchain_dl1ab.py
moralejo: I think it is ok like that... despite the low number of photons

moralejo: AH!
I got you now, silly me...

lstchain/image/modifier.py
```diff
@@ -0,0 +1,83 @@
...
+    camera_geom: camera geometry
+
+    smeared_light_fraction: fraction of the light in a pixel that will be
+    distributed equally among its immediate surroundings, i.e. immediate
+    neighboring pixels. Some light is lost for pixels which are at the
+    camera edge and hence don't have all possible neighbors
+
+    Returns
+    -------
+    Modified (diffuminated) image
```
maxnoe, line 70: Is diffuminated an actually use word?

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -190,17 +190,16 @@ def main():
                     # Add noise in pixels, to adjust MC to data noise levels.
                     # TO BE DONE: in case of "pedestal cleaning" (not used now
                     # in MC) we should recalculate picture_th above!
-                    if (increase_nsb):
                         image = add_noise_in_pixels(rng, image,
```
maxnoe, line 193: One level of indentation to much now here.

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -190,17 +190,16 @@ def main():
...
                                                     extra_noise_in_dim_pixels,
                                                     extra_bias_in_dim_pixels,
                                                     transition_charge,
                                                     extra_noise_in_bright_pixels
                                                     )
-                    if (increase_psf):
-                        image = smear_light_in_pixels(image,
-                                                      camera_geom,
-                                                      smeared_light_fraction)
+                if (increase_psf):
```
maxnoe, line 199: superfluous parentheses

lstchain/image/modifier.py
moralejo: > Is diffuminated an actually used word?

in Spanglish...  :-C

lstchain/image/modifier.py
```diff
@@ -0,0 +1,83 @@
...
+    is determined by the Cherenkov light, and would not be modified by the
+    additional NSB noise (presumably small compared to the C-light)
+
+    Returns
+    -------
+    Modified (noisier) image
+
+    """
+
+    qcopy = image.copy()
```
maxnoe, line 43: This can be simplified to:

```
bright_pixels = image > transition_charge
noise = np.where(bright_pixels, extra_noise_in_bright_pixels, extra_noise_in_dim_pixels)
bias = np.where(bright_pixels, -extra_noise_in_bright_pixels, extra_bias_in_dim_pixels - extra_noise_in_dim_pixels)

image = image + rng.poisson(noise) - bias
```

moralejo: Done

maxnoe: And you make a copy of the image but modify the image itself. Is that intentional?

moralejo:
For the record: the default values of the corrections (note that they are switched off by default) are tuned to 20201120 Crab's data (ZD<30 deg), relative to the standard (low-NSB FOV) Monte Carlo.

maxnoe:
@moralejo Is there a way to easily optimize these settings for a given run? E.g. a script that can output the parameters?

moralejo:
> @moralejo Is there a way to easily optimize these settings for a given run? E.g. a script that can output the parameters?

To be done, if the approach works. For the NSB it is pretty simple (at least if we fix the transition). For the PSF, we would have to do it with muons, because Crab is not always there (not sure if muons will give us enough precision). Alternatively we can use the measurements  of the optical spot that we take regularly.

moralejo:
Ok, now the code is much cleaner and easier to understand.

But just tested it and it is no longer working :'-C  Back to draft.

lstchain/image/modifier.py
```diff
@@ -0,0 +1,83 @@
...
+
+    """
+
+    # Move a fraction of the light in each pixel (fraction) into its neighbors,
+    # to simulate a worse PSF:
+    q_smeared = (image * camera_geom.neighbor_matrix *
+                 smeared_light_fraction / N_PIXEL_NEIGHBORS)
+    # Light remaining in pixel:
+    q_remaining = image * (1 - smeared_light_fraction)
+    image = q_remaining + np.sum(q_smeared, axis=1)
```
maxnoe, line 83: This makes a copy and you don't return the new image

maxnoe: return image and call like `image = smear_light_in_pixels(...)`

maxnoe: Alternatively, to mutate image, do `image[:] = q_remaining + `. But I think I prefer the pure function with new assignment.

moralejo:
Ok, tested. Ready.

lstchain/data/lstchain_standard_config.json
```diff
@@ -34,6 +34,16 @@
...
 
+  "image_modifier": {
+    "increase_nsb": false,
+    "extra_noise_in_dim_pixels": 1.5,
+    "extra_bias_in_dim_pixels": 0.6,
+    "transition_charge": 8,
+    "extra_noise_in_bright_pixels": 1.44,
+    "increase_psf": false,
+    "smeared_light_fraction": 0.175
+  },
```
vuillaut, line 45: do we really want to add this config in the standard one?

maxnoe: I think it's to document the options (It is switched off by default). 

That would not be necessary if we'd directly write  a component...

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -82,6 +83,24 @@ def main():
...
+            log.info(f"image_modifier configuration: {imconfig}")
+        extra_noise_in_dim_pixels = imconfig["extra_noise_in_dim_pixels"]
+        extra_bias_in_dim_pixels = imconfig["extra_bias_in_dim_pixels"]
+        transition_charge = imconfig["transition_charge"]
+        extra_noise_in_bright_pixels = imconfig["extra_noise_in_bright_pixels"]
+        smeared_light_fraction = imconfig["smeared_light_fraction"]
+        if (increase_nsb or increase_psf) and args.noimage is False:
+            log.info("NOTE: Using the image_modifier options means images will "
+                     "not be saved.")
+            args.noimage = True
```
maxnoe, line 102: Why force disable this? What If we want to check if the code is working as expected by storing the images and comparing them to real data?

lstchain/data/lstchain_standard_config.json
moralejo: No strong opinion, it is disabled by default, but I can put it in a separate one.

lstchain/scripts/lstchain_dl1ab.py
moralejo: I was afraid of double-application... Any proposal for a safeguard against that?

maxnoe: store the used values in the metadata and fail to apply that is already there in the input file?

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -82,6 +83,24 @@ def main():
     else:
         config = std_config
 
+    increase_nsb = False
+    increase_psf = False
+    if "image_modifier" in config:
+        imconfig = config["image_modifier"]
```
maxnoe, line 89: Why make all of these config objects local variables? Why not just use the config where the function is called?

lstchain/scripts/lstchain_dl1ab.py
maxnoe: > I'd say it is not urgent. 

Then why actively disable a feature that is already implemented? Especially one that would be quite handy to have to check how good this feature is?

lstchain/scripts/lstchain_dl1ab.py
moralejo: Just tried, but I don't see the advantage, the code where the function is called becomes much less readable.
Besides, I tested this as it is - don't underestimate my capacity to introduce a bug while doing an apparently harmless change.

lstchain/scripts/lstchain_dl1ab.py
moralejo: The way it is done is that the whole 'images' thing is removed if not wanted in the output.   So it is not that I removed a feature, the image was never "actively saved" by the script. It is probably easy to do, but not for me, and since I did not need that feature, and time is limited, I did not add it.  It can be done later when one needs it. If you need it now, please go ahead, I am not against it.

moralejo: > I meant testing in the sense of comparing e.g. pixel distributions between data and mc.
> 
> For that, being able to load an image table transformed like this would be helpful, right?

I've done that applying the correction on the fly when you do the comparison.

maxnoe: ah ok, sorry, forgot that the dl1ab script only copies the intial table. Fine with me.

maxnoe (approved):

lstchain/scripts/lstchain_dl1ab.py
moralejo: Ok, thanks!

PR #709, opened 2021-05-06
Update DL3 Tools with energy dependent optimized cuts

chaimain:
Adding an option (for now), to use optimised (for now only g/h separation) cuts for MC and observed data, for the DL3 files.
Optimization is done on a fixed gamma efficiency for each reco energy bin, using pyirf functions.

I have to test this a bit more before making it open PR.


chaimain:
Sorry for the long delay on this PR.

The work on energy-dependent cuts for gammaness is ready. 

However, for cuts on the theta parameter, it is not clear how we should define the ON/OFF region. There is an ongoing work in gammapy (https://github.com/gammapy/gammapy/pull/3684) on energy-dependent 1D spectrum extraction, and once it is resolved/merged there, a separate PR can be made in lstchain to appropriately prepare the DL3 files to have the energy-dependent theta cuts and store the data event lists appropriately.

If there is no other issue on this, can this PR be merged before the Analysis School begins next week, or for the 0.8.3 release?

rlopezcoto:
Hi @chaimain this is a bit long PR that has not been looked at yet, is this absolutely necessary for the school? Otherwise it may be a better idea to leave its merging for afterwards

maxnoe:
> There is an ongoing work in gammapy (gammapy/gammapy#3684) on energy-dependent 1D spectrum extraction, and once it is resolved/merged there, a separate PR can be made in lstchain to appropriately prepare the DL3 files to have the energy-dependent theta cuts and store the data event lists appropriately.

The ongoing work is on how to implement the analysis for the already defined format. The required format will not change as a result of that work.

You just need to include the `RAD_MAX` HDU in the DL3 files instead of the `RAD_MAX` header keyword in the `EVENTS` and irf HDUs

chaimain:
@rlopezcoto It is not absolutely essential, but if it were to be included, it would be great. It is my fault for delaying this one for so long. The school can work fine without this PR.

chaimain:
> > There is an ongoing work in gammapy ([gammapy/gammapy#3684](https://github.com/gammapy/gammapy/pull/3684)) on energy-dependent 1D spectrum extraction, and once it is resolved/merged there, a separate PR can be made in lstchain to appropriately prepare the DL3 files to have the energy-dependent theta cuts and store the data event lists appropriately.
> 
> The ongoing work is on how to implement the analysis for the already defined format. The required format will not change as a result of that work.
> 
> You just need to include the `RAD_MAX` HDU in the DL3 files instead of the `RAD_MAX` header keyword in the `EVENTS` and irf HDUs

True, but I am more concerned about how we use separate ON/OFF event lists, based on the theta cuts we apply. It can be kept up to the user on how they work it through maybe, but I would like it to be at least indicated in the Tool, or shown in an example notebook on how it can be done.

So, I wanted to keep that work separate from this PR

maxnoe:
> True, but I am more concerned about how we use separate ON/OFF event lists, based on the theta cuts we apply.

We don't use separated event ON/OFF event lists at the DL3 level. DL3 files should include all events surviving gammaness cuts without applying a directional cut with respect to any on or off region.

chaimain:
> > True, but I am more concerned about how we use separate ON/OFF event lists, based on the theta cuts we apply.
> 
> We don't use separated ON/OFF event lists at the DL3 level. DL3 files should include all events surviving gammaness cuts without applying a directional cut with respect to any on or off region.
> 
> The work in gammapy is precisely on how off regions are determined and how the theta cuts are applied. This is the domain of gammapy, not the DL3 export.

So, are you implying that the energy-dependent directional cuts should be applied to all the events in MC and observed data?

maxnoe:
> So, are you implying that the energy-dependent directional cuts should be applied to all the events in MC and observed data?

I don't understand the question. The cuts need to be applied on simulated events for calculating the IRFs, but they must not be already applied for simulated test datasets or observed data, as again, this is done later in gammapy and the DL3 events list must not have a directional cut already applied, as this prevents letting gammapy or any other science tool chose off regions.

chaimain:
Ok, I understand this now. 
I was earlier confused, on whether we should apply direction cuts on events in DL3 event lists, as I assumed that All the cuts applied to simulation data for creating IRFs, should be applied to the real data as well.

Thank you for clarifying this. I will amend the PR accordingly.

chaimain:
Summary of the PR:
Including energy-dependent gammaness and theta cuts for IRFs and DL3 Tools.

The corresponding main functions are stored in `lstchain/io/event_selection.py`, which are `optimize_[gh/theta]_cuts` and `apply_optimized_[gh/theta]_cuts`, which in turn use pyirf functions `calculate_percentile_cut` and `evaluate_binned_cut` respectively.

The new config parameters are
- `fixed-gh-efficiency`
- `fixed-theta-containment`

In the Tools, new flags are 
- `optimize_gh` 
-  `optimize_th` (only IRF Tool)

In the IRF Tool, the presence of either flag activates both flags. The separate flags are created to separate the use case in IRF and DL3 Tool (as mentioned earlier, DL3 will not have directional cuts)

For point-like IRFs, as per https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/point_like/, 
- Fixed global theta cut will be stored as a Header keyword in the IRFs
- Energy-dependent theta cuts will be stored as RAD_MAX_2D HDU

Similarly, we store the energy-dependent gammaness cuts table as GH_CUTS HDU, which can be used by the DL3 Tool to get the gammaness cuts for given reco energy edge.
For now, the DL3 Tool will **require** the IRF file to be with the GH_CUTS HDU, to apply energy-dependent gammaness cuts.

The HDU Index file will include these new HDUs when present.

Once [PR#3733](https://github.com/gammapy/gammapy/pull/3733) in gammapy is merged, we can continue  to DL3+ analyses with these new DL3 files.

lstchain/tools/tests/test_tools.py
```diff
@@ -162,13 +163,24 @@ def test_create_dl3_with_config(temp_dir_observed_files, observed_dl2_file):
...
                 "--overwrite",
-                "--optimize-gh",
             ],
             cwd=temp_dir_observed_files,
         )
         == 0
     )
 
+@pytest.mark.private_data
+@pytest.mark.run(after="test_create_irf_full_enclosure_with_config")
```
maxnoe: Remove the `mark.run` stuff, that's from pytest ordering. Use the correct fixtures instead.

docs/examples/dl3_tool_config.json
```diff
@@ -12,7 +12,14 @@
   },
   "DL3FixedCuts": {
```
maxnoe, line 13: This is called "FixedCuts" and now contains both the global cut value and the values used to calculate the energy dependent cuts at the same time. How is it determined, what is used?

docs/examples/dl3_tool_config.json
```diff
@@ -12,7 +12,14 @@
...
     "fixed_gh_cut": 0.7,
+    "fixed_gh_efficiency": 0.9,
     "fixed_theta_cut": 0.2,
+    "fixed_theta_containment": 0.68,
     "allowed_tels": [1]
+  },
+  "DataBinning": {
+    "reco_energy_min": 0.01,
+    "reco_energy_max": 100,
+    "reco_energy_n_bins_per_decade": 5.5
```
maxnoe, line 23: Why 5.5?

docs/examples/irf_tool_config.json
```diff
@@ -12,7 +12,9 @@
   },
   "DL3FixedCuts": {
     "fixed_gh_cut": 0.7,
+    "fixed_gh_efficiency": 0.9,
```
maxnoe, line 15: Same as above in the other config.

lstchain/io/event_selection.py
```diff
@@ -53,6 +56,16 @@ class DL3FixedCuts(Component):
         default_value=0.6,
     ).tag(config=True)
 
+    fixed_gh_efficiency = Float(
```
maxnoe, line 59: The `fixed_` prefix seems unnecessary.

lstchain/io/event_selection.py
```diff
@@ -65,12 +78,101 @@ class DL3FixedCuts(Component):
     ).tag(config=True)
 
     def gh_cut(self, data):
```
maxnoe, line 80: Rename to `apply_global_gh_cut`?

lstchain/io/event_selection.py
```diff
@@ -65,12 +78,101 @@ class DL3FixedCuts(Component):
...
+        """
+        Applying a fixed global gammaness cut on a given data
+        """
         return data[data["gh_score"] > self.fixed_gh_cut]
 
+    def optimize_gh_cuts(
```
maxnoe, line 86: Add an `energy_dependent`?

lstchain/io/event_selection.py
```diff
@@ -65,12 +78,101 @@ class DL3FixedCuts(Component):
...
+            min_value = min_value,
+            max_value = max_value,
+            fill_value = data["gh_score"].max(),
+            percentile = 100 - self.fixed_gh_efficiency,
+            smoothing = smoothing,
+            min_events = min_events,
+        )
+        return gh_cuts
+
+    def apply_optimized_gh_cuts(self, data, gh_cuts):
```
maxnoe, line 112: `energy_dependent` instead of `optimized`. A single cut can also be `optimized`

lstchain/io/event_selection.py
```diff
@@ -65,12 +78,101 @@ class DL3FixedCuts(Component):
...
+
+        data["selected_gh"] = evaluate_binned_cut(
+            data["gh_score"],
+            data["reco_energy"],
+            gh_cuts,
+            operator.ge,
+        )
+        return data[data["selected_gh"]]
+
     def theta_cut(self, data):
```
maxnoe, line 126: `apply_global_theta_cut`

lstchain/io/event_selection.py
```diff
@@ -65,12 +78,101 @@ class DL3FixedCuts(Component):
...
+        """
+        Apply a fixed global theta cut on a given data
+        """
         return data[data["theta"].to_value(u.deg) < self.fixed_theta_cut]
 
+    def optimize_theta_cuts(
```
maxnoe, line 132: `energy_dependent`

lstchain/io/event_selection.py
```diff
@@ -65,12 +78,101 @@ class DL3FixedCuts(Component):
...
+        self, data, energy_bins, min_value=0.05 * u.deg,
+        fill_value=0.32 * u.deg, max_value=0.32 * u.deg,
+        smoothing=None, min_events=10
+    ):
+        """
+        Evaluating an optimized energy-dependent theta cuts, in a given
+        data, with provided reco energy bins, and other parameters to
+        pass to the pyirf.cuts.calculate_percentile_cut function
```
maxnoe, line 140: Note that the very fine binning of the theta cuts used by event display comes from an interpolation. The actual calculation happens in much coarser bins. Using to fine binning will result in too un-smooth cuts.

lstchain/io/event_selection.py
```diff
@@ -65,12 +78,101 @@ class DL3FixedCuts(Component):
...
+            min_value = min_value,
+            max_value = max_value,
+            fill_value = fill_value,
+            percentile = self.fixed_theta_containment,
+            smoothing = smoothing,
+            min_events = min_events,
+        )
+        return theta_cuts
+
+    def apply_optimized_theta_cuts(self, data, theta_cuts):
```
maxnoe, line 158: `energy_dependent` instead of `optimized`

lstchain/io/tests/test_eventselection.py
```diff
@@ -44,16 +44,31 @@ def test_dl3_fixed_cuts():
     temp_cuts.fixed_gh_cut = 0.7
     temp_cuts.fixed_theta_cut = 0.2
     temp_cuts.allowed_tels = [1, 2]
+    temp_cuts.fixed_gh_max_efficiency = 0.8
```
maxnoe, line 47: Again giving both the fixed config and the energy dependent config. These should be clearly separated.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -6,30 +6,24 @@
 The default values are written in the EventSelector and DL3FixedCuts Component
 and also given in some example configs in docs/examples/
 
+For using energy-dependent gammaness cuts, use the flag optimize_gh
```
maxnoe, line 9: again: I'd prefer calling it `energy_dependent` instead of `optimized`

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -103,6 +107,11 @@ class DataReductionFITSWriter(Tool):
         help="DEC position of the source"
     ).tag(config=True)
 
+    optimize_gh = traits.Bool(
```
maxnoe, line 110: But this is *applying* the cuts, not calculating them. Why is a switch needed at all?

I think the DL3 tool should just look into the IRF file and if it finds the `GH_CUTS` table, apply the energy dependent cuts.

Do we also store the non-genergy-dependent cut in the IRF file as header? We should, and then this script can also use that information in case of a global cut.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -159,16 +172,38 @@ def setup(self):
 
         self.log.debug(f"Output DL3 file: {self.output_file}")
 
+        if self.optimize_gh and self.input_irf:
+            try:
+                QTable.read(self.input_irf, hdu="GH_CUTS")
+            except KeyError:
+                raise ToolConfigurationError(
+                    f"{self.input_irf} does not have GH CUTS HDU, or "
```
maxnoe, line 180: Why "or", it's the same thing, right?

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -137,6 +154,16 @@ class IRFFITSWriter(Tool):
         default_value=False,
     ).tag(config=True)
 
+    optimize_gh = traits.Bool(
```
maxnoe, line 157: Again, I would call it `energy-dependent` not `optimized`

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -181,13 +218,19 @@ def setup(self):
...
+                f"{filename} is not a correct compressed FITS file name"
+                "(use .fits or .fits.gz)."
+                )
 
         if self.input_proton_dl2 and self.input_electron_dl2 is not None:
             self.only_gamma_irf = False
         else:
             self.only_gamma_irf = True
 
+        if self.optimize_gh or self.optimize_th:
```
maxnoe, line 231: No! One can use the one without the other. It's perfectly valid to apply energy dependent gammaness cut but use a fixed theta squared cut and vice versa!

maxnoe: If these are needed to be the same, why use two options? For that case, just use a single `--energy-dependent` flag

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -259,25 +302,58 @@ def start(self):
...
         source_offset_bins = self.data_bin.source_offset_bins()
 
+        gammas = self.event_sel.filter_cut(gammas)
+        gammas = self.fixed_cuts.allowed_tels_filter(gammas)
+
+        if self.optimize_gh:
+            self.gh_cuts_gamma = self.fixed_cuts.optimize_gh_cuts(
+                gammas, reco_energy_bins, min_value=0.1, max_value=0.95
+            )
+            gammas = self.fixed_cuts.apply_optimized_gh_cuts(gammas, self.gh_cuts_gamma)
```
maxnoe, line 318: Something called `fixed_cuts` is applying an energy dependent cut

maxnoe (changes requested):

docs/examples/dl3_tool_config.json
chaimain: I named it, as we use either a _fixed_ global cut or energy-dependent cuts, with _fixed_ efficiency. If this is flawed logic, or is misleading, I will rename it to `DL3Cuts`?

docs/examples/dl3_tool_config.json
chaimain: This was arbitrary choice made long time ago. I will change it to 5 as per the CTA standard.

docs/examples/dl3_tool_config.json
maxnoe: In the first case, it's a property of the cut itself, in the other case it is a property of how the cuts are determined. So yes, I think better remove the "Fixed".

docs/examples/irf_tool_config.json
chaimain: Just an arbitrary choice again. Would you prefer some other value?

lstchain/io/event_selection.py
chaimain: Ok. Will amend it

docs/examples/dl3_tool_config.json
chaimain: Understood. Thanks!

lstchain/io/event_selection.py
chaimain: Ok, will change it.

lstchain/io/event_selection.py
chaimain: Ok, will amend it.

docs/examples/irf_tool_config.json
maxnoe: I was not talking about the value, but the name of the configuration item as in the comment above on the `DL3FixedCuts`

lstchain/io/event_selection.py
chaimain: Ok. I will add the final comment to the description

lstchain/io/tests/test_eventselection.py
chaimain: Will do so.

lstchain/tools/lstchain_create_dl3_file.py
chaimain: I think, letting the user have the option to choose, whether to use energy-dependent or global cuts while creating DL3 Tool is good. 

The IRF Tool does add the header value of global `GH_CUT` and `RAD_MAX` when used.
If the user-provided IRF does not contain the `GH_CUTS` table, or a header value of the global `GH CUT` applied, then the Tool will raise an error.

I have not included a check on whether the IRF has the global `GH_CUT` to use by default, global gammaness cut when creating the DL3 file. I will fix that.

lstchain/tools/lstchain_create_dl3_file.py
chaimain: Yes. I just wanted to add an explanation on the GH_CUTS HDU, as it is not defined by GADF.

lstchain/tools/lstchain_create_dl3_file.py
maxnoe: > I think, letting the user have the option to choose, whether to use energy-dependent or global cuts while creating DL3 Tool is good.

But that's not possible here, since the IRF file has to have the same settings, otherwise you create and invalid combination of `EVENTS` table with the IRFs. 

The user can only choose when creating the IRFs, not when creating the DL3 Events file. At this point, the same settings as used for the IRF calculation **must** be applied.

lstchain/tools/lstchain_create_irf_files.py
chaimain: > No! One can use the one without the other. It's perfectly valid to apply energy dependent gammaness cut but use a fixed theta squared cut and vice versa!

Oh! I didn't realise that that is the case. So, I will just rename these 2 flags, and have them be independent of each other's value.

Would `energy-dependent-gh` and `energy-dependent-theta` be a better substitute for the names?

docs/examples/irf_tool_config.json
chaimain: Ok

lstchain/tools/lstchain_create_dl3_file.py
chaimain: Ok. I will make the changes accordingly.

chaimain:
The DL3 Tool will require the IRF file be provided, and include either,
- global gammaness cut be stored as header `GH_CUT` in all IRF HDUs, or
- energy-dependent gammaness cuts be stored as `GH_CUTS` HDU

such that the Tool can take the particular gammaness cut/s and apply it to the DL2 event list.
 
 The IRF Tool will also store similar information for theta cuts as either `RAD_MAX` header value or HDU.

The flags for using energy-dependent cuts in IRF Tool are `energy-dependent-gh` and `energy-dependent-theta`. The names of variables and functions have been adapted accordingly.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -159,16 +154,45 @@ def setup(self):
 
         self.log.debug(f"Output DL3 file: {self.output_file}")
 
+        try:
+            QTable.read(self.input_irf, hdu=1).meta["GH_CUT"]
```
maxnoe, line 158: There is no nead to read the full table to just look at the fits header.
Since we also need at least one of the two, it's enough here to check the header keyword.

So just use:

```python
from astropy.io import fits

with fits.open(self.input_irf) as hdul:
    self.energy_dependent_gh_cut = "GH_CUT" not in hdul["EFFECTIVE AREA"].header
```

To get the information whether the IRFs are pointlike or not, you can look at:

```

with fits.open(self.input_irf) as hdul:
    aeff_hdu = hdul['EFFECTIVE_AREA']
    self.energy_dependent_gh_cut = 'GH_CUT' not in aeff_hdu.header
    self.is_pointlike = aeff_hdu.header['HDUCLAS3'] == 'POINT-LIKE'
    if self.is_pointlike:
        self.energy_dependent_theta_cut = 'RAD_MAX' not in aeff_hdu.header
```

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -159,16 +154,45 @@ def setup(self):
...
         self.data = read_data_dl2_to_QTable(str(self.input_dl2))
-        ## To reduce the table columns further, add a selection of columns to be read and used.
         self.effective_time, self.elapsed_time = get_effective_time(self.data)
         self.run_number = run_info_from_filename(self.input_dl2)[1]
 
         self.data = self.event_sel.filter_cut(self.data)
-        self.data = self.fixed_cuts.gh_cut(self.data)
-        self.data = add_icrs_position_params(self.data, self.source_pos)
+
+        if self.energy_dependent_cuts:
```
maxnoe, line 179: This applies both energy dependent theta and gh cuts, although only one of the two cuts coul be energy dependent.

lstchain/tools/lstchain_create_dl3_file.py
maxnoe: Sorry, this is the create dl3 file, so no theta cut needed.

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -411,9 +517,33 @@ def start(self):
             )
             self.log.info("PSF HDU created")
 
+        if self.energy_dependent_gh:
+            gh_header = self.hdus[1].header.copy()
+            gh_header["HDUCLAS2"] = "GH_CUTS"
```
maxnoe, line 522: This is not defined in GADF, please don't add headers that make it look like it is a specified GADF HDU

maxnoe (changes requested):
This looks good, some small changes left.

lstchain/tools/lstchain_create_dl3_file.py
chaimain: Thanks! Will amend this

lstchain/tools/lstchain_create_irf_files.py
chaimain: As I need to fill the DL3 Index table, can I fill the headers for this HDU in the following way,
`"CREATOR" = pyirf {pyirf.__version__}
"HDUDOC" = "N/A"
"HDUVERS" = "N/A"
"HDUCLASS" = "N/A"
"HDUCLAS1" = "RESPONSE"
"HDUCLAS2" = "GH_CUTS"
"HDUCLAS4" = "GH_CUTS_2D"
`
Or some other values?

chaimain: And, maybe we should add this HDU information in GADF? If it seems ok with you, I can at least open an issue there

maxnoe: Why do you need to fill this into the hdu index table? It is not a GADF HDU and thus not used by gammapy

maxnoe: And creator should be `lstchain {lstchain.__version__}`

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -90,9 +104,19 @@ class IRFFITSWriter(Tool):
...
+        --global-theta-cut 0.2
         --irf-obs-time 50
+
+    Or use energy-dependent cuts based on a gamma efficiency:
+    > lstchain_create_irf_files
+        -g /path/to/DL2_MC_gamma_file.h5
+        -o /path/to/irf.fits.gz
+        --point-like (Only for point_like IRFs)
+        --energy-dependent-gh
+        --energy-dependent-theta
```
moralejo, line 117: Is there an advantage in having these? It would be enough to activate the E-dependent cuts if the --gh-efficiency and/or --theta-containment options below are used, right?

maxnoe: Yes, but this would need to change the defaults of the configuration to have them all `None` and activate the corresponding setting by giving it.

And I think there is a bug in traitlets that doesn't allow a `Float(allow_none=True)`:

https://github.com/ipython/traitlets/issues/637

I fixed it for `ctapipe_io_lst` here:

https://github.com/cta-observatory/ctapipe_io_lst/blob/e78acdc43ef4c4c95c83ec5100b7c6373af9df86/ctapipe_io_lst/event_time.py#L26

chaimain: The config file may have the options, but it will be up to the user to pass these separate flags to use the energy-dependent cuts

lstchain/io/event_selection.py
```diff
@@ -19,7 +19,7 @@ class EventSelector(Component):
     taken as inputs and filter_events() is used on a table of events
     called in with the Component.
 
-    For event_type, we choose the sub-array trigger of 32, which is for 
+    For event_type, we choose the sub-array trigger of 32, which is for
     gamma-ray event candidate, as per the latest CTA R1 Event Data Model.
```
moralejo, line 23: Not gamma-ray candidate (except "in a very loose way"), but shower candidate or "cosmics" trigger.
I did not understand why the 33 - this is just like an upper cut? Not needed, because filters are min <=  x <= max

Also, you can import ctapipe.instrument.EventType  and use EventType.SUBARRAY.value instead of 32

lstchain/io/event_selection.py
```diff
@@ -64,13 +77,105 @@ class DL3FixedCuts(Component):
...
+        max_value=0.99, smoothing=None, min_events=10
+    ):
+        """
+        Evaluating energy-dependent gammaness cuts, in a given
+        data, with provided reco energy bins, and other parameters to
+        pass to the pyirf.cuts.calculate_percentile_cut function
+        """
+        # To be sure to have the efficiency in between 0 and 100
+        if self.gh_efficiency < 1:
+            self.gh_efficiency *= 100
```
moralejo, line 97: I don't understand this. A setting is being changed without even reporting it. Surely less than 1% efficiency is a strange choice, but one would try it during some test, and would end up inadvertently with a much larger efficiency.

chaimain: No, the values in the config is entered in %, [0,1), and it is just multiplied by 100 to use in the function.

lstchain/tools/lstchain_create_irf_files.py
maxnoe: Yes, we understand the current situation. We are proposing to change that.

The issue seems to be fixed in traitlets 5.1 we are using now.

lstchain/io/event_selection.py
```diff
@@ -64,13 +77,105 @@ class DL3FixedCuts(Component):
...
+        """
+        Evaluating an optimized energy-dependent theta cuts, in a given
+        data, with provided reco energy bins, and other parameters to
+        pass to the pyirf.cuts.calculate_percentile_cut function.
+
+        Note: Using too fine binning will result in too un-smooth cuts.
+        """
+        # To be sure to have the efficiency in between 0 and 100
+        if self.theta_containment < 1:
+            self.theta_containment *= 100
```
moralejo, line 146: Same comment as before. Does not seem a reasonable thing to do, and if you want to spot a suspicious input setting (too low efficiency or containment) I would do it upon reading the configuration file, not in the function which calculates the cuts.

lstchain/io/tests/test_eventselection.py
```diff
@@ -38,33 +38,69 @@ def test_event_selection():
...
     temp_cuts.allowed_tels = [1, 2]
+    temp_cuts.gh_max_efficiency = 0.8
+    temp_cuts.theta_containment = 68
 
     temp_data = QTable({
-        "gh_score": u.Quantity(np.arange(0.1, 1.1, 0.1)),
-        "theta": u.Quantity(np.arange(0., 1., 0.1), unit=u.deg),
-        "tel_id": u.Quantity([1, 1, 2, 2, 1, 2, 1, 3, 4, 5])
+        "gh_score": u.Quantity(np.tile(np.arange(0.35, 0.85, 0.05), 3)),
+        "reco_energy": u.Quantity(np.logspace(-2.301, 1.699, 30), unit=u.TeV),
```
moralejo, line 52: Why -2.301  to  1.699 ? (instead of -2.3, 1.7,  I mean)

lstchain/io/event_selection.py
maxnoe: Than multiply it *always* by 100 **without** changing the value stored in the class.

lstchain/io/tests/test_eventselection.py
maxnoe: I guess copy paste from the EventDisplay definitions.

chaimain: Nothing particular, just wanted the values to be 'slightly more' close to a nice integral value. it should work fine with either

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -259,25 +310,59 @@ def start(self):
...
+
+        if self.energy_dependent_gh:
+            self.gh_cuts_gamma = self.cuts.energy_dependent_gh_cuts(
+                gammas, reco_energy_bins, min_value=0.1, max_value=0.95
+            )
+            gammas = self.cuts.apply_energy_dependent_gh_cuts(
+                gammas, self.gh_cuts_gamma
+            )
+            self.log.info(
+                f"Using gamma efficiency of {self.cuts.gh_efficiency}"
```
moralejo, line 330: % missing in message

maxnoe: Not if we adapt our comment of not changing the value.

moralejo (changes requested):
Looks good to me, I just suggested  a few minor things.

lstchain/io/tests/test_eventselection.py
maxnoe: That should also be 31 bin edges to get 30 bins in energy. And if you were aiming for "five bins per decade", that is only 4 decades so you'd need only 20 bins.

Better use `np.geomspace` to get human readable values here:

`np.geomspace(5 * u.GeV, 50 * u.TeV, 31)`

lstchain/io/event_selection.py
chaimain: Ok.

lstchain/io/tests/test_eventselection.py
moralejo: > Nothing particular, just wanted the values to be 'slightly more' close to a nice integral value. it should work fine with either

I always fear such numbers that in any plot will "look like" exactly -2.3 and 1.7,  but are not. Then, you will anyway not have integer energy numbers (in GeV - is that what you meant?) for all bin edges, given the log scale... Also, if I see such non-round numbers in the code I immediately think there is a reason behind them.  So it is somewhat confusing.

lstchain/io/event_selection.py
maxnoe: No, you won't need to add 1, since filter is using `lower <= value <= upper`, so having the filter just be `[EventType.SUBARRAY.value, EventType.SUBARRAY.value]` is fine:

https://github.com/cta-observatory/cta-lstchain/blob/74e35aebfe6a6f94ba2083e5cd342355668f9565/lstchain/reco/utils.py#L529

lstchain/io/event_selection.py
chaimain: I change the value to be provided in the %, [0, 100], and no need to multiply/change the value in the class.

docs/examples/dl3_tool_config.json
```diff
@@ -12,9 +12,9 @@
   },
   "DL3Cuts": {
     "global_gh_cut": 0.7,
-    "gh_efficiency": 0.9,
+    "gh_efficiency": 90,
```
maxnoe, line 15: That's not what I meant. I think having it in `[0, 1]` in the config file makes sense!

Just passe `percentile=100 * (1 - self.gh_efficiency)` when calling pyirf

chaimain: Ok. I will do the same for `theta_containment`

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -159,16 +160,43 @@ def setup(self):
 
         self.log.debug(f"Output DL3 file: {self.output_file}")
 
+        try:
+            with fits.open(self.input_irf) as hdul:
+                self.use_energy_dependent_cuts = (
+                    "GH_CUT" not in hdul["EFFECTIVE AREA"].header
+                )
+        except KeyError:
```
maxnoe, line 168: This KeyError won't be raised here now, but probably only later when applying the cuts

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -159,16 +160,43 @@ def setup(self):
...
-        ## To reduce the table columns further, add a selection of columns to be read and used.
         self.effective_time, self.elapsed_time = get_effective_time(self.data)
         self.run_number = run_info_from_filename(self.input_dl2)[1]
 
         self.data = self.event_sel.filter_cut(self.data)
-        self.data = self.fixed_cuts.gh_cut(self.data)
-        self.data = add_icrs_position_params(self.data, self.source_pos)
+
+        if self.use_energy_dependent_cuts:
+            self.energy_dependent_gh_cuts = QTable.read(
```
maxnoe, line 184: Here.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -159,16 +160,43 @@ def setup(self):
...
+            self.data = self.cuts.apply_energy_dependent_gh_cuts(
+                self.data, self.energy_dependent_gh_cuts
+            )
+            self.data = add_icrs_position_params(self.data, self.source_pos)
+            self.log.info(
+                "Using gamma efficiency of "
+                f'{self.energy_dependent_gh_cuts.meta["GH_EFF"]}'
+            )
+        else:
+            self.data = self.cuts.apply_global_gh_cut(self.data)
```
maxnoe, line 197: Does this use the `GH_CUT` value in the IRF header or does this use whatever is in the default config? 

It should use the `GH_CUT` value from the IRF.

lstchain/tools/tests/test_tools.py
```diff
@@ -84,6 +84,66 @@ def test_create_irf_full_enclosure_with_config(temp_dir_observed_files, simulate
...
+                f"--input-irf={irf_file}",
+                "--source-name=Crab",
+                "--source-ra=83.633deg",
+                "--source-dec=22.01deg",
+                "--overwrite",
+            ],
+            cwd=temp_dir_observed_files,
+        )
+        == 0
+    )
```
maxnoe, line 144: Please at some tests for the produced output. E.g. that the file can be read using `Observation.read` from gammapy and that all expected data are there.

maxnoe (changes requested):
Three smallish things left, we are getting there!

lstchain/tools/lstchain_create_dl3_file.py
chaimain: Ah yes! Thanks for spotting this

lstchain/tools/lstchain_create_dl3_file.py
chaimain: Is it better to not explicitly mention any Error in this case?

chaimain: Ok. Thanks for clarifying

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -194,6 +194,9 @@ def start(self):
                 f'{self.energy_dependent_gh_cuts.meta["GH_EFF"]}'
             )
         else:
+            self.cuts.global_gh_cut = QTable.read(
```
maxnoe, line 197: Again, there is no need to read the full table to access a single header object.

chaimain: Ah right! I will fix it.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -165,7 +165,7 @@ def setup(self):
                 self.use_energy_dependent_cuts = (
                     "GH_CUT" not in hdul["EFFECTIVE AREA"].header
                 )
-        except KeyError:
+        except:
             raise ToolConfigurationError(
```
maxnoe, line 169: The error message here is misleading. As said, the code above will only raise an error if:

* The input irf file does not exist (which I think is checked before)
* The irf file does not have an `EFFECTIVE AREA` HDU
Nothing to do with a `GH CUTS` table

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -90,9 +106,19 @@ class IRFFITSWriter(Tool):
...
+
+    Or use energy-dependent cuts based on a gamma efficiency:
+    > lstchain_create_irf_files
+        -g /path/to/DL2_MC_gamma_file.h5
+        -o /path/to/irf.fits.gz
+        --point-like (Only for point_like IRFs)
+        --energy-dependent-gh
+        --energy-dependent-theta
+        --gh-efficiency 95
+        --theta-containment 68
```
SeiyaNozaki, line 121: these parameters should be 0.95 and 0.68 now?

chaimain: True! Lemme check elsewhere as well

maxnoe (approved):

moralejo (approved):

PR #710, opened 2021-05-06
WIP: changes to use disp norm and disp sign for the direction reconstruction

vuillaut:
This PR is not meant to be merged.
At the moment, this branch is intended to make tests.
(I am opening the PR to let you know if you want to use it for yourself).

On a simple test, this is technically working:
(sorry for the black labels, this is `reco_alt` vs `mc_alt`)

alt_mig_matrix


vuillaut:
disp_norm_importances


disp_sign_importances

maxnoe:
Interesting that length plops up so prominently for the sign. Didn't for my test using aict-tools:

disp_performance-5

lstchain/reco/dl1_to_dl2.py
```diff
@@ -153,15 +151,15 @@ def train_disp_sign(train, custom_config={}, predict_feature='disp_sign'):
     """
 
     config = replace_config(standard_config, custom_config)
-    classification_args = config['random_forest_classifier_args']
-    features = config["classification_features"]
-    model = RandomForestClassifier
+    regression_args = config['random_forest_regressor_args']
+    features = config['regression_features']
+    model = RandomForestRegressor
```
maxnoe, line 156: Why are you using a regressor for the sign?

maxnoe: I think the code was correct before

vuillaut: haha, because I didn't think

vuillaut: will change that, thanks

moralejo:
> Interesting that length plops up so prominently for the sign. Didn't for my test using aict-tools:
> 
Strange indeed. Can't see how using a regressor instead of classifier can produce that (if that is the reason). Both trained on diffuse gammas, I assume.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -153,15 +151,15 @@ def train_disp_sign(train, custom_config={}, predict_feature='disp_sign'):
     """
 
     config = replace_config(standard_config, custom_config)
-    classification_args = config['random_forest_classifier_args']
-    features = config["classification_features"]
+    regression_args = config['random_forest_regressor_args']
```
maxnoe, line 154: I think this was all correct before, right? Why rename the configuraition to regressor?

vuillaut: no, the classification features in the config are meant for the gamma/hadron classification (include the reco energy and reco position)

vuillaut: but, you are right for the args, that must be for a classifier!

vuillaut:
trained with the classifier:

disp_sign_importances

the `length` appears less important.
but  don't draw too many conclusions, it's a quick test with ~1000 events

SeiyaNozaki:
Hi @moralejo , ok, I will make a separate PR with an option to switch direction methods, and set disp_norm as default.

Issue #696, opened 2021-04-15
To compare Zen pointing of an event list and IRF before combining them to make a DL3 file


chaimain:
Currently, we are using IRFs made from Zen 20deg MC only, but in the near future, we would have 40deg as well as [interpolated IRFs](https://github.com/cta-observatory/pyirf/pull/141) so that we can combine a more appropriate IRF with a given event list (observed data).
I propose we add a check on the event list, to check the Zenith pointing information and compare the same from IRF (this metadata is not currently there, but can be easily added), and once their difference satisfies some safe conditions, we append them together to create the final DL3 file.

This can be done by either 
1. creating a separate script (or Tool) or,
2. updating the current DL3 Tool.

As discussed in the lst-onsite Slack channel, we currently can have at least a **standard** event list file and then add such script or code, as a check before combining with a given IRF file, to make it a DL3 file (not standard yet).

Adding to the current DL3 Tool can have a small issue - recreating an existing event list and overwriting the file.
This process probably should be left for the 'final' version of the DL3 Tool, once we have appropriate IRFs ready.

Please provide any enhancement to this plan of action or suggestions/feedback.

maxnoe:
I think the approach should be that the create_dl3_file gets all availabe irfs and does the interpolation to the conditions of the run.

PR #711, opened 2021-05-07
Update DL3 Tool to to use pyirf IRF interpolation functions

chaimain:
With the upcoming pyirf [PR#150](https://github.com/cta-observatory/pyirf/pull/150) and the merged [PR#141](https://github.com/cta-observatory/pyirf/pull/141), the DL3 Tool needs to be updated according to Max's suggestion in [#696](https://github.com/cta-observatory/cta-lstchain/issues/696).

As soon as the IO functions in pyirf are merged, I will update the new function `interpolate_irf` in `lstchain/irf/hdu_table.py` and with the new MC DL2 productions of different zeniths (same RF model) are ready, the functions can be tested and then I can open the PR.


chaimain:
I have separated the interpolation functions in `lstchain/irf/interpolate.py`. These functions are:
1. compare_irfs - Compare a given list of IRF files with their metadata, binning and cuts (fixed cuts checked only by checking the assigned metadata). I have to generalize this for energy-dependent cuts as well.
2. load_irf_grid - Compiles a list of column data from the given list of IRFs which are to be used for interpolation
3. interpolate_irf - Computes the interpolation from a given list of IRFs and a dict of target values for interpolation along a particular parameter (zenith, azimuth pointing, etc.)

Currently, only effective area and energy dispersion IRFs are interpolated and written into a final IRF.

The IRF Tool is slightly modified to include necessary metadata values for the interpolation.

The DL3 Tool has the following changes:
1. Instead of providing an input IRF file path, the path where the IRFs are stored (`input_irf_path`) is to be provided, a glob search pattern (`irf_file_pattern`) to look up for either a single IRF or multiple for interpolation and the final filename of the interpolated IRF (`final_irf_file`). This puts the interpolation function as an optional feature.
2. If the interpolation is to be done only on the zenith pointing, a flag `interp_only_zd` is to be added. By default, I have put zenith and azimuth as the interpolation parameters.
3. There are a few checks on the number of IRFs provided. For interpolation, it has to be at least twice the number of interpolation parameters. The given list of IRFs are also compared first to check if the select metadata, event binning and selection are compatible to be used for interpolation.
4. The target values for interpolation is provided from `create_event_list` function and then passed into the `interpolate_irf` function.

For testing these functions, I have temporarily created 2 different IRFs, one with different g/h cut and another with an extrapolated values to check interpolation function. 

This PR still needs further cleaning and more improvements.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,536 @@
...
+
+    Returns the neccessary values with applied functions as need be for the
+    interpolation, for each parameter as a list.
+    """
+    mc_pars = []
+    if "ZEN_PNT" in params_list:
+        if isinstance(data["ZEN_PNT"], u.Quantity):
+            unit = data["ZEN_PNT"].unit
+        else:
+            unit = data.comments["ZEN_PNT"]
```
maxnoe, line 39: From the FITS standard:

> . Nonetheless, software should not depend on units being
expressed in this fashion within a keyword comment, and soft-
ware should not depend on any string within square brackets in
a comment field containing a proper units string.

So we should not parse units from the comments. We should write a well-defined unit and attach that to the value when reading.

maxnoe: The issue you linked has nothing to do with units in comments.

> or use string representation of the quantities?

No. We should store quantities like this:

```
unit = u.s
hdu.header['FOO'] = foo.to_value(unit), f'[{unit:fits}]'
```

And parse like this:
```
foo = u.Quantity(hdu.header['FOO'], unit=u.s
```

So writing the unit as comment for humans looking at the header but using the fixed unit for writing and reading.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -83,53 +94,99 @@ class DataReductionFITSWriter(Tool):
...
+        --overwrite
+
+    Or use a list of IRFs for including only the nearest IRF:
+    > lstchain_create_dl3_file
+        -d /path/to/DL2_data_file.h5
+        -o /path/to/DL3/file/
+        -i /path/to/irf/
+        -p irf*.fits.gz
+        -f final_interp_irf.fits.gz
+        --use-nearest-irf-node True
```
gabemery, line 135: This example doesn't work. The True is not expected since use-nearest-irf-node is a flag.

chaimain: True. Will fix it.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -40,11 +46,16 @@
...
+from lstchain.high_level.hdu_table import (
     add_icrs_position_params,
     create_event_list,
     set_expected_pos_to_reco_altaz,
 )
+from lstchain.high_level.interpolate import (
+    check_in_delaunay_triangle,
+    compare_irfs,
+    interpolate_irf,
+)
```
gabemery, line 58: Combine in from lstchain.high_level import (

gabemery (commented):
First quick pass over the code on my side. I will look more into the complex code later.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -2,13 +2,18 @@
 Create DL3 FITS file from given data DL2 file,
 selection cuts and IRF FITS files.
 
+For an interpolated IRF, based on the data provided by the event list,
```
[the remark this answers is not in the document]
chaimain: I tried a better explanation, and checked for some more comments and updated docstrings in some other places as well.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,565 @@
...
+                fov_offset_bins=fov_off,
+                extname="PSF",
+                **extra_headers
+            )
+
+            irf_interp.append(psf_hdu_interp)
+        except KeyError:
+            log.error("PSF HDU not present for IRF interpolation")
+
+    return irf_interp
```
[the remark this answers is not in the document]
chaimain: If you can point out the sections that I missed to cover in tests, that would be much appreciated!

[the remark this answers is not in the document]
gabemery: If you look at the file in the files changed section you can see warnings. There are more than ten so I didn't comment every time.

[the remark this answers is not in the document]
chaimain: Ah! I understand now.
I think, it comes down to the testing of _All_ kinds of IRFs that we produce and for which this PR affects... and given the order of tests, I can just use 2 IRFs produced by conftest (I have not used the src-dep one yet), which have global theta/gammaness cuts. I will see how much the coverage can be expanded by using some more types of IRFs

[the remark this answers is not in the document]
chaimain: I have updated the tests for covering most (not all) of the IRF types.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+        mc_params_full[i, :] = np.array(mc_pars_full)
+
+    data_val_sel = interp_params(data_pars_sel, data_params)
+    data_val_full = interp_params(data_params, data_params)
+
+    try:
+        tri = Delaunay(mc_params_sel)
+    except ValueError:
+        print("Not enough grid values for Delaunay triangulation")
+        return new_irfs
```
gabemery, line 106: Does this mean that your code prevents any DL3 production when not providing at least 3 IRF files?
If so I would say you need to change this. At least if use_nearest_irf_node is True it should proceed with this method (trivial if one IRF, same as your current implementation if 2).

Also looking in lstchain/tools/lstchain_create_dl3_file.py line 331 it seems like it fills you list with [] if you use a single IRF file. While the following code assume it contains a single path.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+
+
+def interp_params(params_list, data):
+    """
+    From a given list of angular parameters, to be used for interpolation,
+    take values from a given data table/dict.
+
+    Returns the neccessary values with applied functions as need be for the
+    interpolation, for each parameter as a list.
+    """
```
gabemery, line 34:
```
Parameters
----------------

Return
----------
```

Format like in other functions?

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+
+    return new_irfs
+
+
+def get_nearest_az_node(irf_params_sel, index, irf_params_full, target_params_full):
+    """
+    Check to see if a given IRF node overlaps with another node, in the
+    interpolation parameter space, and to select based on the azimuth (hard-
+    coded to be index 2 for now) angle, the nearest node to the target.
+    """
```
gabemery, line 144:
```
Parameters
----------------

Return
----------
```

Format like in other functions?

gabemery: Also can't you avoid the hard coding?

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+    except KeyError:
+        print(f"Effective Area is not present in {irfs[0]}")
+
+    # Comparing the metadata information
+    for k in select_meta:
+        meta = [Table.read(i, hdu="ENERGY DISPERSION").meta[k] for i in irfs]
+        m = np.unique(meta)
+
+        if len(m) != 1:
+            meta_bool *= False
```
gabemery, line 216: Why *= ?

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+    Parameters
+    ----------
+    irfs: List of IRFs to compare
+        'List'
+    Returns
+    -------
+    : Boolean indicating whether the given list of IRFs are comparable
+    """
+    bin_bool = False
+    meta_bool = True
```
gabemery, line 178: Is there a particular reason you default one to True and the other to False? 
I would do both as False if not.
(and thus change != to == l215)

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+        a2, a_ind = np.unique(a, return_index=True)
+
+        if len(a2) == len(params[i]):
+            if (a2[np.argsort(a_ind)] == params[i].value).all():
+                bin_bool = True
+    print(
+        "The other parameter axes data " f"are{' ' if bin_bool else ' not '}comparable"
+    )
+
+    return bin_bool * meta_bool
```
gabemery, line 237: While this likely works, I am not sure the multiplication of booleans is common. Someone more python-wise could comment on this.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+
+
+def load_irf_grid(irfs, extname, interp_col, gadf_irf=True):
+    """
+    From a given list of IRFs, load the list of IRF data values that can be
+    interpolated. For GH_CUTS which is not GADF approved, the HDU is stored
```
gabemery, line 243: What is the "custom" storage mentioned here? I assume it is a specific custom storage, not any custom storage.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+    data, to interpolate over in a selected interpolation parameter space.
+
+    Currently for the single telescope, we are using only cos zenith and
+    sin delta as the interpolation parameters for the IRFs, and not including
+    the contributions from azimuth direction (of the shower) as that becomes
+    significant only when considering a stereo system of telescopes. Hence,
+    we have to we perform the selection criteria only for the selected
+    interpolation parameters.
+
+    Along with IRFs, energy-depedent theta and gammaness cuts, RAD_MAX and
```
gabemery, line 353: energy-depe**n**dent

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+def interpolate_irf(irfs, data_pars, interp_method="linear"):
+    """
+    Using pyirf functions with a list of IRFs and parameters to compare with
+    data, to interpolate over in a selected interpolation parameter space.
+
+    Currently for the single telescope, we are using only cos zenith and
+    sin delta as the interpolation parameters for the IRFs, and not including
+    the contributions from azimuth direction (of the shower) as that becomes
+    significant only when considering a stereo system of telescopes. Hence,
+    we have to we perform the selection criteria only for the selected
```
gabemery, line 350: we have to ~we~ perform

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+        'astropy.io.fits'
+    """
+
+    # Gather the parameters to use for interpolation
+    # Exclude AZ_PNT as target interpolation parameter
+    d = data_pars.copy()
+    d.pop("AZ_PNT", None)
+
+    # For the interpolation parameters only
+    params_sel = [*d.keys()]
```
gabemery, line 380: Same as above, the copy feels useless. Still not sure of a better alternative

gabemery (changes requested):
Reviewed interpolate.py

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+
+
+def check_in_delaunay_triangle(irfs, data_params, use_nearest_irf_node=False):
+    """
+    From a given list of IRFs as grid points used for interpolation, retrieve
+    the Delaunay triangulation list of IRFs, where the simplex includes the
+    target points in data_params.
+
+    If the target point does not exist inside the simplex, the IRF
+    corresponding to the nearest grid point to the target value. This is also
```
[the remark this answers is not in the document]
chaimain: The sentence continues in the next line. There is a typo in the first word. Will fix it.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+    Returns
+    -------
+    irf_list: Revised list of IRFs after the checks.
+        'List'
+    """
+    # Exclude AZ_PNT as target interpolation parameter
+    d = data_params.copy()
+    d.pop("AZ_PNT", None)
+    # For the interpolation parameters only
+    data_pars_sel = [*d.keys()]
```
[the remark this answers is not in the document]
chaimain: I have 2 separate `dicts` because for the IRF interpolation of a single telescope, we only need 2 parameters, `ZEN_PNT` and `B_DELTA` , but we also take into account `AZ_PNT` when distinguishing between overlapping IRF nodes in the interpolation parameter space. To use `get_nearest_az_node`, I need to have the information from these 2 dicts to make the comparison. But, for the main interpolation process, we do not need `AZ_PNT` and hence the main object used for the interpolation process, is without `AZ_PNT` based information.

At the time, I could not find a better or simpler way of doing so.

lstchain/high_level/interpolate.py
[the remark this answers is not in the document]
gabemery: No I was saying the sentence on this line is missing something :
If the target point does not exist inside the simplex, the IRF
    corresponding to the nearest grid point to the target value

lstchain/high_level/interpolate.py
chaimain: In the DL3 Tool, https://github.com/cta-observatory/cta-lstchain/blob/78223b7541e7c20d55c145242927487eb17ed823/lstchain/tools/lstchain_create_dl3_file.py#L287-L311 I already take into account of the user passing only a single IRF, or if the glob search in the IRF directory path results in only a single IRF. In that case, no interpolation process is initiated.

The exception here is based on the minimum requirement on the number of nodes (n+1) to create a convex hull based on the interpolation parameters (n). In the stereo IRF interpolation, this will require at least 4 nodes to create an interpolation grids in 3-parameters space.

lstchain/high_level/interpolate.py
```diff
@@ -0,0 +1,577 @@
...
+    """
+    From a given list of angular parameters, to be used for interpolation,
+    take values from a given data table/dict.
+
+    Returns the neccessary values with applied functions as need be for the
+    interpolation, for each parameter as a list.
+    """
+    mc_pars = []
+    if "ZEN_PNT" in params_list:
+        mc_pars.append(np.cos(u.Quantity(data["ZEN_PNT"], "deg").to_value(u.rad)))
```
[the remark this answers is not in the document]
gabemery: Then the unit should be read instead of hardcoding "deg"

lstchain/high_level/interpolate.py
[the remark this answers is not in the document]
gabemery: From what I understood you don't use dictionaries after. The 'd' dictionary is only used to get the keys in this block. 

This is the one I am talking about not the ones obtained with interp_params later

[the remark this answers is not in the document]
chaimain: Ah ok! I wanted to make some easy steps to remove the `AZ_PNT` information from the `dict` to use it further, but not "only include" `ZEN_PNT` and `B_DELTA` in a general way. The latter seemed more hard-coded that the former... I will re-check for a simpler solution.

lstchain/high_level/interpolate.py
gabemery: Ok so only my issue with the number of nodes being between 2 and n remains. If you use nearest node, even if Delaunay fails you should be able to create DL3

chaimain: Ok, I understand your point. I will fix to cover that option.

lstchain/high_level/interpolate.py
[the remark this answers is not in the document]
gabemery: You are right, the current implementation is better than hard codding. And it is also good for what you do. But if you find a version without intermediate copies I would find it a bit more clear when reading the code (e.g. I had to search the code to check if d was used anywhere else)

Best answer here https://stackoverflow.com/questions/4915920/how-to-delete-an-item-in-a-list-if-it-exists 
gives informations in case we are working with a list but it is probably not adapted here.

lstchain/high_level/interpolate.py
gabemery: Good. Honestly it will probably never happen. But I think you can handle it without making the code more complex

lstchain/high_level/interpolate.py
chaimain: 'Custom' here is just used to define the way of storing (data and metadata) the IRF HDUs, not defined/suggested by GADF. This is the case for `GH_CUTS` in this code.

lstchain/high_level/interpolate.py
chaimain: Here, I went with the idea of making the final boolean be `False` whenever the code sees different metadata values, and so, not do anything when the values match.
I agree, that this function may not be using the best practice of playing with python booleans, but at the time, I could not get a simpler approach.

gabemery: You are right i missed the loop in the logic. It is good then (also for the next comment)

lstchain/high_level/interpolate.py
gabemery: Now understood

lstchain/high_level/interpolate.py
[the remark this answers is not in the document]
chaimain: Thanks for the suggestion! I incorporated a simple solution.

lstchain/high_level/interpolate.py
chaimain: Done!

lstchain/high_level/interpolate.py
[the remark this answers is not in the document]
chaimain: Yes, this was part of an earlier discussion with Max, and I just fixed the angular values to be in degrees for IRF HDU metadata and the `data_params` read in `lstchain/io/io/read_data_dl2_to_QTable` for real data's target information, so as to have the values in good human-readable units and use radian units when performing any trigonometric computation.

Further generalization seems irrelevant to our use-case, but if anyone has a strong opinion on this, it can be done.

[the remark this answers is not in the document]
gabemery: Ok as is for me

lstchain/high_level/tests/test_interpolate.py
```diff
@@ -0,0 +1,359 @@
...
+    irfs_3 = [irf_file_3, irf_file_4]
+    irfs_4 = [irf_file_3, irf_file_3]
+    irfs_5 = [simulated_srcdep_irf_file, simulated_srcdep_irf_file]
+
+    assert compare_irfs(irfs_1) == 0
+    assert compare_irfs(irfs_2)
+    assert compare_irfs(irfs_3) == 0
+    assert compare_irfs(irfs_4)
+    assert compare_irfs(irfs_5)
+
```
gabemery, line 66: I think you lack the comparison of two different irf files that should be equivalent. Here you only get true values for comparison of two of the same file.

lstchain/high_level/tests/test_interpolate.py
```diff
@@ -0,0 +1,359 @@
...
+        az_2 = 2 * az_1
+        del_2 = 1.2 * del_1
+
+        factor_zd = (np.cos(zen_2)) / np.cos(zen_1)
+        factor_del = (np.sin(del_2)) / np.sin(del_1)
+
+        aeff_1["EFFAREA"][0] *= factor_zd
+        aeff_2["EFFAREA"][0] *= factor_zd * factor_del
+
+        aeff_1_meta["ZEN_PNT"] = (zen_2 * 180 / np.pi, "deg")
```
gabemery, line 127: why zen2 here? Typo?

gabemery: Also in similar code blocks below

lstchain/high_level/tests/test_interpolate.py
```diff
@@ -0,0 +1,359 @@
...
+    az_check = round(102.199 * np.pi / 180, 3)
+
+    assert mc_closest_az == az_check
+
+
+def test_interpolate_gh_cuts():
+    from lstchain.high_level.interpolate import interpolate_gh_cuts
+
+    # Similar function as interpolate_th_cuts, hence no need for extra test
+    # linear test case
```
gabemery, line 343: Not sure what is the use of duplicated code like this. You created 2 different functions which do exactly the same thing and only rename (and order?) the call to an external method.

Still if you do it like this you should cover both with test and check they always give the same output. When maintaining the code it will allow to avoid unintended behaviour by modifying one and not the other.

gabemery (commented):
I now when other the vast majority of the code (not always looking into the details).
Once current items are addressed I think the PR will be ready.

lstchain/high_level/tests/test_interpolate.py
chaimain: The idea is that for `irf_1`, only the zenith is changed (doubled), and for `irf_2`, both zenith and delta values are changed. The `factor_*` is for the IRF values that 'should' have appropriate values assuming that the IRF varies in a linear way in this parameter space

lstchain/high_level/tests/test_interpolate.py
chaimain: The `interpolate_rad_max` will be imported from pyirf, and when we update the pyirf dependency, we will not have an extra test for it. The current function is just a placeholder so that users can use the PR, without updating pyirf.

Adding an extra test is not an issue, but it just felt like a redundant effort...

I will fix the typo in the name of the function `interpolate_rad_max` here.

lstchain/high_level/tests/test_interpolate.py
gabemery: Ok. I will read exactly what is happening there later.

lstchain/high_level/tests/test_interpolate.py
gabemery: Ok no problem if it is temporary.

lstchain/high_level/tests/test_interpolate.py
gabemery: While this is not critical, your main use of the function is to check that two different original files are compatible (right?). So if you can implement an associated test with small efforts I would go for it.

lstchain/high_level/tests/test_interpolate.py
```diff
@@ -0,0 +1,382 @@
...
+        zen_1 = u.Quantity(aeff_1_meta["ZEN_PNT"], "deg").to_value(u.rad)
+        az_1 = u.Quantity(aeff_1_meta["AZ_PNT"], "deg").to_value(u.rad)
+        del_1 = u.Quantity(aeff_1_meta["B_DELTA"], "deg").to_value(u.rad)
+
+        zen_2 = 2 * zen_1
+        az_2 = 2 * az_1
+        del_2 = 1.2 * del_1
+
+        factor_zd = (np.cos(zen_2)) / np.cos(zen_1)
+        factor_del = (np.sin(del_2)) / np.sin(del_1)
```
gabemery, line 75: redundant parenthesis

lstchain/high_level/interpolate.py
chaimain: I think, we can save this issue to be resolved later when we try to improve the overall code quality.

gabemery (approved):
At this stage I think the PR is ready for merging.

PR #716, opened 2021-05-18
Interpolate charge and peak time of bad pixels using neighbors

SeiyaNozaki:
For now, all pixels regarded as bad using information of ped+cal run and interleaved ped+cal events during the observation have meaningless charge and peak time value (set zero to waveform).
This PR is to interpolate the charge and peak time of pixels labeled as bad using neighboring pixels, suggested by @tysaito2008. 
In this PR, neighboring pixels are searched and the average value using those neighbors is computed. Then, this value is set as parameters of bad pixels.
Please find the following slides about some quick test:
[20210518_Interpolate_bad_pixels.pdf](https://github.com/cta-observatory/cta-lstchain/files/6502558/20210518_Interpolate_bad_pixels.pdf)

interpolation

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -427,27 +427,28 @@ def r0_to_dl1(
                 if not is_simu:
                     selected_gain = event.r1.tel[telescope_id].selected_gain_channel
                     bad_pixels = event.mon.tel[telescope_id].calibration.unusable_pixels[selected_gain] 
+                    bad_pixels_selected_gain = ((selected_gain == 0) & (bad_pixels[0] == 1)) | ((selected_gain == 1) & (bad_pixels[1] == 1))
 
                     if np.sum(bad_pixels) > 0:
                         telescope = subarray.tel[telescope_id]
                         camera_geometry = telescope.camera.geometry
 
                         # find pixels neighboring bad pixels and label these clusters
-                        bad_pixels_neighbors = dilate(camera_geometry, bad_pixels)
-                        num_bad_pixel_islands, bad_pixel_island_labels = number_of_islands(camera_geometry, failing_pixels_neighbors)
+                        bad_pixels_neighbors = dilate(camera_geometry, bad_pixels_selected_gain)
+                        num_bad_pixel_islands, bad_pixel_island_labels = number_of_islands(camera_geometry, bad_pixels_neighbors)
                         
-                        image = calibrated_event.dl1.tel[telescope_id].image
-                        peak_time = calibrated_event.dl1.tel[telescope_id].peak_time
+                        image = event.dl1.tel[telescope_id].image
+                        peak_time = event.dl1.tel[telescope_id].peak_time
 
                         # compute the average value using neighboring pixels and set it to bad pixels' data
-                        for i_island in range(num_islands):
+                        for i_island in range(num_bad_pixel_islands):
 
                             # island labeled as zeros is other than clusters including bad pixels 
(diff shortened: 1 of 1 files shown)


maxnoe (changes requested):
Please move this functionality into a function doing this interpolation instead of adding more imperative lines to this function.

SeiyaNozaki:
@maxnoe thank you for the comment! I moved the functionality for the interpolation to the corresponding function.

lstchain/reco/r0_to_dl1.py
```diff
@@ -673,3 +650,41 @@ def rescale_dl1_charge(event, scaling_factor):
...
+
+    Parameters
+    ----------
+    
+    event: `ctapipe.containers.ArrayEventContainer`
+    subarray: `ctapipe.instrument.subarray.SubarrayDescription`
+    telescope_id: `int`
+    """
+    selected_gain = event.r1.tel[telescope_id].selected_gain_channel
+    bad_pixels = event.mon.tel[telescope_id].calibration.unusable_pixels[selected_gain]
```
maxnoe, line 667: This looks wrong. You seem to select the gain twice here.

SeiyaNozaki: yeah... you are right... 
Removed the duplication.

lstchain/reco/r0_to_dl1.py
```diff
@@ -644,3 +650,41 @@ def rescale_dl1_charge(event, scaling_factor):
...
+    Parameters
+    ----------
+    
+    event: `ctapipe.containers.ArrayEventContainer`
+    subarray: `ctapipe.instrument.subarray.SubarrayDescription`
+    telescope_id: `int`
+    """
+    selected_gain = event.r1.tel[telescope_id].selected_gain_channel
+    bad_pixels = event.mon.tel[telescope_id].calibration.unusable_pixels 
+    bad_pixels_selected_gain = ((selected_gain == 0) & (bad_pixels[0] == 1)) | ((selected_gain == 1) & (bad_pixels[1] == 1))
```
maxnoe, line 668: `bad_pixels_selected_gain = np.where(selected_gain==0, bad_pixels[0], bad_pixels[1])`

lstchain/reco/r0_to_dl1.py
```diff
@@ -644,3 +650,41 @@ def rescale_dl1_charge(event, scaling_factor):
...
+
+    if np.sum(bad_pixels) > 0:
```
maxnoe, line 670: `if bad_pixels.any():`

lstchain/reco/r0_to_dl1.py
```diff
@@ -644,3 +650,41 @@ def rescale_dl1_charge(event, scaling_factor):
...
+        telescope = subarray.tel[telescope_id]
+        camera_geometry = telescope.camera.geometry
+
+        # find pixels neighboring bad pixels and label these clusters
+        bad_pixels_neighbors = dilate(camera_geometry, bad_pixels_selected_gain)
```
maxnoe, line 675: Why use "dilate" here instead of the neighbor matrix?

maxnoe:
@SeiyaNozaki This is definitely something that should also go into ctapipe as a `Component`.

Could you do that?

SeiyaNozaki:
@maxnoe ok, I will also prepare for ctapipe components
anyway I will update this following your comments before long

SeiyaNozaki:
reopened #991 (dl1ab step) instead of this PR (r0_to_dl1 step)

PR #719, opened 2021-05-27
Introduced Poissonian fluctuations in the psf tuning

moralejo:
Introduced Poissonian fluctuations in the psf tuning, which up to now was just dividing a fixed fraction of charge per pixel equally among neighboring pixels.
This had the bad effect of reducing fluctuations, particularly visible in noise-dominated pixels.  

With old code:
image

One can see a deformation of the noise-dominated region of the MC (to which the psf tuning was applied).

Now the amount of p.e. to be smeared has Poissonian fluctuations, and the distribution among the neighbors follows a multinomial. The deformation is now gone with the proposed modification:

image

The new code is three times slower than the old one (6 to 18 ms per  image), and it is not because of the random number generator, but because I had to use a loop to fill a matrix (see comment in the code, suggestions are welcome).

### DIFF
--- a/lstchain/image/modifier.py
+++ b/lstchain/image/modifier.py
@@ -54,19 +54,22 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
     return image
 
 
-def smear_light_in_pixels(image, camera_geom, smeared_light_fraction):
+def smear_light_in_pixels(rng, image, camera_geom, smeared_light_fraction):
     """
 
     Parameters
     ----------
+    rng: numpy.random.default_rng  random number generator
+
     image: charges (p.e.) in the camera
 
     camera_geom: camera geometry
 
     smeared_light_fraction: fraction of the light in a pixel that will be
-    distributed equally among its immediate surroundings, i.e. immediate
-    neighboring pixels. Some light is lost for pixels which are at the
-    camera edge and hence don't have all possible neighbors
+    distributed among its immediate surroundings, i.e. immediate
+    neighboring pixels, according to Poisson statistics. Some light is
+    lost for pixels which are at the camera edge and hence don't have all
+    possible neighbors
 
     Returns
     -------
@@ -74,12 +77,28 @@ def smear_light_in_pixels(image, camera_geom, smeared_light_fraction):
 
     """
 
-    # Move a fraction of the light in each pixel (fraction) into its neighbors,
-    # to simulate a worse PSF:
-    q_smeared = (image * camera_geom.neighbor_matrix *
-                 smeared_light_fraction / N_PIXEL_NEIGHBORS)
+    # How many p.e. to smear?  Poisson of the smeared light fraction (clipped to a minimum of 0):
+    pe_to_smear = rng.poisson(np.clip(image * smeared_light_fraction, 0, np.inf))
+
+    # How to distribute the smeared charge among neighboring pixels (multinomial):
+    smeared_charges = np.zeros(shape=(len(image), N_PIXEL_NEIGHBORS))
+    for q in np.unique(pe_to_smear):
+        # generate the p.e. smearing patterns for all pixels with q  p.e.'s to be smeared:
+        num_pixels = np.sum(pe_to_smear == q)
+        # in this way we speed things up a lot through vectorization (vs. doing it pixel-wise):
+        smeared_charges[pe_to_smear == q, :] = rng.multinomial(q, N_PIXEL_NEIGHBORS * [1 / N_PIXEL_NEIGHBORS],
+                                                               size=num_pixels)
+
+    q_smeared = np.zeros([len(image), len(image)])
+
+    # The bulk of the execution time is spent here:
+    for pixid, charges in enumerate(smeared_charges):
+        q_smeared[pixid][camera_geom.neighbor_matrix[pixid]] = charges[:camera_geom.neighbor_matrix[pixid].sum()]
+    # Any idea on how to speed this up? (note the number of neighbors is not the same for all pixels!)
+
     # Light remaining in pixel:
-    q_remaining = image * (1 - smeared_light_fraction)
-    image = q_remaining + np.sum(q_smeared, axis=1)
+    q_remaining = image - pe_to_smear
+
+    image = q_remaining + np.sum(q_smeared, axis=0)
 
     return image
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -172,7 +172,7 @@ def main():
         if set(dl1_params_input).intersection(disp_params):
             parameters_to_update.extend(disp_params)
 
-        if increase_nsb:
+        if increase_nsb or increase_psf:
             rng = np.random.default_rng(
                     input.root.dl1.event.subarray.trigger.col('obs_id')[0])
 
@@ -196,7 +196,7 @@ def main():
                                                 transition_charge,
                                                 extra_noise_in_bright_pixels)
                 if increase_psf:
-                    image = smear_light_in_pixels(image,
+                    image = smear_light_in_pixels(rng, image,
                                                   camera_geom,
                                                   smeared_light_fraction)
 


lstchain/image/modifier.py
```diff
@@ -54,32 +54,51 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
...
-    # Move a fraction of the light in each pixel (fraction) into its neighbors,
-    # to simulate a worse PSF:
-    q_smeared = (image * camera_geom.neighbor_matrix *
-                 smeared_light_fraction / N_PIXEL_NEIGHBORS)
+    # How many p.e. to smear?  Poisson of the smeared light fraction (clipped to a minimum of 0):
+    pe_to_smear = rng.poisson(np.clip(image * smeared_light_fraction, 0, np.inf))
+
+    # How to distribute the smeared charge among neighboring pixels (multinomial):
+    smeared_charges = np.zeros(shape=(len(image), N_PIXEL_NEIGHBORS))
+    for q in np.unique(pe_to_smear):
```
maxnoe, line 85: This will introduce the same pattern in all pixels that have the same charge "pe_to_smear", correct? Isn't that dangerous?

lstchain/image/modifier.py
```diff
@@ -54,32 +54,51 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
...
+        # generate the p.e. smearing patterns for all pixels with q  p.e.'s to be smeared:
+        num_pixels = np.sum(pe_to_smear == q)
```
maxnoe, line 87: `np.count_nonzero`

lstchain/image/modifier.py
```diff
@@ -54,32 +54,51 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
...
+        # in this way we speed things up a lot through vectorization (vs. doing it pixel-wise):
+        smeared_charges[pe_to_smear == q, :] = rng.multinomial(q, N_PIXEL_NEIGHBORS * [1 / N_PIXEL_NEIGHBORS],
```
maxnoe, line 89: `np.full(N_PIXEL_NEIGHBORS, 1 / N_PIXEL_NEIGHBORS)` end this can be a global constant.

lstchain/image/modifier.py
moralejo: No, not the same pattern. Each one has its own Poisson parameter, and the generator produces according to it.

moralejo: > No, not the same pattern. Each one has its own Poisson parameter, and the generator produces according to it.

Unless I got it wrong (which might well be!)

maxnoe: You draw a poisson for each pixel, then only loop over the unique number of photons. That means that all pixels sharing the same `pe_to_smear` will have the exact same distribution of those photons into their neighbors, right?

I think you have to draw multinomial for each pixel like this:


```
neighbor_charges = np.multinomial(pe_to_smear, np.full(6, 1/6))
```

lstchain/image/modifier.py
```diff
@@ -54,32 +54,51 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
...
+                                                               size=num_pixels)
+
+    q_smeared = np.zeros([len(image), len(image)])
+
+    # The bulk of the execution time is spent here:
+    for pixid, charges in enumerate(smeared_charges):
+        q_smeared[pixid][camera_geom.neighbor_matrix[pixid]] = charges[:camera_geom.neighbor_matrix[pixid].sum()]
+    # Any idea on how to speed this up? (note the number of neighbors is not the same for all pixels!)
```
maxnoe, line 97: numba

lstchain/image/modifier.py
maxnoe: So, if e.g. two pixels have 10 p.e. to smear, they will smear the exact same amount of light into the same respective neighbors.

maxnoe:
Here is a numba version of the code, which takes ≈ 730µs per image on my computer and should implement the idea as intended by @moralejo 

```python
from numba import njit
import numpy as np
from ctapipe.instrument import CameraGeometry
import matplotlib.pyplot as plt

SMEAR_PROBALITITES = np.full(6, 1 / 6)


@njit(cache=True)
def random_psf_smearer(image, fraction, indices, indptr):
    new_image = np.zeros(len(image), dtype=image.dtype)

    for pixel in range(len(image)):

        if image[pixel] <= 0:
            continue

        to_smear = np.random.poisson(image[pixel] * fraction)

        # remove light from current pixel
        new_image[pixel] = (image[pixel] - to_smear)

        # add light to neighbor pixels
        neighbors = indices[indptr[pixel] : indptr[pixel + 1]]
        n_neighbors = len(neighbors)

        # print(n_neighbors)

        # all neighbors are equally likely to receive the charge
        # we always distribute the charge into 6 neighbors, so that charge
        # on the edges of the camera is lost
        neighbor_charges = np.random.multinomial(to_smear, SMEAR_PROBALITITES)

        for n in range(n_neighbors):
            neighbor = neighbors[n]
            new_image[neighbor] += neighbor_charges[n]

    return new_image



if __name__ == "__main__":
    cam = CameraGeometry.from_name("LSTCam")
    image = np.zeros(len(cam))
    image[0] = 1000
    smeared = random_psf_smearer(image, 0.5, cam.neighbor_matrix_sparse.indices, cam.neighbor_matrix_sparse.indptr)

    # we can assert exactly equal, since we have no charge in edge pixels
    assert image.sum() == smeared.sum()

    diff = smeared - image

    fig, (ax1, ax2) = plt.subplots(1, 2, constrained_layout=True)

    d1 = CameraDisplay(cam, image=image, ax=ax1, cmap='inferno')
    d1.add_colorbar()
    d2 = CameraDisplay(cam, image=smeared, ax=ax2, cmap='inferno')
    d2.add_colorbar()

    for ax  in (ax1, ax2):
        ax.set_xlim(-0.1, 0.1)
        ax.set_ylim(-0.1, 0.2)

    for d in [d1, d2]:
        d.set_limits_minmax(0, image.max())

    plt.show()
```

lstchain/image/modifier.py
moralejo: > So, if e.g. two pixels have 10 p.e. to smear, they will smear the exact same amount of light into the same respective neighbors.

Just for the record: no, I don't think so, because I use vectorization, one set of "smeared pe's"  is generated for each of the pixels having a given charge q.   Look at the size=num_pixels argument of the call to the multinomial.

maxnoe: Ha, right, missed that because of the strange linebreak ;)

maxnoe:
As I now basically contributed the code, someone else should review

lstchain/image/modifier.py
```diff
@@ -53,33 +56,59 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
...
     """
-
     Parameters
     ----------
     image: charges (p.e.) in the camera
 
-    camera_geom: camera geometry
+    indices: camera_geometry.neighbor_matrix_sparse.indices
+
+    indptr: camera_geometry.neighbor_matrix_sparse.indptr
```
rlopezcoto, line 72: what is `indptr`? can we include a more explicit name for it (maybe also change it in ctapipe)?

lstchain/image/modifier.py
```diff
@@ -53,33 +56,59 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
...
-                 smeared_light_fraction / N_PIXEL_NEIGHBORS)
-    # Light remaining in pixel:
-    q_remaining = image * (1 - smeared_light_fraction)
-    image = q_remaining + np.sum(q_smeared, axis=1)
+    new_image = image.copy()
 
-    return image
+    for pixel in range(len(image)):
+
+        if image[pixel] <= 0:
```
rlopezcoto, line 90: why not removing 0 as well?

lstchain/image/modifier.py
```diff
@@ -53,33 +56,59 @@ def add_noise_in_pixels(rng, image, extra_noise_in_dim_pixels,
...
+            continue
+
+        to_smear = np.random.poisson(image[pixel] * fraction)
+
+        if to_smear == 0:
```
rlopezcoto, line 95: how can this be exactly 0 if `image[pixel]` is not 0?

lstchain/image/tests/test_modifier.py
```diff
@@ -0,0 +1,52 @@
...
+    ]))
+
+
+    smeared = random_psf_smearer(image, fraction, neighbor_matrix.indices, neighbor_matrix.indptr)
+    # no charge lost in this case
+    assert image.sum() == smeared.sum()
+
+    # test charge is distributed into neighbors
+    assert np.isclose(smeared[0], (1 - fraction) * image[0], rtol=0.01)
+    assert np.allclose(smeared[1:], image[0] * fraction / 6, rtol=0.1)
```
rlopezcoto, line 37: this one here is working because you are testing it with an array with 7 elements, it kept me confused because I was thinking on `smeared[1:]` as an array of 1854 elements... maybe some test with a real geometry?

rlopezcoto (commented):
what I'm just wondering is that for pixels with small extracted charge that is just coming from random poissonian fluctuations and do not contain any light, you are also distributing this charge amongst the neighbors... This is certainly not correct, but probably not very harmful for these studies.

lstchain/image/modifier.py
moralejo: it is included. Skip for 0 or less...

lstchain/image/modifier.py
moralejo: It tells you where the indices of a given pixel's neighbors are in the array  neighbor_matrix_sparse.indices

maxnoe: `intptr` and `indices` are the two arrays of a `scipy.sparse.csr_matrix` that tell you which values are actually non-zero.

These names won't change if scipy doesn't break everyones code.

lstchain/image/modifier.py
maxnoe: because `to_smear` is drawn from a poissonian distribution

moralejo:
> what I'm just wondering is that for pixels with small extracted charge that is just coming from random poissonian fluctuations and do not contain any light, you are also distributing this charge amongst the neighbors... This is certainly not correct, but probably not very harmful for these studies.

The fluctuations are mainly from NSB, so they are actually p.e. which might also end up in a different pixel in case of a wider psf, so I think it is ok.

lstchain/image/tests/test_modifier.py
maxnoe: Why? This is a well-working unit test that allows as to test much more  strictly than if we would use a much more complicated geometry.

moralejo: @maxnoe : where is the test called, and with what "fraction"?  I can't find it. With the random version the tolerance seems a bit small to me.

maxnoe: @moralejo it is at the top of the test, the test is parameterized to run with fractions 0.2 and 0.5

maxnoe: The tolerance works fine for the high value of charge I injected here.

PR #720, opened 2021-07-02
Dynamic cleaning

pawel21:
I have add code to dynamic cleaning in `lstchain_dl1ab.py`
To run script with dynamic cleaning:
`python $script_path -f $input_file -o $output_file -c $config_file --pedestal-cleaning True --dynamic-cleaning True` 
What do think about my solution? (@moralejo , @rlopezcoto , @jsitarek )


pawel21:
I close due to failing test (my bug with branch)

PR #724, opened 2021-07-23
New calibration data tree and onsite scripts

FrancaCassol:
This PR includes the following points:

1. It introduces a pixel calibration data tree ("/fefs/data/rea/monitoring/PixelCalibration") which will contain all the present pixel calibration files (both for drs4 and PMTs) 
2. It simplifies the onsite scripts used to fill this tree, so as to easily permit an automatic call of the scripts online 
3. It adds two new scripts:  one to reconstruct the filter scan data, the other to fit them in order to obtain the systematic noise to be used in the calibration procedure
4. It includes the systematic noise in the F-factor calibration formula

A schematic description of the data tree and of the scripts is given in this note: [CameraCalibrationNote.pdf](https://github.com/cta-observatory/cta-lstchain/files/6869019/CameraCalibrationNote.pdf)


lstchain/data/filters_trasmission.dat
```diff
@@ -0,0 +1,37 @@
...
+53 0.00042
+54 0.00022
+55 0.00005
+56 0.00001
+61 0.00050
+62 0.00015
+63 0.00007
+64 0.00004
+65 0.00001
+66 0.00000
```
maxnoe, line 37: exactly 0? Or do we need more digits here?

FrancaCassol: For the camera is practically zero, but I can indeed add some digit

lstchain/tools/lstchain_fit_intensity_scan.py
```diff
@@ -240,6 +240,73 @@ def finish(self):
...
+
+                    # plot the fit results and residuals for four arbitrary  pixels
+                    fig = plt.figure((chan+1)*10, figsize=(11, 22))
+                    fig.suptitle(f"{channel[chan]} channel", fontsize=25)
+
+                    pad = 0
+                    for pix,i in enumerate([0,600,1200,1800]):
+                        pad+=1
+                        plt.subplot(4, 2, pad)
+                        plt.tight_layout()
```
maxnoe, line 267: Run tight layout once, just before saving the plot

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -25,6 +25,9 @@
                       type=int, nargs="+")
 
 version,subversion=lstchain.__version__.rsplit('.post',1)
+optional.add_argument('-f', '--filters_list', help="Filter list (same order than run list)",
```
maxnoe, line 28: same order as run list

FrancaCassol:
Hi,
I finalised this code with several improvements.

The code now works for creating the calibration data tree. 

For @morcuended:
You can find a tree example in 
/fefs/aswg/workspace/franca.cassol/data/real/monitoring/PixelCalibration 

For creating calibration files without systematic corrections (as it is now) one has to send the calibration script with the option 
--sys_date 0  

The output calibration file will have then the name "no_sys_corrected_calibration..." (I wanted to be sure not have confusion among the two cases). By default the script looks for the last available systematic corrections file before the run.

In order to test with this just copy my directory 
/fefs/aswg/workspace/franca.cassol/data/real/monitoring/PixelCalibration/ffactor_systematics
which contains correction files from three filter scans (for the moment I will take care for the production of these files)

Pay attention that in the case we apply the systematic corrections, the EventSource calibration scaling factors (calib_scale_high_gain, calib_scale_low_gain) must be changed to [1.088,1.004] in order to scale only for the integration window size.  I hope we can move soon to this way of proceeding

I updated the note [CameraCalibrationNote.pdf](https://github.com/cta-observatory/cta-lstchain/files/6945169/CameraCalibrationNote.pdf) and added more details on the scripts' logic.

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -0,0 +1,155 @@
...
+config_file = args.config
+sys_date = args.sys_date
+calib_dir=f"{base_dir}/monitoring/PixelCalibration"
+
+
+def main():
+
+    if shutil.which('srun') is None:
+        print(">>> This script needs a slurm batch system. Stop")
+        return
```
maxnoe, line 76: `sys.exit(1)` to exit with an error code.

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -0,0 +1,155 @@
...
+
+    print(f"\n--> Start reconstruct runs {run_list} and sub-runs {sub_run_list}")
+
+    # verify config file
+    if not os.path.exists(config_file):
+        raise IOError(f"Config file {config_file} does not exists. \n")
```
maxnoe, line 82: Raising an exception in the main method does not really make sense. Print error and `sys.exit(2)`

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -0,0 +1,155 @@
...
+sub_run_list = args.sub_run_list
+config_file = args.config
+sys_date = args.sys_date
+calib_dir=f"{base_dir}/monitoring/PixelCalibration"
+
+
+def main():
+
+    if shutil.which('srun') is None:
+        print(">>> This script needs a slurm batch system. Stop")
```
maxnoe, line 75: Use logging instead of prints

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -125,14 +126,15 @@ def main():
                     print(f"--> Create directory {log_dir}\n")
                     os.makedirs(log_dir, exist_ok=True)
 
-                # job file
-                job_file = f"{log_dir}/{run}_{sub_run}.job"
+                # job file              
+                now = datetime.now().strftime("%d-%m-%Y-%H:%M:%S")
```
maxnoe, line 130: Please use ISO format `%Y-%m-%dT%H:%M:%S` or just 

```
In [7]: datetime.now().isoformat(sep='T')
Out[7]: '2021-09-15T15:25:39.910310'

In [8]: datetime.now().replace(microsecond=0).isoformat(sep='T')
Out[8]: '2021-09-15T15:25:43'
```

moralejo:
Thanks a lot for these developments. I went through the changes & new scripts an to me this seems to be in  a good shape for merging. But probably it is better to wait for green light from @morcuended after testing it on-site. @morcuende, did you have a chance to test it? This should simplify the calibration part of lstosa.

morcuended:
on it

morcuended (changes requested):
Apart from the small comments, the thing that should be changed in my opinion is how the lstchain version is obtained. Especially when you are working with a fixed tag, the sub-version will not be available and the script will fail.

Side note, I've not been able to produce ffactor_systematics files with `lstchain/scripts/onsite/onsite_create_ffactor_systematics_file.py` but I guess it is a matter of setting properly the configuration file.

lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py
```diff
@@ -66,8 +66,10 @@
 source_config = {
     "LSTEventSource": {
         "max_events":args.max_events,
+        "apply_drs4_corrections": False
```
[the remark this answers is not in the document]
FrancaCassol: It is false because here we want calculate the corrections so they must not be applied.  
There is a warning because this trailets concerns a next version of ctapipe_io_lst, but since it does not damage I added it already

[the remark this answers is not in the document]
maxnoe: I just made the release of the new `ctapipe_io_lst` version, so you could update the requirement here

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -22,40 +29,58 @@
...
 required.add_argument('-r', '--run_number', help="Run number if the flat-field data",
                       type=int, required=True)
-required.add_argument('-p', '--pedestal_run', help="Run number of the drs4 pedestal run",
-                      type=int, required=True)
-optional.add_argument('-v', '--version', help="Version of the production",
-                      type=int, default=0)
+optional.add_argument('-p', '--pedestal_run', help="Pedestal run to be used. If None, it looks for the pedestal run of the date of the FF data.",
+                      type=none_or_str)
+
+version,subversion=lstchain.__version__.rsplit('.post',1)
```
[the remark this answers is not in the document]
maxnoe: Why remove the subversion at all?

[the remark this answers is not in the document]
morcuended: subversion is not used anywhere in the code, it is just to unpack directly the version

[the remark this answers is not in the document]
maxnoe: Yes, so the effective is that the subversion is not used for the path, it is removed. I Argumente tagt we don't want to do that.

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -0,0 +1,166 @@
...
+                # job file              
+                now = datetime.now().replace(microsecond=0).isoformat(sep='T')
+                job_file = f"{log_dir}/run_{run}_subrun_{sub_run}_date_{now}.job"
+
+                with open(job_file, "w") as fh:
+                    fh.write("#!/bin/bash\n")
+                    fh.write("#SBATCH --job-name=%s.job\n" % run)
+                    fh.write("#SBATCH --output=log/run_%s_subrun_%s_date_%s.out\n" % (run,sub_run,now))
+                    fh.write("#SBATCH --error=log/run_%s_subrun_%s_date_%s.err\n" % (run,sub_run,now))
+                    fh.write("#SBATCH -A dpps\n")
```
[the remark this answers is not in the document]
maxnoe: This should be an option for the script then probably

lstchain/scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
morcuended: Ah! Now I follow you. You propose to simply leave `version=lstchain.__version__` to make explicit that a non-tagged version is used. I agree with that.

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -69,61 +94,90 @@ def main():
         path, date = input_dir.rsplit('/', 1)
```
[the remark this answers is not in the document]
FrancaCassol: Hi @morcuended, 

sorry, I do not understand your comment, what do you mean?

lstchain/scripts/onsite/onsite_create_drs4_pedestal_file.py
```diff
@@ -21,8 +22,9 @@
 
 required.add_argument('-r', '--run_number', help="Run number with drs4 pedestals",
                       type=int, required=True)
+version,subversion=lstchain.__version__.rsplit('.post',1)
```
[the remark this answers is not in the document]
FrancaCassol: Ops, I just modified in a different way, let me know if it is fine

lstchain/scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
morcuended: Just a comment, nothing to be changed: the only way right now to know which date corresponds to which run number is by looking at the directory tree (what it is done here) or to look into the CaCo db.

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -22,40 +31,77 @@
...
 required.add_argument('-r', '--run_number', help="Run number if the flat-field data",
                       type=int, required=True)
-required.add_argument('-p', '--pedestal_run', help="Run number of the drs4 pedestal run",
-                      type=int, required=True)
-optional.add_argument('-v', '--version', help="Version of the production",
-                      type=int, default=0)
+optional.add_argument('-p', '--pedestal_run', help="Pedestal run to be used. If None, it looks for the pedestal run of the date of the FF data.",
+                      type=none_or_str)
+
+version=lstchain.__version__.rsplit('.post',1)[0]
```
morcuended, line 37: What Max was proposing is to keep the `.post...` part explicitly.

maxnoe: Yes, please just use `__version__` so that test / development tests cannot be confused with the normal productions.

FrancaCassol: Sorry, but this is exactly what I do not want, because it demands to recreate a new calibration tree at each commit. 
I think this solution is good both for the official tree  than for the debugging (head) code

morcuended: @FrancaCassol made me realize that it would be very difficult in the debugging phase because one should reproduce a new tree subdirectory at each commit.

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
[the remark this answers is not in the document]
FrancaCassol: Actually is this needed?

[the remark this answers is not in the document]
morcuended: I do not think so. It will just assume your group account if not specified. The line can be removed.

lstchain/scripts/onsite/onsite_create_calibration_file.py
maxnoe: Then I would propose to just remove the commit part, not the post part.

lstchain/scripts/onsite/onsite_create_drs4_time_file.py
```diff
@@ -0,0 +1,129 @@
...
+ped_run = args.pedestal_run
+prod_id = args.prod_version
+stat_events = args.statistics
+base_dir = args.base_dir
+sub_run = args.sub_run
+config_file = args.config
+if config_file is None:
+    config_file = os.path.join(os.path.dirname(__file__), "../../data/onsite_camera_calibration_param.json")
+
+max_events = 1000000
```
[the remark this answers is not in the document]
FrancaCassol: yes, it is not needed here, I remove it

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -69,100 +114,232 @@ def main():
...
         print(f"\n--> Output file {output_file}")
-        if os.path.exists(output_file) and ff_calibration == 'yes':
+
+        log_file = f"{output_dir}/log/{output_name}.log"
+        print(f"\n--> Log file {log_file}")
+
+        if os.path.exists(output_file):
             if query_yes_no(">>> Output file exists already. Do you want to remove it?"):
                 os.remove(output_file)
+                os.remove(log_file)
```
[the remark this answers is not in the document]
FrancaCassol: Actually in slurm it gives an immediate error, I wonder, is there a way to know if a script is running in slurm?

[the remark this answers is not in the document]
morcuended: You can use either `squeue` or `sacct` (with the `-j` if you know the job_id)

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -0,0 +1,166 @@
...
+                    cmd = f"srun onsite_create_calibration_file -r {run} " \
+                        f"-p {ped_run} -v {prod_id} --sub_run {sub_run} " \
+                        f"-b {base_dir} -s {stat_events} --output_base_name {output_base_name} " \
+                        f"--filters {filters} --sys_date {sys_date} " \
+                        f"--config {config_file} --time_run {time_run}"
+
+                    if no_sys_correction:
+                        cmd += " --no_sys_correction"
+
+                    fh.write(cmd)
```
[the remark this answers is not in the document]
FrancaCassol: Yes, it would be nice, do you perhaps know how to do it?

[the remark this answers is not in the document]
maxnoe: The problem is that you run a batch command and the submission was successful, the job itself failed.

To keep track of which jobs failed and why, you need a much more complex system, that e.g. stores submitted jobs in a database and periodically checks if the job succeeded or not.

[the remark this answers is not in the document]
FrancaCassol: yes, and with slurm is not so easy because the error files are always produced and some time with not meaningful errors

lstchain/scripts/onsite/onsite_create_calibration_file.py
FrancaCassol: what does the post number mean?

morcuended: I think it is the commit hash.

FrancaCassol: So I do not like to have it in the tree either

maxnoe: The part after `post` is the number of commits since the tag and then comes `git<hash>`.

I was proposing to keep the `post` part, not the part after that. So that any development version is e.g.
```
0.7.5.post
```

Note, that with the switch  to `setuptools_scm` in #737, this would be:

```
0.7.6.dev
```

FrancaCassol: I think it is an unuseful complication for debugging. It does not concerns the official tree (which will be alway with a release), and developer people do not need to add a "post" to the personal tree to remember that they are working with a development release. Also it could be I like to use part of the official tree but if my tree has a different version name I can not do it.

lstchain/scripts/onsite/onsite_create_calibration_file.py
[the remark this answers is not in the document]
maxnoe: The `srun` echos the job id, so you get it when starting the job. But this is again somethign that I would rather see in the onsite analysis than in lstchain, this goes rather deep into cluster management, not analysis.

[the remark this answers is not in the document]
FrancaCassol: I found an other way, hope it is fine ;-)

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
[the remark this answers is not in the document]
maxnoe: You should make sure that stdout and stderr are going into the same file and that exit codes are correctly propagated in the scripts. E.g. bash scripts should always have `set -euo pipefail` at the start.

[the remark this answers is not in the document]
FrancaCassol: why in the same file?

lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py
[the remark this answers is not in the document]
FrancaCassol: Are you sure we want to do it here?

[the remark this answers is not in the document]
maxnoe: If you set an option that is only valid for the new version, yes. If not, don't set the option and set it in the update PR

lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -235,16 +235,20 @@ def main():
         log_file = f"{output_dir}/log/{output_name}.log"
         print(f"\n--> Log file {log_file}")
 
+        yes = False
         if os.path.exists(output_file):
-            if query_yes_no(">>> Output file exists already. Do you want to remove it?"):
+            if os.getenv('SLURM_JOB_ID') is None:
```
morcuended, line 242: I think this variable can be only interpreted inside a script launched with sbatch. Here it will be always None.

maxnoe: indeed. This is the current johb id, not the one of a job launches using `sbatch` and friends

FrancaCassol: yes, and it will be not None when the script is running in batch (this is the idea)

maxnoe: But the check is to make sure that in SLURM, the script is not asking for user input.

However, there is a better way: adding the conventional `--yes` option that will prevent dialogs and answers them so that the script continues automatically.

FrancaCassol: ?

maxnoe: It is convention that many programs follow, to add a `-y, --yes` command line option that will prevent any dialogs.

e.g. like this:

```
parser.add_argument('-y', '--yes', help='Do not ask interactively for permissions, assume yes')
```

and then for example later:

```
overwrite = args.yes or query_yes_no("Overwrite?")
```

maxnoe: The sbatch script can then add the `--yes` option to prevent asking for any user input

FrancaCassol: I see, never met this option. But the present implementation has the advantage to automatically switch off the dialog in case of batch (without recalling to use the option),  what is the problem with it?

lstchain/scripts/lstchain_data_create_drs4_pedestal_file.py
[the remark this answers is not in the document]
FrancaCassol: Ok, I take it out

lstchain/scripts/onsite/onsite_create_calibration_file.py
morcuended: Both ways are fine. Please note that there are other (onsite) scripts with the same problem though.

FrancaCassol: right

FrancaCassol: I added also the yes option, it can indeed be useful in case one want to directly overwrite in the interactive case

FrancaCassol:
Dear @maxnoe and @morcuended, 
what is missing to finalize this PR? 
I need to make further developments for the automatic monitoring of the calibration data and I would like to do it after this PR is approved. Thanks!

lstchain/scripts/onsite/onsite_create_calibration_file.py
morcuended: I'm fine with the version line as it is

lstchain/tools/lstchain_fit_intensity_scan.py
```diff
@@ -0,0 +1,364 @@
...
+    ).tag(config=True)
+
+    fit_initialization = List(
+        [[100.0, 0.001], [6.0, 0.001]],
+        help='Fit parameters initalization [gain (ADC/pe), B term] for HG and LG'
+    ).tag(config=True)
+
+    fractional_variance_error = Float(
+        0.02,
+        help='Costant fractional error assumed for the y fit coordinate (variance)'
```
jsitarek, line 86: Costant ==> Constant
However why do you use the constant uncertainty? You can compute it easily from the variance itself and number of values.

lstchain/tools/lstchain_fit_intensity_scan.py
```diff
@@ -0,0 +1,364 @@
...
+    def finish(self):
+        """
+        write fit results in h5 file and the check-plots in pdf file
+        """
+
+        gain = np.ma.array(self.fit_parameters.T[0], mask=self.fit_error.T)
+        quadratic_term = np.ma.array(self.fit_parameters.T[1], mask=self.fit_error.T)
+
+        # give to the badly fitted pixel a median value for the B term
+        median_quadratic_term = np.ma.median(quadratic_term,axis=0)
```
jsitarek, line 235: I suggest here to add an extra check - if >~5-10% of the pixels have their fit wrong the whole script should end with an error.

lstchain/tools/lstchain_fit_intensity_scan.py
```diff
@@ -0,0 +1,364 @@
...
+                        plt.xlabel(f'Q-ped [ADC]')
+                        plt.hlines(0, 0, np.max(sig), linestyle='dashed', color="black")
+
+                    plt.tight_layout()
+                    pdf.savefig()
+
+
+
+def quadratic_fit(t, b=1, c=1):
+    f2 = 1.222
```
jsitarek, line 329: it is not nice to have this F2 value hardcoded here (I think it only affects the gain, while you want to compute first of all the quadratic term, but still, the gain itself is also put in the output, so the value should be consistent with what is used everywhere else)

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -152,21 +181,14 @@ def calculate_calibration_coefficients(self, event):
         calib_data.time_max = ff_data.sample_time_max
         calib_data.n_pe = n_pe
 
-        # find signal median of good pixels
+        # find signal median of good pixels (FF factor=<npe>/npe)
         masked_npe = np.ma.array(n_pe, mask=calib_data.unusable_pixels)
         npe_signal_median = np.ma.median(masked_npe, axis=1)
```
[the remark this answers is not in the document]
FrancaCassol: yes, I will add a comment

lstchain/tools/lstchain_fit_intensity_scan.py
FrancaCassol: I prefer to leave the possibility to add some systematic contribution. By the way, which formula would you use?

lstchain/tools/lstchain_fit_intensity_scan.py
FrancaCassol: I prefer to added a plot of the fitted B values over the camera, this will give a clear idea of the pixels that were badly  fit (there are also the text warning)

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -141,9 +158,21 @@ def calculate_calibration_coefficients(self, event):
         # Extract calibration coefficients with F-factor method
         # Assume fixed excess noise factor must be known from elsewhere
 
+        numerator = ff_data.charge_std ** 2 - ped_data.charge_std ** 2
+        denominator = self.squared_excess_noise_factor  * (ff_data.charge_median - ped_data.charge_median)
+        gain = np.divide(numerator, denominator, out=np.zeros_like(numerator), where=denominator != 0)
+
+        # correct for the quadratic term if possible
+        if self.quadratic_term is not None:
```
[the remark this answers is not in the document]
jsitarek: as you prefer. B=0 I only suggested because then you can directly use one formula (that would not work with "none"), if you prefer to make it in two steps it can also leave as it is now

lstchain/tools/lstchain_fit_intensity_scan.py
jsitarek: sigma_error = sigma / sqrt(2N - 2), where N is the number of measurements from which sigma is computed. 
see e.g.: https://math.stackexchange.com/questions/1259383/calculating-uncertainty-in-standard-deviation
I tested it with a simple toy MC and it works perfect well for a simple gaussian distribution and > 10 measurements. 0.02 that you are using would correspond to ~1250 events. I guess this is more or less in the correct ballpark for the statistics that you are using, but if you have large differences in statistics between different filter wheel positions the fit might be more stable/precise if you use proper uncertainties

lstchain/tools/lstchain_fit_intensity_scan.py
FrancaCassol: yes, my problem was how to pass a constant parameter to the fitted function. I found a solution

PR #725, opened 2021-07-28
Add tool to extract drs4 baseline and spike heights

maxnoe:
Note that this does not work yet, it needs a new version of `ctapipe_io_lst` (providing the get_spike_A_positions method)  

Advantages over the current tool:

1. Skip events with unknown `last_readout_time` so that the timelapse correction is properly applied for all samples
2. By default skip 10 samples in front and 1 at the end (current script 11 int front, none at the back)
3. Skips spikes for baseline estimation
4. Estimates the 3 spike heights
5. Provides count, mean and std per gain, pixel, capacitor for baseline, spike samples 1, 2, and 3
6. Output is a compressed hdf5 file for now, we can discuss the best format / structure inside

### DIFF
--- a/lstchain/scripts/onsite/onsite_create_calibration_file.py
+++ b/lstchain/scripts/onsite/onsite_create_calibration_file.py
@@ -155,7 +155,7 @@ def main():
         # search the pedestal file of the same date
         if ped_run is None:
             # else search the pedestal file of the same date
-            file_list = sorted(Path(f"{ped_dir}/{date}/{pro}/").rglob(f'drs4_pedestal*.0000.fits'))
+            file_list = sorted(Path(f"{ped_dir}/{date}/{pro}/").rglob(f'drs4_pedestal*.0000.h5'))
             if len(file_list) == 0:
                 raise IOError(f"No pedestal file found for date {date}\n")
             if len(file_list) > 1:
@@ -165,7 +165,7 @@ def main():
 
         # else, if given, search a specific pedestal run
         else:
-            file_list = sorted(Path(f"{ped_dir}").rglob(f'*/{pro}/drs4_pedestal.Run{ped_run:05d}.0000.fits'))
+            file_list = sorted(Path(f"{ped_dir}").rglob(f'*/{pro}/drs4_pedestal.Run{ped_run:05d}.0000.h5'))
             if len(file_list) == 0:
                 raise IOError(f"Pedestal file from run {ped_run} not found\n")
             else:
--- a/lstchain/scripts/onsite/onsite_create_drs4_pedestal_file.py
+++ b/lstchain/scripts/onsite/onsite_create_drs4_pedestal_file.py
@@ -86,7 +86,7 @@ def main():
             os.makedirs(log_dir, exist_ok=True)
 
         # define output file
-        output_file = f"{output_dir}/drs4_pedestal.Run{run:05d}.0000.fits"
+        output_file = f"{output_dir}/drs4_pedestal.Run{run:05d}.0000.h5"
 
         if os.path.exists(output_file):
             remove = False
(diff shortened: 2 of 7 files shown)


FrancaCassol:
Hi @maxnoe,  
this PR is fine for me. 

So the idea is to use the output file also to (offline) correct the spikes? 
Could you please recall on how many events do you need to run it and how long does it take?

maxnoe:
@FrancaCassol 20k Event's take 5:30 minutes on the IT Cluster, 53k takes 16 minutes. 

I don't know yet what is sufficient to calculate the spike heights. That will depend on whether we need to extract the values only for the whole camera or for each pixel or even for each capacitor. I am not sure about that.

Looking at first plots, I guess we should go for "by pixel". For that, 20k might be sufficient, but again, I did not do detailed studies.

maxnoe:
I also now ran this on a pedestal run without pre-calibration by EVB to see if this reproduces the right values (instead of the mostly noise offset of the differential calibration).

Looks good:

One pixel for the new script's output:
drs4_baseline

Scatter plot of all pixels, gains, caps compared to the current master script:
offset_comparison

jsitarek:
Hi @maxnoe 
it is somewhat worrying that we have some capacitors that reach the baseline of 10-20. Those will likely be clipped by no negative fluctuations. It would be good to understand if they happen in a single pixel, and how many in total are there. If there is a considerable number we might want to shift the baseline up a bit.

pawel21:
Hi Max,
Good idea with new tool.
> 2\. By default skip 10 samples in front and 1 at the end (current script 11 int front, none at the back)
But, inside the code current script skip 2 last sample (function '_fill_pedestal_event_jit' line 75 from calib/camera/drs4.py )

jsitarek:
Hi @maxnoe 

I went through your code, looks fine to me. The only comment is that it will not work correctly if skip_samples_front is set to 0 or 1, so you might want to add a check on this (either way the default is 10 and we do not want to run with anything less then 3)

maxnoe:
> skip_samples_front is set to 0 or 1

I am confused, why?

SeiyaNozaki:
> it is somewhat worrying that we have some capacitors that reach the baseline of 10-20.
This ticket is related to the very low baseline:
https://forge.in2p3.fr/issues/44184
https://forge.in2p3.fr/attachments/download/70243/20210325_PedestalTableCheck.pdf
I guess pixel 901 &1611 (LG) have relatively small offsets.

jsitarek:
> 
> 
> > not work correctly if skip_samples_front is set to 0 or 1
> 
> I am confused, why?

Hi @maxnoe 
If I understand correctly get_spike_A_positions returns the position only in range 0 to ROISIZE-1, right ? (I cannot find this function in your PR https://github.com/cta-observatory/ctapipe_io_lst/blob/gain_selected_input/ctapipe_io_lst/calibration.py)
if so, if you have a spike starting one slice before the ROI it will extend to the first 2 slices as well, but it will not show up in this function.

maxnoe:
Ah okay. I wouldn't really call this "not working", but you are right, I should remove the restriction in the spike position functoin and also allow this to return values in  [-2, 40] so all positions that affect the 40 samples can be returned.

pawel21 (dismissed):
Code looks very good for me

moralejo:
> Code looks very good for me

Good, can you approve, @pawel21 ?

maxnoe:
@moralejo He did, but I had to rebase master again. Let's wait until we really want to merge this with the approvals, otherwise we will have to reapprove every time this is updated. (And I see it's outdated again....)

pawel21 (approved):

maxnoe:
I thought we didn't want to merge this immediately???

moralejo:
Hi, I think we decided to get this into the 0.8.0 release, so it is ok, right?

maxnoe:
I was under the impression it was asked we perform more checks before merging, but fine with me.

PR #726, opened 2021-07-29
Add a display module to save images and video of events

juanjq:
A module (`event_displayer.py`) to filter events, and calibrate and save images and video of the events. There is also a notebook (`example_event_displayer.ipynb`) with an explanation and examples of how it works.


maxnoe:
Hi, 

Is this the code to create the videos send around by @moralejo? If yes, great job! 

Looking at the code, I think this needs a lot of cleanup though, to be useful for inclusion in lstchain.

To be included in the library for example means that we cannot just change the script to change input values. We need to use the configuration system for this, and for example give input / output files on the command line or as configuration.

Also, matplotlib can be more efficient and directly store videos, when you use the `FuncAnimation` class.



Here is a script I wrote a while ago for creating a video of  a single event using the ctapipe configuration system (a `Tool`):

from ctapipe.coordinates import TelescopeFrame
from ctapipe.io import EventSource
from ctapipe.calib import CameraCalibrator
from ctapipe.core import Tool
from ctapipe.core.traits import Float, Path, Int

import matplotlib.pyplot as plt
import astropy.units as u
import numpy as np
from matplotlib.animation import FuncAnimation
from tqdm import tqdm

```
class ShowerAnimation(Tool):
    tel_id = Int(default_value=1, help="Telescope to show").tag(config=True)
    output = Path(help="Output file").tag(config=True)
    fps = Float(default_value=10, help="FPS for the final video").tag(config=True)
    # samples_per_second = Float(default_value=10).tag(config=True)

    aliases = {
        ('i', 'input'): 'EventSource.input_url',
        ('o', 'output'): 'ShowerAnimation.output',
    }

    def setup(self):
        self.source = EventSource(
            parent=self,
            allowed_tels={self.tel_id, }
        )
        self.subarray = self.source.subarray
        self.cam = self.subarray.tel[self.tel_id].camera.geometry
        # self.calib = CameraCalibrator(self.subarray)


        self.event = next(iter(self.source))
        self.r1 = self.event.r1.tel[self.tel_id].waveform

        self.frames = self.r1.shape[1]

        # figsize / dpi will make FullHD video
        self.fig = plt.figure(figsize=(19.2, 10.8), dpi=100)
        self.ax = self.fig.add_axes([0, 0, 1, 1])
        self.ax.set_axis_off()

        self.disp = CameraDisplay(self.cam, cmap='inferno')
        self.disp.add_colorbar()
        self.progress = tqdm(total=self.frames)


    def init_ani(self):
        self.disp.image = self.r1[:, 0]
        self.disp.set_limits_minmax(self.r1.min(), self.r1.max())
        return self.disp.pixels,

    def update_ani(self, frame):
        self.progress.update(1)
        self.disp.image = self.r1[:, frame]
        return self.disp.pixels,

    def start(self):

        ani = FuncAnimation(
            fig=self.fig,
            func=self.update_ani,
            init_func=self.init_ani,
            frames=self.frames,
            interval=1000 / self.fps,
        )
        ani.save(self.output, fps=self.fps, dpi=100)
        # plt.show()

if __name__ == '__main__':
    ShowerAnimation().run()
```



You can find more about how the config system and the tools work here:
https://cta-observatory.github.io/ctapipe/api/ctapipe.core.Component.html#ctapipe.core.Component
https://cta-observatory.github.io/ctapipe/api/ctapipe.core.Tool.html#ctapipe.core.Tool
https://cta-observatory.github.io/ctapipe/examples/Tools.html

moralejo:
Thanks for the feedback @maxnoe. Certainly the code can be simplified and improved. 
About the input files, @juanjq , I thought it would be possible to change them in the notebook - indeed changing them in the .py file is not the right way.  But for using this from a notebook, I do not think we need a configuration file, etc.

About matplotlib's FuncAnimation, @maxnoe what do you mean? The class is being used already, right?

maxnoe:
> About matplotlib's FuncAnimation, @maxnoe what do you mean? The class is being used already, right?

Sorry, didn't see it. I ony saw the different `savefig` call and assumed the frames were stored as images.

juanjq:
Hi @moralejo , @maxnoe, thanks for the suggestions.
After talking about it this morning with @moralejo , I decided to change the input files for the data to the notebook. Now for using the code you only have to change this directories in the notebook, i supose that this could be done in a more general way, but it works.

Also if is there something at the code to be optimised always it could be changed, the parts at the code are separated and commented.

PR #728, opened 2021-09-10
Recover metadata in DL1 files

Bultako:
This PR recovers metadata added to DL1 files during the initial `r0_to_dl1` process, that were lost when applying `lstchain_dl1ab.py` script.

### DIFF
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -29,7 +29,7 @@ from lstchain.io import get_dataset_keys, auto_merge_h5files
 from lstchain.io.config import get_cleaning_parameters
 from lstchain.io.config import get_standard_config
 from lstchain.io.config import read_configuration_file, replace_config
-from lstchain.io.io import dl1_params_lstcam_key, dl1_images_lstcam_key
+from lstchain.io.io import dl1_params_lstcam_key, dl1_images_lstcam_key, read_metadata, write_metadata
 from lstchain.io.lstcontainers import DL1ParametersContainer
 from lstchain.reco.disp import disp
 from lstchain.image.modifier import smear_light_in_pixels, add_noise_in_pixels
@@ -164,6 +164,7 @@ def main():
         nodes_keys.remove(dl1_images_lstcam_key)
 
     auto_merge_h5files([args.input_file], args.output_file, nodes_keys=nodes_keys)
+    metadata = read_metadata(args.input_file)
 
     with tables.open_file(args.input_file, mode='r') as input:
         image_table = input.root[dl1_images_lstcam_key]
@@ -271,6 +272,8 @@ def main():
                     params[ii][p] = u.Quantity(dl1_container[p]).value
 
             output.root[dl1_params_lstcam_key][:] = params
+
+    write_metadata(metadata, args.output_file)
 
 
 if __name__ == '__main__':


lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -164,6 +164,7 @@ def main():
         nodes_keys.remove(dl1_images_lstcam_key)
 
     auto_merge_h5files([args.input_file], args.output_file, nodes_keys=nodes_keys)
```
vuillaut, line 166: This is actually performed by the `smart_merge_h5files` rather than the `auto_merge_h5files` - and as a bonus, you get a quick check on the metadata to see if we are merging consistent files.

I think however that the `smart_merge_h5files` is causing issues when merging a lot of files if 
`metadata0.SOURCE_FILENAMES` gets too long. Here:
https://github.com/cta-observatory/cta-lstchain/blob/5032edb033a3d01bc6db45b930ad62d2df3a302a/lstchain/io/io.py#L329

I can't find again the limit (imposed by hdf5 I think) but a quick test gives me a limit at 65500 characters for a single attribute.
Maybe you can implement that limit and use the smart merge?

vuillaut (changes requested):

lstchain/scripts/lstchain_dl1ab.py
vuillaut: I overlooked the fact that the `smart_merge` has been designed to merge MC files, while `dl1ab` is not limited to MC, so that won't work actually.

vuillaut: Maybe you can just add the `check_metadata` before merging to make sure the file created is consistent?
But the `SOURCE_FILENAMES` issue remains, you are keeping only one filename here.

Bultako: @vuillaut 
I think in this case we are not merging two existing files where we can check if their metadata are equal, we are creating a new hdf5 file from content stored in another one and we want to recover also the metadata. Actually `smart_merge_h5files` receives as input a list of more than one already existing files. 

Maybe we could refactor a bit all these functions, i.e. add a call to metadata copying and/or checking  in `auto_merge_h5files`? and remove `smart_merge_h5files` having a unique *smart* function `auto_merge_h5files` to copy and/or merge files of any kind? I think `merging_check` should already work with non MC files, since there are already `if subarray_info0.name == "MonteCarloArray":` code checking this.

vuillaut: Ha! My bad, you are absolutely right, we are using the merge to actually just create a new file.
In that case, I withdraw my comment, this is perfectly fine.
As for the refactoring, maybe that can go in another PR if you want to work on it, this one is perfectly fine as it is.

vuillaut (approved):

moralejo:
Hi @Bultako, I was just trying lstchain_dl1ab.py on real data and there seems to be a problem. The output DL1 file is produced, but then the program gives these messages:

Metadata SOURCE_FILENAMES does not exist in file ../orig/dl1_LST-1.Run02975.0080.h5
Metadata LSTCHAIN_VERSION does not exist in file ../orig/dl1_LST-1.Run02975.0080.h5
Metadata CTAPIPE_VERSION does not exist in file ../orig/dl1_LST-1.Run02975.0080.h5
Metadata CONTACT does not exist in file ../orig/dl1_LST-1.Run02975.0080.h5

and never ends... It keeps "running" (doing nothing, the output file is not growing) until one kills it. Upon killing it, the output files are readable (could not check in detail of they are fully ok).

Given the messages about metadata I guess this may have something to do with the changes in this PR. Can you please have a look?

moralejo:
The lstchain_dl1ab script works on MC, by the way, I only found this issue with real data.

Bultako:
Hi @moralejo 
It is indeed weird to have such never-ending process when running `lstchain_dl1ab.py` script. The warning messages are ok when the input file does not have these metadata, but this should not prevent you from ending the script correctly. if you are running `lstchain` in a local env you can always comment the two lines below in `lstchain_dl1ab.py` script and see if we can blame on the changes of this PR.

https://github.com/cta-observatory/cta-lstchain/blob/a84c82b1d3a18b711f7a4a09ad464484a115c109/lstchain/scripts/lstchain_dl1ab.py#L180 

https://github.com/cta-observatory/cta-lstchain/blob/a84c82b1d3a18b711f7a4a09ad464484a115c109/lstchain/scripts/lstchain_dl1ab.py#L307

In the meantime, I will try to reproduce this issue with those scripts.
I will let you know.

PR #731, opened 2021-09-21
Dynamic cleaning usable in r0_to_dl1

moralejo:


lstchain/scripts/lstchain_mc_r0_to_dl1.py
```diff
@@ -42,7 +43,11 @@
                     help='Path to a configuration file. If none is given, a standard configuration is applied',
                     default=None
                     )
-
+parser.add_argument('--dynamic-cleaning', action='store',
+                    type=lambda x: bool(strtobool(x)),
```
maxnoe, line 47: This is not needed. Just use `action='store_true'`

lstchain/scripts/lstchain_mc_r0_to_dl1.py
```diff
@@ -42,7 +43,11 @@
...
+                    dest='dynamic_cleaning',
+                    help='Boolean. True to use dynamic cleaning',
+                    default=False)
```
maxnoe, line 50: With `action='store_true'`, no default is required.

lstchain/scripts/lstchain_mc_r0_to_dl1.py
```diff
@@ -42,7 +43,11 @@
                     help='Path to a configuration file. If none is given, a standard configuration is applied',
                     default=None
                     )
-
+parser.add_argument('--dynamic-cleaning', action='store',
+                    type=lambda x: bool(strtobool(x)),
+                    dest='dynamic_cleaning',
```
maxnoe, line 48: dest is not needed, since the same value is auto-generated from your `--dynamic-cleaning` argument.

lstchain/scripts/lstchain_mc_r0_to_dl1.py
```diff
@@ -42,7 +43,11 @@
...
+                    help='Boolean. True to use dynamic cleaning',
```
maxnoe, line 49: Replace with "If passed, dynamic cleaning is used".

moralejo: Waaait... I was just trying if I could push code properly :'-D

moralejo:
Besides, I realized I'll have to move the code in the dl1ab script to a module into image.cleaning, so that we can use the same in r0_to_dl1. So there is quite a way to go, will undraft this when finished.

lstchain/scripts/lstchain_mc_r0_to_dl1.py
```diff
@@ -20,6 +20,7 @@
 from pathlib import Path
 import logging
 import sys
+from distutils.util import strtobool
```
rlopezcoto, line 23: this new import is not used anywhere, right?

rlopezcoto (commented):
For the time being it is ok because we need to make a new release, but it would be great to have test units for the dynamic cleaning (I think there are none at the moment).

vuillaut:
Hi
Sorry I did no follow the implementation of the dynamic cleaning in dl1ab but wouldn't it be easier to define a function
`dynamic_cleaning(geom, image, **kwargs)`.
Then one can simply read the type of cleaning with its args from the config and apply in a transparent way
`cleaning_method(geom, image, **kwargs)` with no if conditions everywhere in the code?

Or is there something I am missing?

moralejo:
> Hi
> Sorry I did no follow the implementation of the dynamic cleaning in dl1ab but wouldn't it be easier to define a function
> `dynamic_cleaning(geom, image, **kwargs)`.
> Then one can simply read the type of cleaning with its args from the config and apply in a transparent way
> `cleaning_method(geom, image, **kwargs)` with no if conditions everywhere in the code?
> 
> Or is there something I am missing?

I  don't know... we use ctapipe's image cleaning, plus, on top of it, custom lstchain _additions_ (the "pedestal-rms" cleaning - that can only be used in dl1ab - and then this dynamic cleaning)

rlopezcoto (approved):

maxnoe:
> I don't know... we use ctapipe's image cleaning, plus, on top of it, custom lstchain additions (the "pedestal-rms" cleaning - that can only be used in dl1ab - and then this dynamic cleaning)

We now have this structure in at least two files:

```
if pedestal_cleaning:
   # update cleaning levels
do_normal_cleaning

if apply_time_cleaning:
   ...

if dynamic_cleaning:
  ...
```


Having a single function `lst_cleaning` that does these checks internally cleans up the code and makes sure the different places where the cleaning is performed actually do the same (and don't get out of sync).

Even better would be a `Component` properly using the config system...

moralejo:
> 
> Having a single function `lst_cleaning` that does these checks internally cleans up the code and makes sure the different places where the cleaning is performed actually do the same (and don't get out of sync).
> 
> Even better would be a `Component` properly using the config system...

Sounds good, but I cannot do that on short-term. If anyone has time straightaway, and wants to do it, I am more than happy (but better in a separate PR I think, so I'll merge this one).  If not, we can defer that improvement and move on with releasing - I would like to confirm as soon as possible (though I have little doubt) that this works fine with the full analysis of reference Crab data.

PR #733, opened 2021-09-22
Refactor and simplify merging of HDF5 files

Bultako:
This PR simplifies and improves merging of HDF5 files by having a single `auto_merge_h5files` function. This function checks that HDF5 files to merge are compatible for merging by looking at their metadata as well as simulation info and histograms for MC files. 

These checks were also performed by similar function `smart_merge_h5files`, which has been removed. Though this function could merge *only* a list of more than one file, `auto_merge_h5files` function can also create a new file from the content of another one, which is how it is used in `lstchain_dl1ab.py` script. 

`auto_merge_h5files` also keeps existing global metadata in HDF5 files as well as potential existing attributes of their internal tables.

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -209,17 +209,16 @@ def stack_tables_h5files(filenames_list, output_filename="merged.h5", keys=None)
 
 
 def auto_merge_h5files(
-        file_list,
-        output_filename='merged.h5',
-        nodes_keys=None,
-        merge_arrays=False,
-        filters=HDF5_ZSTD_FILTERS,
-        progress_bar=False,
+    file_list=None,
+    output_filename="merged.h5",
+    nodes_keys=None,
+    merge_arrays=False,
+    filters=HDF5_ZSTD_FILTERS,
 ):
     """
     Automatic merge of HDF5 files.
-    A list of nodes keys can be provided to merge only these nodes.
-    If None, all nodes are merged.
+    A list of nodes keys can be provided to merge only these nodes. If None, all nodes are merged.
+    It may be also used to create a new file output_filename from content stored in another file.
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


rlopezcoto (approved):

lstchain/io/io.py
```diff
@@ -195,10 +194,17 @@ def stack_tables_h5files(filenames_list, output_filename="merged.h5", keys=None)
         merged_table.write(output_filename, path=k, append=True)
 
 
-def auto_merge_h5files(file_list, output_filename='merged.h5', nodes_keys=None, merge_arrays=False, filters=HDF5_ZSTD_FILTERS):
+def auto_merge_h5files(
+    file_list=None,
```
maxnoe, line 198: Just remove the default. It is a required Argument.

lstchain/io/io.py
```diff
@@ -209,6 +215,12 @@ def auto_merge_h5files(file_list, output_filename='merged.h5', nodes_keys=None,
     filters
     """
 
+    if file_list is None:
```
maxnoe, line 218: These two live are not needed of you make filelist required.

lstchain/io/io.py
```diff
@@ -209,6 +215,12 @@ def auto_merge_h5files(file_list, output_filename='merged.h5', nodes_keys=None,
...
+        file_list = []
+    if len(file_list) > 1:
+        file_list = merging_check(file_list)
+    assert len(file_list) > 0, "Please, provide a non empty file_list parameter"
```
maxnoe, line 222: Don't use assertions in non-test code.  Raise a ValueError

lstchain/io/io.py
```diff
@@ -209,6 +215,11 @@ def auto_merge_h5files(file_list, output_filename='merged.h5', nodes_keys=None,
     filters
     """
 
+    if type(file_list) != list:
```
maxnoe, line 218: This is needlessly pedantic. Everything that is an iterable of paths should be taken here. E.g. like this, `auto_merge_h5files(Path().glob("*.h5"))` will throw an error for no reason.

Just do `file_list = list(file_list)` at the start.

Bultako: You're right.
Done.

rlopezcoto (approved):

PR #734, opened 2021-09-25

moralejo:
Best MC tuning parameters (in dl1ab stage) to match the Nov'2020 Crab observations (as used in ICRC 2021 performance)

### DIFF
--- a/lstchain/data/lstchain_dl1ab_tune_MC_to_Crab_config.json
+++ b/lstchain/data/lstchain_dl1ab_tune_MC_to_Crab_config.json
@@ -1,12 +1,12 @@
 {
   "image_modifier": {
     "increase_nsb": true,
-    "extra_noise_in_dim_pixels": 1.5,
-    "extra_bias_in_dim_pixels": 0.6,
+    "extra_noise_in_dim_pixels": 1.25,
+    "extra_bias_in_dim_pixels": 0.55,
     "transition_charge": 8,
-    "extra_noise_in_bright_pixels": 1.44,
+    "extra_noise_in_bright_pixels": 2.05,
     "increase_psf": true,
-    "smeared_light_fraction": 0.175
+    "smeared_light_fraction": 0.125
   },
 
   "tailcut": {


rlopezcoto:
it seems that several tests are failing because of this change... any idea why?

moralejo:
I have no idea, I did not know a test used that config. Or is it again some network issue?

maxnoe:
@moralejo I think this is due to a new scikit-learn version that changed / removed deprecated options. It seems the problem is the Random Forest regressor config:         ` TypeError: __init__() got an unexpected keyword argument 'min_impurity_split'`.

https://scikit-learn.org/stable/whats_new/v1.0.html

maxnoe:
> Deprecated since version 0.19: min_impurity_split has been deprecated in favor of min_impurity_decrease in 0.19. The default value of min_impurity_split has changed from 1e-7 to 0 in 0.23 and it will be removed in 1.0 (renaming of 0.25). Use min_impurity_decrease instead.

Why was this option even added? It was deprecated before lstchain was even created.

moralejo:
But my latest PR surely has nothing to do with it... and everything worked right before, when we made the last release. Pretty strange, right?

maxnoe:
@moralejo No, the new sklearn version was just released. This is just the first run of the CI system that installed the new version, nothing strange about that.

moralejo:
Ok, so this is because the version is not fixed in the environment (just a minimum), right? Similar to the problem I had a couple of days ago with astropy, that you tracked down to this: https://github.com/astropy/astropy/issues/12196
I assume this is intended, so we are aware of changes that affect us.

maxnoe:
> I assume this is intended, so we are aware of changes that affect us.

Yes exactly, the unit test for the current development should run on the most current versions, as these are the ones we usually want to use. The exception is ctapipe, as ctapipe upgrades are more work at the moment.

For production, the versions must me fixed, e.g. by creating an initial environment from `the environment.yaml` and then exporting the environment (which will contain fixed versions of all packages, also the transitive dependencies)

rlopezcoto (approved):
now tests are passing after #735

PR #735, opened 2021-09-27
Remove deprecated and now removed sklearn option from config

maxnoe:


rlopezcoto:
so this option wasn't really affecting any of the trainings anyways, right?

maxnoe:
No, it was a deprecated option that was set to `None` and was replaced 5 years ago by the `min_impurity_decrease` option.

rlopezcoto (approved):
ok, thanks @maxnoe, let's get this fixed not to interfere with other PRs...

PR #747, opened 2021-10-26
Store configuration execution metadata in HDF5 files

Bultako:
This PR stores the config execution of scripts as metadata attributes of the HDF5 files and tables produced in the R0->calibration->DL1->DL1AB->DL2 workflow.

The scripts involved in this workflow are:
- `/scripts/lstchain_data_create_time_calibration_file.py`
- `/tools/lstchain_create_calibration_file.py`
- `/scripts/lstchain_data_r0_to_dl1.py`
- `/scripts/lstchain_dl1ab.py`
- `/scripts/lstchain_dl1_to_dl2.py`

The global metadata stored at the root level of the HDF5 files, as well as into each of the tables/containers produced, account for the versions of the software used in the execution of the script that produces de file i.e. `CTAPIPE_VERSION`, `LSTCHAIN_VERSION`

The configuration parameters used for each script/tool are stored as an attribute `config` of the tables/containers produced. 

These metadata may be easily retrieved with `pytables` library and `vitables` software, as described in the python snippets and screenshots below.

```python
with tables.open_file(filename) as file:
   ...:     print(file.root._v_attrs["LSTCHAIN_VERSION"])

0.7.4.post16+git0deb44e
```

```python
with tables.open_file(filename) as file:
   ...:    config = yaml.safe_load(file.root.dl1.event.telescope.parameters.LST_LSTCam.attrs["config"])
print(config["tailcut"])

{'picture_thresh': 6,
 'boundary_thresh': 3,
 'keep_isolated_pixels': False,
 'min_number_picture_neighbors': 2,
 'use_only_main_island': False,
 'delta_time': 2}
```

Captura de pantalla 2021-10-26 a las 10 41 44

In the case where, during one of the workflow steps, a specific table is directly retrieved from one previous existing
 file and copied into the new produced HDF5 file, then the software versions and the config execution metadata are kept untouched.


rlopezcoto (approved):

moralejo:
Hi, 

a question came to my mind during the presentation at the LST analysis session - I am not sure if commenting on a close PR is a proper way to proceed (please let me know):

The question is the following:
```
with tables.open_file(filename) as file:
   ...:    config = yaml.safe_load(file.root.dl1.event.telescope.parameters.LST_LSTCam.attrs["config"])
```

this returns the whole contents of the config file used when the program which produced the file was run, right?
But typically for each stage we use .json files which contain many more parameters than those used by the stage in question.
For example, the pulse integrator settings are irrelevant for running the lstchain_dl1ab script, so in the script's output we would end up with values for those parameter which may be different from those actually used to calibrate the data. Is that the case? And if so, how can that be solved? I each stage we should remove from the config all settings (from the config file) that are not actually used by the program, and replace them by the corresponding settings of the provenance of the previous stage, but I am not sure if that is technically possible. 

Or perhaps this is already taken into account, can you clarify, @Bultako ?

morcuended:
I think this goes in the direction of not having only one all-purpose config file.

Bultako:
@moralejo 

The config is recorded only in the **table** that has been produced.

There are other tables in a HDF5 file coming from lower levels of the processing, for example those present in the `/monitoring` key (i.e. calibration, flatfield, pedestal). These tables also contain their own config metadata untouched, the config used when these tables were produced.

Bultako:
We could follow two different approaches to get rid of useless config params.
- use specific json files for each task/process.
- filter out the config file before saving it as metadata.

The second option seems easier to implement now, but we have to be very careful to not filter out params that are actually involved in the process. 

In this PR I took the decision to register the whole config present in memory when the script is executed, but record it only in the table is produced.

maxnoe:
> I am not sure if commenting on a close PR is a proper way to proceed (please let me know):

Better make an issue ;)

moralejo:
Ok, continues in #795

PR #745, opened 2021-10-23
Function to obtain NSB tuning parameters

moralejo:
This function reads a real data and an MC file and returns the parameters needed to tune the MC NSB to that of the data.

PR #755, opened 2021-10-29

moralejo:
Modify values for the NSB tuning of the MC to the Crab field (as of 20201120). Using now the values obtained from #745. The additional noise in bright pixels is smaller (previous one was tuned based on MC-data comparison of the noise in muon ring events, but I think this is not correct due to the likely presence of light from the shower in real data ring events - whereas MC ring events only contain light from the muon.


LukasNickel (approved):
Do you have some plots about how the event distribution change with that?

LukasNickel:
> Sorry for slow reply. Do you mean the noise distribution?:

I mean old and new fine tuning values actually.
Your argumentation makes sense to me, but I was wondering how big the impact of the change is and whether its an improvement over the old values. Could be that the values are worse, but the approach is more correct for the future of course, just curious in case you had plots produced already.

> By the way , there is perhaps a hint of the data having a weaker afterpulsing branch than the MC?

Interesting for sure. The MC distribution looks narrower, but then has that afterpulse which the data barely shows...

moralejo:
> 
> I mean old and new fine tuning noise distributions actually. Your argumentation makes sense to me, but I was wondering how big the impact of the change is and whether its an improvement over the old values. 

The change of the added noise value is only significant  for the bright pixels. And the added noise is very small compared to the intrinsic poissonian fluctuations of the signal in those bright pixels, hence the impact in later analysis will be irrelevant. Indeed I did not see any change in the  MC gamma vs. data excess comparisons.

The value is not obtained from the charge distribution in interleaved pedestals in real data DL1a  (since the pulses are extracted with the biased pulse integrator), but from the info in the monitoring table filled by the calibration, which uses an unbiased pulse integrator. So I do not have a charge distribution to compare.

The old value was not based on the noise observed (with unbiased integrator) on pedestal events, but on the noise seen on muon ring events (which use a global peak integrator, hence unbiased for pixels with no signal) in pixels far from the ring. I suspect, though, that in those mu-ring events the noise is higher due to the parent shower from which the muon comes, and/or muon light focused far from the ring by misaligned mirrors.

LukasNickel (approved):

moralejo:
For some reason @LukasNickel 's review is not enough to merge the PR. @rlopezcoto, can you please approve it? And any idea why this is needed?

maxnoe:
You need write access to the target branch for a review to count towards the required approvals. 

Currently, the master branch is configured to only allow write access for @vuillaut, @moralejo, @rlopezcoto, @maxnoe

maxnoe (approved):

rlopezcoto:
> You need write access to the target branch for a review to count towards the required approvals.
> Currently, the master branch is configured to only allow write access for @vuillaut, @moralejo, @rlopezcoto, @maxnoe, so these are the only reviewers that count

In the past, when the target branch was `upstream/master` and we were doing the PRs from a fork, everybody with write access in the repo could approve PRs, why is this not the case anymore?

rlopezcoto:
> > You need write access to the target branch for a review to count towards the required approvals.
> > Currently, the master branch is configured to only allow write access for @vuillaut, @moralejo, @rlopezcoto, @maxnoe, so these are the only reviewers that count
> 
> In the past, when the target branch was `upstream/master` and we were doing the PRs from a fork, everybody with write access in the repo could approve PRs, why is this not the case anymore?

ok, the problem was that @ LukasNickel did not have Write access to the cta-lstchain repo, that is why his review did not count. I just added him.

PR #759, opened 2021-10-31
Set disp_norm method as a default direction reconstruction

SeiyaNozaki:
In this PR, disp_norm method implemented in #710 is set as a default direction reconstruction method.
The direction reconstruction method can be switched modifying `disp_method` in config file (`disp_vector` or `disp_norm_sign`).
In addition, some less-important input parameters for random forest are also removed as suggested by @moralejo.
(The selected input parameters were used here: https://indico.cta-observatory.org/event/3627/contributions/30999/attachments/20183/28084/20211020_SourceIndependentImprovement.pdf)


lstchain/data/lstchain_src_dep_config.json
```diff
@@ -148,7 +150,7 @@
     "dist"
   ],
 
-  "particle_classification_features": [
+  "disp_classification_features": [
```
moralejo, line 153: What does "from_source" mean in this context? We don't know the source position yet at this point.
Ah wait, this is source-dependent... then why disp reconstruction??

lstchain/data/lstchain_standard_config.json
```diff
@@ -133,74 +133,54 @@
...
+    "leakage_intensity_width_2"
   ],
 
-  "disp_classification_features": [
+  "particle_classification_features": [
     "log_intensity",
     "width",
     "length",
-    "x",
-    "y",
```
moralejo, line 192: Is removing x and y really ok? We lose any info related to the image position on the camera...
In @misabelber's tests r, phi and psi were gone, but x and y were kept.

The one case in which I guess it may be good to omit them is in the disp classification, because if they are used, the training with inner gammas would certainly introduce a bias towards reconstructing background also close to the FOV center.

moralejo (commented):
Thanks a lot for this, will try it as soon as possible. Only have some questions on the used features.

lstchain/data/lstchain_standard_config.json
SeiyaNozaki: ok, x and y are back now.

lstchain/data/lstchain_src_dep_config.json
SeiyaNozaki: Now I'm wondering if the information of reconstructed disp can be used for source-dependent analysis (especially at high energy ranges).
So I want to keep the disp reconstruction procedure even for source-dependent analysis.
(But input parameters would be irrelevant indeed for the moment...)

moralejo: ok!

lstchain/data/lstchain_standard_config.json
maxnoe: I never used x and y for training of the disp regression / sign classification and I think when I tried it out it had even detrimental effects (less generalization). This should be easy to test though. I'd rather keep `r` and discard `x` and `y` for disp norm, intuitively.

moralejo: Now they are only in E-regression and particle classification, not on disp (though I actually meant to keep them in the disp regression - and not in classification - but that would also introduce some bias when training with inner-FOV-gammas, so let's keep it like it is now for the default).

moralejo (approved):
I tried this on Crab and BL Lac data, and  everything looks good (as you @SeiyaNozaki  showed), significantly better than with the current disp-dxdy. Also the comparison of parameters between MC gammas and Crab excess from Nov'20 is very satisfactory.

PR #760, opened 2021-11-02
Store file names of merged files as an array

Bultako:
This PR intends to solve an issue present when merging a long list of filenames. The list of filenames merged are now stored as an attribute at the root level of the HDF5 file. Due to an internal limit of HDF5 of 64k in size, the value of this `SOURCE_FILENAMES` attribute may be cropped for long lists of long filenames. 

See also comments:
- https://github.com/cta-observatory/cta-lstchain/pull/728#discussion_r705991091
- https://github.com/cta-observatory/cta-lstchain/pull/733#discussion_r737478114
- https://github.com/cta-observatory/cta-lstchain/pull/733#issuecomment-950879905
- https://github.com/cta-observatory/cta-lstchain/pull/747#issuecomment-954741336

This PR stores the list of filenames considered in a merging as an array `filenames` within a key `source_filenames`


lstchain/io/io.py
```diff
@@ -767,23 +774,20 @@ def check_metadata(metadata1, metadata2):
...
     """
-    if source:
-        input_url = source.input_url
 
     metadata = MetaData()
     metadata.LSTCHAIN_VERSION = lstchain.__version__
     metadata.CTAPIPE_VERSION = ctapipe.__version__
     metadata.CTAPIPE_IO_LST_VERSION = ctapipe_io_lst.__version__
     metadata.CONTACT = "LST Consortium"
-    metadata.SOURCE_FILENAMES.append(os.path.basename(input_url))
```
rlopezcoto, line 786: is this variable not used anymore?

maxnoe: This was the point of this PR, right? To no store it in the metadata, because it might be to large.

rlopezcoto: but for single files shouldn't this still be used?

rlopezcoto (approved):

lstchain/io/io.py
Bultako: @rlopezcoto
This variable will not be used anymore.
For single files this var contained just the name of the file, this is useless.

PR #761, opened 2021-11-03
Add docs

rlopezcoto:
Re-live of #503 to start filling out some documentation for lstchain. 
We do not use travis-ci anymore, so I am not sure how to perform an automatic deployment of the documentation with github-ci (maybe we could use travis only for the documentation?), ideas?


maxnoe:
Just copy the gh actions config of ctapipe?

morcuended:
Thanks for reviving this. If we want to publish the docs using GitHub pages (like it is done in ctapipe) I would make the deployment via GitHub actions. If we want to publish it in Read The Docs, the deployment is done externally once you synchronize the repository.

If we go for the first option we can replicate for example the second action in https://github.com/cta-observatory/ctapipe/blob/master/.github/workflows/ci.yml

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+author = u'LST Collaboration'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = u'0.6.0'
+# The full version, including alpha/beta/rc tags.
+release = u'0.6.0'
```
morcuended, line 59: use actual `lstchain.__version__`

docs/index.html
```diff
@@ -0,0 +1,110 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
morcuended, line 1: I think this compiled file has to be removed

maxnoe: No, it's the homepage of the docs.

maxnoe: Or it will be. But it looks like generated output, which should not be in `docs/` but in `docs/build`? I'm a bit confused.

rlopezcoto: yes, thanks, done.

docs/conf.py
rlopezcoto: done

rlopezcoto:
> Thanks for reviving this. If we want to publish the docs using GitHub pages (like it is done in ctapipe) I would make the deployment via GitHub actions. If we want to publish it in Read The Docs, the deployment is done externally once you synchronize the repository.
> 
> If we go for the first option we can replicate for example the second action in https://github.com/cta-observatory/ctapipe/blob/master/.github/workflows/ci.yml

thanks for the suggestions @morcuended and @maxnoe, I copied the part regarding `docs` from https://github.com/cta-observatory/ctapipe/blob/master/.github/workflows/ci.yml , but I think I am still missing some stuff.

morcuended:
@rlopezcoto, I think two PR are mixed up #761 and #503. You pushed the latest changes to #503.

rlopezcoto:
> @rlopezcoto, I think two PR are mixed up #761 and #503. You pushed the latest changes to #503.

Right, thanks, now I pushed them here (#761)

rlopezcoto:
We do not have any `Makefile`, therefore `make doc` does not work, any alternative or shall we add one?

docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
morcuended, line 1: There is a Makefile

rlopezcoto: I mean in the initial `cta-lstchain` folder, that is the one targeted in the ci deployment, the same way as:
https://github.com/cta-observatory/ctapipe/blob/master/Makefile

maxnoe:
> We do not have any Makefile, therefore make doc does not work, any alternative or shall we add one?

Yes, just `cd docs && make html`

docs/Makefile
morcuended: Ah, right. Now I see it. I've always used directly `make html` instead.

rlopezcoto:
Missing sphinx in the environment, are these three:
https://github.com/cta-observatory/ctapipe/blob/4573e88b0015cdf2a9cccd4ba28e722c2256fa16/environment.yml#L36-L38
needed?

what about https://github.com/cta-observatory/ctapipe/blob/63e0df4337c18386aab3dea0d0654c5ec02b447a/setup.py#L29-L39 ?

morcuended:
> Missing sphinx in the environment, are these three: https://github.com/cta-observatory/ctapipe/blob/4573e88b0015cdf2a9cccd4ba28e722c2256fa16/environment.yml#L36-L38 needed?
> 
> what about https://github.com/cta-observatory/ctapipe/blob/63e0df4337c18386aab3dea0d0654c5ec02b447a/setup.py#L29-L39 ?

I had the comment in draft for several hours...

I do not know if there is any difference between installing sphinx stuff via `environment.yaml` or `setup.py` but if you are doing `pip install -e .[docs]` you are expected to have docs requirements in `setup.py`. I see that sphinx is not being installed yet in the gh action. So I would add the requirements to `setup.py`

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- General configuration ------------------------------------------------
+
+# If your documentation needs a minimal Sphinx version, state it here.
+#
+ needs_sphinx = '1.5'
```
morcuended, line 28: remove whitespace at the beginning of the line

docs/index.rst
```diff
@@ -0,0 +1,27 @@
+.. lstchain documentation master file, created by
+   sphinx-quickstart on Tue Aug 11 14:07:30 2020.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+========================================================================
+ Repository for the high level analysis of the LST (:mod:`cta-lstchain`)
```
morcuended, line 7: Do we consider lstchain as high-level analysis?

morcuended (changes requested):
I can build locally the web without having to define the requirements in `setup.py`. Not sure why is still failing.

morcuended:
I think the problem is that the conda environment is not created in the first place for the docs GitHub action. It is created for the test stage but the prerequisites are being treated independently.

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+version = lstchain.__version__
+# The full version, including alpha/beta/rc tags.
+release = lstchain.__version__
+
+# The language for content autogenerated by Sphinx. Refer to documentation
+# for a list of supported languages.
+#
+# This is also used if you do content translation via gettext catalogs.
+# Usually you set "language" from the command line for these cases.
+language = None
```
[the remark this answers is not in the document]
rlopezcoto: already done in https://github.com/cta-observatory/cta-lstchain/pull/761/commits/23b5bc91bbf0088824dc7cdfa507084f8583fa4f

docs/conf.py
rlopezcoto: already done in https://github.com/cta-observatory/cta-lstchain/pull/761/commits/23b5bc91bbf0088824dc7cdfa507084f8583fa4f

docs/conf.py
```diff
@@ -0,0 +1,158 @@
...
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'lstchain'
+copyright = u'2021, LST Collaboration'
```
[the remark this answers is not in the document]
rlopezcoto: thanks, changed now

docs/index.rst
rlopezcoto: no, I changed it now

rlopezcoto:
thanks a lot @morcuended for getting this properly set in the setup, I was doing some mess there, hopefully now should be ok

docs/conf.py
```diff
@@ -0,0 +1,168 @@
...
+
+conf = ConfigParser()
+conf.read([os.path.join(os.path.dirname(__file__), "..", "setup.cfg")])
+setup_cfg = dict(conf.items("metadata"))
+
+# -- General configuration ------------------------------------------------
+
+# If your documentation needs a minimal Sphinx version, state it here.
+#
+needs_sphinx = '1.5'
```
maxnoe, line 36: We require 3.5 in `setup.py`.

Can you test if it works with the current release? 4.2? We had some issues with it in ctapipe I was not yet able to resolve, but maybe this does not affect lstchain.

docs/conf.py
```diff
@@ -0,0 +1,168 @@
...
+# If true, `todo` and `todoList` produce output, else they produce nothing.
+todo_include_todos = False
+
+
+# -- Options for HTML output ----------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
```
maxnoe, line 95: We use the read the docs theme for ctapipe and you also added that theme to the requirements here.

docs/conf.py
```diff
@@ -0,0 +1,168 @@
...
+#  dir menu entry, description, category)
+texinfo_documents = [
+    (master_doc, 'lstchain', u'lstchain Documentation',
+     author, 'lstchain', 'One line description of project.',
+     'Miscellaneous'),
+]
+
+
+import sphinx_rtd_theme
+html_theme = 'sphinx_rtd_theme'
```
maxnoe, line 167: Ah! this is duplicated with the above

docs/conf.py
morcuended: I tried with the latest 4.3 and it worked.

docs/conf.py
morcuended: good catch

docs/conf.py
morcuended: changed by rtd theme

maxnoe (changes requested):
I was surprised that there were no issues at all. Turns out, the docs do not include anything but the index page, no API docs for lstchain.

Do we want to merge it like this and add the API docs later?

maxnoe:
This is basically everything:
Screenshot-2021-11-10T20:59:23+01:00

docs/conf.py
rlopezcoto: that line was directly copied from:
https://github.com/cta-observatory/ctapipe/blob/a2d5d084c0b9fba68646b09cfd20ae3f733fff4e/docs/conf.py#L39

maybe it needs to be updated there as well

rlopezcoto:
what do you mean @maxnoe, a template for the documentation?

morcuended:
I'm working on a minimal API documentation

maxnoe:
@rlopezcoto I don't understand what you mean with template.

I mean including the automatically generated documentation frome module, class and function docststrings we already have in the code.

rlopezcoto:
sorry, I did not get what you meant, it seems that @morcuended is already taking care of the implementation

morcuended:
I added minimal API documentation. I had to fix quite some docstrings to make it work without warnings.

We have to think about what to do with the notebooks, we could move all of them to `docs/tutorial` or leave them where they are and use for example `nbsphinx_link`.

morcuended (dismissed):

rlopezcoto:
thanks @morcuended ! Let's see if we can get an external review and get this merged.
Regarding the notebooks, I think it makes sense that they are moved into `docs/tutorial`, but since almost none of them are working at the moment, we could think about doing that when updating them for the next LST Analysis school.

rlopezcoto:
so can we move forward with this @maxnoe or is there anything that should be implemented at this stage?

maxnoe:
@rlopezcoto I found some packages missing in the `environment.yml` for the docs build and updated the `contribute.rst` to include downloading the test data and running the tests also on the test data.

maxnoe (approved):

morcuended:
Just added a new line after `.. code::` to make the docs compile.

rlopezcoto:
> Just added a new line after `.. code::` to make the docs compile.

Thanks @morcuended and @maxnoe, we need another approval and then we can merge

maxnoe (approved):

vuillaut (approved):
Looks good to me.

Issue #758, opened 2021-10-29
lstchain_create_calibration_file currently broken for simulations


maxnoe:
The tool contains much branching to support both simulations and observations, but is currently not working on simulations since at the very beginning, an `LSTEventSource` is assumend:

https://github.com/cta-observatory/cta-lstchain/blob/d955efa6da3e22ae6b4d1fcbeb8fb686de5c976a/lstchain/tools/lstchain_create_calibration_file.py#L92-L94

Afterwards, it tests in a very crude way if the event source is actually an `LSTEventSource`, but this is too late:
https://github.com/cta-observatory/cta-lstchain/blob/d955efa6da3e22ae6b4d1fcbeb8fb686de5c976a/lstchain/tools/lstchain_create_calibration_file.py#L97

This should also just be `isinstance(self.eventsource, LSTEventSource)`, not this substring finding.

FrancaCassol:
Hi @maxnoe,
indeed, I will have a look to fixe this

FrancaCassol:
Hi @maxnoe,
I solved the problems above, but to make the script compatible with MC data we need either to be able to switch off the MC R1 gain selection and calibration or to use the R0 waveform, but in this case I must  remove the pedestals that are not available in ctapipe 0.10.5, do we have a plan for  moving to ctapipe 0.11.0?

maxnoe:
See #763

PR #765, opened 2021-11-05
Make calibration tool simulation compatible

FrancaCassol:
Hi, 
This PR updates the calibration tool in order to make it again compatible with MC data (see [https://github.com/cta-observatory/cta-lstchain/issues/758](https://github.com/cta-observatory/cta-lstchain/issues/758) )


lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -164,28 +188,26 @@ def start(self):
...
-                # if flat-field event: no calibration  TIB for the moment,
-                # use a cut on the charge for ff events and on std for rejecting Magic Lidar events
+                # if flat-field event
+                # use a cut on the charge for MC events if trigger not defined
                 elif event.trigger.event_type==EventType.FLATFIELD or (
-                        self.simulation and np.median(np.sum(event.r1.tel[tel_id].waveform[0], axis=1))
-                        > self.processor.minimum_hg_charge_median
-                        and np.std(np.sum(event.r1.tel[tel_id].waveform[1], axis=1))
-                        < self.processor.maximum_lg_charge_std):
+                        self.simulation and
```
rlopezcoto, line 208: what about the cut to get rid of MAGIC LIDAR events?

maxnoe: The tagging is now performed by the event source. The event source should not tag LIDAR events as FF so just filtering for the trigger type should be ok.

maxnoe: I don't understand why a cut is needed on simulated events. We have perfect information there, so why is there some additional cut needed?

FrancaCassol: Because simulated FF events can be produced also with the LightEmission code (that is what I used in the past) and the trigger is not set there (but could be is it different now?)

maxnoe: I never worked with the light emission code, so you are probably right that this is still needed then. Maybe ask Konrad if there is a possibility to override the trigger type when using this?

rlopezcoto (approved):

PR #766, opened 2021-11-05
Update calibration scripts to new lstchain version name: post --> dev

FrancaCassol:


morcuended (approved):

maxnoe (changes requested):
I want to voice my concerns here again! We should not manipulate `__version__` for this. Experts that want to test stuff can pass a fixed version like `--version=my-test-3`, that's why the option is there. The default should just be `__version__` without anything removed.

FrancaCassol:
Hi @maxnoe, 
the problem is that the charge calibration files depend on other calibration files (time, baseline, systematics) and you do not want to reproduce all of them each time you make a commit.
A solution that I can propose is that the last file version is always linked to a "pro" directory, that is: if one creates a  new calibration file with a new lstchain version (commit), a new directory is produced with the full version name and it is linked to the directory "pro". The script then will look for the pro directory for all the other calibration data that it needs  (e.g.  time, pedestal, factor,...). So OSA will always search for calibration file in the "pro" directory (which is link to last version used for the specific calibration production). Obviously, in case of reproduction, the calibration files must be reproduced coherently if necessary. But this will allow to avoid reproducing each time the calibration files when it is known that no changes in the new lstchain release are performed at that level.  What do you think?

maxnoe:
What is the problem with just passing `--version=something-fixed-for-all-the-test`?

FrancaCassol:
Yes, for the test it could be, even if it asks to reproduce anyway all the files in the chain or to use files in a general "test" directory for which you do not know the version anymore. But actually my proposal would be useful overall for the case of real releases: you can do/test the production of  old calibration runs without reproducing the  f-factors corrections, time and pedestal files if you know that nothing changed for them (that will be often the case). Or even more generally, if we know that nothing changed in the calibration part, we can simply avoid to reproduce for the new release the calibration files, we continue to use the one in "pro" (linked to the last useful production release). The idea is to dissociate the DL1 production from the calibration production, but keeping in the tree the name of the used released

PR #769, opened 2021-11-06
Update lstchain_dl1ab.py

moralejo:
Fix issue introduced in #747


lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -203,13 +203,14 @@ def main():
         with tables.open_file(args.output_file, mode='a') as output:
             params = output.root[dl1_params_lstcam_key].read()
 
-            for tab in [output.root[dl1_params_lstcam_key], output.root[dl1_images_lstcam_key]]:
+            tabs = [output.root[dl1_params_lstcam_key]]
```
maxnoe, line 206: Why do we store this config in each of the tables? Why not just at the file root?

moralejo: Question for @Bultako I think... But we can just fix the problem here and leave that for another PR, right?

maxnoe: I think this is unnecessary duplication of the same info and this introduced this issue in the first place. `output.root.attrs` will always exist.

maxnoe (approved):

maxnoe (dismissed):

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -203,13 +203,14 @@ def main():
...
+            if not args.noimage:
+                tabs.append(output.root[dl1_images_lstcam_key])
+            for tab in tabs:
                 # need container to use lstchain.io.add_global_metadata and lstchain.io.add_config_metadata
                 for k, item in metadata.as_dict().items():
                     tab.attrs[k] = item
                 tab.attrs["config"] = str(config)
```
Bultako, line 213: I think that the image table is not produced during the execution of this script, but we are just copying the image table that has been produced with `lstchain_data_r0_to_dl1.py` script, so the image table has already its own provenance config stored in the execution of a previous script. So I would not reassign  the config to the image table. 

I would just **not consider** the image table, I would remove the loop in tabs and consider only the parameters table, as below:

```python
    # need container to use lstchain.io.add_global_metadata and lstchain.io.add_config_metadata
    for k, item in metadata.as_dict().items():
        output.root[dl1_params_lstcam_key].attrs[k] = item
        output.root[dl1_params_lstcam_key].attrs["config"] = str(config)
```

Bultako (commented):
good catch @moralejo, I left some comments in between the lines.

lstchain/scripts/lstchain_dl1ab.py
maxnoe: Letting the monitoring tables have their own provenance information at the node level does not mean not putting the information on how the hole file was create at the root.

Bultako: Sure.
In fact, the root level has already the global metadata info about software versions used in producing the **HDF5 file**, while the keys/tables have config and the software versions used in producing the **tables**

Bultako (approved):

PR #770, opened 2021-11-08
Set the drs4 offset to zero because the drs4 baseline script uses the r1 waveform

FrancaCassol:


maxnoe (approved):

FrancaCassol:
Let's merge this so that Daniel can produce a test sample of DL1?

PR #771, opened 2021-11-08
Add meaningful signed parameters to DL2

moralejo:
Calculate appropriate signs for skewness and time_gradient, referred to the reconstructed source position.


moralejo:
Ah damn... we need those before, for training. Back to draft.

moralejo:
But other quantities reconstructed in the dl1 => dl2 step like reco_disp_norm are used in the g/h separation RF, so it should be possible to use also these signed_time_gradient and signed_skewness

rlopezcoto (commented):
Apart from the pep8 formatting (space between mathematical operators and variables), it looks good

rlopezcoto (dismissed):

moralejo:
I realized that we can use ctapipe's camera_to_shower_coordinates (as Seiya does for source-dep analysis) instead of creating a new function. Results are identical, will do.

rlopezcoto (approved):

PR #772, opened 2021-11-11
Bugfix and cleanup on source-dep analysis scripts

SeiyaNozaki:
This PR solved an issue ( See #698 ) on source-dependent analysis, and add some i/o functions for source-dependent parameters.
After this PR, I will implement DL3 source-dep tool and re-implement the function to calculate srcdep parameters just before RF training/DL2 generation (mentioned in #704).


maxnoe:
> After this PR, I will implement DL3 source-dep tool

How do you plan to do this? GADF does not currently have a way for source dependent analysis (each event must have a single RA/DEC coordinate) and gammapy does not support it either.

SeiyaNozaki:
Hi @maxnoe 

Please see the following slides about source-dependent DL3 analysis which I plan to implement.
https://indico.cta-observatory.org/event/3511/contributions/29306/attachments/19602/27144/20210531_SourceDependentAnalysisUpdate.pdf

Points are:
- check which gammaness is the highest one with each assumed source position (on or off), and select this gammaness and assumed source position for each event
- apply gammaness and alpha cut on each event (also for IRF production)
- for survived events, assign the assumed source position to "reco_position" for gammapy analysis
- only 1D analysis in gammapy is assumed, prepare reasonable ON/OFF region to obtain those survived events

lstchain/io/io.py
```diff
@@ -1159,3 +1161,58 @@ def merge_dl2_runs(data_tag, runs, columns_to_read=None, n_process=4):
...
+
+
+def get_srcdep_params(filename, key):
+    """
+    get srcdep parameter data frame
+
+    Parameters
+    ----------
+    filename: str - path to the HDF5 file
+    key: `str` - multi index key correspoinding to an expected source position (e.g. 'on', 'off_180')
```
rlopezcoto, line 1200: correspoinding -> typo

rlopezcoto (commented):
the definition of the angles for the calculation of the source-dependent parameters for the `off` is now in the config file, right? where is this taken into account in the new `get_srcdep_params` function defined?

SeiyaNozaki:
Hi @rlopezcoto 
>the definition of the angles for the calculation of the source-dependent parameters for the off is now in the config file, right? 
yes

>where is this taken into account in the new get_srcdep_params function defined?
This function is just intended to extract already saved source-dependent parameters in DL1/DL2 files.

rlopezcoto (approved):

PR #787, opened 2021-11-12
Use lists of arguments for subprocess calls

maxnoe:


lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -146,21 +146,33 @@ def main():
...
-                        cmd += "--yes "
+                        cmd.append("--yes")
 
                     if no_sys_correction:
-                        cmd += "--no_sys_correction "
+                        cmd.append("--no_sys_correction")
 
-                    fh.write(cmd)
+                    # join command together with newline, line continuation and indentation
+                    fh.write(" \\\n  ".join(cmd))
```
morcuended, line 172: nice

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
```diff
@@ -146,21 +146,33 @@ def main():
                     fh.write("#SBATCH --mem-per-cpu=10G\n")
                     fh.write("#SBATCH -D %s \n" % output_dir)
 
-                    cmd = f"srun onsite_create_calibration_file -r {run} " \
-                        f"-p {ped_run} -v {prod_id} --sub_run {sub_run} " \
-                        f"-b {base_dir} -s {stat_events} --output_base_name {output_base_name} " \
-                        f"--filters {filters} --sys_date {sys_date} " \
-                        f"--config {config_file} --time_run {time_run} "
+                    cmd = [
```
morcuended, line 149: Should we add `set -eux -o pipefail` at the beginning of the job file? Provided the errors are propagated

morcuended (approved):

lstchain/scripts/onsite/onsite_create_calibration_files_with_batch.py
maxnoe: Since there is only a single command run in this script, it will not really make a difference.

PR #788, opened 2021-11-13
Function to calculate cleaning picture threshold from noise in data

moralejo:
Looks for a picture threshold such that not more than a given fraction of pixels in the camera will be above it, once the "clean_with_pedestal_threshold" condition is applied.


lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -34,10 +38,10 @@ def get_threshold_from_dl1_file(dl1_path, sigma_clean):
     Function to get picture threshold from dl1 from interleaved pedestal events.
     Return modified picture threshold for tailcut cleaning method.
     Allow cleaning the most noisy pixels (for example around the star location).
-    Threshold for each pixel is define as:
+    Threshold for each pixel is defined as:
         threshold = pedestal_bias + sigma * pedestal_std.
     Recommended threshold for cleaning:
-        galactic source: picture_thresh=8, boundary_thresh=4, sigma=3
+        galactic source: picture_thresh=8, boundary_thresh=4, sigma=2.5
```
LukasNickel, line 44: Why did that change now?

lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -79,3 +83,78 @@ def get_unusable_pixels(dl1_path, pedestal_id):
                                                       HIGH_GAIN,
                                                       :] == True)
     return unusable_pixels
+
+
+def find_safe_threshold_from_dl1_file(dl1_path, config_file=None,
```
LukasNickel, line 88: Wouldnt it be simpler to directly give the parameters to this function and avoid the standard config dance below?

lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -79,3 +83,78 @@ def get_unusable_pixels(dl1_path, pedestal_id):
...
+
+    # find the value new_threshold above which a fraction max_fraction of
+    # pixels lies:
+    index = int(len(threshold_sorted) * (1 - max_fraction))
+    new_threshold = threshold_sorted[index]
+
+    # Return the first integer value above new_threshold (to avoid too many
+    # different cleaning settings in different runs):
+
+    return np.floor(1 + new_threshold)
```
LukasNickel, line 160: This just rounds up, right?
`np.ceil` would do that as well.

lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -79,3 +83,78 @@ def get_unusable_pixels(dl1_path, pedestal_id):
...
+    max_fraction: maximum fraction of camera pixels that are allowed to get a
+    picture threshold above the base one. That is, we calculate here the base
+    picture threshold that will ensure that the condition is fulfilled
+
+    Returns: scalar
+    the value of the picture threshold that has to be used in data and
+    MC to ensure that no more than max_fraction of the camera gets an
+    increased value via tailcuts_clean_with_pedestal_threshold
+
+    -------
```
LukasNickel, line 133: That should be under `Returns` or not?

lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -79,3 +83,78 @@ def get_unusable_pixels(dl1_path, pedestal_id):
...
+
+    # Obtain the picture thresholds of pixels based on the "clean with
+    # pedestal threshold" method:
+    pic_threshold = get_threshold_from_dl1_file(dl1_path, sigma)
+    threshold_sorted = np.sort(pic_threshold)
+
+    # find the value new_threshold above which a fraction max_fraction of
+    # pixels lies:
+    index = int(len(threshold_sorted) * (1 - max_fraction))
+    new_threshold = threshold_sorted[index]
```
LukasNickel, line 155: Maybe `np.percentile` could improve this. Using `interpolation=lower/upper` should yield the desired result and avoid the need to sort the array completely

LukasNickel (commented):
How do we want to use this?
Should this maybe be part of the datachecker scripts?

lstchain/calib/camera/pixel_threshold_estimation.py
moralejo: It was simply wrong, we use 2.5 always. Also, that should not depend on galactic or extra-gal field, since you can have a bright star in an EG field. What has to depend on the field (and this is what is intended here) is the "base" cleaning, that is what will be also applied to MC for good matching. Because in a bright field the 2.5*sigma condition would often dominate over the base cleaning we use for dark fields, hence making many "holes" in the camera.

lstchain/calib/camera/pixel_threshold_estimation.py
moralejo: I think that this will make life easier to lstosa and lstmcpipe

lstchain/calib/camera/pixel_threshold_estimation.py
moralejo: Performance is not an issue here, it is not a large array. I prefer not to touch it, because I am very able to make it stop working even with such an easy change...

lstchain/calib/camera/pixel_threshold_estimation.py
```diff
@@ -79,3 +83,77 @@ def get_unusable_pixels(dl1_path, pedestal_id):
...
+
+    # Return the first integer value above new_threshold (to avoid too many
+    # different cleaning settings in different runs):
+
```
rlopezcoto, line 158: why is it a problem to get different cleaning settings in different runs? with this `ceil` we can also have very different settings if we have `new_threshold` = 8.9 for one run and 9.1 for the next one, even in the case they are not that different

moralejo: > why is it a problem to get different cleaning settings in different runs? with this `ceil` we can also have very different settings if we have `new_threshold` = 8.9 for one run and 9.1 for the next one, even in the case they are not that different

Yes, of course, the rounding does not avoid steps. The comment refers to rounding as opposed to using a floating-point value.

It is a problem for everything which is not just "produce run-wise IRFs". For that, indeed, we don't care if each run has different settings because it will have different IRF, and the corresponding MC can be produced with the same cleanings. But for other things, like e.g. data-MC comparisons, we have to put together many runs, and compare to a given MC set (the same MC for different reals runs, like in case of data taken on consecutive nights on the Crab in the same zenith range), and definitely we want to have an homogeneous analysis. The use of integer cleanings means that at least of conditions are the same (NSB etc) chances are that the bulk of the data for a given source and observation direction are processed in a homogeneous way.

moralejo:
> How do we want to use this? Should this maybe be part of the datachecker scripts?

The idea, as discussed in the Munich low-E workshop, was to process each run with the lowest possible cleaning. So this should be used by lstosa and lstmcpipe to decide on how to perform the post-DL1a analysis.

But after thinking about it, I am not for implementing this in short timescale. On one hand, @SeiyaNozaki did not see any improvement in analyzing BL Lac flare data with 6-3 cleaning vs. 8-4 (the latter seems ok for about all the LST1 data). And the BL Lac one is a sample where a potential difference would clearly show up. On the other hand, I foresee complications from having non-homogeneous analysis settings for the same source on different nights, or even within a night. So perhaps this decision on the cleaning should not be done on a run-wise basis, but for a whole dataset, and for off-line ("level C") analysis, not for the on-site, automatic analyses.

rlopezcoto (approved):

PR #789, opened 2021-11-16
Add sphinx.ext.githubpages to docs config

rlopezcoto:
For proper docs deployment, add a .nojekyll file

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx.ext.intersphinx",
+    'sphinx.ext.githubpages', 
     "numpydoc",
     "nbsphinx",
     "sphinx_automodapi.automodapi",


maxnoe:
Shouldn't this be in the root of the resulting build? Does putting it into the `docs/` foder achieve this?

rlopezcoto:
> Shouldn't this be in the root of the resulting build? Does putting it into the `docs/` foder achieve this?

I thought it was enough to include it here, it was working when I did the original PR in my fork, but maybe something else was changed?

maxnoe:
Is the `.nojekyll` inside `docs/build/_html` when you build locally?

rlopezcoto:
no, it is not there, but I already pushed a `.nojekyll` file to the `gh-pages` branch... will it be removed the next time the docs are deployed?

maxnoe:
Yes, `gh-pages` is an orphan branch with only one commit that is overwritten each time the docs are deployed.

maxnoe:
ctapipe has no `.nojekyll` and the deploy works, so what is the problem here?

This looks ok, doesn't it?

https://cta-observatory.github.io/cta-lstchain/

morcuended:
I think `.nojekyll` is not needed. The webpage is working already.

rlopezcoto:
No, it was working because I added the `.nojekyll` file by hand to the `gh-pages` branch:
https://github.com/cta-observatory/cta-lstchain/tree/gh-pages

I just removed it and rebuilt it, you can check now, it does not look nice:

Screenshot 2021-11-16 at 16 24 47
 
`ctapipe` (and apparently also `pyirf`) have `.nojekyll`:
https://github.com/cta-observatory/pyirf/blob/gh-pages/.nojekyll
https://github.com/cta-observatory/ctapipe/blob/gh-pages/.nojekyll

in the `gh-pages`, but I'm figuring out how it got there.

maxnoe:
It has to be the `"sphinx.ext.githubpages"` extension

rlopezcoto:
> It has to be the `"sphinx.ext.githubpages"` extension:

Right, thanks!

morcuended (approved):

PR #798, opened 2021-11-25
Pedestal finder

moralejo:
Implements one of the solutions described in 
https://github.com/cta-observatory/ctapipe_io_lst/issues/123

To identify interleaved pedestal events in data with unreliable event tagging.


moralejo:
Please consider this is a script for a one-time action: produce the event_id's for the wrongly tagged data.
I have tested it (and am running it over the relevant data, after which I'll test the output), but it makes no sense to e.g. make unit tests for this.
Besides this, we will need to make the event source able of reading the lists of event id's produced by this script, and to use them to set the right event_type. Then we will be able to process the whole LST1 data set with the proper (star-safe) cleaning.

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env python
+
+"""
+This script has to be launched (with no arguments) on a directory containing
```
maxnoe, line 4: You mean "in a directory"? As in the CWD is taken? That's not good for e.g. running it via osa or in parallel on lots of directories. Please add an input directory argument.

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
...
+import numpy as np
+import pandas as pd
+
+from pathlib import Path
+from ctapipe.io import read_table
+from lstchain.paths import parse_dl1_filename
+
+parser = argparse.ArgumentParser(description="Interleaved Pedestal Finder")
+
+parser.add_argument(
```
maxnoe, line 23: Make it an argument and don't provide a default. I think this should in the end be something like `<base>/monitoring/pedestal_ids/<date>/`

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
...
+
+    args = parser.parse_args()
+    output_dir = args.output_dir.absolute()
+    output_dir.mkdir(exist_ok=True)
+
+    files = glob.glob('dl1_LST-1.Run?????.????.h5')
+    files.sort()
+    if not files:
+        raise IOError("No input dl1 files found")
+    files.sort()
```
maxnoe, line 39: Sort is called twice

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
...
+
+    for file in files:
+        run_number = parse_dl1_filename(os.path.basename(file)).run
```
maxnoe, line 42: Parse filename once, then use the result:

```
run_info = parse_dl1_filename(file)  # no need to call basename, IIRC
run, subrun = run_info.run, run_info.subrun
```

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
...
+            approximate_frequency = 100
+
+        table = read_table(file, '/dl1/event/telescope/parameters/LST_LSTCam')
+        all_times = np.array(table['dragon_time'])
+        # probable interleaved flat-field events (just in case tag is
+        # faulty!):
+        ffmask = ((table['intensity'] > 3e4) &
+                  (table['concentration_pixel'] < 0.005))
+
+        pedmask = np.array(len(table)*[False])
```
maxnoe, line 57: `np.zeros(len(table), dtype=bool)`

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Approximate interleaved pedestal frequency in Hz:
+        approximate_frequency = 50
+        if run_number > 2708:
+            approximate_frequency = 100
+
+        table = read_table(file, '/dl1/event/telescope/parameters/LST_LSTCam')
+        all_times = np.array(table['dragon_time'])
+        # probable interleaved flat-field events (just in case tag is
+        # faulty!):
```
maxnoe, line 53: Why not rely on the event type for the FF events here? It should be tagged correctly by the source, or not?

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
...
+                                                              'Hz')
+        print('  Maximum intensity:', np.nanmax(intensity[pedmask]), 'pe')
+        df = pd.DataFrame({'event_id': event_id[pedmask]})
+        output_file = Path(output_dir,
+                           'pedestal_ids_Run{:0>5}.{:0>4}.h5'.format(run_number,
+                                                                     subrun_index))
+        df.to_hdf(output_file, key='interleaved_pedestal_ids', mode='w')
+
+
+def find_pedestals(timestamps, aprox_frequency=50):
```
maxnoe, line 90: typo, `approx`, maybe better `expected_frequency`?

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,173 @@
...
+        best_bins = bins
+
+    tmod = (timestamps + best_phase) % best_period
+
+    ibin = np.argmax(best_contents)
+    # check the two neighboring bins, which may also contain pedestals.
+    # Include them if content is at least 10% of the max:
+    ifirst = ibin
+    ilast = ibin
+    if best_contents[ibin - 1] > 0.1 * best_contents[ibin]:
```
maxnoe, line 162: This will error if bin 0 or bin N is the largest bin. I think you want circular boundary conditions for this, right? 

so 

```
bin_before = (best_bin + len(best_contents) - 1) % len(best_contents)
bin_after = (best_bin + 1) % len(best_contents)
```

maxnoe (changes requested):
* Please add an option for the input directory

> Please consider this is a script for a one-time action: produce the event_id's for the wrongly tagged data.

If we find an issue, we have to rerun it. If we later find new runs with missing pedestal tags, we have to rerun it. If we have a data loss for some reason, we have to rerun it.

Adding a test with the test data really is not much work and does no harm.

lstchain/scripts/lstchain_find_pedestals.py
moralejo: It is an argument, do you mean "a required argument"?

lstchain/scripts/lstchain_find_pedestals.py
moralejo: *should*,  yes...  It is the grammatical tense that makes me tense. I really prefer a straight test here.

lstchain/scripts/lstchain_find_pedestals.py
moralejo: yes, thanks, will do!

moralejo:
Thanks a lot @maxnoe, I agree with your suggestions, but cannot implement them immediately. Switch to draft, will do asap.

lstchain/scripts/lstchain_find_pedestals.py
maxnoe: Sorry, unclear, positional argument, yes.

lstchain/scripts/lstchain_find_pedestals.py
FrancaCassol, line 4: I think it would be better to create a tool instead of a script

moralejo: What would be the advantage of making it a tool? It will not be part of the pipeline, it just produces input files for it, that will be fixed forever for any given run.

moralejo:
> Adding a test with the test data really is not much work and does no harm.

@maxnoe I added a test but it fails, because the needed dl1 files are produced not by the tests under lstchain/scripts/tests, but by those in lstchain/tests, which are run later. Any idea of how to solve this? 
Anyway, the test one can make with the dl1 files we have is just a very silly one, not really testing any actual functionality of the script. So I'd be perfectly happy with skipping it.

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -257,6 +257,15 @@ def test_lstchain_dl1_to_dl2(simulated_dl2_file):
     assert "reco_src_x" in dl2_df.columns
     assert "reco_src_y" in dl2_df.columns
 
+@pytest.mark.private_data
+def test_lstchain_find_pedestals(temp_dir_observed_files):
```
morcuended, line 261: I think you also should add the fixture to the actual files `observed_dl1_files` to assure the files are also produced
```
def test_lstchain_find_pedestals(temp_dir_observed_files, observed_dl1_files):
```

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,186 @@
...
+
+parser.add_argument('output_dir', metavar='output-dir', type=Path,
+                    help='Path where to store the output hdf5 files'
+                    )
+
+def main():
+
+    args = parser.parse_args()
+
+    files = glob.glob(args.srcdir+'dl1_LST-1.Run?????.????.h5')
```
morcuended, line 35: Use `Path` or at least `os.path(args.srcdir, 'dl1_LST-1.Run?????.????.h5')` as argument for glob.

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,186 @@
...
+    if not files:
+        raise IOError("No input dl1 files found")
+    files.sort()
+
+    output_dir = args.output_dir.absolute()
+    output_dir.mkdir(exist_ok=True)
```
morcuended, line 41: I would add the argument `parents=True` to make sure all parents directories are created.

morcuended (changes requested):

lstchain/scripts/lstchain_find_pedestals.py
maxnoe: `path.join`. But I agree. Add `type=Path` to the `argpase.add_argument` and use `args.srcdir.glob(...)`

moralejo:
All done

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,183 @@
...
+    output_dir = args.output_dir.absolute()
+    output_dir.mkdir(exist_ok=True, parents=True)
+
+    for file in files:
+        run_info = parse_dl1_filename(file)
+        run_number, subrun_index = run_info.run, run_info.subrun
+
+        # Approximate interleaved pedestal frequency in Hz:
+        approximate_frequency = 50
+        if run_number > 2708:
```
morcuended, line 46: Maybe a bit more explanation about this as the date or the hardware change.

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,183 @@
...
+        ffmask = ((table['intensity'] > 3e4) &
+                  (table['concentration_pixel'] < 0.005))
+
+        pedmask = np.zeros(len(table), dtype=bool)
+        pedmask[~ffmask] = find_pedestals(all_times[~ffmask],
+                                          approximate_frequency)
+
+        # Now remove the 10 brightest events (might be cosmics accidentally
+        # falling in the time windows determined by time_pedestals). the
+        # expected value of cosmics is smaller that, so we are probably
```
morcuended, line 62: Incomplete sentence?

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,183 @@
...
+
+def find_pedestals(timestamps, expected_frequency=50):
+    """
+    Parameters
+    ----------
+    timestamps: ndarray, series of timestamps of LST1 events from a sub-run
+    expected_frequency: expected approximate frequency (Hz) of the interleaved
+    pedestals for the epoch of the observations
+
+    Returns
```
morcuended, line 101: missing `-------` below Returns

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,183 @@
...
+    distributed (and with the given approximate frequency)
+
+    The function finds the set of events which are most evenly distributed
+    in time with the approximate frequency expected_frequency. This should be the
+    interleaved pedestals, which are very regularly spaced in time to better
+    (typically) than microsec precision. Better to pass an array of
+    timestamps in which flatfield events (easily recognizable) have been
+    excluded.
+
+    -------
```
morcuended, line 112: actually, move this line below the Returns

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,183 @@
...
+        ilast = bin_after
+
+    minval = best_bins[ifirst]
+    maxval = best_bins[ilast+1]
+
+    if ilast >= ifirst:
+        return ((tmod > minval) & (tmod < maxval))
+    else:
+        return ((tmod < maxval) | (tmod > minval))
+
```
morcuended, line 181: Add extra blank line

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,183 @@
...
+parser.add_argument('-d', '--input-dir', dest='srcdir',
+                    type=Path, default='./',
+                    help='path to the directory of the DL1 files'
+                    )
+
+parser.add_argument('output_dir', metavar='output-dir', type=Path,
+                    help='Path where to store the output hdf5 files'
+                    )
+
```
morcuended, line 28: Extra blank line

morcuended (commented):
Just some minor things

lstchain/scripts/lstchain_find_pedestals.py
moralejo: ?? It is finished below: "we are probably removing some pedestals, but it does not harm."

morcuended: I meant "The expected value of cosmics is smaller that, so ..." (smaller than?)

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,186 @@
...
+        print('  Rate of identified pedestals: {:.3f}'.format(pedmask.sum() /
+                                                              (all_times[-1] -
+                                                               all_times[0])),
+                                                              'Hz')
+        if pedmask.sum() == 0:
+            continue
+
+        print('  Maximum intensity:', np.nanmax(intensity[pedmask]), 'pe')
+
+        df = pd.DataFrame({'event_id': event_id[pedmask]})
```
maxnoe, line 88: Please also add the `obs_id` column, so we can check if the correct file was given.

lstchain/scripts/lstchain_find_pedestals.py
```diff
@@ -0,0 +1,186 @@
...
+        output_file = Path(output_dir,
+                           'pedestal_ids_Run{:0>5}.{:0>4}.h5'.format(run_number,
+                                                                     subrun_index))
+        df.to_hdf(output_file, key='interleaved_pedestal_ids', mode='w')
```
maxnoe, line 92: Please use the `format="table"` format for the `to_hdf` call, this makes it so it can be read without pandas.

maxnoe: Trying this out, it still creates a group.... and a fixed table dataset with hard coded name "table".

So better use

```
with tables.open_file(output_file, "w") as f:
    f.create_table("/", "interleaved_pedestal_ids", obj=df.to_records(index=False), filters=DEFAULT_FILTERS)
```

lstchain/scripts/lstchain_find_pedestals.py
moralejo: The name of the output file (one per subrun) uses the same pattern as the input file, so unless there is already a mismatch in the DL1  file between obs_id and Run in the file name.

maxnoe (dismissed):

maxnoe (approved):

PR #812, opened 2021-12-13
Add --r0-dir option to onsite scripts, fixes #809

### LINKED ISSUE #809: Add separate argument for input base dir in onsite scripts
The onsite scripts use the output base directory option also for finding the input files. That's not good for custom tests. We should add an additional r0 base option so one can use the r0 files from the standard structure and put output files into a custom place.

maxnoe:


morcuended (approved):

maxnoe:
This does not create a different tree. It allows reading the r0 data from the original tree when using a different base dir

lstchain/scripts/onsite/onsite_create_drs4_pedestal_file.py
```diff
@@ -51,7 +52,7 @@ def main():
 
     try:
         # verify input file
-        file_list=sorted(Path(f"{base_dir}/R0").rglob(f'*{run:05d}.0000*'))
+        file_list=sorted(args.r0_dir.rglob(f'*{run:05d}.0000*'))
```
rlopezcoto, line 55: space before and after `=`

rlopezcoto (approved):

morcuended (approved):

PR #813, opened 2021-12-13
Fix types of arguments in onsite calibration script

maxnoe:
Giving run numbers explicitly did not work due to the wrong type of the arguments

### DIFF
--- a/lstchain/scripts/onsite/onsite_create_calibration_file.py
+++ b/lstchain/scripts/onsite/onsite_create_calibration_file.py
@@ -18,11 +18,6 @@ import lstchain
 import subprocess
 import pymongo
 
-def none_or_str(value):
-    if value == "None":
-        return None
-    return value
-
 # parse arguments
 parser = argparse.ArgumentParser(description='Create flat-field calibration files',
                                  formatter_class = argparse.ArgumentDefaultsHelpFormatter)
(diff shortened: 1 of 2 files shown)


lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -56,7 +54,7 @@ def none_or_str(value):
 optional.add_argument('--sub_run', help="sub-run to be processed.", type=int, default=0)
 optional.add_argument('--min_ff', help="Min FF intensity cut in ADC.", type=float)
 optional.add_argument('--max_ff', help="Max FF intensity cut in ADC.", type=float)
-optional.add_argument('-f','--filters', help="Calibox filters", type=none_or_str)
+optional.add_argument('-f','--filters', help="Calibox filters")
```
rlopezcoto, line 57: no `type` given?

FrancaCassol:
I am a bit surprised, I am using since months these scripts. 
Which arguments did you give?

maxnoe:
I gave explicit run numbers for time and pedestal run. The code did not convert them to int but they were used in the format string with `{:05d}`, so the exception "Invalid format for type str" was raised. You can try it on master, but the types are just wrong for the arguments. For the code to work, the run numbers have to be converted to integers.

lstchain/scripts/onsite/onsite_create_calibration_file.py
maxnoe: Yes, not needed. Filters is a string.

FrancaCassol:
Indeed, I though this part was corrected some time ago. Too many branches probably.

FrancaCassol (approved):

PR #814, opened 2021-12-14
Update to ctapipe_io_lst 0.13, including new spike correction method

maxnoe:


lstchain/tests/test_lstchain.py
```diff
@@ -16,9 +16,9 @@
 test_r0_path = test_data / 'real/R0/20200218/LST-1.1.Run02008.0000_first50.fits.fz'
 test_r0_path2 = test_data / 'real/R0/20200218/LST-1.1.Run02008.0100_first50.fits.fz'
 test_drs4_r0_path = test_data / 'real/R0/20200218/LST-1.1.Run02005.0000_first50.fits.fz'
-test_calib_path = test_data / 'real/calibration/20200218/v05/calibration.Run2006.0000.hdf5'
-test_drs4_pedestal_path = test_data / 'real/calibration/20200218/v05/drs4_pedestal.Run2005.0000.fits'
-test_time_calib_path = test_data / 'real/calibration/20200218/v05/time_calibration.Run2006.0000.hdf5'
+test_calib_path = test_data / 'real/monitoring/PixelCalibration/LevelA/calibration/20200218/v0.7.6.dev568+gdecfb58c/calibration_filters_52.Run02006.0000.h5'
+test_drs4_pedestal_path = test_data / 'real/monitoring/PixelCalibration/LevelA/drs4_baseline/20200218/v0.7.6.dev568+gdecfb58c/drs4_pedestal.Run02005.0000.h5'
+test_time_calib_path = test_data / 'real/monitoring/PixelCalibration/LevelA/drs4_time_sampling_from_FF/20191124/v0.7.6.dev568+gdecfb58c/time_calibration.Run01625.0000.h5'
```
rlopezcoto, line 21: why did the run number for the time calib change?

download_test_data.sh
```diff
@@ -24,6 +24,7 @@ wget \
 	-R "*.html*,*.gif" \
 	--no-host-directories --cut-dirs=1 \
 	--no-parent \
+	--level=inf \
```
rlopezcoto, line 27: what is the purpose of `--level=inf`?

maxnoe: wget has a default limit of 5 for recursing into directories. With the new data tree, we have a deeper structure than that

lstchain/tests/test_lstchain.py
maxnoe: That run is the magical, blessed run that shall be used for time calibration, according to @FrancaCassol

rlopezcoto: and why it was `Run2006` for the tests before?

maxnoe: Because that's the pedcal run in the same night as the test data. But @FrancaCassol insists that from now on, only the 1625 run should be used for all data.

FrancaCassol: Not for all data, it depends on the period, but  for the git tests you can use any reasonable run if you  want

lstchain/scripts/lstchain_data_r0_to_dl1.py
```diff
@@ -12,7 +12,7 @@
 $> python lstchain_data_r0_to_dl1.py
 --input-file LST-1.1.Run02030.0000.fits.fz
 --output-dir ./
---pedestal-file drs4_pedestal.Run2028.0000.fits
+--pedestal-file drs4_pedestal.Run2028.0000.h5
```
morcuended, line 15: Why DRS4 pedestal file has now h5 format? I thought EVB team needed FITS files for the online correction.

maxnoe: EVB needs a binary file that just contains the baseline values as uint16. They create it with a small python script from the FITS files and it will be as easy to create it from the h5 file.

rlopezcoto: change introduced in #725

rlopezcoto (approved):

lstchain/tests/test_lstchain.py
maxnoe: @FrancaCassol What is a "period" in this context, where is that information stored and how can it be accessed by analyzers such as osa?

From my limited knowledge, I assume this only has to change when pixel clusters are changed in the camera, is there a list of such interventions in terms of run numbers / dates and the corresponding drs time files that should be used for each of these periods?

PR #815, opened 2021-12-14
Use mamba to create conda env in CI

maxnoe:
Should be faster, let's see


maxnoe:
Installation time down from 4 minutes to 2:40

morcuended (approved):

PR #816, opened 2021-12-15
Fix gammapy version to 0.18.2

maxnoe:


environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - astropy~=4.2
   - pip
   - ctapipe~=0.11
-  - gammapy>=0.18
+  - gammapy=0.18
```
rlopezcoto, line 11: Shouldn't this be as in setup.py:
`'gammapy~=0.18.2'`
?

maxnoe: No, conda version work differently. This is essentially the same as `==0.18.*` in setup.py terms. It will install the latest 0.18 version, which is 0.18.2 which also won't change since gammapy does not do bugfix releases for older versions.

rlopezcoto (approved):

PR #817, opened 2021-12-15
Fix drs4 pedestal plots

maxnoe:
Fix for the issue reported by @morcuended in slack

Plots for the test run are looking like this again now:
Screenshot_20211215_121827

### DIFF
--- a/lstchain/visualization/plot_drs4.py
+++ b/lstchain/visualization/plot_drs4.py
@@ -206,7 +206,7 @@ def plot_pedestals(data_file, pedestal_file, run=0, plot_file=None, tel_id=1, of
                         label=label,
                     )
 
-                    label = "baseline correction \n + dt corr + interp. spikes"
+                    label = "baseline correction \n + dt corr + corrected spikes"
 
                     plt.step(
                         t,
--- a/lstchain/visualization/plot_drs4.py
+++ b/lstchain/visualization/plot_drs4.py
@@ -180,8 +180,6 @@ def plot_pedestals(data_file, pedestal_file, run=0, plot_file=None, tel_id=1, of
 
             pix = 0
             pad = 420
-            offset_value = reader.r0_r1_calibrator.offset.tel[tel_id]
-
             # plot corrected waveforms of first 8 events
             for i, ev in enumerate(reader):
                 for chan in np.arange(2):


morcuended (dismissed):

FrancaCassol:
Are we sure we want to re-calculate the offset value? 
I think we want to see the results with respect to what is supposed to be

maxnoe:
> I think we want to see the results with respect to what is supposed to be

But we don't have that information anywhere, do we?

FrancaCassol:
It was an input parameter

maxnoe:
Yes, but how was it input? We just don't have the knowledge. The input parameter by the way was not used anymore in the current version of the code. It was overriden in the middle of the function by looking the value up in the event source, this is why I removed it.

I can also put it back.

FrancaCassol:
> Yes, but how was it input? We just don't have the knowledge from the files. The input parameter by the way was not used anymore in the current version of the code. It was overriden in the middle of the function by looking the value up in the event source, this is why I removed it.

and what was wrong with that?

We want to arrive to release 0.8 as soon as possible, please minimise any change that is not strictly necessary

maxnoe:
As you might remember, as you actually approved the corresponding PR, we removed the offset from the event source as it was double subtracted from both drs4 baseline and r0 waveform. So the `eventsource.r0_r1calibrator.offset` is 0.

FrancaCassol:
Then leave the input parameter as it was before

maxnoe:
> Then leave the input parameter as it was before

As it was before, it would be broken. We need to change *something* to make the plot work again. Leaving the offset hard coded to 400 does not seem like a good idea to me. Especially since @jsitarek requested to change it here #725.

FrancaCassol:
I do not think we are going to change so often the offset, but this is your worry, one can simply add an argument to the onsite script.

maxnoe:
Ok, I changed to using the supplied offset instead of the one used by the source.

I don't really understand why you think it's better to have the value passed by the user instead of calculating it from the input files.

The worst thing that can happen is that the drs4 pedestal file is broken, thus a wrong offset is used and thus the plot looks weird. Which is actually the reason the plot exists, to see if the drs4 pedestal file is ok.

FrancaCassol:
Simply, for non experts it will be easier to see that data are far from the expected offset

FrancaCassol (dismissed):

moralejo:
A minor thing: "interp spikes" should now be "corrected spikes" or something similar, right?

rlopezcoto:
>A minor thing: "interp spikes" should now be "corrected spikes" or something similar, right?
I included that. @FrancaCassol or @moralejo could you please re-approve?

FrancaCassol (approved):

PR #819, opened 2021-12-15
NSB injection in MC R1 waveform

gabemery:
An option to inject NSB pulses in existing Monte-Carlo simulation at the level of calibrated waveforms (R1) is implemented. 
It : 
  - extracts the simulation NSB from the analysed file
  - find the missing NSB compared to the target data file comparing the pedestal variance (or take it from the config)
  - randomly assign a number of pulses to inject in each waveform using Poisson statistic
  - randomly draw an exact injection time (uniform) and normalisation (using the spe normalisation pdf)
  - Modify the waveform accordingly accounting for AC coupling by lowering the baseline uniformly

It is controlled by adding the following block in the configuration file:
  "waveform_nsb_tuning": {
    "nsb_tuning":true,                   #false to turn it off, is ignored for real data
    "nsb_tuning_ratio": 0.89,        #Additional fraction of the original NSB to be added, if not given will be extracted from the target_data file
    "target_data": "",                     # File from which the NSB correction should be extracted. To avoid running it for each MC file, using the nsb_tuning_ratio input with pre-extracted value will often be preferred 
    "spe_location":"",                    # File containing the single photo-electron normalisation pdf
    "pulse_template_location": ""  # file containing the spe pulse template in the form "time pulseLG pulseHG" or "time pulse"
  }

The following shows a couple of results obtained with prod5_trans 80 MC taking a gamma diffuse and a proton simtel file (run1). The target data here is the run 2969, subrun 0000 of the Crab Nebula observation of November 20th 2020. They are produced using another branch from the likelihood method development where it was first developed. Assuming no error during the copy of the changes to the current branch it should be equivalent but dedicated tests of this version will be needed. The NSB increase is by ~72%.

Current injection looks as below on a couple of waveforms (without and with signal, event_id 11003  of prod5_trans80 gamma_20deg_180deg_run1___cta-prod5-lapalma_4LSTs_MAGIC_desert-2158m_mono_cone6.simtel.gz) : 
image
image

The title contains the pixel number and number of injected pulses in a time window starting 20 samples before the waveform (relative to the 0 of the pulse template which is ~10 ns before its maximum).

The effect on integrated charge is shown below (on the next event of the same file, id 16705)
image

A first check of the validity of the correction was done by comparing the distribution of all extracted charges of all pixels in the low intensity range. The comparison below shows : a non modified gamma diffuse MC file, the same file after tuning to our target data set, a tuned proton file, and the target dataset. The NSB correction was independently extracted for both MC file during the processing.

image

I look at the low intensity pixels since they are dominated by pure background pixels. Higher intensities contain signal so spectral normalisation as a function of the particle type would be needed. It is clearly visible here that the correction improves the data/MC agreement in the charge extracted for low intensity pixels.

### DIFF
--- a/lstchain/analysis/__init__.py
+++ b/lstchain/analysis/__init__.py
@@ -1,0 +1,5 @@
+from .normalised_pulse_template import NormalizedPulseTemplate
+
+__all__ = [
+    'NormalizedPulseTemplate',
+]
--- a/lstchain/analysis/normalised_pulse_template.py
+++ b/lstchain/analysis/normalised_pulse_template.py
@@ -1,0 +1,193 @@
+import numpy as np
+from scipy.interpolate import interp1d
+
+
+class NormalizedPulseTemplate:
+    """
+        Class for handling the template for the pulsed response of the pixels
+        of the camera to a single photo-electron in high and low gain.
+    """
+
+    def __init__(self, amplitude_HG, amplitude_LG, time, amplitude_HG_err=None,
+                 amplitude_LG_err=None):
+        """
+            Save the pulse template and optional error
+            and create an interpolation.
+        Parameters
+        ----------
+        amplitude_HG/LG: array
+            Amplitude of the signal produced in a pixel by a photo-electron
+            in high gain (HG) and low gain (LG) for successive time samples
+        time: array
+            Times of the samples
+        amplitude_HG/LG_err: array
+            Error on the pulse template amplitude
+        """
+        self.time = np.array(time)
+        self.amplitude_HG = np.array(amplitude_HG)
+        self.amplitude_LG = np.array(amplitude_LG)
+        if amplitude_HG_err is not None:
+            assert np.array(amplitude_HG_err).shape == self.amplitude_HG.shape
+            self.amplitude_HG_err = np.array(amplitude_HG_err)
+        else:
+            self.amplitude_HG_err = np.zeros(self.amplitude_HG.shape)
+        if amplitude_LG_err is not None:
+            assert np.array(amplitude_LG_err).shape == self.amplitude_LG.shape
+            self.amplitude_LG_err = np.array(amplitude_LG_err)
+        else:
+            self.amplitude_LG_err = self.amplitude_LG * 0
+        self._template = self._interpolate()
+        self._template_err = self._interpolate_err()
+
+    def __call__(self, time, gain, amplitude=1, t_0=0, baseline=0):
+        """
+            Use the interpolated template to access the value of the pulse at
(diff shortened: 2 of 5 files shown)


moralejo:
Hi @gabemery, thanks for this very useful tool!  

I did not check the code, but from the description of the procedure, I think it does everything it has to do. 

How is the test done? You mention gamma and proton MC, but you should compare to noise-only MC (e.g. plotting only pixels which have no p.e.'s from Cherenkov light, if that info is available), so it should not matter if it is gamma or proton.

In the test you show the resulting noise seems marginally higher, the difference being surely unimportant, but just out of curiosity: can it be further tuned? (I mean manually - I understand the automatic approach may have this kind of small differences, and they are irrelevant)

I think it would be good to add the example configuration for Crab (disabled by default) to the standard json file.
Then, as for the availability of the reference files for the pulse shape and gain fluctuations, shall we add them together with the code, at least those we use in the coming major MC production?

gabemery:
Hi @moralejo , In the test, the code is run on either the gamma or proton file without pre-computed correction to the NSB. It is instead determined using a new function 'calculate_required_additional_nsb' (which is largely a copy of your function calculate_noise_parameters, I plan to try to reduce duplicated code here if possible). Then, taking the output dl1 files, I load the charge in each pixels (in image) for all events and create an histogram. Here two things are done : first I limit to 10 p.e. to look mostly at pure noise pixels, second I normalise the distributions to account for the different number of events and noise pixels in each files. 

As you said, looking only at pure noise pixels would remove the need to look only at low intensity. It could probably be done by applying a cleaning on data and MC (mimicking the use of only signal less pixels in MC with data will be harder, but possible).

I am not sure I see this high resulting noise you mention. After tuning, the ratio of data/MC distributions of low intensity pixels are visually the same up to ~6 p.e. Then there is a bit more data but this can be signal. Either removing those or using a realistic renormalisation of the MC could solve this discrepancy.
It is possible to manually tune quite freely by using the "nsb_tuning_ratio" parameter in the config file (its presence deactivate the automatic procedure). The extraction method can be called in external scripts to get the needed correction. I actually think it would be the best workflow in the end to extract the needed correction first (averaging it over the full target dataset instead of a single file, and maybe also averaging over multiple MC files). Then give it in the config instead of re-extracting it 11000 times.

Yes I think the reference pulse and gain fluctuation files should be included. I saw that the pulse is already in lstchain/data/oversampled_pulse_LST_8dynode_pix6_20200204.dat  but without the time information. From the commit it seems it will be removed in 0.8.0 so it could be replaced by the version which works with my code.

PR #824, opened 2021-12-16
Fix pyflakes check in CI and fix issues

maxnoe:
The yml was missing a line continuation character and that is why pyflakes was not actually run in the CI.

Fixed and also fixed some issues that were found now.

### DIFF
--- a/lstchain/scripts/onsite/onsite_create_calibration_file.py
+++ b/lstchain/scripts/onsite/onsite_create_calibration_file.py
@@ -116,8 +116,8 @@ def main():
     print(f"\n--> Input file: {input_file}")
 
     # find date
-    input_dir, name = os.path.split(os.path.abspath(input_file))
-    path, date = input_dir.rsplit('/', 1)
+    input_dir, _ = os.path.split(os.path.abspath(input_file))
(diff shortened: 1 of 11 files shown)


lstchain/scripts/onsite/onsite_create_calibration_file.py
```diff
@@ -154,7 +154,7 @@ def main():
     # search the pedestal file of the same date
     if ped_run is None:
         # else search the pedestal file of the same date
-        file_list = sorted(Path(f"{ped_dir}/{date}/{pro}/").rglob(f'drs4_pedestal*.0000.h5'))
+        file_list = sorted(Path(f"{ped_dir}/{date}/{pro}/").rglob('drs4_pedestal*.0000.h5'))
```
rlopezcoto, line 157: what is the difference computation-wise between using :
`f'drs4_pedestal*.0000.h5'` and `'drs4_pedestal*.0000.h5'`

maxnoe: An f string without something to format is just a string

rlopezcoto:
tests are not passing... some connection problem?

FrancaCassol (approved):

PR #835, opened 2021-12-30
Cleanup code

rlopezcoto:
Apologies for the long PR. I wanted to do a cleanup of the code (did not want to mess with the tests, at least for now) before the school, so I optimized and re-arranged imports and reformatted the code. Tests are passing and I checked as many things as possible, but I would appreciate if people could have a look and report some things that may seem critical to them.


rlopezcoto:
if we want to clean up the code, now it would be the time to review and merge this PR, otherwise conflicts will start to pile up again...

maxnoe (approved):
Let's merge this. I think some of these style changes look a bit off, but that can be easily fixed later.

Issue #839, opened 2022-01-03
source_filenames is not updated


moralejo:
When merging DL1 files produced with the DL1ab script I get the error below, because the "source_filenames" array already exists in the input file. This should be overwritten, since what we want (I think) is the input files of the latest processing - we might keep all earlier history, but if there is only one source_filenames it should contain the inputs of the program which produced the file in question.

```
  File "/home/abelardo.moralejo/software/cta-lstchain/lstchain/scripts/lstchain_merge_hdf5_files.py", line 92, in <module>
    main()
  File "/home/abelardo.moralejo/software/cta-lstchain/lstchain/scripts/lstchain_merge_hdf5_files.py", line 83, in main
    auto_merge_h5files(
  File "/fefs/home/abelardo.moralejo/software/cta-lstchain/lstchain/io/io.py", line 301, in auto_merge_h5files
    sources = file.create_group("/", "source_filenames", "List of files merged")
  File "/fefs/aswg/workspace/abelardo.moralejo/miniconda3/envs/lst-dev/lib/python3.8/site-packages/tables/file.py", line 939, in create_group
    return Group(parentnode, name,
  File "/fefs/aswg/workspace/abelardo.moralejo/miniconda3/envs/lst-dev/lib/python3.8/site-packages/tables/group.py", line 235, in __init__
    super(Group, self).__init__(parentnode, name, _log)
  File "/fefs/aswg/workspace/abelardo.moralejo/miniconda3/envs/lst-dev/lib/python3.8/site-packages/tables/node.py", line 255, in __init__
    parentnode._g_refnode(self, name, validate)
  File "/fefs/aswg/workspace/abelardo.moralejo/miniconda3/envs/lst-dev/lib/python3.8/site-packages/tables/group.py", line 524, in _g_refnode
    raise NodeError(
tables.exceptions.NodeError: group ``/`` already has a child node named ``source_filenames``
```

PR #844, opened 2022-01-03
Set list of source filenames to last used

Bultako:
This PR fixes issue #839 setting the value of the `root.source_filenames.filenames`array to the last list of filenames used when merging files.

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -298,11 +298,10 @@ def auto_merge_h5files(
         source_filenames.append(str(file))
 
     with open_file(output_filename, mode="a") as file:
-        try:
-            file.root.source_filenames._f_remove(recursive=True)
-        except tables.NoSuchNodeError:
-            sources_group = file.create_group("/", "source_filenames", "List of files merged")
-            file.create_array(sources_group, "filenames", source_filenames, "List of files merged")
+        if "/source_filenames" in file.root:
+            file.remove_node("/", "source_filenames", recursive=True)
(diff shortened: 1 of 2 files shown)


lstchain/io/io.py
```diff
@@ -298,8 +298,11 @@ def auto_merge_h5files(
         source_filenames.append(str(file))
 
     with open_file(output_filename, mode="a") as file:
-        sources = file.create_group("/", "source_filenames", "List of files merged")
-        file.create_array(sources, "filenames", source_filenames, "List of files merged")
+        try:
+            file.root.source_filenames._f_remove(recursive=True)
+        except tables.NoSuchNodeError:
+            sources_group = file.create_group("/", "source_filenames", "List of files merged")
```
maxnoe, line 304: This now only creates the source filenames if it was already there.

I think you should switch to:

```
if "/source_filenames" in f.root:
    f.remove_node("/", "source_filenames", recursive=True)

sources_group = file.create_group("/", "source_filenames", "List of files merged")
 file.create_array(sources_group, "filenames", source_filenames, "List of files merged")
```
```

Bultako: @maxnoe 
I guess you mean that data is written only if the node `root.source_filenames` **does not** exist in the file, so no data is written if `root.source_filenames` is already there. You're right, I've fixed the code.

moralejo (approved):
Looks good to me. Is it tested on some newly-merged (or DL1ab'ed ) file?

PR #845, opened 2022-01-03
Relax DL1 check limit on rate of >30 pe pulses

moralejo:


lstchain/scripts/longterm_dl1_check.py
```diff
@@ -944,6 +944,12 @@ def plot(filename='longterm_dl1_check.h5', batch=False, tel_id=1):
                                )
     r30_lowlim = reference_rate_above_30 * (1 - rate_above_30_tolerance)
     r30_upplim = reference_rate_above_30 * (1 + rate_above_30_tolerance)
+    
+    # The outermost pixels, on the camera corners, just because of their position, have 
+    # lower values of this rate of >30 pe pulses. We just set a looser lower limit 
+    # requirement (we just halve the value) on pixels with id >= 1800, to avoid lots of 
+    # unnecessary warnings:
+    r30_lowlim[:, 1800:] = 0.5 * r30_lowlim[:, 1800:]
```
maxnoe, line 952: The camera geometry has a mask for the camera boundary pixels:

```
cam.get_border_pixel_mask(width=1)
```

moralejo: > The camera geometry has a mask for the camera boundary pixels:
> 
> ```
> cam.get_border_pixel_mask(width=1)
> ```

It is not exactly the edge pixels (regardless of the width value), but those close to the 6 corners, that have lower values. I now changed the number of affected pixels:

image

This is just to avoid many useless warnings (and it is not worth to make now a more complicated parametrization of the expected value for just those few pixels), or a distance-dependent mask (because the cut has to be displayed on the "value vs. pixel index" plots)

lstchain/scripts/longterm_dl1_check.py
```diff
@@ -944,6 +948,11 @@ def plot(filename='longterm_dl1_check.h5', batch=False, tel_id=1):
                                )
     r30_lowlim = reference_rate_above_30 * (1 - rate_above_30_tolerance)
     r30_upplim = reference_rate_above_30 * (1 + rate_above_30_tolerance)
+    
+    # The outermost pixels, on the camera corners, just because of their position, have 
+    # lower values of this rate of >30 pe pulses. We just set a looser lower limit for
+    # them, to avoid lots of unnecessary warnings:
+    r30_lowlim[:, pixel_index_limit:] = 0.5 * r30_lowlim[:, pixel_index_limit:]
```
rlopezcoto, line 955: is the limit is usually ~half of the inner ones or it is just a very relaxed one? I am afraid we may lose some real problems with those pixels because of this condition.

moralejo: The limit is half the value it would have for those pixels. It would still detect a pixel severely malfunctioning (and this is just one of many tests, surely a significant problem with a pixel would show up in other tests). All these limits will have to be refined with experience - and with the improvement of the telescope stability.

rlopezcoto (approved):

PR #847, opened 2022-01-06
Fix compression options and source files for dl1ab, fixes #846

### LINKED ISSUE #846: Weird change of dl1ab script output
After merging #841 I tried the dl1ab script on an MC file and noticed two strange things:
- size of output is more than twice than it was with the previous version. This is probably related to compression level, which was 5 before bt default and is now 0
- the source_filenames group is missing (note that I also merged #844)

maxnoe:


moralejo:
Thanks!!

PR #849, opened 2022-01-07
Fix indentation in add_source_files

maxnoe:
For the case of giving a path instead of an `pytables.File`, the context manager would have already closed the file


moralejo (approved):
Ah, right, thx! Will merge and test.

maxnoe:
It shouldn't affect anything in current lstchain, as we give an open file as Argument everywhere.

PR #852, opened 2022-01-12
Update requirements in setup.py and environment.yml

maxnoe:
We don't rely on pytest-runner and pytest-ordering anymore and pandas and seaborn were missing from the conda env, resulting in these packages being installed from pypi when setting up the env.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -29,5 +29,5 @@ dependencies:
   - seaborn
   - pip:
       - ctapipe_io_lst~=0.13.2
-      - ctaplot
+      - ctaplot~=0.5.5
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -1,8 +1,7 @@
 name: lst-dev
 channels:
-  - default
-  - cta-observatory
   - conda-forge
+  - default
```
rlopezcoto, line 4: why changing the order?

maxnoe: To give priority to the conda-forge packages, which tend to be more up to date and are actually open source

rlopezcoto (approved):

PR #857, opened 2022-01-13
Downgrade mpl to 3.4 for gammapy, pin pyparsing to 2.4

maxnoe:


rlopezcoto (dismissed):
`v0.8.3.post1` or `v0.8.4`?

rlopezcoto (approved):

maxnoe:
If the issue reported by @moralejo here #858 is real, let's fix that as well and make it 0.8.4

PR #860, opened 2022-01-13
Use parallel tests in CI

maxnoe:
Trying if using pytest-xdist is possible and leads to a speed up.

The current test structure of lstchain is not optimal for parallelization by loadscope (think one py file) since most of the time is passed on `test_lstchain.py`.

Running with four threads locally went from 6:50 to 5:30 minutes for all tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
         # we currently only support python 3.7 since traitlets 5 requires 3.7
         # and protozfits does not support python 3.8.
         python-version: [3.7, 3.8]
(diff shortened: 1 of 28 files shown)


maxnoe:
Test time went from 12 to 9 minutes.

We can probably do better by splitting up the giant test files ('test_lstchain', 'test_lstchain_scripts') into smaller files, but that's for another PR.

rlopezcoto:
This PR contains changes on `ImageExtractor` that are not related to parallel tests on CI, right?

rlopezcoto (approved):

PR #861, opened 2022-01-13
Add parents=True to create output dir for dl1 output

maxnoe:


maxnoe:
I tried to run it with an output directory structure that was not yet existing. I don't see why we should only support creating the last directory of a tree and not the whole tree.

rlopezcoto (approved):

moralejo (approved):

PR #863, opened 2022-01-14
Exit early if output file already exists and filter out NaturalNameWarnings

maxnoe:
The warnings were created by the astropy meta data tables, which we have no control over, so a warning is just annoying.


moralejo (dismissed):
Thanks! Looks good to me, will merge & test

moralejo (approved):

PR #870, opened 2022-01-17
Fix muon script for real data

rlopezcoto:
The muon script was not working to analyze real data, this fixes it

### DIFF
--- a/lstchain/scripts/lstchain_dl1_muon_analysis.py
+++ b/lstchain/scripts/lstchain_dl1_muon_analysis.py
@@ -133,6 +133,10 @@ def main():
             dummy_times[:] = np.nan
             parameters['dragon_time'] = dummy_times
 
+        # fill MC energies with -1 in case they do not exist (like in real data):
+        if 'mc_energy' not in parameters.keys():
+            parameters['mc_energy'] = np.ones_like(parameters['event_id']) * -1
+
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_dl1_muon_analysis.py
```diff
@@ -133,6 +133,10 @@ def main():
             dummy_times[:] = np.nan
             parameters['dragon_time'] = dummy_times
 
+        # fill MC energies with -1 in case they do not exist (like in real data):
+        if 'mc_energy' not in parameters.keys():
+            parameters['mc_energy'] = np.ones_like(parameters['event_id']) * -1
```
maxnoe, line 138: Please use nan, energy a floating point value: `np.full(len(parameters), np.nan)`

rlopezcoto: we fill -1 here:
https://github.com/cta-observatory/cta-lstchain/blob/94f3170d72caa169a0afbad7d79adcad31d97cf6/lstchain/reco/r0_to_dl1.py#L609

I recall some problems with `nan` there, but I do not remember which ones... in any case, -1 is already an invalid value for `mc_energy` and there can be no confusion.

moralejo (approved):

PR #874, opened 2022-01-18
Refactor command line arguments of scripts

maxnoe:
* Remove strtobool from distutils,
  use action='store_true', action='store_false' instead.

* Changes default of pedestal cleaning in dl1ab to true, use
  `--no-pedestal-cleaning` to disable

### DIFF
--- a/lstchain/io/data_management.py
+++ b/lstchain/io/data_management.py
@@ -3,7 +3,7 @@
 import os
 import shutil
 import sys
-from distutils.util import strtobool
+
 
 __all__ = [
     'check_and_make_dir',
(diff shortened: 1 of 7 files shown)


rlopezcoto:
apart from tests not passing at the moment:

> * Remove strtobool from distutils,
>   use action='store_true', action='store_false' instead.

why did we need to remove `strtobool` to implement a custom solution?

> * Changes default of pedestal cleaning in dl1ab to true, use
>   `--no-pedestal-cleaning` to disable

also, the change of the default of the pedestal cleaning may affect data processing, is this ok for LSTOSA @morcuended ?

maxnoe:
> also, the change of the default of the pedestal cleaning may affect data processing, is this ok for LSTOSA @morcuended ?

This was discussed during the analysis school, and I had the impression that this was uncontroversial

maxnoe:
> why did we need to remove strtobool to implement a custom solution?

I don't understand the question.

rlopezcoto:
> The reason to remove `strtobool` is that `distutils` is deprecated and will be removed.

ok, thanks for clarifying!

maxnoe:
@morcuended @vuillaut @garciagenrique  This changes a couple of command line flags that would need adaption in anything that calls these scripts, e.g. osa and lstmcpipe

morcuended:
Noted

rlopezcoto (approved):

PR #875, opened 2022-01-19
DL3 tools for source-dependent analysis

SeiyaNozaki:
This PR is to adapt DL3 tools to source-dependent analysis.
The feature is:
- alpha cut is used for IRF
- for DL3 tools, alpha cut is already applied at this stage (theta cut is applied in gammapy analysis for source-indep analysis)
- for surviving events, an assumed source position (both on and off) is set as a reconstructed position to adapt the gammapy analysis

This is a draft PR since I saw DL3-related PRs are prepared and this change would be better to implement it after other PRs are merged.
Anyway any comments are welcome.


lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -151,14 +164,48 @@ def setup(self):
...
+                time = data_temp['dragon_time']
+                obstime = Time(time, scale='utc', format='unix')
+                expected_src_x = data_temp['expected_src_x'] * u.m
+                expected_src_y = data_temp['expected_src_y'] * u.m
+                focal = 28 * u.m
+                pointing_alt = data_temp['pointing_alt']
+                pointing_az  = data_temp['pointing_az']
+                expected_src_altaz = camera_to_altaz(expected_src_x, expected_src_y, focal, pointing_alt, pointing_az, obstime=obstime)
+                data_temp["reco_alt"] = expected_src_altaz.alt
+                data_temp["reco_az"]  = expected_src_altaz.az
```
chaimain, line 203: Probably better to create a separate function like add_icrs_position_params in lstchain/irf/hdu_table.py, so as to keep the Tool simple

chaimain:
Apart from updating test_tools for these changes, also update the docs/example/[irf, dl3]_tool_config.json

chaimain:
I think, #709 would be merged soon, and then you can add energy-dependent alpha cuts as well (saving the direction cuts to RAD_MAX HDU)

maxnoe:
@chaimain No. `RAD_MAX` is for theta squared cuts. There is no support at all in the GADF data model for applying alpha cuts and doing source dependent analysis. 

Please don't fill invalid data into files that claim to be GADF compliant.

chaimain:
> @chaimain No. `RAD_MAX` is for theta squared cuts. There is no support at all in the GADF data model for applying alpha cuts and doing source dependent analysis.
> 
> Please don't fill invalid data into files that claim to be GADF compliant.

Yes, I stand corrected.

docs/examples/dl3_tool_config_srcdep.json
```diff
@@ -0,0 +1,18 @@
+{
```
chaimain, line 1: You do not need to add separate config files to just add alpha_cut. you can just add to the existing ones, as the Tool will use the values as required by the particular functions or usage.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -173,31 +190,74 @@ def setup(self):
 
     def start(self):
 
-        self.data = read_data_dl2_to_QTable(str(self.input_dl2))
-        self.effective_time, self.elapsed_time = get_effective_time(self.data)
-        self.run_number = run_info_from_filename(self.input_dl2)[1]
+        if not self.source_dep:
+            self.data = read_data_dl2_to_QTable(str(self.input_dl2))
```
maxnoe, line 194: If `read_data_dl2_to_QTable` does not support giving a `Path`, we should fix that function, not convert to string here.

lstchain/tools/lstchain_create_dl3_file.py
```diff
@@ -173,31 +190,74 @@ def setup(self):
...
+            self.effective_time, self.elapsed_time = get_effective_time(self.data)
+            self.run_number = run_info_from_filename(self.input_dl2)[1]
```
maxnoe, line 196: You have the same code later in the loop. Just move outside the if here and remove from the loop code.

maxnoe (changes requested):
Two minor comments and one question:

Can we separate the source dependent and independent case into two functions returning the events table?

lstchain/tools/lstchain_create_dl3_file.py
chaimain: the `read_data_dl2_to_QTable` and `read_mc_dl2_to_QTable` uses `pandas.read_hdf` to read the files. It supports string or `os.PathLike` arguments for filename. So, either we convert all arguments to string within this function, or make it `os.PathLike`, by `pathlib.PurePath` right?

maxnoe: The only reason to add `str()` here is that `read_data_dl2_to_QTable` does not support passing a `Path`. If it does support that, just remove `str`. If it doesn't support it, make it support it.

lstchain/tools/lstchain_create_dl3_file.py
SeiyaNozaki: thanks, done

lstchain/tools/lstchain_create_dl3_file.py
SeiyaNozaki: it seems only `get_dataset_keys` , which is used to check if source-dependent parameters are included or not, does not support `Path` in `read_data_dl2_to_QTable` and `read_mc_dl2_to_QTable`. I added `str()` for `get_dataset_keys` in these functions

maxnoe: perfect, thanks!

maxnoe: Wait, I am confused, `get_dataset_keys` should already support `Path`.

Current master:

```
In [1]: from lstchain.io.io import get_dataset_keys

In [2]: from pathlib import Path

In [3]: path = Path('../data/gamma_20deg_0deg_run101___cta-prod5-paranal_desert-2147m-Paranal-dark_merged.DL1.h5')

In [4]: get_dataset_keys(path)
Out[4]: 
['/simulation/service/shower_distribution',
 '/simulation/event/subarray/shower',
 '/simulation/event/telescope/images/tel_001',
 '/simulation/event/telescope/images/tel_002',
 '/simulation/event/telescope/images/tel_003',
 '/simulation/event/telescope/images/tel_004',
```

What is the error you are getting?

SeiyaNozaki: true... it seems I got some errors by what I introduced to check step by step...
so no need to use str anywhere!

lstchain/tools/tests/test_tools.py
```diff
@@ -85,6 +85,36 @@ def test_create_irf_full_enclosure_with_config(temp_dir_observed_files, simulate
     )
 
 
+def test_create_irf_point_like_srcdep(
+    temp_dir_observed_srcdep_files, simulated_srcdep_dl2_file):
```
chaimain, line 89: Just a matter of completion, place the closing bracket in the next line

lstchain/tools/tests/test_tools.py
```diff
@@ -212,6 +242,38 @@ def test_create_dl3_with_config(temp_dir_observed_files, observed_dl2_file):
...
+                "--source-dec=22.01deg",
+                "--source-dep",
+                "--overwrite",
+            ],
+            cwd=temp_dir_observed_srcdep_files,
+        )
+        == 0
+    )
+
+    hdulist = fits.open(temp_dir_observed_srcdep_files / "dl3_LST-1.Run02008.0000.fits.gz")
```
chaimain, line 270: Use from `lstchain.paths import dl2_to_dl3_filename` to get the name of the DL3 file, for more uniform code

chaimain (approved):

maxnoe (approved):

PR #876, opened 2022-01-19
Correct sample time estimation

FrancaCassol:


lstchain/calib/camera/flatfield.py
```diff
@@ -278,7 +278,7 @@ def calculate_time_results(
                                              pixel_median > self.time_cut_outliers[1])
 
         return {
-            'sample_time': (trigger_time - time_start).value / 2 *u.s,
+            'sample_time': (trigger_time + time_start).value / 2 *u.s,
```
maxnoe, line 281: I don't understand what this is supposed to do. Why do you take the mean between two astropy times like this?
Or are the `..._time` here not astropy times? If they are not, what are they?

If you want to have the midpoint, it would correctly be something like this:

```
start_time + 0.5 * (trigger_time - start_time)
```

to get a proper astropy time back and not a value in seconds:

```
In [1]: from astropy.time import Time

In [2]: t0 = Time('2020-01-01T20:00')

In [3]: t1 = Time('2020-01-01T21:00')

In [4]: t0 + 0.5 * (t1 - t0)
Out[4]: <Time object: scale='utc' format='isot' value=2020-01-01T20:30:00.000>
```

maxnoe: See also the error in the CI you are getting:

> E           astropy.time.core.OperandTypeError: Unsupported operand type(s) for +: 'Time' and 'Time'

maxnoe: The sum of two points in time is not something well-defined. Which is why astropy does not allow it.

The difference is a `TimeDelta` and you can add a `TimeDelta` to a `TimeDelta` or a `TimeDelta` to a `Time`

lstchain/calib/camera/flatfield.py
```diff
@@ -278,7 +278,7 @@ def calculate_time_results(
                                              pixel_median > self.time_cut_outliers[1])
 
         return {
-            'sample_time': (trigger_time + time_start).value / 2 *u.s,
+            'sample_time': (time_start +(trigger_time - time_start) / 2).value *u.s,
```
maxnoe, line 281: Please do not use `.value` here. What do you want to store here? If you want the time, just leave it as a time and the ctapipe table writer will convert it when writing / reading.

If you need a specific format for some reason, access that (e.g. unix or unix_tai) but not `.value`.

rlopezcoto:
there are some tests still not passing @FrancaCassol, it seems that the `to_value` is called for an array instead of an astropy quantity

maxnoe (dismissed):

maxnoe:
@FrancaCassol we need to merge master again to run the tests with ctapipe 0.12

maxnoe (approved):

PR #879, opened 2022-01-21
Fix rad max format

maxnoe:


rlopezcoto (approved):

chaimain:
With the restructuring of the tests in test_tools, it would be good to also do the same for test_create_dl3. Maybe I can add to this PR itself?

chaimain (approved):

maxnoe:
@chaimain We removed all the pytest ordering stuff. Why did you add it back? And why does runing the different IRF tests depent on the other?

maxnoe:
Edit: it seems I missed a couple of locations where pytest-ordering was still supposed to be used. 

I will fix those in a different PR. I'll revert your last commit.

chaimain:
Ah! I think I didn't check the new change.

> Edit: no point in reverting, let's just fix the other ordering remnants in another PR

Ok.

PR #882, opened 2022-01-21
Remove pytest ordering remnants

maxnoe:


lstchain/irf/hdu_table.py
```diff
@@ -169,7 +169,7 @@ def create_hdu_index_hdu(
             "HDU_TYPE": "events",
             "HDU_CLASS": "events",
             "FILE_DIR": str(os.path.relpath(fits_dir, hdu_index_file.parent)),
-            "FILE_NAME": file,
+            "FILE_NAME": str(file),
```
rlopezcoto, line 172: why is it necessary to explicitly transform the input into a string?

maxnoe: Because it is written into a fits file and having it as `Path` created an error as the dtype is `object`, which astropy cannot convert to something that can be stored as fits.

maxnoe: We want to support giving the files as `Path` but for storing into fits, it must be converted to `str` here.

rlopezcoto (approved):

PR #885, opened 2022-01-24
Remove duplicated image parametrization and fix #580

### LINKED ISSUE #580: Concentration features are calculated on all pixels, not just the ones that survive cleaning
I don't know if it's worth to change this before switching to ctapipe-stage1.

maxnoe:


lstchain/reco/r0_to_dl1.py
```diff
@@ -121,12 +121,52 @@ def setup_writer(writer, subarray, is_simulation):
...
+    hillas = hillas_parameters(geom_selected, image_selectecd)
+
+    # Fill container
+    dl1_container.fill_hillas(hillas)
+
+    # convert ctapipe's width and length (in m) to deg:
+
+    for key  in ['width', 'width_uncertainty', 'length', 'length_uncertainty']:
+        value = getattr(dl1_container, key)
+        setattr(dl1_container, key, _camera_distance_to_angle(value, focal_length))
```
vuillaut, line 149: shouldn't this conversion be done in ctapipe directly?
Having them in deg rather than meters is really interesting when there are several telescopes types.

vuillaut: Ha, I saw https://github.com/cta-observatory/cta-lstchain/issues/886 afterwards
Does it mean these lines will be removed once moving to ctapipe 0.12?

maxnoe: See comment above in the function doing that ;)

vuillaut (approved):
The comment about width/length units is not blocking

PR #888, opened 2022-01-25
Add source-dependent parameters during RF training or dl1_to_dl2 process

SeiyaNozaki:
continued from #704 

This PR makes it possible to add source-dependent parameters during the RF training or DL1_to_dl2 process, so no need to add these parameters by another script explicitly.

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -55,7 +55,7 @@ __all__ = [
     'extract_simulation_nsb',
     'extract_observation_time',
     'get_dataset_keys',
-    'get_srcdep_index_keys',
+    'get_srcdep_wobble_angles',
     'get_srcdep_params',
     'get_stacked_table',
     'global_metadata',
@@ -1262,9 +1262,9 @@ def merge_dl2_runs(data_tag, runs, columns_to_read=None, n_process=4):
     return observation_time, df
 
 
-def get_srcdep_index_keys(filename):
+def get_srcdep_wobble_angles(filename):
     """
-    get index column name of source-dependent multi index columns
+    get wobble angles of source-dependent multi index columns
 
     Parameters
     ----------
@@ -1272,7 +1272,7 @@ def get_srcdep_index_keys(filename):
 
     Returns
     -------
-    source-dependent index names
+    wobble angles for source-dependent parameters
     """
     dataset_keys = get_dataset_keys(filename)
 
@@ -1282,6 +1282,9 @@ def get_srcdep_index_keys(filename):
     elif dl1_params_src_dep_lstcam_key in dataset_keys:
         data = pd.read_hdf(filename, key=dl1_params_src_dep_lstcam_key)
 
+    else:
+        raise IOError('File does not contain source-dependent parameters')
+
     if not isinstance(data.columns, pd.MultiIndex):
         data.columns = pd.MultiIndex.from_tuples(
             [tuple(col[1:-1].replace('\'', '').replace(' ', '').split(",")) for col in data.columns])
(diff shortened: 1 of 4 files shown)


lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -141,13 +140,21 @@ def main():
...
+
+        # if not, source-dependent parameters are added now
+        else:
+            data_srcdep = pd.concat(dl1_to_dl2.get_source_dependent_parameters(
+                data, config, focal_length=focal_length), axis=1)
+
         dl2_srcdep_dict = {}
         srcindep_keys = data.keys()
-        srcdep_index_keys = get_srcdep_index_keys(args.input_file)
+        srcdep_index_keys = data_srcdep.columns.levels[0]
```
rlopezcoto, line 154: why are these read in a different way as the source independent ones?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -20,10 +20,11 @@
 
 from . import disp
 from . import utils
-from ..io import standard_config, replace_config
+from ..io import standard_config, replace_config, get_dataset_keys, get_srcdep_params
```
rlopezcoto, line 23: I would break this line

lstchain/io/io.py
```diff
@@ -1289,14 +1289,15 @@ def get_srcdep_index_keys(filename):
...
-def get_srcdep_params(filename, key):
+def get_srcdep_params(filename, key=None):
     """
     get srcdep parameter data frame
 
     Parameters
     ----------
     filename: str - path to the HDF5 file
     key: `str` - multi index key corresponding to an expected source position (e.g. 'on', 'off_180')
+    If it is not specidied, all keys are loaded
```
rlopezcoto, line 1300: typo: specified

lstchain/io/io.py
```diff
@@ -1314,7 +1315,10 @@ def get_srcdep_params(filename, key):
         data.columns = pd.MultiIndex.from_tuples(
             [tuple(col[1:-1].replace('\'', '').replace(' ', '').split(",")) for col in data.columns])
 
-    return data[key]
+    if key is not None:
```
rlopezcoto, line 1318: if `dl1_params_src_dep_lstcam_key` and `dl2_params_src_dep_lstcam_key` are not in dataset_keys, the `data` variable does not get initialize and continuing will raise an error, maybe you could add:

```python3
else:
    raise IOError('File does not contain source-dependent parameters')
```

lstchain/scripts/lstchain_dl1_to_dl2.py
SeiyaNozaki: it was a bit confusing indeed...
On this line, assumed wobble angles, not parameter keys,  are loaded from source-dep data frame.
Those angles are saved as level 0 of the data frame, and each parameter key is saved as level 1 of the data.
I modified the name of this parameter (srcdep_index_keys -> srcdep_wobble_angles).

lstchain/reco/dl1_to_dl2.py

lstchain/io/io.py
SeiyaNozaki: thanks for spotting, corrected.

lstchain/io/io.py
SeiyaNozaki: ok, added

lstchain/scripts/lstchain_dl1_to_dl2.py
rlopezcoto: but is this really wobble angles or off positions? we could take data with 2 or 4 different wobble angles and take calculate the off src-dependent parameter from 1, 3, 5 or any number of OFF positions

SeiyaNozaki: ok, you're right. It should be independent of wobble angle (observation conditions).
It would be better to use 'srcdep_assumed_positions' or something?
This column contains 'on', 'off_090', 'off_180', 'off_270' for example in case that n_off=3 is assumed.

rlopezcoto: yes, `srcdep_assumed_positions` sounds good

SeiyaNozaki: ok, changed.

rlopezcoto (approved):

PR #892, opened 2022-01-28
Correct config of interleaved event processing

FrancaCassol:
The config was not successfully read but the Components, therefore the interleaved events were processed with default trailet values (e.g. charge integration over only 7 samples, which explains the difference on the number of pe with respect to the beginning of night calibration). Also the f-factor systematic correction trailet was not included in the config file


FrancaCassol:
see #893

PR #895, opened 2022-01-28
Add DL3+ notebooks

chaimain:
Adding 5 notebooks:
- MC DL2 to IRF (source-independent, fixed global and energy-dependent gammaness and theta cuts) 
- DL2 to DL3 (source-independent, fixed global and energy-dependent gammaness cuts)
- DL3 to DL4 (using gammapy v0.19 and saving DL4 as OGIP files)
- plot LC from DL4 (OGIP files)
- plot SED from DL4 (OGIP files)

The energy-dependent cuts procedures follows #709


rlopezcoto:
After v0.9 we should try to make all notebooks work with the current version of lstchain

lstchain/scripts/lstchain_merge_run_summaries.py
```diff
@@ -0,0 +1,140 @@
+from pathlib import Path
```
maxnoe, line 1: This is a script from another PR and does not belong here

maxnoe (changes requested):
Why are there notebooks for things that are supposed to be done with the tools, e.g. `create_dl3_events_file` and `create_irfs`?

chaimain:
> Why are there notebooks for things that are supposed to be done with the tools, e.g. `create_dl3_events_file` and `create_irfs`?

To just show the workflow of the Tools.

maxnoe:
@chaimain But the notebooks duplicate code from the tools.

chaimain:
> @chaimain But the notebooks duplicate code from the tools.
Yes. I do not understand, how I can explain the workflow of the Tools, without doing so in a jupyter notebook. Can you elaborate more on this?

maxnoe:
@chaimain You show how you run the tools and explain what they do shortly in prose? Have a look in the inputs/outputs?

We really should not duplicate the code in notebooks that is supposed to be run via command line tools.

People who really want to understand the code need to look at the code, we shouldn't repeat it in a notebook

chaimain:
> @chaimain You show how you run the tools and explain what they do shortly in prose? Have a look in the inputs/outputs?
> 
> We really should not duplicate the code in notebooks that is supposed to be run via command line tools.
> 
> People who really want to understand the code need to look at the code, we shouldn't repeat it in a notebook


Ok. I understand it now. Thanks!
I will fix the notebooks accordingly.

chaimain:
@maxnoe should the `--help-all` section be run in the notebook?

rlopezcoto:
as commented at the beginning of the discussion, after the new release we should update all notebooks to make them work with the current version of lstchain and some of them like these ones should not contain the information that is better explained elsewhere (as in gammapy notebooks)

rlopezcoto (approved):

PR #896, opened 2022-01-31
Do not redo link to pro directory if already correct

FrancaCassol:


maxnoe (changes requested):
Could you maybe put that code into a function

```
def create_pro_link(output_dir):
```

It is duplicated 4 times and you now had to make a change at 4 locations in exactly the same way.

FrancaCassol:
Hi @maxnoe,

where do you want to put this function?

maxnoe:
@FrancaCassol maybe create a new module `lstchain/onsite.py`? 

I wanted to also refactor the rest of the duplicated onsite scripts, e.g. by introducing functions like:

```
def find_drs4_baseline_run(run_id, base):
    ...
```

FrancaCassol:
I see, do we want to do it now or we better keep it for later?

maxnoe:
We can also do it in one go later...

maxnoe (approved):

rlopezcoto:
wait, I was modifying it after talking to @FrancaCassol

FrancaCassol:
If there are other changes I think it is better to do all later.
I would merge it as it is now

rlopezcoto:
>@FrancaCassol maybe create a new module lstchain/onsite.py?

I created instead a `path.py` module inside `scripts/onsite`, does that sound good?

FrancaCassol:
I liked better the onsite.py idea that is more general.

lstchain/scripts/onsite/paths.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+def create_pro_symlink(output_dir, prod_id):
+    pro = "pro"
+    pro_dir = f"{output_dir}/../{pro}"
+    if os.path.exists(pro_dir) and os.readlink(pro_dir) is not output_dir:
+        os.remove(pro_dir)
+        os.symlink(prod_id, pro_dir)
+    print("\n--> Use symbolic link pro")
```
maxnoe, line 9: The print shouldn't be here

lstchain/scripts/onsite/paths.py
```diff
@@ -0,0 +1,9 @@
+import os
```
maxnoe, line 1: I think this should not go into a `scripts` subfolder, maybe just `lstchain.onsite`?

lstchain/scripts/onsite/paths.py
rlopezcoto: removed

lstchain/scripts/onsite/paths.py
rlopezcoto: ok, I moved the file and put it where you initially suggested

lstchain/scripts/onsite/onsite_create_drs4_pedestal_file.py
```diff
@@ -80,10 +80,10 @@ def main():
 
     # update the default production directory
     if pro_symlink:
-        pro = "pro"
+        #pro = "pro"
```
maxnoe, line 83: Please just remove instead of commenting out. Since the drs4 pedestal creation only needs raw data as input and not other calibration products, it is just not needed here.

rlopezcoto: done

maxnoe (approved):

PR #897, opened 2022-02-01
Add pytest for source-dependent analysis

SeiyaNozaki:
This PR implemented pytest codes for source-dependent analysis (MC and real data).
I'd like to implement it before #875 , then I can also add additional pytest for source-dependent DL3 production.

### DIFF
--- a/lstchain/io/tests/test_config.py
+++ b/lstchain/io/tests/test_config.py
@@ -4,6 +4,9 @@ from lstchain.io.config import get_cleaning_parameters
 def test_get_standard_config():
     config.get_standard_config()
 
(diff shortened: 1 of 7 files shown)


lstchain/io/tests/test_config.py
```diff
@@ -4,6 +4,9 @@
 def test_get_standard_config():
     config.get_standard_config()
 
+def test_get_srcdep_config():
+    config.get_srcdep_config()
```
maxnoe, line 8: Please add a little check to make sure what this function returns is actually what you expect.

maxnoe (approved):

PR #902, opened 2022-02-03
Improved checks of UCTS & TIB trigger tags

moralejo:


lstchain/scripts/lstchain_longterm_dl1_check.py
```diff
@@ -1748,7 +1817,13 @@ def trigtag_mismatches(table, tag_value):
...
                 # number of events in the subrun which have that
                 # trigger type.
-                if trigtype[0] == 0:
+
+                # skip the 'unknown' cases, we are counting here just the
+                # wrong tags:
+                if trigtype[0] == 0: # TriggerBits.UNKNOWN:
+                    continue
+
+                if trigtype[1] == 0:
```
rlopezcoto, line 1826: what would be the value of `trigtype[0]` in this case (I assume 1) and why?

moralejo: trigtype[0] is the trigger tag value, and trigtype[1] the number of events in the subrun that have that value. There are always 10 "trigtype" per subrun, but if in the subrun there are e.g. just  3 different values of the trigger value, only the first three trigtype are used. From the fourth element all the trigtype's will be [0, 0] :
```
[0, 2]
[1, 2500]
[32, 45000]
[0, 0]
[0, 0] 
[0, 0]
[0, 0] 
[0, 0]
[0, 0] 
[0, 0] 

```

0 is also the code for UNKNOWN trigger bits. So in the above example there were 2 events with UNKNOWN,  2500 events with 1  and 45000 events with 32.

As soon as we find a trigtype[1] == 0  it means that there are no more values of the trigger tag to be processed, for this subrun, so we move on to the next.

rlopezcoto (approved):

PR #904, opened 2022-02-03
Lhfit implementation

gabemery:
First implementation of the likelihood reconstruction method in the main lstchain repository.

Main additions : 
- Class and function to perform the likelihood analysis on R1 waveforms in lstchain/reco/reconstructor.py
- A dedicated get_dl1_lhfit function in lstchain/reco/r0_to_dl1.py
- A math module containing functions used for the reconstruction : lstchain/image/pdf.py

Additionally, the r0_to_dl1 function and LST DL1 container are extended to use the method and a dedicated config file is added.

Edit : 
lstchain/image/pdf.py is not included anymore.
Current version uses numba CC with ahead of time compilation to improve the code speed.
The code was vastly re-organised since the start of the PR (using traitlet, one fitter object creation,..)


lstchain/image/pdf.py
```diff
@@ -0,0 +1,63 @@
...
+        Width of the normal distribution
+
+    Returns
+    -------
+    log_pdf: float or array-like
+        Log of the evaluation of the normal law at x
+
+    """
+    pi = np.pi
+    assert pi
```
maxnoe, line 25: Okay, story time: What terrible thing happened that you added a check that `numpy.pi` is not 0?

gabemery: I implemented the use of numexpr to improve the code speed by around 10% but the variables used in numexpr.evaluate are not considered as used by pyflakes. So the CI fails due to pi here and log_gauss (in reconstructor.py l637).
So this is kind of a dirty hack that I would be happy to remove but didn't find a solution yet (also tried adding inline comment NOQA but it didn't work)

maxnoe: AAAAH, ok. Makes sense, kind of. A better solution would be to pass the `local_dict` option, which is also safer:

```
ne.evaluate("-(x - mean) ** 2 / (2 * sigma ** 2) - log((sqrt(2 * pi) * sigma))", local_dict=dict(x=x, mean=mean, sigma=sigma, pi=np.pi))
```

NOQA only works with `flake8` I think, not pyflakes.

gabemery: Thanks I will implement this.

lstchain/io/lstcontainers.py
```diff
@@ -123,6 +123,42 @@ class DL1ParametersContainer(Container):
     trigger_time = Field(None, "trigger time")
     event_type = Field(None, "event type")
 
+    # Parameters obtained with the likelihood reconstruction method
+    lhfit_call_status = Field(-10, "Status of the processing of the event "
```
maxnoe, line 127: Please don't add further fields to this already monstrous container. Make a new container for the likelihood output.

lstchain/reco/r0_to_dl1.py
```diff
@@ -167,6 +168,7 @@ def get_dl1(
     telescope_id,
     dl1_container=None,
     custom_config={},
+    lhfit_args={}
```
maxnoe, line 171: Don't use mutable default arguments, use None.

lstchain/reco/r0_to_dl1.py
```diff
@@ -257,6 +261,45 @@ def get_dl1(
                 dl1_container=dl1_container,
             )
 
+            if 'lh_fit_config' in config.keys():
```
maxnoe, line 264: Break this out into a function, please don't add more to the spaghetti here.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,944 @@
...
+
+
+class DL0Fitter(ABC):
+    """
+        Base class for the extraction of DL1 parameters from R1 events
+        using a log likelihood minimisation method.
+
+    """
+
+    def __init__(self, waveform, error, sigma_s, geometry, dt, time_shift, n_samples,
```
maxnoe, line 27: Same comment as above. A fitter should be instantiated once per run in the script and only take options in the `__init__` method. Then it should take an event in `__call__` taking all options that change per event from there.

See e.g. how the HillasReconstructor in ctapipe works:

https://github.com/cta-observatory/ctapipe/blob/bc78946d752680189ffc49c1f848b6765b8dc6c4/ctapipe/reco/hillas_reconstructor.py#L153-L180

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,944 @@
...
+                                   psi=psi,
+                                   rl=rl)
+        mu = ne.evaluate("exp(log_mu)", local_dict=dict(log_mu=log_mu))
+        mu[mu <= 0] = 1e-320
+
+        # We reduce the sum by limiting to the poisson term contributing for
+        # more than 10^-6. The limits are approximated by 2 broken linear
+        # function obtained for 0 crosstalk.
+        # The choice of kmin and kmax is currently not done on a pixel basis
+        mask_LL = (mu <= self.n_peaks/1.096 - 47.8)
```
maxnoe, line 587: Please don't use magic numbers in the code (also below)

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,944 @@
...
+            Compute quantities used at each iteration of the fitting procedure.
+        """
+        self.n_peaks = n_peaks
+        photoelectron_peak = np.arange(n_peaks, dtype=int)
+        photoelectron_peak[0] = 1
+        log_factorial = np.log(photoelectron_peak)
+        log_factorial = np.cumsum(log_factorial)
+        self.log_factorial = log_factorial
+
+    def log_pdf(self, charge, t_cm, x_cm, y_cm,
```
maxnoe, line 532: Looking at this, I am not surprised this is very slow. This function needs to be compiled. Using numba would probably be easiest.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,944 @@
...
+            else:
+                log_pixel_pdf_HL = log_gaussian(signal[~mask_LL], mu_hat, sigma_hat)
+            n_points_HL = np.sum(weight[~mask_LL])
+        else:
+            log_pixel_pdf_HL, n_points_HL = np.asarray([0]), 0
+
+        log_poisson = np.expand_dims(log_poisson.T, axis=1)
+        log_pixel_pdf_LL = ne.evaluate("log_poisson + log_gauss",
+                                       local_dict=dict(log_poisson=log_poisson, log_gauss=log_gauss))
+        if self.use_weight:
```
maxnoe, line 661: You have many `if self.use_weight` statements here. It looks like it would be much better to just have two functions for the weighted and unweighted cases.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,944 @@
...
+    image : array_like
+        Charge in each pixel
+
+    Returns
+    -------
+    (x_cm, y_cm): `astropy.units.Quantity`
+
+    """
+    unit = geom.pix_x.unit
+    pix_x = u.Quantity(np.asanyarray(geom.pix_x, dtype=np.float64)).value
```
maxnoe, line 938: `geom.pix_x.to_value(unit)`

maxnoe (changes requested):
I think for this to be useful, it needs to be cleaned up and the likelihood needs to be compiled, probably using numba.

The fitter class should follow the ctapipe api for reconstructors, i.e. taking the `subarray` and configuration options only in `__init__` and an `event` in `__call__` that performs the actual computation returning a container containing the fit results:

Usage in r0_to_dl1 should be like this:
```
# create once from subarray and config
fitter = LikelihoodFitter(source.subarray, config=config)

# call for each event / tel_id returning a container.
for event in source:
    ....
    for tel_id in event.d1.tel:
        hillas_likelihood =  fitter(event, tel_id)
```


I also don't think it is a good idea to have that amount of plotting code bundled with the implementation of the calculation.

maxnoe:
This library providing numba accelerated probability densities might be relevant here:

https://github.com/HDembinski/numba-stats

lstchain/reco/r0_to_dl1.py
gabemery: @maxnoe Should I also solve this for "custom_config"? From what I see it implies adding a check for None in at least "replace_config" or a if custom_config is None: custom_config={} at the start of the get_dl1 function.

lstchain/reco/r0_to_dl1.py
```diff
@@ -267,6 +310,167 @@ def get_dl1(
...
+    Returns
+    -------
+    DL1ParametersContainer
+    """
+    lh_fit_config = custom_config['lh_fit_config']
+
+    waveform = calibrated_event.r1.tel[telescope_id].waveform
+
+    time_shift = None
+    if dl1_container.mc_type == -9999:  # Find time shift correction for data
```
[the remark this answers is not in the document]
gabemery: This is the default value, which is not updated for real observation data file. In case of MC is is dependent of the particle type. Previously I passed the is_simu flag as an additional parameter.
Switching to :
if calibrated_event.simulation is None:
as it may be easier to understand.

lstchain/reco/r0_to_dl1.py
gabemery: Separated the likelihood reconstruction from get_dl1()
Now the call to get_dl1_lh_fit is in a new function apply_lh_fit itself used in r0_to_dl1 after the hillas parameter extraction.

lstchain/reco/r0_to_dl1.py
```diff
@@ -316,6 +319,7 @@ def get_dl1_lh_fit(
     geometry,
     telescope_id,
     dl1_container,
+    lhfit_container,
```
maxnoe, line 343: How can the lhfit container be an input here?

gabemery:
Large code refactoring performed. Only one class is introduced in reconstructor.py which inherits the ctapipe.core.Component class and makes use of traitlets. The fitter is created once in r0_to_dl1() and called for each events. The fitter's predict() method now returns a dedicated DL1LikelihoodParametersContainer.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+    sigma_s = Float(1, help='', allow_none=False).tag(config=True)
+    crosstalk = Float(0, help='', allow_none=False).tag(config=True)
+    sigma_space = Float(4, help='', allow_none=False).tag(config=True)
+    sigma_time = Float(3, help='', allow_none=False).tag(config=True)
+    time_before_shower = Float(10, help='', allow_none=False).tag(config=True)
+    time_after_shower = Float(20, help='', allow_none=False).tag(config=True)
+    use_weight = Bool(False, help='', allow_none=False).tag(config=True)
+    no_asymmetry = Bool(False, help='', allow_none=False).tag(config=True)
+    use_interleaved = Path(None, help='', allow_none=True).tag(config=True)
+    telescope_id = Int(1, help='', allow_none=False).tag(config=True)
```
maxnoe, line 33: Why do you require a `telescope_id` here?

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+        self.error_parameters = None
+        self.fcn = None
+
+        poisson_peaks = np.arange(self.n_peaks, dtype=int)
+        poisson_peaks[0] = 1
+        log_factorial = np.log(poisson_peaks)
+        self.log_factorial = np.cumsum(log_factorial)
+
+    def __call__(self, event, dl1_container):
+        self.event_id = event.index.event_id
```
maxnoe, line 77: Why are these stored on the class? After `__call__` is over, they should not be needed anymore.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+        start_x_cm, start_y_cm = init_centroid(dl1_container,
+                                               self.geometry[hillas_signal_pixels],
+                                               self.image[hillas_signal_pixels],
+                                               self.no_asymmetry
+                                               )
+
+        waveform = event.r1.tel[self.telescope_id].waveform
+
+        time_shift = None
+        if event.simulation is None:  # Find time shift correction for data
```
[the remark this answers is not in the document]
maxnoe: I think you should do this for all events

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+
+        self.n_pixels, self.n_samples = waveform.shape
+        self.times = np.arange(0, self.n_samples) * self.dt
+        selected_gains = event.r1.tel[self.telescope_id].selected_gain_channel
+        self.is_high_gain = (selected_gains == 0)
+
+        v = dl1_container.time_gradient
+        psi = dl1_container.psi.to(u.rad).value
+        if v < 0:
+            if psi >= 0:
```
maxnoe, line 103: Why are you changing psi here?

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+                                 't_cm': dl1_container.intercept - self.template_time_of_max,
+                                 'v': np.abs(v),
+                                 'psi': psi,
+                                 'length': start_length,
+                                 'wl': dl1_container.wl,
+                                 'rl': 0.0
+                                 }
+
+        if self.start_parameters['length'] <= 0.02:
+            self.start_parameters['length'] = 0.02
```
maxnoe, line 121: You can just do `start_length = max(..., 0.02)` above to simplify this code.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+        if self.start_parameters['wl'] == 0:
+            self.start_parameters['wl'] = 0.01
+        if np.isnan(self.start_parameters['t_cm']):
+            self.start_parameters['t_cm'] = 0.
+        if np.isnan(self.start_parameters['v']):
+            self.start_parameters['v'] = 40
+
+        t_max = self.n_samples * 1
```
maxnoe, line 129: * 1

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+
+        self.times = (np.arange(0, self.n_samples) * self.dt)[self.mask_time]
+        if time_shift is None:
+            self.time_shift = spatial_zeros
+        else:
+            self.time_shift = time_shift[self.mask_pixel]
+
+        self.p_x = self.pix_x[self.mask_pixel]
+        self.p_y = self.pix_y[self.mask_pixel]
+        self.pix_area = self.geometry.pix_area.to(u.m ** 2).value[self.mask_pixel]
```
maxnoe, line 174: Apply mask first, then `to_value`

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+            self.error = np.ones(self.data.shape) * std
+        else:
+            std = np.std(self.data[~self.mask_pixel])
+            self.error[self.error <= std / 2] = std
+            self.error = self.error[..., None] * np.ones(self.data.shape)
+
+        self.data = np.delete(self.data, filter_pixels, axis=0)
+        self.data = np.delete(self.data, filter_times, axis=1)
+        self.error = np.delete(self.error, filter_pixels, axis=0)
+        self.error = np.delete(self.error, filter_times, axis=1)
```
maxnoe, line 193: `__call__` does not seem to perform the fit?

lstchain/reco/r0_to_dl1.py
```diff
@@ -267,6 +267,45 @@ def get_dl1(
...
+
+    """
+    # Applied to all simulated event and all physic events
+    if dl1_container.mc_type > -1000 or (event.trigger.event_type == EventType.SUBARRAY):
+        # Don't fit if the cleaning used in the seed parametrisation didn't select any pixels
+        if dl1_container.n_pixels <= 0:
+            lhfit_container = DL1LikelihoodParametersContainer(lhfit_call_status=0)
+        else:
+            try:
+                fitter(event=event, dl1_container=dl1_container)
```
maxnoe, line 297: This should just be `lhfit_container = fitter(event=event, dl1_container=dl1_container)`. I.e. `__call__` should perform the fit and return the result.

lstchain/reco/reconstructor.py
gabemery: It is used to select a couple of information (select geometry for the pixel positions, the sampling frequency). Later when going to stereo, It may be useful to select a subarray for a full run. And a second event wise list of id would also be needed. 
I assumed it was faster to save, e.g. the pixels position before masking (mask per event) and the sampling frequency instead of using the telescope id at each event. I can change this if I am wrong.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,977 @@
...
+        if self.use_interleaved is None:  # test to include interleaved correction on pedestal, NOT FUNCTIONAL
+            self.pedestal_std = None
+        else:
+            _, self.pedestal_std = get_bias_and_std(config['lh_fit_config']['use_interleaved'])
+        self.end_parameters = None
+        self.error_parameters = None
+        self.fcn = None
+
+        poisson_peaks = np.arange(self.n_peaks, dtype=int)
+        poisson_peaks[0] = 1
```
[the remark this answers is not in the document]
gabemery: I am creating a log(factoria(k)) array. 
0! = 1! = 1 and n! = 1*2*3*...*n = 1*1*2*3*...*n
When going to the log, we have a cumulative sum of log( i < n ) excluding log(0). But log(0!) = 0 = log(1). So overriding 0 by 1 allows to apply the log to the array and gives the correct log_factorial after applying the cumulative sum.

lstchain/reco/reconstructor.py
gabemery: It is actually not used in __call__ but only for the plotting function to save the figure.
As we discussed before, we don't save the information needed for this plots and they are very useful to identify issues in development. So I would like to keep them for now.
One possibility I though about would be to import them from a dedicated file taking the fitter as input.

Now that I think about it, passing a string to TimeWaveformFitter.plot_debug() instead of always using the event id could also work. And this would remove the need to save it in the fitter.

lstchain/reco/reconstructor.py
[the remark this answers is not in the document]
gabemery: I was not aware, this will be fixed.

lstchain/reco/reconstructor.py
gabemery: A choice was made in the original code implementation ( before I joined the project) to use a positive time_gradient and a 2 pi rotation for psi. My interpretation is that the sign of the time_gradient is not physical.

maxnoe: Makes sense! Should be explained in a comment...

lstchain/image/pdf.py
```diff
@@ -0,0 +1,62 @@
...
+    sigma: float
+        Width of the normal distribution
+
+    Returns
+    -------
+    log_pdf: float or array-like
+        Log of the evaluation of the normal law at x
+
+    """
+    return ne.evaluate("-(x - mean) ** 2 / (2 * sigma ** 2) - log((sqrt(2 * pi) * sigma))",
```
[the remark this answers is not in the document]
gabemery: Re-written large part of the likelihood computation using numba CC

gabemery:
Re-wrote the likelihood computation using numba CC to have compiled versions of the main computations.
The most time consuming process is then the interpolation of the pulse template for each sample. Using 'linear' instead of 'cubic' divides by 2 the total processing time and should be correct if the pulse template in the eventsource is well sampled.

Non-modified feature which were commented on before:
 - 'magic' number in the low/high luminosity separation and Poisson term reduction : Alternative not found yet. Is it a problem for this merge?
 - Plotting code : For dev purpose, plotting code is provided. If requested for merging it can be removed and only provided as a patch in a separate location.

Current issue linked to the numba CC function : Code is not compiled when installing lstchain. 
It is compiled before the test (using a try, except block). I had to add a call to the function imported in the try to the test so that no unused import are present. (see lstchain/tests/test_lstchain.py)
 Since the fitter is imported in r0_to_dl1, even if the likelihood reconstruction is not used the import of the compiled function occurs. To avoid errors due to no-compilation in this case I put them in a try block. (see lstchain/reco/reconstructor.py)
The except block gives a log.warning every time. This may not be the good behaviour for standard processing so I am thinking of removing it.

gabemery:
Check of the performance at high level of the new version were performed. No errors are encountered and the results are similar to before. Important improvements in sensitivity, angular resolution and energy resolution and bias are seen in the new processing using a similar RF as what what used in lstchain 0.7.3. I checked its origin and it is coming from the inclusion of time shift correction on MC events.
Below are performance plots using blue as the lstchain v0.7.3 performances, cyan for the new version without the time shift correction, green for the new version with the time shift correction and finally red for the new version with a new type of RF (provided config, using disp_norm_sign and different set of parameters).
image
image
image
image

lstchain/reco/reconstructor.py
maxnoe: Since both are just a dictionary lookup, it won't be much faster to do something like

```
self.sampling_frequency
```
instead of
```
self.subarray.tel[tel_id].camera.readout.sampling_frequency 
```

especially considering, that you are performing a pretty expensive likelihood computation, this will be completely negligible. So yes, please do not limit to setting the telescope id via the config. 

If there is something really expensive to be done per telescope, it can be cached.

gabemery: So you are saying to pass and additional argument and to move all this to __call__ ? :
        self.geometry = self.subarray.tel[self.telescope_id].camera.geometry
        self.pix_x = self.geometry.pix_x.to_value(u.m)
        self.pix_y = self.geometry.pix_y.to_value(u.m)
        self.pix_area = self.geometry.pix_area.to_value(u.m ** 2)
        self.r_max = np.sqrt(self.pix_x ** 2 + self.pix_y ** 2).max()
        self.focal_length = subarray.tel[self.telescope_id].optics.equivalent_focal_length
        readout = self.subarray.tel[self.telescope_id].camera.readout
        sampling_rate = readout.sampling_rate.to_value(u.GHz)
        self.dt = (1.0 / sampling_rate)
and this (i forgot to pass the telescope_id here, I will change it)
        self.template = NormalizedPulseTemplate.load_from_eventsource(subarray.tel[1].camera.readout)
        self.template_time_of_max = self.template.compute_time_of_max()

It seems like a waste of resources and I don't understand what would be the advantage.

maxnoe (commented):
I went through the code again and made a lot of comments. The code still has a lot of hard coded magic numbers and behavior that is not clear to me and is not explained in comments/docstrings.

The reconstructor class should not store event data on the class and it should not restrict to a single telescope (just take a tel_id on `__call__`) like ctapipe methods.

Most of the configuration parameters should probably be per telescope.


Much of the likelihood does still happen in pure python code. The whole likelihood should probably be implemented using numba compiled code to make the performance acceptable. What is blocking that? Looking at the code, it should be possible.

lstchain/reco/r0_to_dl1.py
```diff
@@ -267,6 +268,44 @@ def get_dl1(
...
+    dl1_container: DL1ParametersContainer
+    fitter: TimeWaveformFitter
+
+    Returns
+    -------
+    DL1LikelihoodParametersContainer
+
+    """
+    # Applied to all simulated event and all physic events
+    if dl1_container.mc_type > -1000 or (event.trigger.event_type == EventType.SUBARRAY):
```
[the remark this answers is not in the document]
gabemery: I honestly don't remember. If `EventType` was always filled it is probably a mistake. If not it could be because part of this was written in lstchain 0.6.3. I will do the modification in the next commit.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,951 @@
...
+logger = logging.getLogger(__name__)
+
+try:
+    from lstchain.reco.log_pdf_CC import log_pdf_ll as log_pdf_ll
+    from lstchain.reco.log_pdf_CC import log_pdf_hl as log_pdf_hl
+    from lstchain.reco.log_pdf_CC import asygaussian2d as asygaussian2d
+except ImportError:
+    logger.warning('WARNING: Compiled likelihood reconstruction numbaCC functions missing.'
+                   ' If you plan to use the likelihood reconstruction you should run:\n'
+                   'lstchain/scripts/numba_compil_lhfit.py')
```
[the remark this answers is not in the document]
gabemery: I tried to integrate the compilation following the page you show but didn't manage to make it work. The issue was that this was not applicable to wheel based installation from what I could gather. A possible solution may be to have my code outside of lstchain and import it as a separated module (I guess removing current dependencies to lstchain, but this should be easy from what I see). If you have a solution It would be nice.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,951 @@
...
+                                 'charge': dl1_container.intensity,
+                                 't_cm': dl1_container.intercept - self.template_time_of_max,
+                                 'v': np.abs(v),
+                                 'psi': psi,
+                                 'length': start_length,
+                                 'wl': max(dl1_container.wl, 0.01),
+                                 'rl': 0.0
+                                 }
+
+        if np.isnan(self.start_parameters['t_cm']):
```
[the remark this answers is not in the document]
gabemery: I will have a deeper look, first checking if it still occurs with the current version and if so when.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,951 @@
...
+                                          + 1.0 * start_length),
+                                 'y_cm': (start_y_cm.to_value(u.m)
+                                          - 1.0 * start_length,
+                                          start_y_cm.to_value(u.m)
+                                          + 1.0 * start_length),
+                                 'charge': (dl1_container.intensity * 0.25,
+                                            dl1_container.intensity * 4.0),
+                                 't_cm': (-10, t_max + 10),
+                                 'v': (v_min, v_max),
+                                 'psi': (-np.pi * 2.0, np.pi * 2.0),
```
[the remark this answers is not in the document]
gabemery: To avoid boundary effects for seeds near +-pi

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,951 @@
...
+            self.error_parameters = m.errors.to_dict()
+
+        except (KeyError, AttributeError, RuntimeError):
+            self.error_parameters = {key: np.nan for key in self.names_parameters}
+            pass
+
+    @staticmethod
+    def log_pdf_l(mu, waveform, error, crosstalk, sig_s, templates, factorial, kmin, kmax, weight):
+        """Calls the compiled function for low luminosity pixels log_pdf."""
+        return log_pdf_ll(np.float32(mu), np.float64(waveform), np.float64(error), np.float64(crosstalk),
```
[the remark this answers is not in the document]
gabemery: Using numba AOC requires to give the type of the variables in the function definition. I thus apply the conversions here to ensure that no errors arise. The choice between 32 and 64 was only lightly motivated, I will see what is best.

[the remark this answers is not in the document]
maxnoe: Waveform will (or should) be 32-bits. There is certainly not 64bits of information n there.

lstchain/visualization/camera.py
```diff
@@ -111,3 +112,24 @@ def overlay_hillas_major_axis(display, hillas, **kwargs):
                       hillas.y + x * np.sin(hillas.psi.to(u.rad).value),
                       **kwargs,
                       )
+
+
+def display_array_camera(image, camera_geometry, axes=None, **kwargs):
```
rlopezcoto, line 117: what is the difference between this and the camera display from ctapipe? In any case this is a very general function that should be implemented there and not in lstchain

setup.py
```diff
@@ -48,9 +48,11 @@ def find_scripts(script_dir, prefix):
         'eventio>=1.5.1,<2.0.0a0',  # at least 1.1.1, but not 2
         'gammapy~=0.19.0',
         'h5py',
+        'iminuit>=2',
         'joblib',
         'matplotlib~=3.5',
         'numba',
+        'numexpr',
```
rlopezcoto, line 55: is `numexpr` really needed (not called anywhere)?

lstchain/reco/tests/test_r0_to_dl1.py
```diff
@@ -16,13 +20,12 @@ def test_rescale_dl1_charge():
     for tel_id in tel_ids:
         np.testing.assert_allclose(event.dl1.tel[tel_id].image, images[tel_id]*rescaling_factor)
 
+
 def test_r0_to_dl1_nsb_tuning(tmp_path, mc_gamma_testfile):
     from lstchain.reco.r0_to_dl1 import r0_to_dl1
```
rlopezcoto, line 25: if you move all the imports to the beginning, maybe you can also move this one

lstchain/reco/reconstructorCC.py
```diff
@@ -0,0 +1,95 @@
...
+        n_pixels, n_samples = waveform.shape
+        log_pixel_pdf_hl = np.empty((n_pixels, n_samples))
+        for i in range(n_pixels):
+            for j in range(n_samples):
+                mean = mu[i] / (1 - crosstalk[i]) * templates[i, j]
+                sigma = (mu[i] / (1 - crosstalk[i]) / (1 - crosstalk[i]) / (1 - crosstalk[i])
+                         * templates[i, j] * templates[i, j])
+                sigma = np.sqrt((error[i] ** 2) + sigma)
+                log_pixel_pdf_hl[i, j] = (-(waveform[i, j] - mean) * (waveform[i, j] - mean) / 2.0 / sigma / sigma
+                                          - np.log(np.sqrt(2 * 3.141592653589793) * sigma))
```
rlopezcoto, line 54: `np.pi`

lstchain/reco/reconstructorCC.py
```diff
@@ -0,0 +1,95 @@
...
+                    sigma = np.sqrt(error[i] * error[i] + sigma)
+                    gauss = 1 / (np.sqrt(2 * 3.141592653589793) * sigma) * np.exp(
+                        -(waveform[i, j] - mean) * (waveform[i, j] - mean) / 2.0 / sigma / sigma)
+                    sum_k += poisson + gauss
+                if sum_k <= 0:
+                    return -np.inf
+                sum += weight[i, j] * np.log(sum_k)
+        return sum
+
+    @cc.export('log_pdf_hl', 'f8(f4[:],f8[:,:],f8[:],f8[:],f8[:,:],f4[:,:])')
```
rlopezcoto, line 42: this is very cryptic, some comments would help a lot to understand the purpose of these lines

lstchain/reco/reconstructorCC.py
```diff
@@ -0,0 +1,95 @@
...
+        for i in range(n_pixels):
+            for j in range(n_samples):
+                sum_k = 0
+                for k in range(n_k):
+                    poisson = (mu[i] * pow(mu[i] + (kmin + k) * crosstalk[i], (kmin + k - 1)) / factorial[k]
+                               * np.exp(-mu[i] - (kmin + k) * crosstalk[i]))
+                    mean = (kmin + k) * templates[i, j]
+                    sigma = (kmin + k) * ((sig_s[i] * templates[i, j]) ** 2)
+                    sigma = np.sqrt(error[i] * error[i] + sigma)
+                    gauss = 1 / (np.sqrt(2 * 3.141592653589793) * sigma) * np.exp(
```
rlopezcoto, line 34: np.pi

lstchain/reco/reconstructorCC.py
```diff
@@ -0,0 +1,95 @@
...
+    Functions pool:
+    -log_pdf_ll
+    -log_pdf_hl
+    -asygaussian2d
+    """
+
+    cc = CC('log_pdf_CC')
+    cc.verbose = True
+
+    @cc.export('log_pdf_ll', 'f8(f4[:],f8[:,:],f8[:],f8[:],f8[:],f8[:,:],f4[:],i4,i4,f4[:,:])')
```
rlopezcoto, line 18: same as below

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,951 @@
...
+        Returns
+        -------
+        cam_display: `ctapipe.visualization.CameraDisplay`
+            Camera image using matplotlib
+
+        """
+
+        params = self.end_parameters
+        rl = 1 + params['rl'] if params['rl'] >= 0 else 1 / (1 - params['rl'])
+        mu = asygaussian2d(np.float32(params['charge']
```
rlopezcoto, line 822: why using explicitly `np.float32`? is this really needed?

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,951 @@
...
+
+        cam_display = display_array_camera(mu,
+                                           camera_geometry=self.geometry)
+        if save:
+            cam_display.axes.get_figure().savefig('event/' + ids +
+                                                  '_model.png')
+            plt.close()
+        return None if save else cam_display
+
+    def plot_waveforms(self, image, axes=None, save=False, ids=''):
```
rlopezcoto, line 841: all the plotting code should go in different classes, probably into `visualization`

lstchain/reco/r0_to_dl1.py
```diff
@@ -267,6 +268,44 @@ def get_dl1(
...
+        # Don't fit if the cleaning used in the seed parametrisation didn't select any pixels
+        if dl1_container.n_pixels <= 0:
+            lhfit_container = DL1LikelihoodParametersContainer(lhfit_call_status=0)
+        else:
+            try:
+                lhfit_container = fitter(event=event, dl1_container=dl1_container)
+            except Exception as err:
+                logger.exception("Unexpected error encountered in likelihood reconstruction")
```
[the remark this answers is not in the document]
gabemery: I tested and it behaves as you said. I removed 2 `logger.exception`.

lstchain/reco/reconstructor.py
gabemery: I have removed this, and instead modified the compiled function signature to use the type appearing naturally in the code.
What is needed is to have the same type in the function definition (f4 == float32, f8 == float64,...). So assuming that default typing is always the same this local conversion is not needed.

lstchain/reco/reconstructor.py
gabemery: It is now the case

lstchain/reco/reconstructorCC.py
gabemery: Added explanation in the first docstring. I will redo each function docstring next.

lstchain/visualization/camera.py
gabemery: It was a small helper function that is indeed not very helpful. It simply called the ctapipe CameraDisplay and returned it.
I removed it.

lstchain/reco/reconstructor.py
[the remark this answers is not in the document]
gabemery: So this occurs when the cleaning selects 2 pixels only. It may not be worth it to recover such events, I will do some tests.

lstchain/reco/reconstructor.py
```diff
@@ -0,0 +1,951 @@
...
+            m.limits[key] = val
+        m.print_level = print_level
+        m.errordef = 0.5
+        m.simplex().migrad()
+        self.end_parameters = m.values.to_dict()
+        self.fcn = m.fval
+        try:
+            self.error_parameters = m.errors.to_dict()
+
+        except (KeyError, AttributeError, RuntimeError):
```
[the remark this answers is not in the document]
gabemery: It doesn't seem to do so, I will thus try removing the try-except block

mexanick:
> Unsolved questions:
> 
> * 'magic numbers' in bounds (reconstructor.py l178-201) : I don't really see an elegant solution but it could be passed somehow in the config file (lot of parameters)

The best solution would be to encode these parameters as component members and set the defaults as you do for other parameters. You do not need to populate always the config with all of them, but you keep the way to change them without changing the source code.

> * No compilation at installation : not sure if possible, and if it is I didn't find how

It is definitely possible. Please see https://numba.pydata.org/numba-doc/latest/reference/aot-compilation.html#numba.pycc.CC.distutils_extension 

You can configure the extension and compile/install it through the `setup.py`

> * nan values in seed temporal parameters : I found the origin, but kept my current handling of the issue (adding a comment)

gabemery:
Compilation is now done during installation and bound control parameters are now configurable.

mexanick:
@gabemery please extend the test coverage to satisfy the project standards

maxnoe:
@mexanick please note that numba code is not contributing to the reported coverage value, so as this PR has large amounts of numba code, the actual coverage is higher than the reported number

mexanick:
@maxnoe in this case we perhaps need to tune somehow the coverage reporting? Otherwise it would be difficult to meet the requirements or justify the merge while not meeting them?

mexanick:
Also I believe numba code can potentially be covered by tests as well through the direct imports of python code instead of compiled one. Although this would certainly be tests duplication

vuillaut:
I did not know we had coverage _requirements_

maxnoe:
@vuillaut the codecov Integration will show up red if the PR decreases the coverage or the diff coverage is below the project coverage, I think that's what @mexanick referred to as requirement.

However, that's just codecovs default and I think for better or worse we didn't enforce this

environment.yml
```diff
@@ -25,6 +26,7 @@ dependencies:
   - sphinx_rtd_theme
   - sphinx-argparse
   - nbsphinx
+  - numba::numba
```
maxnoe, line 29: Why install numba from the numba namespace? It's on conda forge

gabemery: I did this because the pip install project.toml uses a different source for the installation. In my test pip used 0.52 while conda-forge used 0.51. For consistency I though it would be better to get the latest version for both (or enforce a specific version for both).
This line can be removed otherwise.

gabemery:
Is this PR ready for merging or are further changes/tests needed before? @maxnoe @rlopezcoto

lstchain/reco/tests/test_r0_to_dl1.py
```diff
@@ -16,13 +20,11 @@ def test_rescale_dl1_charge():
     for tel_id in tel_ids:
         np.testing.assert_allclose(event.dl1.tel[tel_id].image, images[tel_id]*rescaling_factor)
 
+
 def test_r0_to_dl1_nsb_tuning(tmp_path, mc_gamma_testfile):
-    from lstchain.reco.r0_to_dl1 import r0_to_dl1
-    from lstchain.io import standard_config
-    import os
-    config = standard_config
+    config = deepcopy(standard_config)
```
rlopezcoto, line 25: why is this `deepcopy` needed?

lstchain/reco/r0_to_dl1.py
```diff
@@ -569,6 +616,17 @@ def r0_to_dl1(
                     containers=[event.index, dl1_tel, extra_im]
                 )
 
+                if lhfit_fitter is not None:
+                    lhfit_container = apply_lh_fit(event, telescope_id, dl1_container, lhfit_fitter)
+                    # Plotting code for development purpose only, will disappear in final release
+                    if lhfit_fitter.verbose >= 2 and lhfit_container["lhfit_call_status"] == 1:
+                        plot_debug(lhfit_fitter, event, telescope_id, dl1_container, str(event.index.event_id))
```
rlopezcoto, line 623: is this `plot_debug` really necessary? it could break some tests in the future just because of including this plotting here.

rlopezcoto (commented):
thanks @gabemery for all the effort in the improvement of the code and separation of the needed functionalities to try to have the likelihood code as separated as possible from the standard pipeline processing and sorry for the delay in the review. 
I left a couple of final comments, but the code looks already in a shape that should be acceptable to be merged with the main lstchain code

lstchain/reco/tests/test_r0_to_dl1.py
maxnoe: The tests mutates the config, so to not change this standard_config object for every following test, the deepcopy is there

environment.yml
maxnoe: Please use the standard version from conda-forge, we should avoid having too many channels in the environment file.

maxnoe: It should also be updated very quicklyu after releases.

rlopezcoto (dismissed):

gabemery:
Thanks for the approval @rlopezcoto and for all the reviewers comments. I don't have the authorization for it so if no one else wanting to comment remains, can someone perform the merge?

environment.yml
maxnoe: @gabemery this comment is not yet addressed.

maxnoe (changes requested):

PR #908, opened 2022-02-07
Refactor onsite scripts

maxnoe:
* Use `Path` in `create_pro_symlink`, add test
* Factor out functions to find R0 subruns, pedestal files, time calibration files, run summaries and add tests

### DIFF
--- a/lstchain/onsite.py
+++ b/lstchain/onsite.py
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+from .paths import parse_calibration_name
+
 DEFAULT_BASE_PATH = Path('/fefs/aswg/data/real')
 DEFAULT_R0_PATH = DEFAULT_BASE_PATH / 'R0'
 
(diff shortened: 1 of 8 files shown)


FrancaCassol:
Hi, I suggest to postone this to after the release production

maxnoe:
> Hi, I suggest to postone this to after the release production

But it fixes at least two issues we found trying to run the current version. The link creation (which I also fixed in a single change here #913) and the `Trash` situation

maxnoe:
I tested that I can produce all calibration files with the exception of the filter scan / systematics file for the test data and the crab sample night 2021-11-20.

If you could test running it on a calibration filter scan, we could solve these issues now.

FrancaCassol:
Already that we are refactoring, in recent DPPS meeting and documentation I saw that concerning the data, it is used the term "Cat-A" instead of "LevelA". Should we better change it now?

maxnoe:
That is a real change! Here I just improve the code quality by refactoring out functions and fixing some issues.

I would make the change you propose in a different PR.

FrancaCassol:
Ok, let me check the filter scan

moralejo:
> Already that we are refactoring, in recent DPPS meeting and documentation I saw that concerning the data, it is used the term "Cat-A" instead of "LevelA". Should we better change it now?

I agree with changing that before the 0.9.2 release

FrancaCassol (approved):

PR #909, opened 2022-02-07
Fix rst syntax to fix docs build

maxnoe:


maxnoe:
An until now not reported error in the docs made the CI fail

PR #912, opened 2022-02-10
Adapt charge limits to -2% change in calibration

moralejo:
From v0.8 onwards, due to some fixes, calibrated charges of real data are 2% smaller. Here we adapt some limits of the DL1 data check.


moralejo:
I know, I should eventually move all the DL1 check limits to a config file...

rlopezcoto (approved):

PR #913, opened 2022-02-11
Fix link logic in create_pro_symlink, make link replacement atomic

maxnoe:
@FrancaCassol reported problems with the filter scan jobs that might be due to our LDAP problems or due to a race condition when the symlink is replaced.

Due to a wrong comparison (`is not` instead of `!=`) the link was always recreated, which might lead to problems if something tries to read from it while another job is recreating it.

Here I make the check properly that no action is performed if the link is already correct and for the case it really needs to be modified, the change happens atomically by first creating a temporary link and then replacing the existing one for the temporary link.

### DIFF
--- a/lstchain/onsite.py
+++ b/lstchain/onsite.py
@@ -6,17 +6,21 @@ def create_symlink_overwrite(link, target):
     '''
     Create a symlink from link to target, replacing an existing link atomically.
     '''
+    target = target.resolve()
+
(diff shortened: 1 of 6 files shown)


FrancaCassol (approved):

FrancaCassol (approved):

rlopezcoto:
this is solved in #908

PR #916, opened 2022-02-15
Solved DL1 check issues

moralejo:
Script did not work in case some runs had no flatfield events or no muon rings.


morcuended (approved):
Worked fine for the problematic nights

PR #918, opened 2022-02-16
Change LevelA directory name so to be compliant with CTAO documentation

FrancaCassol:


moralejo (approved):

Mitsunari-Takahashi:
When you change the structure of the relevant data directories, could you announce it in advance?

PR #922, opened 2022-02-16
Fix azimuth & RA range

moralejo:
Fixed wrong range of RA and telescope azimuth in datacheck plots (was -pi, pi, now is 0, 2pi)

### DIFF
--- a/lstchain/scripts/lstchain_longterm_dl1_check.py
+++ b/lstchain/scripts/lstchain_longterm_dl1_check.py
@@ -30,6 +30,9 @@ import tables
 import warnings
 
 from astropy.table import Table
+from astropy.coordinates import Angle
+import astropy.units as u
+
 from bokeh.io import output_file as bokeh_output_file
 from bokeh.io import show, save
 from bokeh.layouts import gridplot, column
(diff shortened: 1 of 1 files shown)


maxnoe (changes requested):
Use `astropy.coordinates.Angle.wrap_at` to get a known range for input angles, e.g.:
```

In [1]: from astropy.coordinates import Angle
   ...: import astropy.units as u
   ...: import numpy as np
   ...: Angle(np.arange(-360, 361, 90) * u.deg).wrap_at(180 * u.deg)  # or 360 * u.deg
Out[1]: <Angle [   0.,   90., -180.,  -90.,    0.,   90., -180.,  -90.,    0.] deg>

In [2]: 
   ...: Angle(np.arange(-360, 361, 90) * u.deg).wrap_at(360 * u.deg)  # or 360 * u.deg
Out[2]: <Angle [  0.,  90., 180., 270.,   0.,  90., 180., 270.,   0.] deg>
```

moralejo:
> Use `astropy.coordinates.Angle.wrap_at` to get a known range for input angles, e.g.:
> 
> ```
> 
> In [1]: from astropy.coordinates import Angle
>    ...: import astropy.units as u
>    ...: import numpy as np
>    ...: Angle(np.arange(-360, 361, 90) * u.deg).wrap_at(180 * u.deg)
> Out[1]: <Angle [   0.,   90., -180.,  -90.,    0.,   90., -180.,  -90.,    0.] deg>
> 

The angles in question are not astropy quantities, so the code would become unnecessarily clumsy I think.

maxnoe:
`Angle(mean_ra, u.rad).wrap_at('360d').rad`

morcuended (approved):

maxnoe (approved):
Why ra in deg and az in rad?

moralejo:
> Why ra in deg and az in rad?

For the telescope az, alt I just kept radians because this is what we have in the DL1 files for the event-wise values, no other reason. I show them in degrees in the plots, though.

PR #923, opened 2022-02-18
Clarify docstring for no-image in dl1a

vuillaut:
I am not sure to understand why we don't want to write the images in this case?
But at least the docstring could mention it.
Condition at line 121 is useless.


maxnoe:
@moralejo had the strong opinion that modified images should not be stored, if I remember correctly

garciagenrique (approved):

PR #937, opened 2022-03-07
Update of a source-dependent config file

SeiyaNozaki:
This PR is just an update of the source-dependent configuration file following the latest update of the standard one.
I also changed the default `n_off_wobble` to `1` from `3`.

### DIFF
--- a/lstchain/data/lstchain_src_dep_config.json
+++ b/lstchain/data/lstchain_src_dep_config.json
@@ -1,7 +1,7 @@
 {
   "source_config" : {
     "EventSource": {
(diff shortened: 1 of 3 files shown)


rlopezcoto:
tests are based on having three off positions as default, that is why they are not passing, maybe we could adapt them to make them pass for any number of selected off positions?

SeiyaNozaki:
thank you @rlopezcoto ! Now it should work with any number of n_off_wobble

rlopezcoto (approved):
Great, looks good, thanks @SeiyaNozaki !

PR #940, opened 2022-03-10
Correct the R0 baseline subtraction in case of simulation events

FrancaCassol:
At present we subtract the sim_tel calibration pedestals (pixel per pixel) in order to obtain the uncalibrated R1 with baseline = 0 but the sim_tel calibration pedestals are affected by sim_tel calibration errors (supposed errors on the calibration values) that are then propagated to the waveform. In this PR the camera median pedestals are subtracted instead, which are around the orginal pedestal shift,  since the calibration errors are gaussian. In future releases, we could think to use the meta sim_tel information to directly subtracted the shift

### DIFF
--- a/lstchain/tools/lstchain_create_calibration_file.py
+++ b/lstchain/tools/lstchain_create_calibration_file.py
@@ -151,10 +151,10 @@ class CalibrationHDF5Writer(Tool):
 
                 # if simulation use not calibrated and not gain selected R0 waveform
                 if self.simulation:
-                    event.r1.tel[tel_id].waveform = (
-                            event.r0.tel[tel_id].waveform.astype(np.float32)
-                            -np.median(event.mon.tel[tel_id].calibration.pedestal_per_sample, axis=1)[...,np.newaxis, np.newaxis]
-                    )
+                    # estimate offset of each channel from the camera median pedestal value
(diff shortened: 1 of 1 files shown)


lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -153,7 +153,7 @@ def start(self):
                 if self.simulation:
                     event.r1.tel[tel_id].waveform = (
                             event.r0.tel[tel_id].waveform.astype(float)
-                            - event.mon.tel[tel_id].calibration.pedestal_per_sample[..., np.newaxis]
+                            -np.median(event.mon.tel[tel_id].calibration.pedestal_per_sample, axis=1)[...,np.newaxis, np.newaxis]
```
maxnoe, line 156: A comment would be good, something like:

```
# estimate offset of each channel from pedestal per pixel values
offset = np.mean(event.mon.tel[tel_id].calibration.pedestal_per_sample, axis=1).round()
event.r1.tel[tel_id].waveform = event.r0.tel[tel_id].waveform.astype(np.float32) - offset[:, np.newaxis, np.newaxis]
```

Not that the waveform should be `np.float32` not `float`, which is the same as `np.float64`

maxnoe (changes requested):

maxnoe (approved):

PR #942, opened 2022-03-13
DL3 energy-dependent alpha cut

SeiyaNozaki:
This PR implements energy-dependent alpha cut for the IRF/DL3 production.
The same solution as energy-dependent g/h cuts is basically used for the energy-dependent alpha cut.
(create a separate temporary header to save energy-dependent alpha cuts in IRF)
Since the alpha cut is already applied to observational DL3 data, there is no need to modify gammapy.

### DIFF
--- a/lstchain/io/event_selection.py
+++ b/lstchain/io/event_selection.py
@@ -72,6 +72,11 @@ class DL3Cuts(Component):
         default_value=0.2,
     ).tag(config=True)
 
+    alpha_containment = Float(
+        help="Percentage containment region for alpha cuts",
+        default=0.68,
+    ).tag(config=True)
+
     global_alpha_cut = Float(
         help="Global selection cut for alpha",
         default_value=20,
@@ -112,11 +117,6 @@ class DL3Cuts(Component):
         )
         return gh_cuts
 
-    def apply_global_alpha_cut(self, data):
-        """
-        Applying a global alpha cut on a given data
-        """
-        return data[data["alpha"].to_value(u.deg) < self.global_alpha_cut]
 
     def apply_energy_dependent_gh_cuts(self, data, gh_cuts):
         """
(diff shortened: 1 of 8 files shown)


lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -484,11 +508,17 @@ def start(self):
                         'deg'
                     )
             else:
-                extra_headers["AL_CUT"] = (
-                    self.cuts.global_alpha_cut,
-                    'deg'
-                )
-
+                if self.energy_dependent_alpha:
+                    extra_headers["ALPHA_CONT"] = (
```
chaimain, line 512: I think there is still the 8 character limit on the HIERARCH card. In the latest astropy there is support for [longer keywords](https://docs.astropy.org/en/stable/io/fits/usage/headers.html#hierarch-cards), but we do not currently have the latest astropy.

lstchain/io/event_selection.py
```diff
@@ -178,6 +178,50 @@ def apply_energy_dependent_theta_cuts(self, data, theta_cuts):
...
+        """
+        Applying a given energy-dependent alpha cuts to a data file, along the
+        reco energy bins provided.
+        """
+        
+        data["selected_alpha"] = evaluate_binned_cut(
+            data["alpha"],
+            data["reco_energy"],
+            alpha_cuts,
+            operator.le,
```
chaimain, line 221: The operation should be consistent with the operation we use when applying global alpha cuts. I have made a commit to this fix in PR #943. There is no need to apply the minor fix here.

chaimain (commented):
Sorry for the late review.
It looks good to me in general.

lstchain/tools/lstchain_create_irf_files.py
maxnoe: No, HIERARCH is specifically for longer than 8 character keywords. But it's an extension. We should try to fit in 8 chars.

lstchain/io/event_selection.py
SeiyaNozaki: ok, thanks for the comment!

lstchain/tools/lstchain_create_irf_files.py
SeiyaNozaki: thanks, I forgot it... I changed it to `AL_CONT`

chaimain (dismissed):

chaimain (approved):

PR #943, opened 2022-03-15
Some small improvements to DL3 Tools

chaimain:
Add a flag to not store DL3 as gzipped file and to be able to index DL3 files from nested directories.
Unzipped DL3 files can be used for faster analysis with gammapy, especially when dealing with large datasets or for source-dependent files.


chaimain:
To test the improvement in speed with unzipped DL3 files, I used files with a total of about 1E8 events, 30+ hrs of livetime of observations, and the spectral analysis worked in a few minutes, instead of about an hour.
Gammapy functions when trying to access information from each `Observation` object, tries to seek the specific DL3 files directly and takes some time to unzip a gzipped file every time.
@morcuended @SeiyaNozaki or others operating with such huge DL3 data, can you kindly confirm this improvement?

morcuended:
Thanks, @chaimain! I confirm the reduction in run time when using unzipped DL3 files in Gammapy. I would create the DL3 files unzipped by default instead of having a `--no-gzip` option.

morcuended:
Agreed with having the `--gzip` options instead.

lstchain/high_level/hdu_table.py
```diff
@@ -141,15 +135,14 @@ def create_hdu_index_hdu(
     base_dir = os.path.commonpath(
         [
             hdu_index_file.parent.absolute().resolve(),
-            fits_dir.absolute().resolve()
+            file_list[0].absolute().resolve()
```
morcuended, line 138: I'd say that `.absolute()` is not needed here. I think `.resolve()` already gives the absolute path.

morcuended (changes requested):

morcuended (dismissed):

morcuended (approved):

PR #949, opened 2022-03-21
Allow selecting N samples for training and testing

### LINKED ISSUE #948: Configurable number of training samples

vuillaut:
Early review is welcome

Fixes #948

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -35,7 +35,12 @@
     "r": [0, Infinity],
     "leakage_intensity_width_2": [0, Infinity]
   },
-  "n_events": null,
+  "n_training_events": {
+    "gamma_regressors": null,
+    "gamma_classifier_train": null,
+    "gamma_classifier_test": 0.2,
+    "proton_classifier": null
+  },
 
   "tailcut": {
     "picture_thresh":8,
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -35,6 +35,7 @@
     "r": [0, Infinity],
     "leakage_intensity_width_2": [0, Infinity]
   },
+  "n_events": null,
 
   "tailcut": {
     "picture_thresh":8,
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -36,10 +36,10 @@
     "leakage_intensity_width_2": [0, Infinity]
   },
   "n_training_events": {
-    "gamma_regressors": null,
-    "gamma_tmp_regressors": null,
+    "gamma_regressors": 1.0,
+    "gamma_tmp_regressors": 0.8,
     "gamma_classifier": 0.2,
-    "proton_classifier": null
+    "proton_classifier": 1.0
   },
 
   "tailcut": {
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -37,8 +37,8 @@
   },
   "n_training_events": {
     "gamma_regressors": null,
-    "gamma_classifier_train": null,
-    "gamma_classifier_test": 0.2,
+    "gamma_tmp_regressors": null,
+    "gamma_classifier": 0.2,
     "proton_classifier": null
   },
 
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -496,13 +496,13 @@ def filter_events(
     The Table object will be converted to pandas dataframe and used.
     Each filtering range is applied if the column name exists in the DataFrame so that
     `(events >= range[0]) & (events <= range[1])`
-    Returning filter is converted to a numpy object so that it can be used by both dataframe
-    and table inputs
+    The returned object is of the same type as passed `events`
 
     Parameters
     ----------
     events: `pandas.DataFrame` or 'astropy.table.Table'
     filters: dict containing events features names and their filtering range
+        example : dict(intensity=[0, np.inf], width=[0, np.inf], r=[0, np.inf])
     finite_params: optional, None or list of strings
         extra filter to ensure finite parameters
     n_events: int or float
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -16,6 +16,7 @@ import pandas as pd
 from astropy.coordinates import AltAz, SkyCoord, EarthLocation
 from astropy.time import Time
 from ctapipe.coordinates import CameraFrame
+from sklearn.model_selection import train_test_split
 
 from . import disp
 
@@ -543,16 +544,18 @@ def filter_events(
 
     filter = filter.to_numpy() if isinstance(filter, pd.DataFrame) else filter
 
-    events_df = events_df[filter]
-    if n_events is not None:
-        if n_events < 0:
-            raise ValueError("The number of events to keep cannot be < 0")
-        if 0 <= n_events <= 1:
-            n_events = int(n_events * len(events_df))
-        elif not isinstance(n_events, int):
-            raise ValueError(f"n_events={n_events} > 1 is not a valid value, an integer is expected")
+    events = events[filter]
 
-    return events[filter][:n_events]
+    if n_events is not None and n_events < len(events):
+        # sample n_events random events
+        try:
+            events = train_test_split(events, train_size=n_events)[0]
+        except ValueError as e:
+            raise ValueError(f"n_events={n_events} is not a valid value, check the provided config\n{e}")
+
+    return events
+
+
 
 
 def linear_imputer(y, missing_values=np.nan, copy=True):
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -486,16 +486,17 @@ def expand_tel_list(tel_list, max_tels):
 
 
 def filter_events(
-    events,
-    filters=dict(
-        intensity=[0, np.inf],
-        width=[0, np.inf],
-        length=[0, np.inf],
-        wl=[0, np.inf],
-        r=[0, np.inf],
-        leakage_intensity_width_2=[0, 1],
-    ),
-    finite_params=None,
+        events,
+        filters=dict(
+            intensity=[0, np.inf],
+            width=[0, np.inf],
+            length=[0, np.inf],
+            wl=[0, np.inf],
+            r=[0, np.inf],
+            leakage_intensity_width_2=[0, 1],
+        ),
+        finite_params=None,
+        n_events=None,
 ):
     """
     Apply data filtering to a pandas dataframe or astropy Table.
@@ -511,6 +512,10 @@ def filter_events(
     filters: dict containing events features names and their filtering range
     finite_params: optional, None or list of strings
         extra filter to ensure finite parameters
+    n_events: int or float
+        Number of events to keep.
+        If an integer > 1 is passed this will be the maximum number of events to keep.
+        If a float < 1, this is the ratio of events to keep.
 
     Returns
     -------
@@ -542,7 +547,15 @@ def filter_events(
                 if v > 0:
                     log.warning(f"{k} : {v}")
 
-    return events[filter.to_numpy()]
+    if n_events is not None:
+        if n_events < 0:
+            raise ValueError("The number of events to keep cannot be < 0")
+        if 0 <= n_events <= 1:
+            n_events = int(n_events * len(events_df))
+        elif not isinstance(n_events, int):
+            raise ValueError(f"n_events={n_events} > 1 is not a valid value, an integer is expected")
+
+    return events[filter.to_numpy()][:n_events]
 
 
 def linear_imputer(y, missing_values=np.nan, copy=True):
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -490,7 +490,6 @@ def filter_events(
         events,
         filters=None,
         finite_params=None,
-        n_events=None,
 ):
     """
     Apply data filtering to a pandas dataframe or astropy Table.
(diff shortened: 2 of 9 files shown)


lstchain/reco/utils.py
```diff
@@ -542,7 +541,18 @@ def filter_events(
...
+    events_df = events_df[filter]
+    if n_events is not None:
+        if n_events < 0:
+            raise ValueError("The number of events to keep cannot be < 0")
+        if 0 <= n_events <= 1:
+            n_events = int(n_events * len(events_df))
+        elif not isinstance(n_events, int):
+            raise ValueError(f"n_events={n_events} > 1 is not a valid value, an integer is expected")
+
+    return events[filter][:n_events]
```
maxnoe, line 555: Do we really want to take the first n events? not random sample?

moralejo:
Can you briefly describe how the number n_events is used?  It seems to me that you are using n_events gammas in total, but only a part of them are used in the g/h RF training, together with n_events protons. Probably it would be useful to set the numbers of each species separately.

vuillaut:
> Can you briefly describe how the number n_events is used? It seems to me that you are using n_events gammas in total, but only a part of them are used in the g/h RF training, together with n_events protons. Probably it would be useful to set the numbers of each species separately.

In the present config, `n_events` sets the maximal number of events to be used for each particle, 
So during the training stage:
- a max of `n_events` diffuse gammas will be kept in total, and then used for training as usual
- a max of `n_events` protons will be kept for training

If the same config is applied to `dl1_to_dl2` step, then again, a max of `n_events` will be kept for each particle.

For a more fine-grained description, I can think of something like:
``` 
"n_events": {
     "training_gammas": null,
     "training_protons": null,
}
```
And not allow this option for the inference step.
Otherwise, we will have to add:
``` 
"n_events": {
     "training_gammas": null,
     "training_protons": null,
     "dl1_inference": null
}
```
Not sure this last one is really useful...

moralejo:
I would define this option only for the training stage.
Then, I think it would be better to be able to control what number of events is used in each of the trainings, right? As of now, the g/h training would have fewer gammas than protons, since those used for the energy reco are (rightly) not re-used for the g/h training. 
Actually if the fraction of gammas that is used for E-reco is known, then it is enough to control the total numbers of training g's and p's  separately.

vuillaut:
> I would define this option only for the training stage. 

Ok I will apply these changes.

> Then, I think it would be better to be able to control what number of events is used in each of the trainings, right? As of now, the g/h training would have fewer gammas than protons, since those used for the energy reco are (rightly) not re-used for the g/h training. Actually if the fraction of gammas that is used for E-reco is known, then it is enough to control the total numbers of training g's and p's separately.

The fraction of gamma used for the regressors is known (20%) - it is the same for all regressors, but not configurable at the moment. I will modify the code so we can also control that number.

vuillaut:
For reference, this is the current training process:
```mermaid
graph LR
    GAMMA[gammas] -->|reg training| REG(regressors) --> |dump| DISK
    GAMMA --> S(split)
    S --> |80%| g_train
    S --> |20%| g_test
    g_train --> |reg training| tmp_reg(tmp regressors)
    tmp_reg --- A[ ]:::empty
    g_test --- A
    A --> g_test_dl2
    g_test_dl2 --- D[ ]:::empty
    protons -------- D
    D --> cls(classifier)
    cls--> |dump| DISK
    classDef empty width:0px,height:0px;
```

vuillaut:
New implementation with the config:

```mermaid
graph LR
    GAMMA[gammas] -->|`gamma_regressors`| REG(regressors) --> DISK
    GAMMA --> S(split)
    S --> |`gamma_classifier_train`| g_train
    S --> |`gamma_classifier_test`| g_test
    g_train --> |reg training| tmp_reg(tmp regressors)
    tmp_reg --- A[ ]:::empty
    g_test --- A
    A --> g_test_dl2
    g_test_dl2 --- D[ ]:::empty
    protons -------- |`proton_classifier`| D
    D --> cls(classifier)
    cls--> DISK
    classDef empty width:0px,height:0px;
```

vuillaut:
@moralejo Is this implementation ok?

@maxnoe if you have time to give another look for review...

Also, do we have a doc page explaining the config file?
That'd be useful...

lstchain/tests/test_lstchain.py
```diff
@@ -169,6 +176,24 @@ def test_build_models(simulated_dl1_file, rf_models):
     joblib.dump(cls_disp_sign, rf_models["disp_sign"], compress=3)
 
 
+@pytest.mark.xfail(raises=ValueError)
+def test_fail_build_models(simulated_dl1_file):
+    custom_config = {
+        "n_training_events": {
+            "gamma_regressors": 0.99,
+            "gamma_classifier_train": 0.78,
+            "gamma_classifier_test": 0.31,
```
moralejo, line 185: "gamma_classifier_test": **0.21**   ?

vuillaut:
> > For reference, this is the current training process:
> > We keep all gammas for the final regressors saved on disk. I think the rational was to train the best possible regressors. However, it might introduce a small bias down the road, since during inference, we apply that fully grown RF to the data for regression, before applying the classifier. In my opinion, that is negligible, I am mentioning for reference.
> 
> Are the 20% and 80% labels inverted in the graph above? Or only 20% of the original gamma sample is used in the g/h classifier training?

The graph is correct. Only 20% of the gammas are used for the g/h classifier, which is probably not optimal !
https://github.com/cta-observatory/cta-lstchain/blob/a39a5dff8134ec338dce248209826fc0e3b1287c/lstchain/reco/dl1_to_dl2.py#L375

lstchain/tests/test_lstchain.py
moralejo, line 185: Forget my previous comment :)

lstchain/tests/test_lstchain.py
```diff
@@ -151,14 +152,20 @@ def test_get_source_dependent_parameters_observed(observed_dl1_files):
...
 def test_build_models(simulated_dl1_file, rf_models):
-    from lstchain.reco.dl1_to_dl2 import build_models
     infile = simulated_dl1_file
-
+    custom_config = {
+        "n_training_events": {
+            "gamma_regressors": 0.99,
+            "gamma_classifier_train": 0.78,
+            "gamma_classifier_test": 0.21,
+            "proton_classifier": 0.98
```
moralejo, line 161: gamma_classifier_test seems to me a bit of a confusing name.  This is actually the fraction used in the _training_ of the classifier, right? (it is "test" only from the point of view of the tmp_regressors). 

Also, if we are not using the full gamma statistics, we could use the same training sample for the tmp regressors and the final ones. In that case, wouldn't it be enough, e.g.:

"gamma_regressors": 0.70
"gamma_classifier": 0.30     (could default to 1 -  fraction_gamma_regressors)
"proton_classifier": 1.00

vuillaut:
> gamma_classifier_test seems to me a bit of a confusing name. This is actually the fraction used in the training of the classifier, right? (it is "test" only from the point of view of the tmp_regressors).

I agree but failed to find something more explicit and clear. Any suggestion?

vuillaut:
> Also, if we are not using the full gamma statistics, we could use the same training sample for the tmp regressors and the final ones. In that case, wouldn't it be enough, e.g.:
> "gamma_regressors": 0.70
"gamma_classifier": 0.30 (could default to 1 - fraction_gamma_regressors)
"proton_classifier": 1.00

🤔
Not sure I understand.
The gammas used for the classifier are **not** a subpart of the `gamma_regressors` ones.  

So you could set 70% used for the regressors (`gamma_regressors=0.7`) and then  `gamma_classifier_train=0.6`, ` gamma_classifier_test=0.4` equals to one to use all gammas for the classifer, or any other repartition.

moralejo:
Perhaps I am not understanding. 

If we have enough statistics, the ideal thing would be to (i) train the regressors on some part of the gamma sample; (ii) apply those regressors to another independent part of the gamma sample (perhaps all the rest of it), and to the (desired fraction of the) proton sample; (iii) use the p and g "regressed" samples from (ii) for the classifier training.

In this way the classifier training would be done with events reconstructed in the same way (same regressors) as we will use for the real data and the "MC-IRF" sample.

As of now, the final regressors are not those used in the creation of the classifier training sample. Probably the effect is not large.

vuillaut:
> Perhaps I am not understanding.
> 
> If we have enough statistics, the ideal thing would be to (i) train the regressors on some part of the gamma sample; (ii) apply those regressors to another independent part of the gamma sample (perhaps all the rest of it), and to the (desired fraction of the) proton sample; (iii) use the p and g "regressed" samples from (ii) for the classifier training.
> 
> In this way the classifier training would be done with events reconstructed in the same way (same regressors) as we will use for the real data and the "MC-IRF" sample.
> 
> As of now, the final regressors are not those used in the creation of the classifier training sample. Probably the effect is not large.

Yes, you got it right. That is what I explained in 
https://github.com/cta-observatory/cta-lstchain/issues/949#issuecomment-1076905784_

So, do you suggest that we also change the workflow?

moralejo:
The problem I see with the current workflow is that if you use a large fraction of the gammas (as we do now) for the "tmp regressors" then the gamma statistics for the classifier training is small. On the other hand, if you increase the gamma stats for the classifier, then the tmp regressors will be more and more different (worse) from those we will later use for data (which are created using all gammas). So I think a more clean solution is to go for a single set of regressors that can be used also to create the classifier-training sample. But there may be more opinions. @rlopezcoto ? @maxnoe ?

vuillaut:
Ok, this will decrease the final regressors performance and is a change from what was decided early in the development of lstchain - so this is not just a new feature allowing to configure the number of samples for training.
Note that the number of gammas used to train the classifier would now be configurable and we could set to say 80% (rather than the actual 20%). I think that this would improve the classifier performances, without decreasing the final regressors performances.
Do we have somewhere an up-to-date features importance plot for the classifier? 
Are the regressors outputs so important for the classifier? 


For reference, this would be the new workflow:
```mermaid
graph LR
    GAMMA[gammas] --> Sg(split)
    Sg --> | p | g_train
    Sg --> | 1-p | g_test 

    g_train --> |training| reg(regressors)
    reg --> |dump| DISK
    reg --- A[ ]:::empty
    g_test --- A
    A --> |apply reg| g_test_dl2
    g_test_dl2 --- D[ ]:::empty
    PROTON[protons] -------- D
    D --> |train| cls(classifier)
    cls--> |dump| DISK
    classDef empty width:0px,height:0px;
```

moralejo:
Note that "apply regressors" is missing for the protons in the graph

Probably the importance of regressor outputs in the classifier is not large, after all they are built from the same inputs that go anyway into the classifier. 

If that is true, and for the case in which no regressor outputs are fed to the classifier, we may want to use the full gamma sample for regressors and classifier. Shall we just allow the fractions of gammas (for regressors ad classifier) to add up to more than 1? (if they add up to <=  1, then independent event samples are drawn from the gamma sample;  otherwise we just take the requested fraction for each at random from the original sample).

vuillaut:
This is pending, I am not sure I should completely refactor the training workflow or not at the end?

lstchain/tests/test_lstchain.py
moralejo: Hi, this comment about the options naming still holds.  My "forget my previous comment" refers to another one that I deleted (it was about the test with fractions adding to more than 1, to make it fail)

moralejo:
> This is pending, I am not sure I should completely refactor the training workflow or not at the end?

I think we can do without that for now. But the naming of the config file options is misleading, in my opinion (see https://github.com/cta-observatory/cta-lstchain/pull/949#discussion_r834136957).

vuillaut:
> > This is pending, I am not sure I should completely refactor the training workflow or not at the end?
> 
> I think we can do without that for now. But the naming of the config file options is misleading, in my opinion (see [#949 (comment)](https://github.com/cta-observatory/cta-lstchain/pull/949#discussion_r834136957)).

what about?

```
"gamma_regressors", "gamma_tmp_regressors", "gamma_classifier", "proton_classifier"
```

vuillaut:
I renamed the variables.

Note that the diagram of the workflow will be added to the built documentation, that should help to understand what is what.

lstchain/data/lstchain_standard_config.json
```diff
@@ -35,6 +35,12 @@
     "r": [0, Infinity],
     "leakage_intensity_width_2": [0, Infinity]
   },
+  "n_training_events": {
+    "gamma_regressors": null,
+    "gamma_tmp_regressors": null,
+    "gamma_classifier": 0.2,
+    "proton_classifier": null
+  },
```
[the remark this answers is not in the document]
moralejo: And the written out regressors are those using 100% of the gammas - but the diagram indicates they would be the 80% ones. Or did I misunderstand?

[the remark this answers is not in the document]
maxnoe: Json has no comments

vuillaut:
> Then the default is now:
100% of gammas for (final) regressors
80% of gammas for tmp regressors
20% of gammas for classifier
100% of protons for classifier

> correct? Should we write a comment in the json of what the default means?

That is correct, and this corresponds to the current situation.

I could replace the `null` values with the actual values as standard - that would probably be more explicit.

Max is correct, we cannot comment directly in the json file.

vuillaut:
>  And the written out regressors are those using 100% of the gammas - but the diagram indicates they would be the 80% ones. Or did I misunderstand?

I am sorry @moralejo I think you are mistaking the diagram I made when discussing if we should refactor the training process as well.

The diagram that will be in the doc is this one:
```mermaid
        graph LR
            GAMMA[gammas] -->|#`gamma_regressors`| REG(regressors) --> DISK
            GAMMA --> S(split)
            S --> |#`gamma_tmp_regressors`| g_train
            S --> |#`gamma_classifier`| g_test
            g_train --> tmp_reg(tmp regressors)
            tmp_reg --- A[ ]:::empty
            g_test --- A
            A --> g_test_dl2
            g_test_dl2 --- D[ ]:::empty
            protons -------- |#`proton_classifier`| D
            D --> cls(classifier)
            cls--> DISK
            classDef empty width:0px,height:0px;
```

moralejo:
> > Then the default is now:
> > 100% of gammas for (final) regressors
> > 80% of gammas for tmp regressors
> > 20% of gammas for classifier
> > 100% of protons for classifier
> 
> > correct? Should we write a comment in the json of what the default means?
> 
> That is correct, and this corresponds to the current situation.
> 
> I could replace the `null` values with the actual values as standard - that would probably be more explicit.

yes, I think so, thanks!

moralejo:
About the diagram: ok, I see, sorry for the confusion!
Just make the explicit settings in the default for clarity, and we can move on.

moralejo (approved):

vuillaut:
Thank you for the exchanges and review. I merge.

PR #950, opened 2022-03-22
Update ctapipe_io_lst and add `--flatfield-heuristic` switch to relevant programs

maxnoe:
Adding a switch to disable the flatfield heuristic in the onsite scripts that need it

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -28,6 +28,7 @@ dependencies:
   - pandas
   - pymongo
   - seaborn
+  - jinja2=3.0 # pinned for bokeh 1.0 compatibility
   - pip:
       - ctapipe_io_lst~=0.18.0
       - ctaplot~=0.5.5
--- a/environment.yml
+++ b/environment.yml
@@ -29,6 +29,6 @@ dependencies:
   - pymongo
   - seaborn
   - pip:
-      - ctapipe_io_lst~=0.16.1
+      - ctapipe_io_lst~=0.17.0
       - ctaplot~=0.5.5
       - pyirf~=0.6.0
--- a/environment.yml
+++ b/environment.yml
@@ -29,6 +29,6 @@ dependencies:
   - pymongo
   - seaborn
   - pip:
-      - ctapipe_io_lst~=0.17.0
+      - ctapipe_io_lst~=0.18.0
       - ctaplot~=0.5.5
       - pyirf~=0.6.0
--- a/lstchain/conftest.py
+++ b/lstchain/conftest.py
@@ -135,7 +135,6 @@ def observed_dl1_files(temp_dir_observed_files, run_summary_path):
         "2516351600",
         "--dragon-module-id",
         "132",
-        "--flatfield-heuristic",
     )
 
     run_program(
@@ -164,7 +163,6 @@ def observed_dl1_files(temp_dir_observed_files, run_summary_path):
         test_drive_report,
         '--run-summary-path',
         run_summary_path,
-        "--flatfield-heuristic",
     )
 
     run_program(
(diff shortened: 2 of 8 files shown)


moralejo (dismissed):

maxnoe:
Discussed with @moralejo: We want to switch the default. So no option should be not using the heuristic and for old data we add the switch. However, this needs change of the default in `ctapipe_io_lst` (or adding the config option here everywhere, which is not ideal).

PR in io lst here, let's put this on hold until the new version of the source with the other default is out: https://github.com/cta-observatory/ctapipe_io_lst/pull/145

FrancaCassol:
Hi, should we write somewhere the limiting date & run between the old and the new data
(with/without heuristic option) ?

moralejo:
> Hi, should we write somewhere the limiting date & run between the old and the new data (with/without heuristic option) ?

Do we already have something like that in lstchain, in e.g. the calibration?  I thought that would go in lstosa, but it would probably be safer to have it in lstchain (or io_lst) for cases in which one runs without lstosa. The idea would just be to have a different default before a certain run number, right?

moralejo:
I converted to draft, since we may change the switch  to --fatfield-heuristic  . Also to discuss a possible runnumber - dependent default

maxnoe:
I inverted the default now (`--flatfield-heuristic` to use it, off by default)

moralejo:
Thanks @maxnoe. What about the runnumber-dependent automatic default? In order to have it we would need to have the option to set it both ways (i.e. flatfield-heuristic  true or false). Perhaps it is better to do this at the lstosa level, explicitly using the flatfield-heuristic option for the old data?  @morcuended, it should not be too complicated, right? And there must be similar epoch-dependent configurations in lstosa already.

morcuended:
This flatfield heuristic identification would be only for pre-2022 data, right? Right now in lstosa we are simply using a different lstchain configuration for the r0 to dl1 step with that option set to false. This is currently not automated and it would be the first epoch-dependent configuration that we would use in lstosa as far as I know. There are several run-dependent implementations used in calibration scripts for choosing the time calibration and systematics correction files though.

IMO, it would be safer to do this check at lstchain level since it would be reviewed & approved so less prone to errors. With the command line option, we could implement the run-dependent choice in lstosa as well.

FrancaCassol:
Hi, I agree with @morcuended, lstosa is not the only one that is reading the raw data. Ideally the information should be inside the RunSummary (or the future run data base), with a flag that chose the correct type of FF tagging. But, for the moment, should we not put it simply in ctapipe_io_lst?

maxnoe:
Since the rule is pretty simple (Everything from 2022 onwards shouldn't need it) we could just look at `camera_config.date`, right? Or do we want to have a specific run_id?).

The scheme I would propose is having a default of `None` for the option that means "Check the date". This could then be overridden with either False or True to force using or not using the heuristic.

FrancaCassol:
yes, the camera_config.date will be enough I think


> 



> Since the rule is pretty simple (Everything from 2022 onwards shouldn't need it) we could just look at `camera_config.date`, right? Or do we want to have a specific run_id?).

yes, the date will be enough I think

> The scheme I would propose is having a default of `None` for the option that means "Check the date". This could then be overridden with either False or True to force using or not using the heuristic.

yes, it seems to me a good solution. So, if one wants to force an other choice he can do it.

maxnoe:
PR here: https://github.com/cta-observatory/ctapipe_io_lst/pull/146

FrancaCassol (approved):

maxnoe:
Sorry, wrong keyword, fixed now

FrancaCassol (approved):

PR #955, opened 2022-03-29
Combine the post DL3 notebooks, to solve few minor issues

chaimain:
The LC was not produced appropriately, by providing any custom energy range, as gammapy required the flux points edges to be the same as the "energy" axis (used for the events) provided while creating the SpectrumDataset.

So, to keep the process uniform, I combine the 3 notebooks, back into a single notebook, where the energy range for LC estimation depends on the energy edges of the `energy_axis`. Also, `energy_axis` and `energy_fit_edges` have the same edges.

Other changes are -
- Option of applying custom safe energy masks, in blocks 19 and 20, instead of using `SafeMaskMaker`
- Small fix on displaying (or saving) the various plots of each dataset (counts+excess, exposure and energy dispersion) in block 25, to avoid issues when having large datasets.
- OGIP files are still written, but not used in this example to produce LC and SED. They can be used easily, but for further simplicity, I avoid doing so here.
- After performing the Fit function, using the updated spectral model parameters, for estimating the LC from the dataset
- Storing the Optimization results along with the final model into a file, in blocks 44 and 56
- In block 57, I save the flux points tables of SED and LC data as HDUs in a fits file, and in the final blocks, show how one can use the fits file to reproduce the LC and SED. This fits file is not based on GADF standard, but just a temporary usage, where we can compare the results of different analyses (or sources if need be).

I request analyzers to review and check if there is a need for some more modification or information in this general example notebook.


notebooks/post_DL3_analysis.ipynb
```diff
@@ -0,0 +1,2633 @@
+{
```
gabemery, line 1: <p>typo : not<s>e</s> present</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

gabemery, line 1: <p>i am confused, what is the status of the storage of RAD_MAX?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

gabemery, line 1: <p>What is the interest of having energy_fit_edges? And why create it from a different MapAxis? </p><p>Is it possible (without strange behaviours) to have e_reco_m{in,ax} != e_fit_m{in,ax} ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

gabemery, line 1: <p>I don't see the excluded region in the plot. Is there an issue? Or is it not displayed correctly on reviewnb?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

gabemery, line 1: <p>You should not use the figure range to plot the sed. Use the range used for the fit. Maybe even limit to the used range with events.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='65'/>

gabemery, line 1: <p>What are the two set of points?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='72'/>

gabemery, line 1: <p>Same comment as above. Either store/read the fit range for the SED or use the min and max energies of the flux points (should be the same?).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='79'/>

maxnoe: <p>Current master of lstchain stores it correctly as float in degrees, the version of the school stored the string `"0.2 deg". The notebook should reflect only the current state, not what was before.</p>

gabemery:
Very good notebook but I still had a couple of comments. Some of them need to be solved before merging.

It seems like your build is failing due to "lstchain_longterm_dl1_check" which is weird.

notebooks/post_DL3_analysis.ipynb
chaimain: <p>The energy_fit_edges can be different than energy_axis, in the sense of energy bins only. From gammapy dev - "Gammapy does not do unfolding of spectra (yet). It uses the 'hybrid' approach of computing flux points in reconstructed energy using forward folding and assuming flux points are not strongly correlated".</p><p>So for example, </p><p>if energy_axis.edges = [0.01, 0.1, 1, 10, 100] TeV</p><p>then we can have, </p><p>energy_fit edges.edges = [0.01, 0.1, 1, 10, 100] TeV, or </p><p>energy_fit edges.edges = [0.1, 1, 10, 100] TeV</p><p>And similarly, for LC estimation, we can use [0.1, 10] TeV</p><p>but not something like [0.5, 2] TeV</p><p> </p><p>I created a different MapAxis object for the energy_fit_edges, to have this distinction. But maybe, showing them as the same as energy_axis.edges is misleading.</p>

chaimain: <p>Ah no, the axes limits are small in this case. I will expand it to show the excluded region.</p>

chaimain: <p>I just thought of having a large enough energy range, irrespective of the choice of the energy_fit_edges, to show the SEDs in a 'broad energy range' as a better 'visual' than fixing the plot limits based on the energy_fit_edges or even flux points estimated.</p><p> </p><p>I do not have a strong opinion on this. Can you elaborate more on your point?</p>

chaimain: I missed specifying the `sed_type` in the `plot()`. Fixing it.

gabemery: Ok so if I understood correctly, energy_fit_edges should be a subset of  energy_axis.edges ?
In this case maybe define it as such.

chaimain: I agree that a good comparison can only be made while considering the instrument's constraints. But showing the spectrum of the best fit model of the data and the reference, I think, should be ok to understand the trend of the spectral model used in both cases.

I chose 10 GeV to 40TeV, because, on average, I had events in the DL3 files, between 10-15 GeV and 40-45 TeV. So for just showing the LST1 spectrum, I assume that it is ok to show at least the spectrum, even if we do or do not include the upper limit flux points at the thresholds, to show it in the range we have the 'selected gamma-like' events.

But yes, when making a comparison with other analyses' SEDs, we should only plot within the reasonable energy range.

But, maybe my reasoning is flawed.

gabemery: <p>Spectral Fit will be done in energy edges:\n", energy_fit_edges) </p><p> </p><p>This is wrong, you use this edges only for the flux point computation.</p>

gabemery: And even here the first point actually produced starts at 89 GeV because (i think) you are using dataset filtered by the safe mask anyway.

chaimain: Ok, I will rephrase it. But the flux point computation is indeed done in the provided bins. As the safe mask is applied, the dataset does not fill the energy bins below ~89 GeV, as can be seen in the `flux_points.to_table(...)` block, where those rows have nans for the flux estimation.

gabemery: "the flux point computation is indeed done in the provided bins" -> I would not say it is done here. The method tries to use the first points but we don't provide it the data so it gives NaN. This is independent of the content of your DL3 file.

With this being the case, it seems like this edges definition (subset of the edges used to bin the data, only used to create flux points, and even there actually compute a value only if the safe mask allows) is useless. You could simply run the flux point computation over all intervals and leave the selection to your safemask.

maxnoe:
Shouldn't the original notebooks also be removed?

chaimain:
> Shouldn't the original notebooks also be removed?

Ah Sorry, I thought I had done so.

notebooks/post_DL3_analysis.ipynb
```diff
@@ -0,0 +1,2655 @@
...
```
gabemery, line 1: <p>You should probably integrate only over the valid fit range here. If the 23-89 GeV range is not safe for spectrum extraction it is probably not safe for the light curve. Especially since we extrapolate the spectrum in this range.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/955/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='51'/>

chaimain: Agreed. Will fix it

gabemery (approved):
Latest changes look good. I didn't go over previously reviewed code again. I think you can merge.

PR #957, opened 2022-04-01
IERS B (instead of IERS A) & smarter data filter.

mireianievas:
Two suggested improvements:

* Use astropy-blunded IERS B (lower precision, but allows to skip the need for the buggy IERS A downloading, https://github.com/astropy/astropy/issues/13007 )
* Use less outlier-affected filtering of data distributions, based on the 32-50-68 percentiles instead of hardcoding the cuts or using std deviation.

Tested with Crab Nebula and M87.

### DIFF
--- a/notebooks/data_selection.ipynb
+++ b/notebooks/data_selection.ipynb
@@ -25,6 +25,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# IERS A downloading is somewhat buggy: \n",
+    "# see https://github.com/astropy/astropy/issues/13007\n",
+    "# The next two lines disable IERS A, using lower precision / pre-downloaded IERS B instead."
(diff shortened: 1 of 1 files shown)


moralejo:
Hi @mireianievas,  please do the changes in a branch of the main repo (not on a private fork), it is the current standard way of contributing to lstchain.  If you use your fork the tests cannot access the private data, and fail.

Also, don't mix different things in a PR, please separate the part about the IERS and the smarter data filter in two separate PRs.

rlopezcoto:
@mireianievas thanks for this implementation. As @moralejo told you, to get it merged, one needs to open a new PR from a branch in the main repo to get tests passed. I gave you Write permissions so if you could repeat the PR as described above, we'll get it quickly reviewed and merged. Thanks

PR #960, opened 2022-04-04
Change std outlier condition to be multiple of median std

maxnoe:
The previous condition of a three sigma interval over the pixels
was very restrictive and removed too many valid pixels.

Instead of the interval
[median_of_std - 3 * std_of_std, median_of_std + 3 * std_of_std]
we now use the interval
[1/3 * median_of_std, 3 * median_of_std] by default.


lstchain/calib/camera/pedestals.py
```diff
@@ -34,13 +34,13 @@ class PedestalIntegrator(PedestalCalculator):
 
      """
     charge_median_cut_outliers = List(
-        [-3, 3],
+        [-4, 4],
```
rlopezcoto, line 37: why is this changed 3->4?

maxnoe: Even if we assume that all pixels follow a common distribution, cutting an interval of 3 sigma will always remove ~5 pixels erroneously. That's way to harsh.

With 4 sigma, the expected number of erroneously rejected pixels is 0.1:
```
In [10]: def expected_rejected_pixels(a, b):
    ...:     return (1 - (norm.cdf(b) - norm.cdf(a))) * 1855
    ...: 

In [11]: expected_rejected_pixels(-3, 3)
Out[11]: 5.008121697347684

In [12]: expected_rejected_pixels(-4, 4)
Out[12]: 0.11750030720087512
```

rlopezcoto:
@FrancaCassol could you check if you agree with the modification of the limits so we can get this merged, or is this not needed anymore after all the modifications in the (updated branch)[https://github.com/cta-observatory/cta-lstchain/tree/ctapipe_0.17]?
Maybe we can change the target branch as well instead of the `master` one

FrancaCassol:
mmh, I agree that a 3 sigma cut is too strong (this cut is indeed practically disabled in our config where it is put to 10 sigma), but I would change the number of default sigmas (e.g to 4 or even to 5 sigma), not the cut criteria

PR #961, opened 2022-04-06
Reduce size of random forests in standard config

vuillaut:
Based on https://indico.cta-observatory.org/event/3992/

and cross-check on real-data from @SeiyaNozaki showing no diff in Crab spectra between the two configs.

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -60,7 +60,7 @@
   },
 
   "random_forest_energy_regressor_args": {
-    "max_depth": 50,
+    "max_depth": 30,
     "min_samples_leaf": 2,
     "n_jobs": 4,
     "n_estimators": 150,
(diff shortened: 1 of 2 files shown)


maxnoe (approved):

vuillaut:
- tests are failing for an unrelated issue
- I'd like to have an additional review

nbiederbeck (approved):

vuillaut:
Why do I get this error on `test_create_dl3_energy_dependent_cuts` about gammapy?

chaimain:
> Why do I get this error on `test_create_dl3_energy_dependent_cuts` about gammapy?

The error comes because the diffuse gamma events are very few (10) to create the energy-dependent IRFs with default/example config. Especially the theta values are almost all above the hard-coded max value of 0.32 deg. Will try to fix it.

vuillaut:
> > Why do I get this error on `test_create_dl3_energy_dependent_cuts` about gammapy?
> 
> The error comes because the diffuse gamma events are very few (10) to create the energy-dependent IRFs with default/example config. Especially the theta values are almost all above the hard-coded max value of 0.32 deg. Will try to fix it.

Thank you for looking into this.

chaimain (approved):

vuillaut:
Thank you very much @chaimain !

PR #962, opened 2022-04-07
fix hardcoded viewcone value for point source gammas

vuillaut:
This should fix your issue @SeiyaNozaki

PR #963, opened 2022-04-07
Add new function to check mc type

SeiyaNozaki:
continued from https://github.com/cta-observatory/cta-lstchain/pull/962

I implemented a new function to check mc_type (`point_like`, `diffuse`, `ring_wobble`) based on the suggestion in https://github.com/cta-observatory/pyirf/issues/176
This PR makes it possible to generate point-like IRF with ring-wobble MC properly (especially for source-dependent analysis).

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -30,6 +30,7 @@ from .lstcontainers import (
     ThrownEventsHistogram,
 )
 
+from ctapipe_io_lst import OPTICS
 
 log = logging.getLogger(__name__)
 
@@ -41,6 +42,7 @@ __all__ = [
     'auto_merge_h5files',
     'check_mcheader',
     'check_metadata',
+    'check_mc_type'
     'check_thrown_events_histogram',
     'copy_h5_nodes',
     'extract_simulation_nsb',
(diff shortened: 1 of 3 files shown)


lstchain/io/io.py
```diff
@@ -1131,3 +1133,49 @@ def extract_simulation_nsb(filename):
...
+    max_viewcone = simu_info.max_viewcone_radius.value
+
+    if max_viewcone == 0.0:
+        mc_type = 'point_like'
+
+    elif min_viewcone == 0.0:
+        mc_type = 'diffuse'
+        
+    else:
+        # check if mc_type is 'ring_wobble' or not
```
maxnoe, line 1159: Why do you need to check the file here? Why not just check if `min_viewcone == max_viewcone != 0`?

maxnoe:
```
In [6]: info.min_viewcone_radius
Out[6]: <Quantity 0.3999 deg>

In [7]: info.max_viewcone_radius
Out[7]: <Quantity 0.4001 deg>

```

Any reason they are not exactly identical... CORSIKA does not support it?

That might make things harder in pyirf.. 

Maybe use: 
```

In [11]: (info.max_viewcone_radius - info.min_viewcone_radius) < 0.1 * u.deg
Out[11]: True
```

chaimain (approved):

PR #964, opened 2022-04-08
update to ctaplot v0.6

vuillaut:
Update to use ctaplot v0.6
The main change is the use of astropy units in plots.


lstchain/mc/tests/test_plot_utils.py
```diff
@@ -1,18 +1,18 @@
...
 
 
 def test_sensitivity_plot_comparison():
-
     energy = np.geomspace(0.01, 100, 20) * u.TeV
     e_center = np.sqrt(energy[:-1] * energy[1:])
-    sensitivity = e_center**-1 * 1e-12 / (u.cm**2 * u.s)
-
+    sensitivity = e_center**-1 * 1e-12 * u.TeV**2/(u.cm**2 * u.s)
+    plt.figure()
```
rlopezcoto, line 16: why do you need to call `plt.figure()` twice?

vuillaut: why twice? in the two different tests you mean?
Creating a new figure is necessary in unit tests to make sure the plot is not added to an existing one (from a previous test) that has different axis units, leading to an error.

rlopezcoto: damn, the line in which a different test was defined must have been collapsed... sorry!

rlopezcoto (approved):

PR #969, opened 2022-04-14
Try using sigma clipping in calibration code to get robust mean / std estimation

maxnoe:
This is an early draft, only opened because I couldn't really finish it before leaving for the easter break, feel free to have a look and comment.


lstchain/statistics.py
```diff
@@ -66,3 +66,32 @@ def var(self):
     def std(self):
         '''Get the current sample std. dev. of all tracked statistics'''
         return np.sqrt(self.var)
+
+
+
+def sigma_clipped_mean_std(values, axis=0, max_sigma=4, n_iterations=5):
```
FrancaCassol, line 72: Hi @maxnoe, at the least the max_sigma should be configurable in the pedestal and flatfield components

FrancaCassol: also the number of iterations

FrancaCassol:
Hi @maxnoe, could we finalise this PR?
It proved to be pretty useful

lstchain/calib/camera/pedestals.py
```diff
@@ -267,11 +269,8 @@ def calculate_pedestal_results(self,
     # median over the sample per pixel
     pixel_median = np.ma.median(masked_trace_integral, axis=0)
 
-    # mean over the sample per pixel
-    pixel_mean = np.ma.mean(masked_trace_integral, axis=0)
-
-    # std over the sample per pixel
-    pixel_std = np.ma.std(masked_trace_integral, axis=0)
+    # mean and std over the sample per pixel
+    pixel_mean, pixel_std, _ = sigma_clipped_mean_std(masked_trace_integral, axis=0)
```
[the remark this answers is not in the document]
maxnoe: A simple warning won't be enough I fear, since there will always some ignored values, by design.

Maybe with a threshold? If more than the expected number of entries are ignored? I had to think about that... Or a given percentage of an event's pixels are ignored?

lstchain/statistics.py
maxnoe: Done

lstchain/calib/camera/pedestals.py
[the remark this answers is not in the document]
FrancaCassol: Looking back at the code, after the clipping it would be better to update the pixel mask of the masked_trace_integral with then new clipping mask (which not exploited for the moment)

> A simple warning won't be enough I fear, since there will always some ignored values, by design.

why always? (by the way, perhaps 5 sigma is enough?)

> Maybe with a threshold? If more than the expected number of entries are ignored? I had to think about that... Or a given percentage of an event's pixels are ignored?

If previous and new mask are different, for the moment we could just write as log.debug the number of pixels with rejected events (for the previous and the new mask) and  the % of rejected events?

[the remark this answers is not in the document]
FrancaCassol: let increase to 5 sigma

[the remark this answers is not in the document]
maxnoe: lets use whatever works well for the calibration ;)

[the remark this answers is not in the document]
maxnoe: Saying: it would be great if you could run a couple of tests on runs you know are problematic and see if 5 sigma is enough to get ok results and if this generally works better than e.g. doing all this by-hand cuts

lstchain/statistics.py
```diff
@@ -66,3 +67,56 @@ def var(self):
...
+    else:
+        mask = np.zeros(values.shape, dtype=bool)
+        values = np.ma.array(values, mask=mask, copy=True)
+
+    original_mask = values.mask.copy()
+    mean = values.mean(axis=axis)
+    std = values.std(axis=axis)
+
+    for _ in range(n_iterations):
+        values.mask = original_mask | (np.abs(values - mean) >= (max_sigma * std))
```
FrancaCassol, line 94: Hi @maxnoe,
we should not here add "original_mask = values.mask " ?

lstchain/calib/camera/pedestals.py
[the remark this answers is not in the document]
FrancaCassol: @max,

> Maybe with a threshold? If more than the expected number of entries are ignored? I had to think about that... Or a given percentage of an event's pixels are ignored?

yes, let add a warning of the number of events that have more than 50 pixels outliers , something like: 
```
max_pix_outliers=50
bad_events=list(np.where(np.sum(~mask, axis=2)[:,0]>max_pix_outliers)[0])
n_bad_pixels=np.sum(~mask, axis=2)[bad_events,0]
print(f"{n_bad_pixels.shape[0]} events with > 50 pixel outliers:{n_bad_pixels}") 
```
you can surely program it much better ;-) 

But also we should update the mask of the masked_trace_integral after the clipping

lstchain/statistics.py
FrancaCassol: yes, actually I did not mean that (the original mask is a copy isn't it?), what I meant is that we should keep track in the loop of the masked pixels in the iteration because they are other wise lost in the next one and again re-detected in the successive one and so on.

FrancaCassol: It is ok only to output at the end the final mask

maxnoe: >  what I meant is that we should keep track in the loop of the masked pixels in the iteration because they are other wise lost in the next one and again re-detected in the successive one and so on.

That's exactly the reason why the original mask is `or` ed with the currently detected outliers. Since both mean and std are updated in the loop, it is ok to recompute what are outliers and combine it with the previously excluded pixels.

FrancaCassol: ok

FrancaCassol (dismissed):

FrancaCassol (dismissed):

lstchain/calib/camera/pedestals.py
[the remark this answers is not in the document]
maxnoe: done

FrancaCassol (approved):

PR #973, opened 2022-04-22
Remove duplicate events when generating DL3 files for source-dependent analysis

SeiyaNozaki:
For source-dependent analysis, it can happen that duplicate events are saved in DL3 files since each event has multiple gammaness values depending on the number of assumed source positions. The effect should be small if we use a relatively tight cut. But if we use a looser cut (e.g. efficiency cut like #942), it would be better to take into account the effect.

If duplicate events are survived after gammaness/alpha cut, only one with higher gammaness will be saved.
In this PR, a separate function is prepared and implemented in `lstchain_create_dl3_file.py`


lstchain/io/__init__.py
```diff
@@ -29,6 +29,7 @@
     get_srcdep_assumed_positions,
     get_srcdep_params,
     add_source_filenames,
+    remove_duplicate_events,
```
rlopezcoto, line 32: better to name it `remove_duplicated_events`

lstchain/io/__init__.py
```diff
@@ -60,5 +61,6 @@
     'read_data_dl2_to_QTable',
     'get_srcdep_assumed_positions',
     'get_srcdep_params',
-    'add_source_filenames'
+    'add_source_filenames',
+    'remove_duplicate_events'
```
rlopezcoto, line 65: same as above (and also below)

lstchain/io/__init__.py
maxnoe: Why not duplicate?

maxnoe:
Are you sure it is correct to do this? 

You want to estimate the background as if that was your source position. I think removing duplicate events will make you underestimate the  background.

What I am particularly converted about is that the background rate will depend now implicitly on the number of off positions since more off positions means higher chance of duplicate events/higher chance of higher gamaness in another off region.

moralejo:
So for events whose axes intersect more than one region (on-source or off-source regions) you select the one which makes the event more gamma-like.  Assuming the off regions are equivalent to the on region (like we assume in source-independent analysis) in terms of background acceptance, real background will have in average equal probability of ending up in the ON or in any one of the OFF regions. So it should work.  

The problem is for signal, since gammas will sometimes end up in the off regions. This happens also in source-independent analysis, the psf tail contributes to populate the off regions, and biases the background estimate. For a bright source it may indeed be a problem. The additional complication in source-dependent analysis is that even the number of ON events will be affected by the number of chosen off regions (more chances to get a lower gammaness elsewhere). This would have an equivalence in the source-independent approach: increasing so much the number of off regions that you are forced to reduce their size to avoid overlap - and then you also start losing signal in the on region.

So I think the method should work, but depending on the source brightness, number of considered regions, and sharpness of the Alpha distributions, it can lead to biases.

lstchain/io/__init__.py
rlopezcoto: `duplicate` can be noun, verb or adjective, I think that with `duplicated` it is clearer, but fine

rlopezcoto:
>So I think the method should work, but depending on the source brightness, number of considered regions, and sharpness of the Alpha distributions, it can lead to biases.

maybe we could add a warning here when the number of OFF regions is >3 for people to be aware of the possibility of this effect

rlopezcoto:
@SeiyaNozaki implemented in this PR implemented all the comments from myself, @moralejo and @maxnoe . Even though we agree that it could be problematic if we assume more than 3 OFF positions (it will be very rarely done in single-telescope analysis), a warning is already implemented to make the users aware that they should be careful with the background determination. If nobody opposes, I'll merge it.

rlopezcoto (approved):

PR #974, opened 2022-04-25
Update the event binning for true/reco energy and source offset

### LINKED ISSUE #729: EventSelector defaults not meaningful
The default values defined in the `DataBinning` Class of `io/event_selection.py` are, partially, not meaningful, as they specify a too fine binning. As a result, the PSFs created by lstchain using the default values have to little statistics per bin if used with the current MCs, as can be seen here: [PSF_1000bins.pdf](https://github.com/cta-observatory/cta-lstchain/files/7163106/PSF_1000bins.pdf). With energy upper limits of 100 TeV, they also extend beyond the simulated energies of 50 TeV.

Since these values are used to compute the somewhat "official" IRFs it would be desirable to have a set of parameters that result in good quality IRFs. Decreasing the number of offset bins to eg 100 (instead of 1000) results in this PSF: [PSF_100bins.pdf](https://github.com/cta-observatory/cta-lstchain/files/7163109/PSF_100bins.pdf).

chaimain:
Resolves #729 and also adds support for binning different MC nodes, with varying true energy range, into IRFs. Will be also useful for #711.
If there any other defaults to be modified, please let me know here.

### DIFF
--- a/docs/examples/irf_tool_config.json
+++ b/docs/examples/irf_tool_config.json
@@ -40,8 +40,8 @@
     "bkg_fov_offset_min": 0,
     "bkg_fov_offset_max": 10,
     "bkg_fov_offset_n_edges": 21,
-    "source_offset_min": 0.0001,
-    "source_offset_max": 1.0001,
+    "source_offset_min": 0.001,
+    "source_offset_max": 1.001,
     "source_offset_n_edges": 100
   }
 }
--- a/docs/examples/irf_tool_config.json
+++ b/docs/examples/irf_tool_config.json
@@ -25,11 +25,11 @@
     "allowed_tels": [1]
   },
   "DataBinning": {
-    "true_energy_min": 0.01,
-    "true_energy_max": 100,
+    "true_energy_min": 0.005,
+    "true_energy_max": 200,
     "true_energy_n_bins_per_decade": 5,
-    "reco_energy_min": 0.01,
-    "reco_energy_max": 100,
+    "reco_energy_min": 0.005,
+    "reco_energy_max": 200,
     "reco_energy_n_bins_per_decade": 5,
     "energy_migration_min": 0.2,
     "energy_migration_max": 5,
@@ -42,6 +42,6 @@
     "bkg_fov_offset_n_edges": 21,
     "source_offset_min": 0.0001,
     "source_offset_max": 1.0001,
-    "source_offset_n_edges": 1000
+    "source_offset_n_edges": 100
   }
 }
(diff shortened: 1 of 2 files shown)


docs/examples/irf_tool_config.json
```diff
@@ -25,11 +25,11 @@
...
   "DataBinning": {
-    "true_energy_min": 0.01,
-    "true_energy_max": 100,
+    "true_energy_min": 0.005,
+    "true_energy_max": 200,
     "true_energy_n_bins_per_decade": 5,
-    "reco_energy_min": 0.01,
-    "reco_energy_max": 100,
+    "reco_energy_min": 0.005,
+    "reco_energy_max": 200,
```
RuneDominik, line 32: Is 200 TeV a reasonable upper limit or will this just result in empty bins to high energies?

chaimain: The new MC with higher zenith pointings have a maximum energy of 200 TeV, so this range accommodates them.
We will have empty bins for MC with different sky pointing nodes in either low or high energies, when compared together, for example, in performing IRF interpolations. 

But so far, I have not seen a problem with having empty bins in the IRFs. If you can point to some issues, it would be great to look into them.

RuneDominik (approved):

docs/examples/irf_tool_config.json
```diff
@@ -42,6 +42,6 @@
     "bkg_fov_offset_n_edges": 21,
     "source_offset_min": 0.0001,
     "source_offset_max": 1.0001,
```
RuneDominik, line 44: These should be changed to 0.001 and 1.001 to reflect the 100 bins

RuneDominik (changes requested):

docs/examples/irf_tool_config.json
maxnoe: Why start at a value > 0?

chaimain: I think I had some confusion when I first started with the values and did not correct them, as I had not received any comments about them.
I will change them
```
source_offset_min = 0
source_offset_max = 1
source_offset_n_edges = 101
```
That should be fine for default right?

RuneDominik (approved):

docs/examples/irf_tool_config.json
RuneDominik: They will at least create 100 evenly spaced bins between 0 and 1 deg which sounds reasonable.

moralejo (approved):

PR #976, opened 2022-04-28
Remove obsolete sensitivity function

moralejo:
Fixed sensitivity formula (it seems it is not used anywhere, but just in case)

### DIFF
--- a/lstchain/mc/__init__.py
+++ b/lstchain/mc/__init__.py
@@ -8,7 +8,6 @@ from .sensitivity import (
     bin_definition,
     calculate_sensitivity,
     calculate_sensitivity_lima,
-    calculate_sensitivity_lima_ebin,
     read_sim_par,
     ring_containment,
 )
@@ -17,7 +16,6 @@ __all__ = [
     'bin_definition',
     'calculate_sensitivity',
     'calculate_sensitivity_lima',
-    'calculate_sensitivity_lima_ebin',
     'int_diff_sp',
     'power_law_integrated_distribution',
     'rate',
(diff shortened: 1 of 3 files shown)


rlopezcoto (dismissed):

rlopezcoto:
tests are not passing after this modification, we need to change the values tested as well

moralejo:
> tests are not passing after this modification, we need to change the values tested as well

Sorry for that, let's just remove it, no?

maxnoe:
Do we still need the other function right above?

moralejo:
I would not remove it (in this PR at least), since that one is correct.

maxnoe:
But for what is it needed when we can just use https://github.com/cta-observatory/pyirf/blob/285bca156d9c7d88dbfe7a5af402fde7321924fb/pyirf/sensitivity.py#L95 ?

moralejo:
@rlopez perhaps you still use those functions somewhere? From my side I have no objection to drop them.

rlopezcoto:
> @rlopezcoto perhaps you still use those functions somewhere? From my side I have no objection to drop them.

no, I was not using them, maybe @misabelber, otherwise I think it should be fine to remove them and update any offline code to work with `pyirf` functions that are being better maintained.

rlopezcoto (approved):

moralejo:
Ok, let's merge it like this to remove the buggy formula. @misabelber, can check whether the other functions (or the whole file) superseded by pyirf can be removed, and of so do it in a separate PR?

misabelber:
I think the whole file can be removed, I am not using it for anything right now, as said, it has been superseded by pyirf.

PR #977, opened 2022-05-04
Update Random Forest input features

moralejo:
Include telescope pointing among RF features, for use with the new "all sky" Monte Carlo.

Question: for the particle classification we may want to use signed_skewness and signed_time_gradient instead of their "unsigned" counterparts. The sign is referred to the (event-wise) disp-reconstructed gamma-ray direction.


rlopezcoto:
we should add a check for people using old MC, although if we use the pointing as an input feature with the old MC (only with one pointing), the only effect will probably be a slower running and larger files, but will not affect the results further than that, right?

moralejo:
> we should add a check for people using old MC, although if we use the pointing as an input feature with the old MC (only with one pointing), the only effect will probably be a slower running and larger files, but will not affect the results further than that, right?

I think the difference in speed and tree size cannot be large. The parameters will never be chosen for any split. It will affect the RF meta-parameters (one if which goes as the square root of the number of features), but also that should be hardly noticeable in the performance.

moralejo:
Using now signed versions of the skewness and time gradient parameters. Only in particle  classification RF (because it is not available before).

moralejo:
@SeiyaNozaki would be good to update also the source-dep RF configurations (to include telescope pointing)

SeiyaNozaki:
Hi @moralejo , sorry I overlooked your comment...

Yes, please update also the source-dependent configurations as the standard one.

SeiyaNozaki:
now I'm wondering if it would be better to use delta (or sin(delta)) instead of azimulth in general since test gamma MC samples are chosen in a cos(zd)-sin(delta) space. For the analysis of north-pointing source (e.g. BL Lac), azimuth distribution could be different between training samples and test samples (azimuth distribution of training samples is from -90 to 90, but one of most of the test samples is from 90 to 270).

moralejo:
I don't understand @SeiyaNozaki. Azimuth is defined in 0-360 (N-E-S-W). Why -90 to 90?  
One thing I am a bit unsure about is whether we'll have trouble from the fact that the zenith-azimuth combinations of the pointings along the training lines is in general different  from that of a specific source (for Crab nearly identical, though), and different also from that of the test nodes. I think it should not be a problem (for a given pointing RF should effectively use the parts of the trees which were built with the closest training events), but I am not 100% sure how it will affect performance.

SeiyaNozaki:
Hi @moralejo 
> Azimuth is defined in 0-360 (N-E-S-W). Why -90 to 90?
sorry, please forget it... I just wanted to say north-pointing or south-pointing.
I'm also not sure a jump of azimuth (360->0) would affect the random forest performance with north-pointing training samples.

moralejo:
> 
> sorry, please forget it... I just wanted to say north-pointing or south-pointing. I'm also not sure a jump of azimuth (360->0) would affect the random forest performance with north-pointing training samples.

You are right, we had the same issue when using the angles (psi & phi) related to source position on the camera: there were funny features in the 0/360 line (though global effect on performance was not significant). These "cyclic" variables are rather inadequate for the way RF works. 

Using sin_delta would require to opt for a given reference date for the magnetic field (done already when creating the grid), which is a complication, and not really needed I think. Also, with stereo we will have azimuth-related differences from the array footprint projection, unrelated to B-field.

The best would be to use hour angle, but we would have to include its calculation somewhere and add it to the DL1 (or add it directly in the DL1 to DL2 stage like we do with the signed asymmetry parameters).  

A simpler solution is to use sin(az_tel). We would also have to implement its calculation, but I think it would solve the issue as well, right?

SeiyaNozaki:
> A simpler solution is to use sin(az_tel). We would also have to implement its calculation, but I think it would solve the issue as well, right?

Yes, agree! It would be better to use sin(az_tel) than just az_tel (at least for north pointing sources)

vuillaut:
> > we should add a check for people using old MC, although if we use the pointing as an input feature with the old MC (only with one pointing), the only effect will probably be a slower running and larger files, but will not affect the results further than that, right?
> 
> I think the difference in speed and tree size cannot be large. The parameters will never be chosen for any split. It will affect the RF meta-parameters (one if which goes as the square root of the number of features), but also that should be hardly noticeable in the performance.

I agree that the impact should not be that large. Moreover, if the AllSky production becomes the de-facto standard, then it makes sense that the standard configuration is adapted to that prod.

vuillaut:
> A simpler solution is to use sin(az_tel). We would also have to implement its calculation, but I think it would solve the issue as well, right?

Yes please.

vuillaut:
> @SeiyaNozaki would be good to update also the source-dep RF configurations (to include telescope pointing)

Might not be necessary with #983

vuillaut (approved):
Looks good to me.
Do you want to make a test using `sin_az_tel` before moving that to master?

moralejo:
> 
> Looks good to me. Do you want to make a test using `sin_az_tel` before moving that to master?

Not needed I think, we can do it after merging. The sin change is 100% safe, and the use of signed parameters I have tried already.

PR #983, opened 2022-05-20
Dealing with configs in a better way

### LINKED ISSUE #208: Single config file ?

### LINKED ISSUE #924: dealing with different "standard" configs for MC and telescope data
# The problem

In the standard configuration, we have some parameters that are different for processing MC and telescope data.
e.g: `"apply_integration_correction"` is set to `False` in the standard config but must be turned to `True` for MC.
This is problematic since:
- the config is not really "standard" since we have to change the parameters every time. So anyone re-analysis and taking the standard config actually has a wrong config.
- we will forget and it will bite us

# How to solve it (proposals to discuss)

## A: separate completely the configs
e.g: `standard_mc_config` and `standard_obs_config`
The issue with that proposal is that these two will eventually diverge and we'll have different parameters values in both that should actually be the same.

## B: Base config + specific configs

Instead of have the whole config and more in the standard config file, we could have a base config, on top of which are built the specificities for MC and OBS, in different configs.
As an example with a partial config, we would have:
 - base config = `standard_config.json`:
 ```
"tailcut": {
    "picture_thresh":6,
    "boundary_thresh":3,
    "keep_isolated_pixels":false,
    "min_number_picture_neighbors":2,
    "use_only_main_island":false,
    "delta_time": 2
  },
  "energy_regression_features": [
    "log_intensity",
    "width",
    "length",
    "x",
    "y",
    "wl",
    "skewness",
    "kurtosis",
    "time_gradient",
    "leakage_intensity_width_2"
  ],
```
- `mc_config.json`
```
  "source_config" : {
    "EventSource": {
      "allowed_tels": [1, 2, 3, 4],
      "max_events": null
    },
```

- `obs_config.json`
```
"LSTEventSource": {
      "default_trigger_type": "ucts",
      "allowed_tels": [1],
      "min_flatfield_adc": 3000,
      "min_flatfield_pixel_fraction": 0.8,
      "calibrate_flatfields_and_pedestals": false,
      "EventTimeCalculator": {
        "dragon_reference_counter": null,
        "dragon_reference_time": null
      },
"LSTCalibrationCalculator":{
    "systematic_correction_path": null,
    "squared_excess_noise_factor": 1.222,
    "flatfield_product": "FlasherFlatFieldCalculator",
    "pedestal_product": "PedestalIntegrator",
    "PedestalIntegrator":{
      "sample_size": 10000,
      "sample_duration":100000,
      "tel_id":1,
      "time_sampling_correction_path": null,
      "charge_median_cut_outliers": [-10,10],
      "charge_std_cut_outliers": [-10,10],
      "charge_product":"FixedWindowSum",
      "FixedWindowSum":{
        "window_shift": 6,
        "window_width":12,
        "peak_index": 18,
        "apply_integration_correction": false
      }
    },
```

The `standard_config.json` implements the default values. If you don't include them in the passed config, that's what is used. So you can add only what is necessary in the passed config `mc_config.json` or `obs_config.json`.

## C. Hard code an `if` condition for these parameters
This requires minimal change in the current setup, but of course it hides things behind the curtain.

vuillaut:
I think this should solve #924 (implementing solution B discussed there) #208 and improve the way we deal with configs.

- MC now gets its own config file so we stop forgetting about `apply_integration_correction`
- Users can still work with a limited config file (e.g. containing only an updated `tailcut` config different from the standard one)
- Users can then use the `--update-with` to dump a full configuration from that limited config

@SeiyaNozaki could you have a look at the update srcdep config please?

Fixes #924 
Fixes #208

### DIFF
--- a/lstchain/data/lstchain_dl1ab_tune_MC_to_Crab_config.json
+++ b/lstchain/data/lstchain_dl1ab_tune_MC_to_Crab_config.json
@@ -1,4 +1,15 @@
 {
+  "LocalPeakWindowSum":{
+    "window_shift": 4,
+    "window_width": 8,
+    "apply_integration_correction": true
+  },
+  "GlobalPeakWindowSum":{
+    "window_shift": 4,
+    "window_width": 8,
+    "apply_integration_correction": true
+  },
+
   "image_modifier": {
     "increase_nsb": true,
     "extra_noise_in_dim_pixels": 1.2,
--- a/lstchain/data/lstchain_mc_config.json
+++ b/lstchain/data/lstchain_mc_config.json
@@ -1,0 +1,12 @@
+{
+  "LocalPeakWindowSum":{
+    "window_shift": 4,
+    "window_width": 8,
+    "apply_integration_correction": true
+  },
+  "GlobalPeakWindowSum":{
+    "window_shift": 4,
+    "window_width": 8,
+    "apply_integration_correction": true
+  }
+}
--- a/lstchain/data/lstchain_src_dep_config.json
+++ b/lstchain/data/lstchain_src_dep_config.json
@@ -1,139 +1,4 @@
 {
-  "source_config" : {
-    "EventSource": {
-      "allowed_tels": [1],
-      "max_events": null
-    },
-    "LSTEventSource": {
-      "default_trigger_type": "ucts",
-      "allowed_tels": [1],
-      "min_flatfield_adc": 3000,
-      "min_flatfield_pixel_fraction": 0.8,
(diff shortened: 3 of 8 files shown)


vuillaut:
Note I have also taken the liberty to update `min_samples_leaf = 10` since this is what is used for the new MC prod by default.

SeiyaNozaki (approved):
Thank you @vuillaut !
It looks fine from my side.

PR #984, opened 2022-05-20
compute and use sin of pointing

vuillaut:
Adding to PR #977


lstchain/reco/r0_to_dl1.py
```diff
@@ -550,6 +550,8 @@ def r0_to_dl1(
 
                 dl1_container.az_tel = event.pointing.tel[telescope_id].azimuth
                 dl1_container.alt_tel = event.pointing.tel[telescope_id].altitude
+                dl1_container.sin_alt_tel = np.sin(dl1_container.alt_tel)
```
rlopezcoto, line 553: is `sin_alt_tel` necessary anywhere? In the discussion in #977 we agreed to include `sin_az_tel` because it is a cyclic variable, but `alt` should only go from 0 to 90

moralejo (approved):

PR #987, opened 2022-05-23
Fix timestamps in DL3 files, fixes #986

### LINKED ISSUE #986: Event times in DL3 files are wrong
the `TIME` column needs to contain the actual number of SI seconds since `MJDREF`. When storing just `dragon_time`, which is the unix time stamp in UTC, this neglects the leap seconds:

```
In [1]: from astropy.time import Time

In [2]: epoch = Time('1970-01-01T00:00:00', scale='utc')

In [3]: epoch
Out[3]: <Time object: scale='utc' format='isot' value=1970-01-01T00:00:00.000>

In [4]: epoch.mjd
Out[4]: 40587.0

In [5]: dragon_time = Time.now().unix

In [6]: import astropy.units as u

In [7]: wrong = dragon_time * u.s

In [8]: correct = (Time(dragon_time, format='unix') - epoch).to(u.s)

In [9]: wrong
Out[9]: <Quantity 1.65330599e+09 s>

In [10]: correct
Out[10]: <Quantity 1.65330602e+09 s>

In [11]: wrong - correct
Out[11]: <Quantity -28.99991822 s>


In [12]: epoch + correct
Out[12]: <Time object: scale='utc' format='isot' value=2022-05-23T11:39:47.238>

In [13]: epoch + wrong
Out[13]: <Time object: scale='utc' format='isot' value=2022-05-23T11:39:18.238>
```

same for the `TSTART` / `TSTOP` values.

To increase precision, we should also choose a better MJDREF, e.g. `2018-10-01` to have a date closely before first light.

maxnoe:
Timestamps were wrong by amount of UTC leap seconds before.

### DIFF
--- a/lstchain/high_level/hdu_table.py
+++ b/lstchain/high_level/hdu_table.py
@@ -432,15 +432,18 @@ def create_event_list(
     ev_header["TIME-OBS"] = time_params["time_obs"]
     ev_header["DATE-END"] = time_params["date_end"]
     ev_header["TIME-END"] = time_params["time_end"]
-    ev_header["TSTART"] = time_params["t_start"]
-    ev_header["TSTOP"] = time_params["t_stop"]
+    ev_header["TSTART"] = (time_params["t_start"].value, time_params["t_start"].unit)
+    ev_header["TSTOP"] = (time_params["t_stop"].value, time_params["t_stop"].unit)
     ev_header["MJDREFI"] = time_params["MJDREFI"]
     ev_header["MJDREFF"] = time_params["MJDREFF"]
     ev_header["TIMEUNIT"] = "s"
     ev_header["TIMESYS"] = "UTC"
     ev_header["TIMEREF"] = "TOPOCENTER"
     ev_header["ONTIME"] = elapsed_time
-    ev_header["TELAPSE"] = (time_params["t_stop"] - time_params["t_start"]).to(u.s)
+    ev_header["TELAPSE"] = (
+        (time_params["t_stop"] - time_params["t_start"]).to_value(u.s),
+        u.s
+    )
     ev_header["DEADC"] = effective_time / elapsed_time
     ev_header["LIVETIME"] = effective_time
 
--- a/lstchain/high_level/tests/test_hdus.py
+++ b/lstchain/high_level/tests/test_hdus.py
@@ -88,7 +88,7 @@ def test_get_timing_params():
 
     params = get_timing_params(data)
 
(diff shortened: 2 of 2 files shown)


lstchain/high_level/tests/test_hdus.py
```diff
@@ -72,3 +75,26 @@ def test_create_obs_hdu_index(tmp_path, dl3_file):
...
+    t = Time(['2020-01-01T00:00:00', '2020-01-01T00:00:01', '2020-01-01T00:00:10'])
+
+    data = Table({
+        'event_id': np.arange(len(t)),
+        'dragon_time': t.unix,
+    })
+
+    params = get_timing_params(data)
+
+    # Converting to a single mjd number looses precision but should be better than 0.1 us
```
rlopezcoto, line 91: loses

lstchain/high_level/hdu_table.py
```diff
@@ -34,6 +34,16 @@
 
 wobble_offset = 0.4 * u.deg
 
+LST_EPOCH = Time('2018-10-01T00:00:00', scale='utc')
+
+
+def time_to_fits(time, epoch=LST_EPOCH):
```
rlopezcoto, line 40: why "time_to_fits"?

lstchain/high_level/hdu_table.py
```diff
@@ -420,13 +434,13 @@ def create_event_list(
...
     ev_header["TSTART"] = time_params["t_start"]
     ev_header["TSTOP"] = time_params["t_stop"]
-    ev_header["MJDREFI"] = int(time_params["MJDREF"].mjd)
-    ev_header["MJDREFF"] = time_params["MJDREF"].mjd - int(time_params["MJDREF"].mjd)
+    ev_header["MJDREFI"] = time_params["MJDREFI"]
+    ev_header["MJDREFF"] = time_params["MJDREFF"]
     ev_header["TIMEUNIT"] = "s"
     ev_header["TIMESYS"] = "UTC"
     ev_header["TIMEREF"] = "TOPOCENTER"
     ev_header["ONTIME"] = elapsed_time
```
rlopezcoto, line 442: what is this `elapsed_time` and why is it not filled in `ev_header["TELAPSE"]`?

lstchain/high_level/hdu_table.py
maxnoe: Because ot converts an astropy time to the fits representation

chaimain (approved):

PR #989, opened 2022-06-02
Adding a custom-fov-offset flag for non standard FoV offset IRFs

chaimain:
For using a custom range of FoV offset, while using the new ring-wobble GammaDiffuse MC files, by changing the values of `fov_offset_min`, `fov_offset_max` and `fov_offset_n_edges` in the IRF config file. As per the request of @jurysek

### DIFF
--- a/lstchain/tools/lstchain_create_irf_files.py
+++ b/lstchain/tools/lstchain_create_irf_files.py
@@ -28,6 +28,10 @@ For source-dependent analysis, alpha cut can be used instead of theta cut.
 If you want to generate source-dependent IRFs, source-dep flag should be activated.
 The global alpha cut used to generate IRFs is stored as AL_CUT in the HDU header.
 
+One can also use custom FoV offset for point-like or ring-wobble MC, by using
+the custom_fov_offset flag and providing the range in the config file or
+using the default range.
+
 """
 
 from astropy import table
@@ -113,6 +117,7 @@ class IRFFITSWriter(Tool):
         -g /path/to/DL2_MC_gamma_file.h5
         -o /path/to/irf.fits.gz
         --point-like (Only for point_like IRFs)
+        --custom-fov-offset (For custom FoV offset range)
         --global-gh-cut 0.9
         --global-theta-cut 0.2
         --irf-obs-time 50
@@ -205,6 +210,11 @@ class IRFFITSWriter(Tool):
         default_value=False,
     ).tag(config=True)
 
+    custom_fov_offset = traits.Bool(
+        help="True for using custom FoV offset range",
+        default_value=False,
+    ).tag(config=True)
+
     classes = [EventSelector, DL3Cuts, DataBinning]
 
     aliases = {
@@ -247,6 +257,10 @@ class IRFFITSWriter(Tool):
         "energy-dependent-alpha": (
             {"IRFFITSWriter": {"energy_dependent_alpha": True}},
             "Uses energy-dependent cuts for alpha",
+        ),
+        "custom-fov-offset": (
+            {"IRFFITSWrite": {"custom_fov_offset": True}},
+            "Uses custom FoV offset range",
         ),
     }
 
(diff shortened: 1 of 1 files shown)


chaimain:
> For what is this needed?
> 
> If this is related to the issue reported in the last call that was about analyzing data taken at 0.5° wobble offset, I think it is not the right approach to just increase the width of the offset bins for the 0.4° degree ring wobble mcs.
> 
> Instead, we should probably use diffuse test mc or ring wobble at 0.5 degrees?
> 
> Do we really want to enable "cheating" and using 0.4 degree ring wobble simulations to analyze 0.5 degree wobble observations?

Indeed, it is related to that issue. Maybe @jurysek can better justify this approach over asking for a separate MC production.

jurysek:
> For what is this needed?
> 
> If this is related to the issue reported in the last call that was about analyzing data taken at 0.5° wobble offset, I think it is not the right approach to just increase the width of the offset bins for the 0.4° degree ring wobble mcs.
> 
Yes, we found that while in the v0.9.6 release of lstchain one could specify the offset axis in the config file even for point-like IRFs created from point-like/ring-like MC gammas, this feature disappeared in the dev version. In the new version, the offset range is hardcoded as [0.3, 0.5] deg, which means that gammapy fails in analysis of runs outside this range.

I should also note that using v0.9.6 version of the lstchain_create_irf_files tool to get point-like IRFs from ring-like MC gammas is dangerous because it provides systematically biased Aeffs, which took me a few sleepless nights thinking about why the fluxes reconstructed using the new MC are so low.. That's the reason why I started using the dev lstchain version.

> Instead, we should probably use diffuse test mc or ring wobble at 0.5 degrees?
> 
We already have diffuse testing gamma MC along the path of the source I am interested in, which is enough for the analysis. But we also wanted to compare our results with those obtained using point-like IRFs. I am not sure if a new large-scale production of ring wobble around 0.5 deg is the option..

> Do we really want to enable "cheating" and using 0.4 degree ring wobble simulations to analyze 0.5 degree wobble observations?
We discussed that with @moralejo and he is convinced that acceptance doesn't change that fast close to the center of the FoV, so it should be safe. In general, I think it's always better to have an option to use the IRF on data taken with slightly different offset than hardcode anything.

maxnoe:
> I should also note that using v0.9.6 version of the lstchain_create_irf_files tool to get point-like IRFs from ring-like MC gammas is dangerous because it provides systematically biased Aeffs, which took me a few sleepless nights thinking about why the fluxes reconstructed using the new MC are so low.. That's the reason why I started using the dev lstchain version.

Could you elaborate where this effect comes from?

maxnoe:
> We discussed that with @moralejo and he is convinced that acceptance doesn't change that fast close to the center of the FoV, so it should be safe. In general, I think it's always better to have an option to use the IRF on data taken with slightly different offset than hardcode anything.

I agree. So maybe we just increase the offset range we use for wobble gammas to the range where we think these are applicable? E.g. we use `[0, 1] deg` instead of `[d -0.1, d + 0.1] deg` for ring wobble mcs with 0.4 degrees?

jurysek:
> > I should also note that using v0.9.6 version of the lstchain_create_irf_files tool to get point-like IRFs from ring-like MC gammas is dangerous because it provides systematically biased Aeffs, which took me a few sleepless nights thinking about why the fluxes reconstructed using the new MC are so low.. That's the reason why I started using the dev lstchain version.
> 
> Could you elaborate where this effect comes from?
> 
> Edit: it seems this issue was fixed in a recent update?
>
Not sure about that, I reported to @chaimain and he suggested using the dev version (so he will probably know better), where the issue seems to be fixed - Aeffs in point-like IRFs from ring-like gammas are now close to those I have from the old point-like gammas MC for 20/40 deg zenith angle. 

v0.9.6
aeff_int50_leak10_gh075_th004_point04

dev - same cuts, same DL2 (trained and reco in 0.9.6)
aeff_v097dev_int50_leak10_gh075_th004_point04

chaimain:
> Edit: it seems this issue was fixed in a recent update?
Yes, this is fixed.

jurysek:
> > We discussed that with @moralejo and he is convinced that acceptance doesn't change that fast close to the center of the FoV, so it should be safe. In general, I think it's always better to have an option to use the IRF on data taken with slightly different offset than hardcode anything.
> 
> I agree. So maybe we just increase the offset range we use for wobble gammas to the range where we think these are applicable? E.g. we use `[0, 1] deg` instead of `[d -0.1, d + 0.1] deg` for ring wobble mcs with 0.4 degrees?

Yes, this is also an option.

chaimain:
> > We discussed that with @moralejo and he is convinced that acceptance doesn't change that fast close to the center of the FoV, so it should be safe. In general, I think it's always better to have an option to use the IRF on data taken with slightly different offset than hardcode anything.
> 
> I agree. So maybe we just increase the offset range we use for wobble gammas to the range where we think these are applicable? E.g. we use `[0, 1] deg` instead of `[d -0.1, d + 0.1] deg` for ring wobble mcs with 0.4 degrees?

It can probably help with everything. Will check if it affects anywhere else.

maxnoe:
> Yes, this is fixed.

Just for my interest: what was the underlying issue?

chaimain:
> > Yes, this is fixed.
> 
> Just for my interest: what was the underlying issue? And why is there no issue in this repository describing this?

It was #963 which resolved this issue.

moralejo:
> > > Yes, this is fixed.
> > 
> > 
> > Just for my interest: what was the underlying issue? And why is there no issue in this repository describing this?
> 
> It was #963 which resolved this issue. Probably we need to add more description on this change?

Yes, please explain why mistaking ring wobble with diffuse produced a huge difference in Aeff

rlopezcoto:
this PR was trying to perform a custom hack for sources that were observed at a different-than-simulated wobble offset in the MC, but is now obsolete and can be closed

PR #990, opened 2022-06-05
Apply interleaved calibration factors in dl1ab step

SeiyaNozaki:
The calculation of calibration factors and identification of bad pixels can be done using interleaved pedestal and calibration laser events. However, such information is not applied to the data analysis chain (subrun-wise r0 to dl1 step) since 100,000 laser events are required to update calibration factors in the standard configuration file. Thus, the analysis chain uses the calibration factors computed using the standard ped+cal run taken at the beginning of the night shift for all of the runs.
 
 This PR enables to recalibrate pixel charge and peak time in dl1ab step with calibration factors computed using interleaved ped+cal events. This process is applied to events with `calibration_id`=0 (for now all events have `calibration_id`= 0). Since `dc_to_pe` of `unusable_pixels` is zero, pixels defined unusable by interleaved calibration events have zero charge after this recalibration. So it would be better to treat those bad pixels using interpolation or something (#716 which will be updated soon). In addition, more robust definition of bad pixels in calibration codes is needed using #969 or #971 

@FrancaCassol What do you think?

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -64,6 +64,7 @@
     "threshold": 267,
     "fraction_cleaning_intensity": 0.03
   },
+  "apply_interleaved_cal": true,
 
   "random_forest_energy_regressor_args": {
     "max_depth": 30,
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -269,7 +269,7 @@ def main():
                     
                     if params['calibration_id'] == ORIGINAL_CALIBRATION_ID:
                         flag_selected_gain = np.array(
-                            [selected_gain_channel == HIGHG_AIN, 
+                            [selected_gain_channel == HIGH_GAIN, 
                              selected_gain_channel == LOW_GAIN],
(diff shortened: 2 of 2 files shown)


SeiyaNozaki:
CI failed since `dc_to_pe` of `calibration_id`=1 are all zeros in test dl1 data...

SeiyaNozaki:
Dear @FrancaCassol 

Thank you for your comments.
I agree with you that we can use the estimated gain with higher statistics. I just wanted to start to discuss how to use interleaved calibration events, so your comments are appreciated.

> Perhaps one could just do an updated of the flat-fielding values?

Which parameter do you mention exactly? not `dc_to_pe`?

What I want to implement is :
1) apply calibration coefficient using interleaved if calibration fails when using the camera calibration ped+cal run due to whatever reason. Now if it fails with the initial ped+cal run, those failure pixels are deactivated (zero charge) for all of the events of that day.
2) apply calibration coefficient using interleaved if HV was reduced due to bright star
3) tag `unusable_pixel` using the interleaved events (and interpolation or something should be applied for those bad pixels)

Any comments are welcome!

SeiyaNozaki:
duplicated with #1072, so I close this PR

PR #991, opened 2022-06-05
Interpolate charge and peak time of bad pixels using neighbors in dl1ab step

SeiyaNozaki:
reopened instead of #716 

This PR enables to interpolate `image` and `peak_time` using good neighbors in dl1ab step.
First, clusters of bad pixels are labeled with good neighbors using `number_of_islands` function. The interpolation step can add a weighting based on the distance between a given bad pixel and good neighbors.

The interpolation is important when a proper `unusable_pixels` definition is applied (e.g. #969, #971, #990).

Any comments are welcome!


maxnoe:
@SeiyaNozaki I made an initial PR in ctapipe, as this also came up with the FlashCam in simulations today: https://github.com/cta-observatory/ctapipe/pull/2023/

I would be happy if you can take a look.

SeiyaNozaki:
thanks, I will have a look!

maxnoe:
The ctapipe class is merged and was first released in 0.16.0

SeiyaNozaki:
Since it was already implemented in ctapipe, I close this PR

PR #1003, opened 2022-07-05
Require atleast ctapipe_io_lst 0.18.2

maxnoe:


FrancaCassol (approved):

maxnoe:
I had to add an option to `r0_to_dl1` which I though already existed, to use trigger type TIB since the test run (2008) has occurences of UCTS trigger being 42 (STEREO BIT, SINGLE_PE Bit and PEDESTAL bit set).

After the update, this event was correctly labeled as UNKNOWN, but it was the only one previously tagged pedestal, so the test for the calculate noise parameters failed.

With default_trigger_type="tib", the one actual pedestal event in that file is found again and the test passes with a slight modification of the tested value.

Do we just not look at these old runs? What does osa do with runs like these where TIB would be the better choice?

morcuended:
> Do we just not look at these old runs? What does osa do with runs like these where TIB would be the better choice?

We have been using `default_trigger_type="ucts"` for all the data runs. We do not perform any check to choose between `tib` or `ucts` source of information.

maxnoe:
Ok, for me this mainly means that we should include a more recent file in the test data sample that doesn't have all these now solved issues.

FrancaCassol (approved):

PR #1004, opened 2022-07-08
Add an option to disable disp reconstruction

SeiyaNozaki:
This PR enables to disable disp reconstruction for source-dependent analysis. 
It can be switched on/off through a configuration file.

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -452,18 +452,19 @@ def build_models(filegammas, fileprotons,
         disp_angle = test['psi']  # the source here is supposed to be in the direction given by Hillas
         disp_vector = disp.disp_vector(disp_norm, disp_angle, disp_sign)
 
-    test['reco_disp_dx'] = disp_vector[:, 0]
-    test['reco_disp_dy'] = disp_vector[:, 1]
-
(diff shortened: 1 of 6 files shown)


lstchain/reco/dl1_to_dl2.py
```diff
@@ -597,48 +598,49 @@ def apply_models(dl1,
         disp_angle = dl2['psi']  # the source here is supposed to be in the direction given by Hillas
         disp_vector = disp.disp_vector(disp_norm, disp_angle, disp_sign)
 
-    dl2['reco_disp_dx'] = disp_vector[:, 0]
-    dl2['reco_disp_dy'] = disp_vector[:, 1]
+    if config['disp_method'] != None:
```
vuillaut, line 601: I think it would be clearer to write that:
```
if config['disp_method'] is not None:
    do_common_disp_stuff
    if config['disp_method'] == 'disp_vector':
         do
    elif config['disp_method'] == 'disp_norm_sign':
        do
    else:
         raise ValueError("Unkown disp method")

```

vuillaut (changes requested):

SeiyaNozaki:
Since disp reconstruction is also performed now for source-dep analysis to compute extra parameters, now we don't need this option.

PR #1006, opened 2022-07-13
Add possibility to use a compiled interpolation during R1 NSB injection

gabemery:
Using a new compiled function implemented during the likelihood reconstruction implementation, some time can be gained during NSB injection at waveform level.

Observed time gain:
Processing gamma_20deg_180deg_run1___cta-prod5-lapalma_4LSTs_MAGIC_desert-2158m_mono_cone6.simtel.gz, and adding 52% NSB, processing times (r0_to_dl1) are
 - ~30 seconds without injection
 - ~6 minutes with injection without compiled code
 - ~4 minutes 30 seconds with injection with compiled code
 - ~3 minutes 50 seconds after some improvements and using AOT or njit compiled code

### DIFF
--- a/lstchain/reco/reconstructorCC.py
+++ b/lstchain/reco/reconstructorCC.py
@@ -204,7 +204,7 @@ def linval(a, b, x):
     return y
 
 
-@njit(cache=True)
+@njit()
 @cc.export('template_interpolation', 'f8[:,:](b1[:],f8[:,:],f8,f8,f8[:],f8[:])')
 def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
     """
--- a/lstchain/reco/reconstructorCC.py
+++ b/lstchain/reco/reconstructorCC.py
@@ -241,7 +241,7 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
             # Find the index before the requested time
             a = (times[i, j]-t0)/dt
             t = int(a)
-            if a < size:
+            if t+1 < size:
                 # Select the gain and interpolate the pulse template at the requested time
                 out[i, j] = a_hg[t] * (1. - a + t) + a_hg[t+1] * (a-t) if gain[i] else \
                     a_lg[t] * (1. - a + t) + a_lg[t+1] * (a-t)
--- a/lstchain/reco/reconstructorCC.py
+++ b/lstchain/reco/reconstructorCC.py
@@ -204,9 +204,9 @@ def linval(a, b, x):
     return y
 
 
-@njit()
-@cc.export('template_interpolation', 'f8[:,:](b1[:],f8[:,:],f8,f8,f8[:],f8[:],i8)')
-def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
+@njit(cache=True)
+@cc.export('template_interpolation', 'f8[:,:](b1[:],f8[:,:],f8,f8,f8[:],f8[:])')
+def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
     """
     Fast template interpolator using uniformly sampled base with known origin and step.
     The algorithm finds the indexes between which the template is needed and performs a linear interpolation.
@@ -235,6 +235,7 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
 
     """
     n, m = times.shape
+    size = a_hg.shape[0]
     out = np.empty((n, m))
     for i in range(n):
         for j in range(m):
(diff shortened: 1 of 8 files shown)


gabemery:
@maxnoe  Since there were worries about maintenance of the code with the addition of the likelihood reconstruction code I did this first version which is fully separable from related code. It also allows for a test that both are compatible.

If it is agreed that the compiled code should be available and that the slow option is not needed I can:
 - remove the try block (introduced here and previously added to the lstchain/reco/reconstructor.py)
 - remove the slower version (it is a bit easier to use since less type/shape sensitive but assuming that it is only used for NSB injection the faster one is enough)

What do you think @moralejo , @rlopezcoto ?

moralejo:
I don't understand what you mean by "_is a bit easier to use since less type/shape sensitive but assuming that it is only used for NSB injection the faster one is enough_".   Other than that, it seems to me that keeping only the faster version makes sense (I was not aware by the way it took so long, but I guess it is not a problem since it is just for MC processing).

gabemery:
I saw it took that much time only recently but it is not that surprising since it needs to create and add around 10 pulses per pixels and per events.

What I meant is that the compiled code requires very specific input (selected for the LHfit use case). So the function created to use it needs to receive a 1D array (to transform it into a 2D array) of times for a single pixel and it also need to transform the gain into a 1D array even if it is only one value. 
The current call method uses numpy array and scipy.interpd1d and can easily be used on multiple pixels or a single pixel for example. That's why it was used by both the LHfit and NSB injection before.

A specific, and more complete, compiled function for NSB injection could be added but is out of the scope of this PR which simply aims at using the new available function. 

I will update the code to remove the slower method.

maxnoe:
> AOT is what is used for the likelihood reconstruction. 

The question extends to that as well. There shouldn't be any runtime difference in AoT vs. JiT compilation with numba once the compilation is done. 

Are you sure you don't just measure the compilation time overhead, i.e. do you include `cache=True` in the njit call?

> But it still requires a 2D array of time and an array of gain with how it is written. 

What would be the alternative?

maxnoe:
> I didn't know that cache was needed for the use of the same function in the same code execution. 

No, the cache is used to persist the compilation result on disk for multiple runs of the python interpreter, within one run, also without cache the compilation should only happen once per type combination.

gabemery:
> > Either write a function for each input format or try to find a single algorithm working in all expected case.
> 
> I rather meant what these different expected input formats are. E.g. which use cases are not covered by the currently implemented signature?

Current interpolation function was written assuming n pixels with each having m time samples. Thus it contains 2 loops and assumes the gain information is available for each pixel.

The NSB use case uses this function to interpolate for a single pixel. So using a 1D time sample array and a single gain information. So it needs to wrap it in a function which convert the input to this format.
But for NSB we actually want k new pulses to inject per pixels (k being a random number per pixel based on the average NSB injection required). This could either be done using a 3D version, or keeping the 2D version where n pulses are created with n being the total number of NSB pulses to inject in all pixel. 

In the end I think the current implementation is only a bit weird when interpolating a single pulse. And further improvement of the NSB injection case would be on the code surrounding the interpolation function.

gabemery:
> > Strangely, at least for this application, njit is actually faster.
> 
> I guess this could be a type difference? For the AoT compiled version for one type signature you could get a cast with AoT if types do not match exactly but are convertible (e.g. float32 data for a float64 signature) and an appropriately typed version with `njit`?

Such a small effect seemed to take place : a variable of type int was used in the function requiring a int64. I remove the variable as a parameter and instead evaluated inside the function. It reduce the processing duration from 4m10s to 3m50s in AOT equivalent to njit before caching. But still not as good as njit with cache.
I don't see any other incompatible type.

I also checked AOT vs njit with cache on the LHfit usage. It seems processing duration are similar (1m15s+-5s) but with maybe a few percent improvement with AOT. I would need a more precise measure than having 2 'date' calls in my script, and a bigger test file, to really be sure.

lstchain/reco/reconstructorCC.py
```diff
@@ -204,9 +204,9 @@ def linval(a, b, x):
     return y
 
 
-@njit()
-@cc.export('template_interpolation', 'f8[:,:](b1[:],f8[:,:],f8,f8,f8[:],f8[:],i8)')
-def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
+@njit(cache=True)
```
maxnoe, line 207: Looking at the numba documentation, I never see jit applied to an exported function. What is the effect of applying `njit` to a `cc.export` function?

gabemery: It is needed for call to other functions to work in compiled functions.
I had errors and found this information here https://stackoverflow.com/questions/69783252/numba-ahead-of-time-aot-compiling-with-dependent-functions

gabemery:
Current version should be ready to merge if there are no new comments

maxnoe:
Didn't you say that you found aot to be the same? If that's true, I'd prefer to fully remove the aot compilation. (Could be done in another PR).

gabemery:
I saw a difference but testing on small MC files it may have been random. I will test it more deeply on data files to see if there is a gain with AOT. 
But I would prefer a dedicated PR (which would re-add the cache if AOT is removed).

gabemery:
The likelihood reconstruction code now uses njit instead of AOT compilation. I will create a new branch and PR.

PR #1008, opened 2022-07-20

moralejo:
To avoid confusions, set here as default the values currently used (with the LST-Prod2 MC) to tune the simulations to Crab data. 

The NSB increase is higher than before because in this simulation we use mirror_degraded_fraction (instead of telescope_transmission) to adjust in sim_telarray the light collection efficiency (by a factor 0.784). This will scale the NSB down, so we have to increase the value in the tuning applied in DL1ab. 

As for PSF, we are not tuning it now based on muons: we have now a better PSF in the simulation (based on star images on the star imaging screen), which seems ok to match the Width distributions f gammas,  plus we have now doubts that the muon ring width info is accurate enough to allow fine tuning of the psf.

### DIFF
--- a/lstchain/data/lstchain_dl1ab_tune_MC_to_Crab_config.json
+++ b/lstchain/data/lstchain_dl1ab_tune_MC_to_Crab_config.json
@@ -12,12 +12,12 @@
 
   "image_modifier": {
     "increase_nsb": true,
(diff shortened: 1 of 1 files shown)


vuillaut (approved):

moralejo:
Superseded by #1025

PR #1012, opened 2022-07-30
Update muon_analysis

ferxinii:
With these changes, the fit_muon function now fits a ring correctly even under high noise conditions. This was previously not the case. The changes made are: 
- the min. number of pixel neighbors to survive the tailcuts cleaning is now set to 2 (previously was set to default, which is 0).
- If no tailcuts are indicated as an input, the function will select appropiate tailcuts based on the noise of the image.

Also, in the analyze_muon_event function, when calling the fit_muon function in line 231, the 'tailcuts' input has been removed. This way, when using analyze_muon_event, the tailcuts will be adapted to the image's noise.

### DIFF
--- a/lstchain/image/muon/muon_analysis.py
+++ b/lstchain/image/muon/muon_analysis.py
@@ -101,7 +101,7 @@ def update_parameters(config, n_pixels):
     return params
 
 
-def fit_muon(x, y, image, geom, tailcuts):
+def fit_muon(x, y, image, geom, tailcuts=None):
     """
     Fit the muon ring
 
@@ -113,7 +113,9 @@ def fit_muon(x, y, image, geom, tailcuts):
         Number of photoelectrons in each pixel
     geom : CameraGeometry
     tailcuts :`list`
-        Tail cuts for image cleaning
+        Tail cuts for image cleaning.
+        Default is None, such that the tailcuts are calculated for each image.
+        If tailcuts are an input, those indicated will be used.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


lstchain/image/muon/muon_analysis.py
```diff
@@ -125,13 +127,29 @@ def fit_muon(x, y, image, geom, tailcuts):
     image_clean: `np.ndarray`
         Image after cleaning
     """
+    
+    if tailcuts == None:
+        #we want to quantize the noise of the image. To do so, we will use the negative Q cumulative distribution.
+        negative_Q=np.sort(image[image <= 0])
```
moralejo, line 133: Add spaces around  "="

Also: np.sort is not needed here, because you histogram the values, and only use the bin contents after that.

lstchain/image/muon/muon_analysis.py
```diff
@@ -125,13 +127,29 @@ def fit_muon(x, y, image, geom, tailcuts):
...
+
+        hist, bins = np.histogram(negative_Q,range=(-15,0),bins=30)
+        bins=bins[:-1]
```
moralejo, line 136: This modification of "bins" seems unnecessary

lstchain/image/muon/muon_analysis.py
```diff
@@ -125,13 +127,29 @@ def fit_muon(x, y, image, geom, tailcuts):
...
+
+        normalised_cumulative_hist = np.asarray(np.cumsum(hist)/np.sum(hist))
+
+        idx = (np.abs(normalised_cumulative_hist - 0.318)).argmin() #Find q closest to standard deviation
```
moralejo, line 140: Explain better: we want to get, from a single image, a quantity related to the width of the noise distribution, but only using the negative side of the distribution of pixel charges (because the positive side includes actual signal, i.e. the light from the muon).  So we look for the value of q below which we find 31.8% of the pixels in the image. We consider that "1 sigma" to use it as a reference to determine the image cleaning.  
"dev" is just the absolute value of that (it would correspond to the standard deviation in case the distribution was gaussian).

moralejo (changes requested):

rlopezcoto:
to be able to merge this PR, we need it to pass the tests, so it has to be done not from a fork, but directly from the cta-lstchain repo in the cta-observatory organization. @ferxinii I gave you write access to it, could you repeat this PR from a branch created directly in this repo, please?

lstchain/image/muon/muon_analysis.py
rlopezcoto: done in #1112

lstchain/image/muon/muon_analysis.py
rlopezcoto: transferred to #1112

lstchain/image/muon/muon_analysis.py
rlopezcoto: transferred to #1112

PR #1015, opened 2022-09-16
greater or equal filters in DL3Cuts

vuillaut:
I think a greater *or equal* operator is actually what we want here. For example, if a user requests an efficiency of 100%, I think he expects all events to be included, which is not the case atm.


rlopezcoto:
One of the tests need to be modified for the CI to pass, otherwise, I agree with the proposed changes

rlopezcoto (approved):

PR #1016, opened 2022-09-21
Add alias to disable timelapse correction to lstchain_create_drs4_pedestal_file

### LINKED ISSUE #1014: Create pedestal table using online-dt-corrected data
It would be nice to have an option to create pedestal table using data to which dt correction is already applied by EVB. 

Now dt correction is always applied in this tool (hard-coded), so I think what we need to change is just adding an argument input to switch off dt correction in this tool.

maxnoe:
Fixes #1014 

Adds a `--no-timelapse-correction` flag to the tool to disable the dt correction

@SeiyaNozaki

### DIFF
--- a/lstchain/tools/lstchain_create_drs4_pedestal_file.py
+++ b/lstchain/tools/lstchain_create_drs4_pedestal_file.py
@@ -165,6 +165,11 @@ class DRS4PedestalAndSpikeHeight(Tool):
             "DRS4PedestalAndSpikeHeight.full_statistics",
             "Whether to write the full statistics about spikes or not",
         ),
+        **flag(
+            "timelapse-correction",
+            "LSTR0Corrections.apply_timelapse_correction",
+            "Wether to apply drs4 timelapse correction or not.",
+        ),
     }
 
     classes = [LSTEventSource]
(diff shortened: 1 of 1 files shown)


lstchain/tools/lstchain_create_drs4_pedestal_file.py
```diff
@@ -165,6 +165,11 @@ class DRS4PedestalAndSpikeHeight(Tool):
             "DRS4PedestalAndSpikeHeight.full_statistics",
             "Whether to write the full statistics about spikes or not",
         ),
+        **flag(
+            "timelapse-correction",
+            "LSTR0Corrections.apply_timelapse_correction",
+            "Wether to apply drs4 timelapse correction or not.",
```
SeiyaNozaki, line 171: typo

SeiyaNozaki (dismissed):
Thank you @maxnoe !
I ran this script using online-dt-corrected data just in case and it works well.

lstchain/tools/lstchain_create_drs4_pedestal_file.py
maxnoe: Fixed

maxnoe:
@SeiyaNozaki Sorry for the delay, typo fixed. Please approve again.

rlopezcoto (approved):

PR #1017, opened 2022-09-21
Tool to calculate dt correction parameters

maxnoe:


maxnoe:
At the moment, this only write a giant file with the dt vs charge histograms.

We should add making the actual fits and just write the fit results for each pixel or group of pixels (and maybe some control plots).

Or we add a second tool to actually perform the fits to have that separate from creating the data.

FrancaCassol:
Hi @max, 
- could you please add some doc description in the code (e.g., target, on which data should be run and which is the output ?
- the fitting part (with the control check plots) is indeed necessary to complete the tool goal, should we not wait for it before merging?

maxnoe:
I updated this tool now to work a bit differently. Exporting timelapse data per cell was just to much...

Instead, the tool gets the drs4 baseline calibration file and uses the baseline to correct the individual samples before aggregating the mean / std values for each bin in dt per channel / pixel.

This is very manageable and results look good. I also added a second tool fitting all the pixels using iminuit and storing the fit coefficients. It's very fast. During development it was easier to have separate tools, since the timelapse data extraction runs much longer than the fitting, but we might also just join the two tools and fit everything after the data has been collected directly.

Please have a look @FrancaCassol @SeiyaNozaki

maxnoe:
Example dt curve with fit result:

dt_fit


Histogram of fit coefficients, vertical lines are the values currently used globally for all pixels:

dt_coefficients

maxnoe:
I am closing this here, will continue in the `lstcam_calib` repo in the gitlab: https://gitlab.cta-observatory.org/cta-array-elements/lst/analysis/lstcam_calib/

PR #1018, opened 2022-09-26
Add exceptions when required params are not defined for source-dep analysis

SeiyaNozaki:
In this PR, I added exceptions for source-dependent analysis with real data when some required parameters (`source_name` or `source_ra` & `source_dec`, `observation_mode`) are not defined in a configuration file. It helps analyzers to understand the cause of the error when those parameters are missing.

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -761,12 +761,16 @@ def get_expected_source_pos(data, data_type, config, focal_length=28 * u.m):
             expected_src_pos_y_m = np.zeros(len(data))
 
         # compute source position in camera coordinate event by event for wobble mode
-        if config.get('observation_mode') == 'wobble':
+        elif config.get('observation_mode') == 'wobble':
 
             if 'source_name' in config:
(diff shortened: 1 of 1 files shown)


lstchain/reco/dl1_to_dl2.py
```diff
@@ -768,12 +768,16 @@ def get_expected_source_pos(data, data_type, config, focal_length=28 * u.m):
...
-        if config.get('observation_mode') == 'wobble':
+        elif config.get('observation_mode') == 'wobble':
 
             if 'source_name' in config:
                 source_coord = SkyCoord.from_name(config.get('source_name'))
-            else:
+            elif 'source_ra' and 'source_dec' in config:
                 source_coord = SkyCoord(config.get('source_ra'), config.get('source_dec'), frame="icrs", unit="deg")
+            else:
+                raise NameError(
```
maxnoe, line 778: NameError seems not to be the correct thing to do here... 

KeyError seems more appropriate

SeiyaNozaki: It seems I implemented it locally at that time, but didn't push it to here... now it was done!

rlopezcoto (approved):

PR #1024, opened 2022-10-19
fixed calculation of noise in dim pixels parameter

jsitarek:
Unnecessary square root was added so the noise in dim pixels was defined as standard deviation. While later on it is used as a number of NSB photons from which poissonian number is selected (that adds this square root in the noise). Noise in the brigth pixels was properly defined (without square root)

### DIFF
--- a/lstchain/image/modifier.py
+++ b/lstchain/image/modifier.py
@@ -351,8 +351,8 @@ def calculate_noise_parameters(simtel_filename, data_dl1_filename,
     # The idea is that when a strong signal is present, the biased extractor
     # will integrate around it, and the additional noise is unbiased because
     # it won't modify the integration range.
-    # The noise is defined as the number of NSB photons, i.e. the extra variance, 
-    # rather than standard deviation, of the distribution
+    # The noise is defined as the number of NSB photoelectrons, i.e. the extra
+    # variance, rather than standard deviation, of the distribution
     extra_noise_in_bright_pixels = \
         ((data_median_std_ped_pe**2 - mc_unbiased_std_ped_pe**2) *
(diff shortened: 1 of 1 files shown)


lstchain/image/modifier.py
```diff
@@ -351,6 +351,8 @@ def calculate_noise_parameters(simtel_filename, data_dl1_filename,
     # The idea is that when a strong signal is present, the biased extractor
     # will integrate around it, and the additional noise is unbiased because
     # it won't modify the integration range.
+    # The noise is defined as the number of NSB photons, i.e. the extra variance,
```
moralejo, line 354: "photons" => photoelectrons

jsitarek: done

moralejo:
Thanks for finding this. Indeed the new version works better in getting the same distribution of charges in the relevant range.

This is pre-fix, for a Crab Nebula  run:

image

This is post-fix:

image

The peak itself seems to match a bit worse (see left plots, in linear scale) but above 3 p.e. (what we use in the standard analysis) the match is better. Note that the power-lawish branches above 12 pe, probably due to afterpulses do not match perfectly, but these are not the "target" of this NSB tuning procedure. That may need some separate study.

moralejo (dismissed):

moralejo:
@jsitarek  tests have to be modified...

jsitarek:
> @jsitarek tests have to be modified...

test failed because it was expecting 0.0 and got 0 in the output string
in the old code 0 was used but sqrt made a float out of it.
To reproduce the exact behaviour of the old code I now put 0. under the np.max(...), now it should work ...

moralejo (approved):

PR #1025, opened 2022-10-19
Set correct NSB tuning settings for low-zenith MC

moralejo:
NSB parameters are obtained with the fixed code for finding the NSB-tuning parameters (#1024).  Note: the replaced parameters were very outdated, not obtained with the immediately earlier version, and were not those used for Crab in the analyses over the past few months.

Also switched off obsolete optical PSF tuning.


rlopezcoto (approved):

jsitarek:
looks fine to me. 
Just for the reference, here are the values that I am using for the LST1+MAGIC analysis: noise_dim: 1.56, bias_dim: 0.52, transition: 8, noise_bright: 1.8.
They are quite similar

jsitarek (approved):

PR #1027, opened 2022-10-19
Fix definition of "true disp_sign" parameter

moralejo:


moralejo:
WIP, have to change calls to the modified functions.

lstchain/reco/disp.py
```diff
@@ -46,7 +47,13 @@ def disp(cog_x, cog_y, src_x, src_y):
         else:
             disp_angle = np.arctan(disp_dy/disp_dx)
 
-    disp_sign = np.sign(disp_dx)
+    # disp_sign : indicates in which direction, "positive" or "negative" we must move along the 
+    # reconstructed image axis (with direction defined by the versor cos(hillas_psi), sin(hillas_psi)) 
+    # we must move from cog_x, cog_y to get closest to the true direction (src_x, src_y)
+
+    sqrdist_plus  = (cog_x + disp_norm*np.cos(hillas_psi) - src_x)**2 + (cog_y + disp_norm*np.sin(hillas_psi) - src_y)**2
```
maxnoe, line 54: This is a little hard to read...

Is it the same as here?

https://github.com/fact-project/aict-tools/blob/ee091fc933110ff008e32df02ee12a97e412e8b3/aict_tools/preprocessing.py#L42-L79

maxnoe: Looks different....

moralejo: Same result, but ok, the approach you mention is simpler. Will change it.

lstchain/reco/disp.py
```diff
@@ -36,7 +37,16 @@ def disp(cog_x, cog_y, src_x, src_y):
     """
     disp_dx = src_x - cog_x
     disp_dy = src_y - cog_y
-    disp_norm = np.sqrt(disp_dx**2 + disp_dy**2)
+    
+    disp_norm = disp_dx * np.cos(hillas_psi) + disp_dy * np.sin(hillas.psi)
+    disp_sign = np.sign(disp_norm)
+    disp_norm = abs(disp_norm)
```
moralejo, line 43: Actually, @maxnoe, everyone: is there any reason not to simply leave the sign in "disp_norm", and do the reconstruction with RF of the signed quantity? We would save one RF, and I see no reason why it should be worse than doing norm & sign separately.

maxnoe: We can give it a try. I think I tried it out for FACT and got slightly worse performance, but not sure anymore. Classification plus regression is not using the same loss function as only regression... 

The separate prediction mainly comes from the history of this algorithm I'd say

maxnoe: One thing that came to mind is that a single random forest predicting disp with sign will behave differently in cases where the sign prediction is difficult/impossible. 

The two model version will guess a sign and predict an okish disp value, leading to the "ghost events" on the other side of the true source, very far away.

The single model version will guess values close to 0, leading to ghost events close to the center of gravity, so much closer to the true source position.

This seems like it could be worse for point sources?

Have not tested it though

rlopezcoto: if we go for a reduction in the number of parameters fed to the RF, this may imply a major release instead of a bugfix one. Additionally, since Max is saying that the performance may be slightly worse, I would go for the solving of this bug and opening an issue to keep this improvement in mind, test it after its implementation to see if the performance is the same, and propose it in a new PR to be released in a future major release.

moralejo: Yes, I did not intend it to be part of this fix, of course! Max's argument kind of convinces me, so let's park that for now.

rlopezcoto: great!

lstchain/reco/disp.py
```diff
@@ -114,6 +123,7 @@ def disp_parameters_event(hillas_parameters, source_pos_x, source_pos_y):
              hillas_parameters.y.to(u.m).value,
              source_pos_x.to(u.m).value,
              source_pos_y.to(u.m).value,
+             hillas_parameters.psi.to(u.rad).value
```
rlopezcoto, line 126: why are we using the hillas parameters container only here? wouldn't it make sense to use it also on all the previous calls?

maxnoe: `.to_value(u.rad)`

maxnoe: (Also above)

moralejo: > why are we using the hillas parameters container only here? wouldn't it make sense to use it also on all the previous calls?

You mean passing the container instead of passing as argument the individual parameters, right?
I can agree, but that is not related to the big being fixed here, right?

rlopezcoto: not related to the bug, just a suggestion

lstchain/reco/disp.py
```diff
@@ -36,7 +37,16 @@ def disp(cog_x, cog_y, src_x, src_y):
     """
     disp_dx = src_x - cog_x
     disp_dy = src_y - cog_y
-    disp_norm = np.sqrt(disp_dx**2 + disp_dy**2)
+    
+    disp_norm = disp_dx * np.cos(hillas_psi) + disp_dy * np.sin(hillas_psi)
+    disp_sign = np.sign(disp_norm)
+    disp_norm = abs(disp_norm)
```
vuillaut, line 43: Since this is changing the definition of `disp_norm` and `disp_sign`, shouldn't the reconstruction should be changed accordingly? (I mean applying the reconstructed disp to get the source position)

maxnoe, line 43: This should be `np.abs`

maxnoe: What do you want to change? If anything, this new definition brings it more in line with how it is applied. Using the distance along the shower axis, not the full 2d distance.

lstchain/reco/disp.py
```diff
@@ -123,7 +133,7 @@ def disp_parameters_event(hillas_parameters, source_pos_x, source_pos_y):
     disp_container.sign = d[4]
     disp_container.miss = miss(disp_container.dx.value,
```
maxnoe, line 134: If the result is multiplied with `u.m`, dx and dy must be `u.m`, right? So I would add replace `value` with `to_value(u.m)`, which will save us some headache once we switch to all parameters in telescope frame (degrees) because it will fail instead of silently attach the wrong unit.

lstchain/reco/disp.py
moralejo: No, why? The definition of disp_sign, in what refers to how it is used, is the same (though before its true value was not correctly calculated). For disp_norm there is indeed a small change, it is now the cog-to-source distance projected on the reconstructed image axis. But that is (conceptually) more correct than using directly the distance. In practice it won't change much for gammas, the two things agree for Alpha=0 and are nearly identical for all relevant values of Alpha.

moralejo: > This should be `np.abs`

ok - to make it work also with arrays, right?

vuillaut: >  But that is (conceptually) more correct than using directly the distance. 

Ha yes, you are right.

lstchain/reco/disp.py
moralejo: Can we merge? I'd rather not change the arguments of the functions.

rlopezcoto (approved):

PR #1028, opened 2022-10-21
Remove numba aot compilation, use njit instead

maxnoe:
This brings us back to a pure python distribution, fixing the wheel upload to pypi


vuillaut:
It seems that the installed astropy version (4.3.1) is not compatible with numpy 1.23

vuillaut:
LGTM
What was the motivation to use aot compilation if we have the same performances with jit??

vuillaut:
I will approve but that would be better to have a review from @gabemery as well.

vuillaut (approved):

gabemery (approved):
The changes look correct. This will lead to a slight loss of processing speed if my previous tests are still correct but to avoid large issues it is still worth it to make the change I think. I already approve but will do some tests of the effect on processing time tomorrow or on Monday. And I will update #1006 accordingly.

PR #1030, opened 2022-10-21
Recalculate sin_az_tel

moralejo:
Allows to get DL1 file with the sin_az_tel parameter just running the DL1ab script.

### DIFF
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -328,7 +328,7 @@ def main():
                     dl1_container['disp_angle'] = disp_angle
                     dl1_container['disp_sign'] = disp_sign
 
-                dl1_container['sin_az_tel'] = np.sin(infile.root[dl1_params_lstcam_key].col('az_tel'))
+                dl1_container['sin_az_tel'] = np.sin(params['az_tel'][ii])
                     
                 for p in parameters_to_update:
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -327,6 +328,8 @@ def main():
                     dl1_container['disp_angle'] = disp_angle
                     dl1_container['disp_sign'] = disp_sign
 
+                dl1_container['sin_az_tel'] = np.sin(infile.root[dl1_params_lstcam_key].col('az_tel'))
+
```
vuillaut, line 332: Hum, this is setting a whole column to a single event value (where there should be a float)

vuillaut (changes requested):

lstchain/scripts/lstchain_dl1ab.py
vuillaut: I think you want
```
dl1_container['sin_az_tel'] = np.sin(infile.root[dl1_params_lstcam_key].col('az_tel')[ii])
```

vuillaut: or even
```
dl1_container['sin_az_tel'] = np.sin(params['az_tel'][ii])
```

vuillaut (approved):

PR #1033, opened 2022-10-24
checking if parameters exist before trying to update them

### LINKED ISSUE #1032: Can't apply dl1ab to former lstchain versions files
New parameter `sin_az_tel` is not present in older files.
Applying dl1ab to them fails when we try to recompute it.
```
"lstchain_dl1ab.py", line 334, in main
    params[ii][p] = u.Quantity(dl1_container[p]).value
ValueError: no field of name sin_az_tel
```

vuillaut:
Fixes #1032 

That is the simplest solution at the moment, the alternative being to revert entirely the `sin_az_tel` addition.

### DIFF
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -226,8 +226,12 @@ def main():
             copy_h5_nodes(infile, outfile, nodes=nodes_keys)
             add_source_filenames(outfile, [args.input_file])
 
+            params_node = outfile.root[dl1_params_lstcam_key]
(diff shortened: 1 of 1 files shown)


moralejo:
I think this is the best solution for the time being.

Given that the only reason to recompute sin_az_tel was that it is not present in the current version of DL1 files, it does not make sense to keep it (in this version) in the list of parameters to recompute, right?

rlopezcoto (approved):
let's get this one merged, I think it is perfectly ok if we cannot calculate parameters introduced in later versions in low level data processed with previous ones.

Issue #1021, opened 2022-10-05
lstchain_create_dl3_index_files does not work correctly if it is executed in the same directory as DL3 data


katagirihide:
If it is executed in the same directory, it reads only one file. The same situation was confirmed by Seiya.

(lst-dev) [hideaki.katagiri@cp02 22091601_lst1_crab]$ lstchain_create_dl3_index_files --input-dl3-dir ./ --file-pattern dl3_LST-1.Run*.gz --overwrite
2022-10-04 06:51:30,212 WARNING [ctapipe.FITSIndexWriter] (lstchain_create_dl3_index_files.setup): Overwriting /fefs/aswg/workspace/hideaki.katagiri/work/22091601_lst1_crab/hdu-index.fits.gz
2022-10-04 06:51:30,214 WARNING [ctapipe.FITSIndexWriter] (lstchain_create_dl3_index_files.setup): Overwriting /fefs/aswg/workspace/hideaki.katagiri/work/22091601_lst1_crab/obs-index.fits.gz
Run 2929 does not contain HDU BACKGROUND
Run 2929 does not contain HDU PSF
(lst-dev) [hideaki.katagiri@cp02 22091601_lst1_crab]$ mkdir tmp
(lst-dev) [hideaki.katagiri@cp02 22091601_lst1_crab]$ cd tmp/
(lst-dev) [hideaki.katagiri@cp02 tmp]$ lstchain_create_dl3_index_files --input-dl3-dir ../ --file-pattern dl3_LST-1.Run*.gz --overwrite
2022-10-04 08:49:46,607 WARNING [ctapipe.FITSIndexWriter] (lstchain_create_dl3_index_files.setup): Overwriting /fefs/aswg/workspace/hideaki.katagiri/work/22091601_lst1_crab/tmp/../hdu-index.fits.gz
2022-10-04 08:49:46,608 WARNING [ctapipe.FITSIndexWriter] (lstchain_create_dl3_index_files.setup): Overwriting /fefs/aswg/workspace/hideaki.katagiri/work/22091601_lst1_crab/tmp/../obs-index.fits.gz
Run 2929 does not contain HDU BACKGROUND
Run 2929 does not contain HDU PSF
Run 2930 does not contain HDU BACKGROUND
Run 2930 does not contain HDU PSF
Run 2931 does not contain HDU BACKGROUND
Run 2931 does not contain HDU PSF
Run 2950 does not contain HDU BACKGROUND
[...]

PR #1034, opened 2022-10-25
Add another example message to index DL3 files

chaimain:
To help understand #1021

### DIFF
--- a/lstchain/tools/lstchain_create_dl3_index_files.py
+++ b/lstchain/tools/lstchain_create_dl3_index_files.py
@@ -39,10 +39,17 @@ class FITSIndexWriter(Tool):
         --overwrite
 
     Or if the DL3 files are stored in sub-directories:
-    < lstchain_create_dl3_index_files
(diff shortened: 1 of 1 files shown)


lstchain/tools/lstchain_create_dl3_index_files.py
```diff
@@ -39,10 +39,17 @@ class FITSIndexWriter(Tool):
         --overwrite
 
     Or if the DL3 files are stored in sub-directories:
-    < lstchain_create_dl3_index_files
+    > lstchain_create_dl3_index_files
        -d /path/to/DL3/files/
        -o /path/to/DL3/index/files
-       -p /sub-directory*/dl3*[run_1-run_n]*.fits   
+       -p /sub-directory*/dl3*[run_1-run_n]*.fits
```
maxnoe, line 45: Please also add the quotes here. The pattern should always be quotes.

chaimain: Done

PR #1035, opened 2022-10-25
Implementing suggestions for better evaluation of energy-dependent theta cuts

chaimain:
With taking suggestions from #997 to redefine the "100% range of gamma events" to be used to evaluate percentile cuts on theta value of events, and adding the following -
- An event filter function to take only those with the same `disp_sign` value after reconstruction (applied to only MC).
- New variable in `DL3Cuts` component: `max_theta_range` to create a custom maximum limit of theta value to be considered as "100% of central PSF region", based on which, percentile cuts can be evaluated for each `reco_energy` bins


lstchain/io/event_selection.py
```diff
@@ -43,10 +46,19 @@ class EventSelector(Component):
 
     def filter_cut(self, events):
         """
-        Apply the event filters
+        Apply the standard event filters.
         """
         return filter_events(events, self.filters, self.finite_params)
 
+    def same_disp_sign(self, events):
```
vuillaut, line 53: maybe add `filter` in the function name would be more clear?

lstchain/io/event_selection.py
```diff
@@ -43,10 +46,19 @@ class EventSelector(Component):
...
+        """
+        Apply a mask to have events with the reconstructed disp sign,
+        be the same as the true, to avoid wrong head-tail assignment
+        in the reconstructed image of the MC data.
```
vuillaut, line 57: docstring of input/output would also be useful

lstchain/io/event_selection.py
```diff
@@ -88,6 +100,12 @@ class DL3Cuts(Component):
         default_value=0.32,
     ).tag(config=True)
 
+    max_theta_range = Float(
+        help="Maximum value of theta (deg), approximate for the PSF size," +
+            " to calculate the percentile cut in each energy bin.",
+        default_value=0.7,
```
vuillaut, line 106: why 0.7?

chaimain: It is used in the Performance paper, and I so kept this value as default/nominal. Maybe each analyzer can select a different one for the specific data-MC comparison.

lstchain/io/event_selection.py
chaimain: Will the change in the docstring be sufficient?

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -461,6 +462,7 @@ def start(self):
                 background = self.cuts.apply_global_gh_cut(background)
 
             background = self.event_sel.filter_cut(background)
+            background = self.event_sel.same_disp_sign_filter(background)
```
moralejo, line 465: I don't understand what is the goal of this.

The only case in which we may want to apply the selection "only the events in which the head-tail estimate was correct" is when dealing with gammas, and for the sole purpose of finding an efficiency-based theta cut (or gammaness cut). The idea is to keep after the cut a given fraction of events in "the main peak of the PSF". Of course, in real data we don't know which events have right or wrong head-tail assignment, but it does not matter because (from the MC) we just obtain theta (and gammaness) cut values to be applied to those real data.

In no case we want to apply that "correct head-tail" cut  to the background, because what would it mean anyway? The head-tail finding is trained for gammas, and for most of the background events the outcome of the disp_sign RF will be more or less random, not related to the actual proton direction.

lstchain/io/event_selection.py
chaimain: The first option, use the 0.7 deg max theta value when evaluating and applying an efficiency cut

lstchain/tools/lstchain_create_irf_files.py
chaimain: I did not consider the implication of this selection on the background properly and had just proceeded to keep the event selection uniform. I will correct this.

maxnoe: >. I am not sure why you would not want the IRFs to be created with the MC gamma events with the same disp_sign reconstruction. can you please further clarify on this?

IRFs (the actual IRFs, edisp, aeff and psf) must be computed on exactly the cuts you perform on real data. You must not apply any other cuts. It is ok to remove the wrong disp sign events for the optimization of cuts, but the calculation on IRFs must happen on all events.

chaimain: Ok! I understand now, thanks for the clarification. I will remove the changes made in the IRF Tool, and also in the evaluation of energy-dependent theta cuts function, but keep the `same_disp_sign_filter` function.

rlopezcoto:
@chaimain after discussing in private, I think this PR can be closed, right?

chaimain:
> @chaimain after discussing in private, I think this PR can be closed, right?

Sorry, I couldn't recheck it yet. Will do it today

chaimain:
After discussing with @moralejo, I have removed the `max_theta_range` implementation for limiting the 'max' theta range for evaluating percentile cuts, and as the comparison of `disp_sign` is only essential in this context for the MC data, I moved that functionality to the function `energy_dependent_theta_cuts`, as the default option in the function, using the boolean `use_same_disp_sign`. One can choose not to perform such a selection with this boolean argument.

lstchain/io/event_selection.py
```diff
@@ -186,16 +183,33 @@ def apply_global_theta_cut(self, data):
...
+        data,
+        energy_bins,
+        use_same_disp_sign=True,
+        smoothing=None,
     ):
         """
-        Evaluating an optimized energy-dependent theta cuts, in a given
-        data, with provided reco energy bins, and other parameters to
-        pass to the pyirf.cuts.calculate_percentile_cut function.
+        Evaluating an optimized energy-dependent theta cuts, in a given MC data,
```
moralejo, line 193: Please clarify, "optimized" is a bit misleading. We simply evaluate the cut which has a given efficiency on a given sample of MC gammas (namely, those with correct disp_sign).

lstchain/io/event_selection.py
```diff
@@ -186,16 +183,33 @@ def apply_global_theta_cut(self, data):
...
+        with provided reco energy bins, and other parameters to pass to the
+        pyirf.cuts.calculate_percentile_cut function.
+
+        For MC data, the reconstructed disp_sign may not be the same as the
+        True value, and thus resulting in a distinct angular distribution at
+        larger theta values owing to the events with incorrect disp_sign. This
+        affects the total range over which percentile cuts are evaluated. To
+        fix this by default we apply a mask on the data, so as to only use
+        events with the same disp_sign after reconstruction, for evaluating the
```
moralejo, line 202: I think the rationale for doing this is missing in this explanation.
Due to the events with wrongly reconstructed disp_sign, The PSF is bi-modal, and what we typically want is to keep a given fraction of the central peak.

lstchain/io/event_selection.py
```diff
@@ -89,8 +89,8 @@ class DL3Cuts(Component):
     ).tag(config=True)
 
     fill_theta_cut = Float(
-        help="Fill value of theta cut (deg) in an energy bin with fewer " +
-            "than minimum number of events present",
+        help="Fill value of theta cut (deg) in an energy bin with fewer "
+        + "than minimum number of events present",
         default_value=0.32,
```
moralejo, line 94: A fixed default value is probably not the best approach.
I think on the higher energy end we should use the value for the highest energy bin in which the cut could be determined. And for the lower energies, the cut value for the lowest energy bin where it could be determined.

moralejo: But may be something for a separate PR.

moralejo (changes requested):

lstchain/io/event_selection.py
```diff
@@ -186,16 +183,33 @@ def apply_global_theta_cut(self, data):
...
+        pyirf.cuts.calculate_percentile_cut function.
+
+        For MC events, the disp_sign may be reconstructed incorrectly with
+        respect to the true value, and thus resulting in a bi-modal PSF.
+        For evaluating the energy-dependent theta cuts, we want to consider,
+        only the central region of PSF. To fix this issue, by default, we apply
+        a mask on the data, so as to only use events with the same disp_sign
+        after reconstruction, for evaluating the percentile cut.
+        If the user wishes to not use this method, they can make the boolean
+        False.
```
moralejo, line 204: "Note that in that case, at low energies, where disp_sign determination is pretty uncertain, an efficiency  of 40% or larger will result in a cut which keeps the whole central part of the PSF".

moralejo (approved):

Issue #1009, opened 2022-07-20
Do not use `create_bins_per_decade` for IRF binnings, allow binning per irf


maxnoe:
`create_bins_per_decade` is designed to create *exactly* the number of bins per decade as required by CTA for sensitivity.

As a result, the upper bin edge cannot be chosen freely, but is a result of the lower limit and the n bins per decade.

E.g.:

```
In [7]: create_bins_per_decade(1 * u.TeV, 260 * u.TeV, 5)
Out[7]: 
<Quantity [  1.        ,   1.58489319,   2.51188643,   3.98107171,
             6.30957344,  10.        ,  15.84893192,  25.11886432,
            39.81071706,  63.09573445, 100.        , 158.48931925,
           251.18864315] TeV>
```

```
In [7]: create_bins_per_decade(1 * u.TeV, 260 * u.TeV, 5)
Out[7]: 
<Quantity [  1.        ,   1.58489319,   2.51188643,   3.98107171,
             6.30957344,  10.        ,  15.84893192,  25.11886432,
            39.81071706,  63.09573445, 100.        , 158.48931925,
           251.18864315] TeV>
```

This is not a bug! This is what this function was designed for. However, for IRFs, this restriction doesn't make sense and we should be able to set arbitrary energy limits.

Solution: switch to using `np.geomspace(low, high, n_bins + 1)` for the binning of IRFs instead of `create_bins_per_decade`.

It is probably also needed to specify the number of bins per IRF / cut (we probably don't want to be forced to have rad max, gh cuts, edisp and aeff in the same binning.)

maxnoe:
> np.geomspace(low, high, n_bins + 1)

Or alternatively use the similar gammapy `MapAxis.from_energy_bounds` where the per decade is not exact but is changed to exactly match the limits:

```
In [10]: MapAxis.from_energy_bounds(1 * u.TeV, 200 * u.TeV, 5, per_decade=True).edges
Out[10]: 
<Quantity [  1.        ,   1.55507915,   2.41827118,   3.76060309,
             5.84803548,   9.09415806,  14.14213562,  21.9921403 ,
            34.19951893,  53.18295897,  82.70371084, 128.61081668,
           200.        ] TeV>```

chaimain:
> It is probably also needed to specify the number of bins per IRF / cut (we probably don't want to be forced to have rad max, gh cuts, edisp and aeff in the same binning.)

[...]

PR #1036, opened 2022-10-25
Implementing suggestions for IRF binning

chaimain:
Taking the suggestion from #1009 to replace the pyirf function with `numpy.geomspace` function to create energy bins for IRFs by providing the number of bins required. Currently, the bins are the same for all IRFs and DL3 cuts.


rlopezcoto (commented):
this is going to change the binning of all irfs and will make them incompatible with all the ones produced with previous lstchain versions (at least for the default settings), then I would suggest that this one should be merged whenever we go for a major release to avoid incompatibility problems, opinions?

rlopezcoto:
@chaimain can you change the target branch, instead of the `master` do it in the `ctapipe_v0.17`, please? Everybody else, please be aware that by merging this *any DL3 files produced with lschain < v0.10 will not be compatible with those produced with later versions*

rlopezcoto (approved):

PR #1044, opened 2022-11-05
truly optional config image_modifier

vuillaut:
I suppose the intent was to have these optional, since a default value was defined.
However, the reading from the config was mandatory, which is a contradiction.

This PR fixes this.


rlopezcoto (dismissed):

rlopezcoto:
sorry, I went too fast, it seems that there is a problem with the location of the variables definition and tests are not passing

rlopezcoto (approved):

PR #1051, opened 2022-11-30
Patches to calibration code

FrancaCassol:
Small improvements and corrections to the present calibration code


rlopezcoto:
tests are not passing, it seems that there are some problems with an f-string, I'll review when this is fixed

rlopezcoto (approved):

Issue #1056, opened 2022-12-08
fixing joblib version?


vuillaut:
Hi,
@FrancaCassol has had some issues lately with loading RFs from disk, see below.
```
Traceback (most recent call last):
 File "/fefs/aswg/workspace/franca.cassol/software/miniconda3/envs/cta_dev/bin/lstchain_dl1_to_dl2", line 8, in <module>
  sys.exit(main())
 File "/fefs/aswg/workspace/franca.cassol/software/cta-lstchain/lstchain/scripts/lstchain_dl1_to_dl2.py", line 248, in main
  models_dict = {'reg_energy': joblib.load(Path(args.path_models, 'reg_energy.sav'))}
 File "/fefs/aswg/workspace/franca.cassol/software/miniconda3/envs/cta_dev/lib/python3.8/site-packages/joblib/numpy_pickle.py", line 587, in load
  obj = _unpickle(fobj, filename, mmap_mode)
 File "/fefs/aswg/workspace/franca.cassol/software/miniconda3/envs/cta_dev/lib/python3.8/site-packages/joblib/numpy_pickle.py", line 506, in _unpickle
  obj = unpickler.load()
 File "/fefs/aswg/workspace/franca.cassol/software/miniconda3/envs/cta_dev/lib/python3.8/pickle.py", line 1212, in load
  dispatch[key[0]](self)
KeyError: 1
```

This is due to a different version of joblib in the env. that created the RF (joblib v1.2) and the one used to load the RF (joblib v1.1).

In theory, I'd argue that the same env should be used throughout the whole analysis, to avoid errors and for provenance issues.
However, in practice, that would often mean a lot of re-computing, especially to retrain RF in case of minor patches (the case here, between lstchain v0.9.9 and v0.9.11). And since the analysis is done in several steps anyway, one could argue that the provenance is not broken: MC has been done with xx and data analysis with yy.

Enough for the philosophical talk, technically, I think we should fix joblib version and update it only when doing a major bump (that would always require redoing RF anyway).

PR #1057, opened 2022-12-08
fixing joblib version to 1.2

vuillaut:
Fixing #1056

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - matplotlib=3.5
   - notebook
   - iminuit>=2
-  - joblib==1.2
+  - joblib~=1.2.0
   - toml
   - protozfits=2.0
   - protobuf=3.20
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - matplotlib=3.5
   - notebook
   - iminuit>=2
-  - joblib~=1.1
+  - joblib==1.1
   - toml
   - protozfits=2.0
   - protobuf=3.20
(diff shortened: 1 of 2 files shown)


maxnoe (changes requested):
@vuillaut No! This is the wrong solution. There are files already produced with joblib 1.2, which are not readable with 1.1. So this would force everyone to use 1.1, which is incompatible with some files produced.

The correct solution is to pin 1.2

environment.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - matplotlib=3.5
   - notebook
   - iminuit>=2
-  - joblib
+  - joblib~=1.1
```
maxnoe, line 18: This means ">=1.1,<2.0"  so it does not in fact pin to 1.1

environment.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - matplotlib=3.5
   - notebook
   - iminuit>=2
-  - joblib
+  - joblib==1.2
```
maxnoe, line 18: I think this will pin to `1.2.0`, we should pin to `1.2.x` I think. `joblib=1.2` here

setup.py
```diff
@@ -48,7 +48,7 @@ def find_scripts(script_dir, prefix):
         'gammapy~=0.19.0',
         'h5py',
         'iminuit>=2',
-        'joblib',
+        'joblib==1.2',
```
maxnoe, line 51: and `~=1.2.0` here

environment.yml
vuillaut: does it? 🤔 
Damn, ok changing

maxnoe (approved):

PR #1064, opened 2022-12-20
Added init function to set the histograms' binnings

moralejo:
This is needed because default values (now set to None) made the docs build fail


moralejo:
A better solution was found!

PR #1065, opened 2022-12-22
URGENT: Add change of filter wheels' name in Mongo DB

FrancaCassol:
On December 5th there was a change of the Mongo DB filter wheel names.
This PR should solve the present  problem to access the filter wheel number with the calibration script.
In order to have it active onsite, one should create a new release

However, **notice that** the change was corrupted till the 21st of December, so if no correction of the DB will be performed, the period from 5/12/2022 till 21/12/2022 will remain without automatic access to the filter wheel number

### DIFF
--- a/lstchain/scripts/onsite/onsite_create_calibration_file.py
+++ b/lstchain/scripts/onsite/onsite_create_calibration_file.py
@@ -13,7 +13,7 @@ import os
 import subprocess
 import sys
 from pathlib import Path
-
+from astropy.time import Time
 import pymongo
 
 import lstchain
@@ -98,7 +98,6 @@ optional.add_argument(
         " Should be used only for data from before 2022"
     )
 )
-
 
 
 def main():
@@ -255,6 +254,10 @@ def main():
 
 def search_filter(run, database_url):
     """read the employed filters form mongodb"""
+
+    # there was a change of Mongo DB data names on 5/12/2022
+    NEW_DB_NAMES_DATE = Time("2022-12-04T00:00:00")
+
     filters = None
     try:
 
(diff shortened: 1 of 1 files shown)


morcuended (approved):
The hotfix is fine with me for the time being. However, we should probably change to fetch this information from the TCU database instead at some point.

maxnoe:
This is not how databases are supposed to work!

If the schema is changed, a schema migration needs to happen changing also the old entries to the new format.

Who is responsible for filling these values? Why were these keys changed? How are we supposed to learn that this happened? And why were the keys not updated for older entries?

morcuended:
> Hi @morcuended, @maxnoe, I got the information on how to fetch the data from the TCU data-base. I can modify this PR in order to used the TCU DB starting from December 2022 (for previous period better we use the Caco DB, for which we know already the (in)stability and that was the only one filled in the first data taking period), what do you think?

You propose to access different DB based on date? I would suggest that until we have the CaCo-TCU DBs sorted out we stick with the CaCo with the current hotfix for the new entry naming.

> If the schema is changed, a schema migration needs to happen changing also the old entries to the new format.
Who is responsible for filling these values? Why were these keys changed? How are we supposed to learn that this happened? And why were the keys not updated for older entries?

I have no answer to none of these questions. We used to fetch this information from the CaCo DB (camera team responsible) Now we are advised to get this information from TCU observation summaries DB (TCU team responsible). I guess the proper way to go would be to dump old CaCo entries into the new TCU DB and use a unique entry name for the filter wheels. Then we would just have to access only TCU DB regardless of the date.

FrancaCassol:
> 
> You propose to access different DB based on date? I would suggest that until we have the CaCo-TCU DBs sorted out we stick with the CaCo with the current hotfix for the new entry naming.

I agree, then @rlopezcoto, @moralejo, @maxnoe  could this PR rapidly approved and the new release installed otherwise the online calibration check plots are not produced (see report of commissioning meeting of today)

PR #1066, opened 2023-01-11
Use numba njit code to inject NSB in R1 waveforms

gabemery:
New version of #1006 done in a more recent version of lstchain (the diff ended up being the same, but it cleans up the commit history). Significantly accelerate NSB injection on waveforms compared to the current implementation using scipy.interpd1d every time.

### DIFF
--- a/lstchain/reco/reconstructorCC.py
+++ b/lstchain/reco/reconstructorCC.py
@@ -183,7 +183,7 @@ def linval(a, b, x):
 
 
 @njit(cache=True)
-def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
+def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
     """
     Fast template interpolator using uniformly sampled base with known origin and step.
     The algorithm finds the indexes between which the template is needed and performs a linear interpolation.
@@ -192,7 +192,7 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
     ----------
     gain: boolean 1D array
         Gain channel used per pixel
-    times: float64 1D array
+    times: float64 2D array
         Times of each waveform samples
     t0: float64
         Time of the first value of the pulse templates
@@ -202,8 +202,6 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
         Template values for the high gain channel
     a_lg: float64 1D array
         Template values for the low gain channel
-    size: int64
-        Number of element in a_hg and a_lg
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


lstchain/reco/reconstructorCC.py
```diff
@@ -212,18 +210,19 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
...
     n, m = times.shape
+    size = a_hg.shape[0]
     out = np.empty((n, m))
     for i in range(n):
         for j in range(m):
             # Find the index before the requested time
             a = (times[i, j]-t0)/dt
             t = int(a)
-            if a < size:
+            if t+1 < size:
```
rlopezcoto, line 220: why was this modified in one unit?

lstchain/reco/reconstructorCC.py
```diff
@@ -192,7 +192,7 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg, size):
     ----------
     gain: boolean 1D array
         Gain channel used per pixel
-    times: float64 1D array
+    times: float64 2D array
```
rlopezcoto, line 195: did the variable `times` change its shape or was the description wrong before?

lstchain/reco/reconstructorCC.py
maxnoe: Use <= instead of the awkward +  1 <

gabemery: Sorry I was wrong in what I was saying before since a<size also guaranties that t is less than size. The change is now quite old so I didn't fully remember. 
But we use both the t and t+1 element in the following code. So the error occurred when t+1 was equal to size. Since the array for interpolation was usually binned in fractions of nano second it was a rare bug I saw quite late. (there is no a==size issue).

I made the decision on checking on the value of t instead of a in the fix because this is what is used in the array afterwards.

The needed criteria is thus that t+1 is the index of an array element (so < size = len(array)).

@maxnoe <= wouldn't work, it goes in the wrong direction. It would be +2 <=

maxnoe: ah, right, sorry...

rlopezcoto (approved):

gabemery:
I was told that a problematic behaviour was observed with this branch in an extreme injection case (nominal NSB X 32). I will try to reproduce/understand the issue. Since users are still limited for this feature I propose to wait for this check before merging.

gabemery:
2 Issues is fixed:
 - generally important : add check that we don't interpolate before the start of the provided template when injecting pulses (0 < t+1 in reconstructorCC.py)
 - small improvement for small injection, important for larger ones : extended the pre-signal time window for pulse injection by 5 samples to better account for tails in the pulse (for information, the pulse peaks at ~9ns and previous version injected pulses starting up to 20ns before the data. Now it is 25ns. I tried 30ns without seeing effects)

PR #1071, opened 2023-02-06
Changes for application of calibration tool to interleaved events and moon data

FrancaCassol:
This PR  add some feature to the calibration tool:
- It add an outlier criteria based on the number of estimated pe of FF events. This permits to keep pixels that have a lower HV but that can be in any case well calibrated and used
- It add the possibility to calibrate the low gain scaling the HG results. This is useful in the case of low HV due to high moon, where the low gain intensity could be not sufficient for the F-factor method.
- It modify the logic of the calibration, based on the new possibility to process in series several subruns

All this permit to apply the calibration tool to the interleaved events, which is (provisionally) useful in the case of changing moon condition during the night. The complete treatment of interleaved events for creating the Cat-B coefficient will be the object of a coming PR


lstchain/calib/camera/calibration_calculator.py
```diff
@@ -195,10 +215,33 @@ def calculate_calibration_coefficients(self, event):
...
+        npe_outliers = (
+            np.logical_or(npe_deviation < self.npe_median_cut_outliers[0] * npe_signal_median[:,np.newaxis],
+                          npe_deviation > self.npe_median_cut_outliers[1] * npe_signal_median[:,np.newaxis]))
 
+        # calibration unusable pixels are an OR of all masks
+        calib_data.unusable_pixels = np.logical_or(unusable_pixels, npe_outliers)
+        
+        # give to the unusable pixels the median camera value for the dc_to_pe and pedestal
+        # (these are the starting data for the Cat-B calibration)        
+        dc_to_pe_masked=np.ma.array(calib_data.dc_to_pe, mask=calib_data.unusable_pixels)
```
rlopezcoto, line 229: `dc_to_pe_masked = np.ma.array(calib_data.dc_to_pe, mask=calib_data.unusable_pixels)`

lstchain/data/lstchain_standard_config.json
```diff
@@ -237,6 +237,7 @@
 
   "LSTCalibrationCalculator":{
     "systematic_correction_path": null,
+    "npe_median_cut_outliers": [-4,4],
```
rlopezcoto, line 240: where are these numbers coming from?
Also: `[-4, 4]`

lstchain/data/lstchain_standard_config.json
```diff
@@ -261,7 +262,7 @@
       "tel_id":1,
       "time_sampling_correction_path": null,
       "charge_product":"LocalPeakWindowSum",
-      "charge_median_cut_outliers": [-0.5,0.5],
+      "charge_median_cut_outliers": [-0.9,2],
```
rlopezcoto, line 265: why did this change, for the application of the cut also to moon data? we should in this case consider a couple of cases because for standard data we may be missing some outliers because of a stronger cut.

lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -127,12 +125,7 @@ def setup(self):
 
             if self.eventsource.r0_r1_calibrator.drs4_pedestal_path.tel[tel_id] is None:
                 raise IOError("Missing (mandatory) drs4 pedestal file in trailets")
-
-            # remember how many events in the files
-            self.tot_events = len(self.eventsource.multi_file)
```
rlopezcoto, line 132: not needed anymore?

lstchain/data/lstchain_standard_config.json
FrancaCassol: From statistical considerations, but you are right I am now working with a more conservative 5 sigma cut, i.e : a pixels is tagged as unusable if the estimated number of phe is 5 sigma far from the median camera value (keep in mind that de camera median dispersion is  7%).

lstchain/data/lstchain_standard_config.json
FrancaCassol: Yes.  For that I added the above cut on the number of estimated photo-electrons (npe_median_cut_outliers)

lstchain/tools/lstchain_create_calibration_file.py
FrancaCassol: I took it out because it was no more calculated with the new multifile reader, and honestly with the reading of more subruns it is no more necessary (we will now discard the very last calibration sample which has in any case a lower statistics, hence it is better not to use it)

lstchain/data/lstchain_standard_config.json
rlopezcoto: and that one is applied exactly how to differentiate between moon and dark data?

lstchain/tools/lstchain_create_calibration_file.py
rlopezcoto: but the field will be empty, right?

rlopezcoto: `self.tot_events`

lstchain/data/lstchain_standard_config.json
rlopezcoto: ok, but based on what? could you please add some comments on how these numbers were obtained?

lstchain/data/lstchain_standard_config.json
rlopezcoto: so shall they be changed to 5?

FrancaCassol: Done

lstchain/tools/lstchain_create_calibration_file.py
FrancaCassol: this filed is not more there actually

lstchain/data/lstchain_standard_config.json
FrancaCassol: These numbers were obtained on general considerations and experience, they are much less strict that what we had before (thanks to the cut on n_pe),  they ask that a pixel have a signal of at least 10% the camera median because otherwise the cat-B calibration can not be performed with enough precision  (recall that high NSB reduces the gain of a factor ~ 6, for which we an safely perform a Cat-B calibration ) and they must have a signal not  higher than two times the camera median, this is just to detect faulty pixels with uncontrolled high signal (the HV FF is 3% and in case of moon we have now also a FF low HV table). I am afraid one can not put comments in a json file. One should write an internal note with all this inside, let's say that before we start the next mass production I plan to confirm all these cuts with a study on dark and moon data.

rlopezcoto (approved):

PR #1077, opened 2023-02-13
Extra srcdep parameters and limit the offset of gamma MC for training

SeiyaNozaki:
This PR implements extra two parameters for source-dep analysis defined in the following slides.
Any comments are welcome, especially the naming of the parameters...!
https://indico.cta-observatory.org/event/4631/contributions/38096/attachments/23113/33203/20230213_srcdep_update.pdf

In addition, this PR changes the default value of `src_r_cut of MC gamma for the RF training.


lstchain/reco/dl1_to_dl2.py
```diff
@@ -478,10 +478,13 @@ def build_models(filegammas, fileprotons,
         # Apply the temporary disp norm regressor and sign classifier to the test set
         disp_norm = tmp_reg_disp_norm.predict(test[config['disp_regression_features']])
         disp_sign = tmp_cls_disp_sign.predict(test[config['disp_classification_features']])
+        disp_sign_proba = tmp_cls_disp_sign.predict_proba(test[config['disp_classification_features']])
```
maxnoe, line 481: There is no need to call predict and predict proba, that's wasteful. Since we now want the proba, you should change the `disp_sign` calculation to use the proba to not evaluate the random forest twice:


```
column = list(tmp_cls_disp_sign.classes_).index(1)
disp_sign = np.where(disp_sign_proba[:, col] > 0.5, 1, -1)
```

SeiyaNozaki: Thanks for the comment! Actually, I just copy & paste the codes of gammaness computation, so this part can also be updated to reduce computation time
https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/reco/dl1_to_dl2.py#L652-L653

rlopezcoto:
> CI failed since only a single event survives after off-axis cut (src_r < 1 degree) ... Is it possible to include several low off-axis events in the test MC gamma dataset?

I think it was @maxnoe the one creating those datasets, right? Would it be possible to create one including more inner events now that we have seen an improvement in the performance of src-dependent analysis? 
Otherwise we could modify the tests, @SeiyaNozaki aren't there any events surviving because of the cuts (just relax them) or overall without applying any additional constraints?

rlopezcoto:
so @maxnoe would it be possible to get a different test dataset with more off-axis events or shall we change the tests for them to pass after these conditions are applied?

maxnoe:
> so @maxnoe would it be possible to get a different test dataset with more off-axis events or shall we change the tests for them to pass after these conditions are applied?

No sure what simulations you are talking about, but I can't remember creating any.

If you need a new small test file with specific settings, you should probably ask the simulation team to create it, like I did here:
https://github.com/cta-observatory/lst-sim-config/issues/51

SeiyaNozaki:
ah I see Max's comment now, but now I prepared fake DL1 MC gamma file (like fake proton MC) by changing `src_x` and `src_y` to zeros to keep events after src_r cut. So now it should work with the current test data.

rlopezcoto (approved):

PR #1078, opened 2023-02-14
update ctapipe to 0.18

maxnoe:
Try updating to 0.18, needs release of the source first, so just draft for now

### DIFF
--- a/lstchain/image/muon/muon_analysis.py
+++ b/lstchain/image/muon/muon_analysis.py
@@ -15,8 +15,7 @@ from ctapipe.image.muon import (
     MuonIntensityFitter,
     MuonRingFitter
 )
-from ctapipe.image.muon.features import ring_containment
-from lstchain.ctapipe_compat import ring_completeness
+from ctapipe.image.muon.features import ring_containment, ring_completeness
 
 from lstchain.image.muon import plot_muon_event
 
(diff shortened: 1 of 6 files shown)


lstchain/image/muon/muon_analysis.py
```diff
@@ -274,8 +271,8 @@ def analyze_muon_event(subarray, tel_id, event_id, image, good_ring_config, plot
             x[dist_ringwidth_mask], y[dist_ringwidth_mask],
             image[dist_ringwidth_mask],
             muonringparam.radius,
-            muonringparam.center_x,
-            muonringparam.center_y,
+            muonringparam.center_fov_lon,
```
rlopezcoto, line 274: aren't the units of these coordinates changing? I do not exactly recall, but I think that `muonringparam.center_x` was in `m` and  `muonringparam.center_fov_lon` looks like to be in degrees. The second one also sounds like sky coordinates instead of camera ones, but it may not be the case

rlopezcoto (commented):
the new `ctapipe_io_lst` was released so this PR should be ready for review, right?
now that we are on it... what about ctapipe v0.19? did many things change in the new version so it would be too much work to adapt the code to it?

lstchain/image/muon/muon_analysis.py
maxnoe: No, this was inconsistent before in ctapipe. it was already in deg before, but named x and y.

maxnoe:
> did many things change in the new version so it would be too much work to adapt the code to it?

No, basically nothing changed, only the new `FlashCamExtractor`

rlopezcoto:
> > did many things change in the new version so it would be too much work to adapt the code to it?
> 
> No, basically nothing changed, only the new `FlashCamExtractor`

Great, I see that you opened https://github.com/cta-observatory/ctapipe_io_lst/pull/176 and I guess another one in lstchain will follow, shall we get this PR merged and evaluate later the one on ctapipe v0.19?

rlopezcoto (approved):

PR #1086, opened 2023-04-28
code snippets in docs

vuillaut:
Add a code snippets section in the docs.


rlopezcoto (approved):

morcuended:
Looks good to me. I would also add the way of accessing the configuration used to produce a given file:

```python
with tables.open_file(filename) as file:
    config = yaml.safe_load(file.root.dl1.event.telescope.parameters.LST_LSTCam.attrs["config"])
print(config)
```

and maybe also which DRS4 baseline correction and calibration coefficients files were used since they are also stored as metadata in the file.

vuillaut:
> Looks good to me. I would also add the way of accessing the configuration used to produce a given file:
> 
> ```python
> with tables.open_file(filename) as file:
>     config = yaml.safe_load(file.root.dl1.event.telescope.parameters.LST_LSTCam.attrs["config"])
> print(config)
> ```
> 
> and maybe also which DRS4 baseline correction and calibration coefficients files were used since they are also stored as metadata in the file.

If I merge, could you add this info in another please?

morcuended:
Sure, go ahead with the merging and I will include the additional snippet
in another PR.

PR #1087, opened 2023-05-02
More complete doc on Analysis steps and configs to use

vuillaut:


docs/introduction.rst
```diff
@@ -35,40 +35,104 @@ Analysis steps
 
 R1 to DL1
 ---------
-Usage of
 
-``lstchain.scripts.lstchain_data_r0_to_dl1``
+MC data
+^^^^^^^
```
chaimain, line 40: Not sure if this sub-section will be easily understood on Sphinx. Maybe just writing in a single section on both MC and real data, or listing them, might be more useful. There might be better ways to make it easily readable on Sphinx.

Maybe referencing the `lst_analysis_workflow` would be helpful to understand the sub-sections where there are distinct procedures for MC and real data, and when they are similar.

This is also applicable to the `DL2 to DL3` section

docs/introduction.rst
```diff
@@ -35,40 +35,104 @@ Analysis steps
...
+
+Here is an example configuration file for the IRF creation step.
+
+.. toggle:: 
+
+    .. include:: examples/dl3_tool_config.json
+        :code: json
+
+
+Post DL3 analysis
```
chaimain, line 134: This has to be a major section, as we proceed towards DL4, DL5 stages

chaimain (changes requested):
Some simple changes are requested.

Do we intend to use the example Jupyter notebooks on the documentation (which of course, implies that we update all the relevant notebooks!)?

docs/introduction.rst
```diff
@@ -35,40 +35,104 @@ Analysis steps
...
+
+Here is an example configuration file for the IRF creation step.
+
+.. toggle:: 
+
+    .. include:: examples/irf_tool_config.json
+        :code: json
+
+
+Events creation
```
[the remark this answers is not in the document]
vuillaut: right

docs/introduction.rst
vuillaut: Capture d’écran 2023-05-04 à 11 57 53

vuillaut: It looks ok to me this way, no?
I added a sentence with link to the workflows.

vuillaut:
> Do we intend to use the example Jupyter notebooks on the documentation (which of course, implies that we update all the relevant notebooks!)?

That's a good idea, outside of this PR.
The notebooks could be executed and turned to HTML by the CI:
- making sure they are maintained
- serving as doc

docs/introduction.rst
chaimain: Yes, it looks ok!

chaimain (approved):

PR #1088, opened 2023-05-02
Dvr pixel selector

moralejo:
This scripts uses DL1 files to identify (event-wise) pixels likely containing a signal, and creates hdf5 files with this information for use in the raw data volume reduction.


moralejo:
There are no tests for this script as of now, it requires to have DL1 files with the DL1a images. Are they produced? At the end of the tests the created DL1 files only have DL1b.

moralejo:
Obviously I messed something up with the history... Will try to fix it.

PR #1089, opened 2023-05-02
Dvr pixel selector

moralejo:
This scripts uses DL1 files to identify (event-wise) pixels likely containing a signal, and creates hdf5 files with this information for use in the raw data volume reduction.

~~There are no tests for this script as of now, it requires to have DL1 files with the DL1a images. Are they produced? At the end of the tests the created DL1 files only have DL1b.~~

### DIFF
--- a/lstchain/scripts/lstchain_dvr_pixselector.py
+++ b/lstchain/scripts/lstchain_dvr_pixselector.py
@@ -165,7 +165,7 @@ def main():
     fraction_of_survival = 1.
     target_nevents = 5000 # number of events for the calculation (to speed up!)
     charges_cosmics = charges_data[cosmic_mask]
-    event_jump = int(charges_cosmics.shape[0] / target_nevents)
+    event_jump = int(charges_cosmics.shape[0] / target_nevents) + 1
     charges_cosmics_sampled = charges_cosmics[::event_jump, :]
 
     while fraction_of_survival > max_pix_survival_fraction:
--- a/lstchain/scripts/lstchain_dvr_pixselector.py
+++ b/lstchain/scripts/lstchain_dvr_pixselector.py
@@ -56,8 +56,18 @@ class RunSummary(Container):
 
 def main():
     args = parser.parse_args()
-
     summary_info = RunSummary()
+
+    # The pixel selection will be applied only to "physics triggers"
+    # (showers), not to other events like interleaved pedestal and flatfield
+    # events:
+    event_types_to_be_reduced = [EventType.SUBARRAY.value,
+                                 EventType.UNKNOWN.value,
+                                 EventType.HARDWARE_STEREO.value,
+                                 EventType.DAQ.value]
+    # (Note: some of these types are not yet in use). Chances are that pixels
+    # labeled as "unknown" are cosmics, so we include them among the types to
+    # be reduced.
 
     # By default use the provided picture threshold to keep pixels:
     min_charge_for_certain_selection = args.picture_threshold
@@ -68,9 +78,9 @@ def main():
     summary_info.picture_threshold = args.picture_threshold
 
     # Value will be modified, if a too high average fraction (
-    # > max_survival_fraction) of pixels (and FIRST neighbors) survive
-    # with that picture threshold
-    max_survival_fraction = 0.15
+    # > max_pix_survival_fraction_in_ped) of pixels (and FIRST neighbors)
+    # survive, in pedestal events, with that picture threshold
+    max_pix_survival_fraction_in_ped = 0.05
 
     # The new value, if survival fraction turns out to be too high, will be the
     # charge above which a fraction ped_cut_fraction of pedestal charges lie.
@@ -115,6 +125,9 @@ def main():
     summary_info.number_of_events = number_of_events
 
     event_type_data = data_parameters['event_type']
+    # event_type: physics trigger (32) interleaved flat-field(0) pedestal (2),
+    # unknown(255)  are those currently in use in LST1
+
     cosmic_mask   = event_type_data == EventType.SUBARRAY.value
     pedestal_mask = event_type_data == EventType.SKY_PEDESTAL.value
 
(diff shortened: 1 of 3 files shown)


gabemery:
> There are no tests for this script as of now, it requires to have DL1 files with the DL1a images. Are they produced? At the end of the tests the created DL1 files only have DL1b.

This test seems to write data https://github.com/cta-observatory/cta-lstchain/blob/78223b7541e7c20d55c145242927487eb17ed823/lstchain/tests/test_lstchain.py#L38
You may be able to chain your tests after this one.


I am not convinced by the current script usage. Running twice the same script, the second time using the output of the first and adding a flag feel non intuitive. Maybe use a "processing_step" input parameter which can be 'settings', 'pixelmask' or 'both' depending on if you want to run step one, two or both directly?

moralejo:
Thanks @gabemery . Now the script is steered in this way:

lstchain_dvr_pixselector --action compute_dvr_settings ... (which is the default)

or

lstchain_dvr_pixselector --action create_pixel_masks ...

I think it is more clear in this way.
As for the test, I will try, thanks!

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+                    help='Picture threshold (p.e., default: %(default)s)')
+parser.add_argument('--action', dest='action', type=str,
+                    default='compute_dvr_settings',
+                    help='compute_dvr_settings (default) or create_pixel_masks')
+parser.add_argument('-n', '--number-of-rings', dest='number_of_rings',
+                    type=int, default=1,
+                    help='Number of rings around picture pixels (default: %('
+                         'default)s)')
+
+class PixelMask(Container):
```
rlopezcoto, line 81: shouldn't we define all these containers elsewhere? I mean, not in the script that is running the DVR

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+    event_id = Field(-1, 'event id')
+    event_type = Field(-1, 'event_type')
+    number_of_pixels_after_standard_cleaning = Field(-1, 'number_of_pixels_after_standard_cleaning')
+    highest_removed_charge = Field(np.float32(0.), 'highest_removed_charge')
+    pixmask = Field(None, 'selected pixels mask')
+    
+class RunSummary(Container):
```
rlopezcoto, line 88: same as above

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+    # pixels too survive.
+    summary_info.picture_threshold = args.picture_threshold
+
+    # Value will be modified, if a too high average fraction (
+    # > max_pix_survival_fraction) of pixels (and FIRST neighbors)
+    # survive, in shower events, with that picture threshold. The value is
+    # computed excluding the 10% noisiest pixels in the camera, because we
+    # want it to be stable when e.g. stars get in and out of the camera,
+    # or when changing wobbles (star field changes)
+    max_pix_survival_fraction = 0.1
```
rlopezcoto, line 156: if this is gonna be hardcoded, I would define it at the beginning

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+    # Number of "rings" of pixels to be finally kept around pixels which are
+    # above min_charge_for_certain_selection
+    number_of_rings = args.number_of_rings
```
rlopezcoto, line 161: why not:
`summary_info.number_of_rings = args.number_of_rings`
?
I do not see where you use the variable `number_of_rings` later

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+            # construction they are all identical. These are the settings
+            # used in the execution that resulted in the creation of the
+            # DVR_settings_LST-1.Run*.h5 file:
+            summary_info.number_of_rings = dvr_settings['number_of_rings'][0]
+            summary_info.picture_threshold = dvr_settings['picture_threshold'][0]
+
+        print('Output file:', output_file)
+
+        data_parameters = read_table(dl1_file,
+                                     '/dl1/event/telescope/parameters/LST_LSTCam')
```
rlopezcoto, line 212: this is the `dl1_params_lstcam_key`
https://github.com/cta-observatory/cta-lstchain/blob/e41d81d1041f9e700d5190df095d7415cd29dd60/lstchain/io/io.py#L79

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+
+        # Read some useful extra info from the file:
+        subarray_info = SubarrayDescription.from_hdf(dl1_file)
+        focal_length = subarray_info.tel[1].optics.equivalent_focal_length
+        print('Focal length:', focal_length)
```
rlopezcoto, line 217: do we need to print this or is it some debugging left?

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+        camera_geom = subarray_info.tel[1].camera.geometry
+        print('Camera geometry:', camera_geom)
```
rlopezcoto, line 219: same as above

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+
+        # Time between first and last timestamp:
+        summary_info.elapsed_time = (data_parameters['dragon_time'][-1] -
+                                     data_parameters['dragon_time'][0])
+
+        summary_info.mean_zenith = np.round(90 -
```
rlopezcoto, line 225: we are using altitude everywhere in CTA, I would suggest to use `mean_altitude` instead of zenith

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+        found_event_types = np.unique(event_type_data)
+        print('Event types found:', found_event_types)
+        if EventType.SUBARRAY.value not in found_event_types:
+            print('No shower event (SUBARRAY type) found in file! SKIPPING IT!')
+            continue
+
+        cosmic_mask   = event_type_data == EventType.SUBARRAY.value  # showers
+        pedestal_mask = event_type_data == EventType.SKY_PEDESTAL.value
+
+        data_images = read_table(dl1_file, "/dl1/event/telescope/image/LST_LSTCam")
```
rlopezcoto, line 244: `dl1_images_lstcam_key`
https://github.com/cta-observatory/cta-lstchain/blob/e41d81d1041f9e700d5190df095d7415cd29dd60/lstchain/io/io.py#L80

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+
+        data_calib = read_table(dl1_file, "/dl1/event/telescope/monitoring/calibration")
```
rlopezcoto, line 246: https://github.com/cta-observatory/cta-lstchain/blob/e41d81d1041f9e700d5190df095d7415cd29dd60/lstchain/io/io.py#L77

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -0,0 +1,516 @@
...
+            min_charge_for_certain_selection  = np.floor(meanq)
+            # we do not have to calculate it from scratch, we get the
+            # value (same for all subruns) from the DVR_settings_LST*.h5 file,
+            # just averaging the subrun-wise values and taking the closest
+            # (lower) integer
+
+        summary_info.min_charge_for_certain_selection = min_charge_for_certain_selection
+
+        # Cuts to identify promising muon ring candidates:
+        mucan_event_list = np.where(cosmic_mask &
```
rlopezcoto, line 295: these hardcoded values should be defined elsewhere (at the beginning of the script, maybe?) with a comment on why they are chosen

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: It is not to be used anywhere else in lstchain. The program which reads the files in is necessarily a C++ program, outside of lstchain.

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: It is in the call to get_selected_pixels

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: This is already early inside the main, I put it there because it is related to the charge threshold for pixel selection which is read at that point from the arguments. 
I would not like to make it a configurable parameter, just for simplicity.

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: Not needed, I'll remove.

lstchain/scripts/lstchain_dvr_pixselector.py
rlopezcoto: everywhere code-wise, that is a completely different world than real conversations :-P
Fine with me.

lstchain/scripts/lstchain_dvr_pixselector.py
rlopezcoto: but are you saying that the same variable is defined twice? this would not be good, but maybe I am misunderstanding... can you point me where is the line you are referring to, please?

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -143,6 +143,18 @@ def test_observed_dl1_validity(observed_dl1_files):
     )
     np.testing.assert_allclose(dl1_df["dragon_time"], dl1_df["trigger_time"])
 
+@pytest.mark.private_data
+def test_dvr_file_validity(observed_dl1_files):
```
rlopezcoto, line 147: shouldn't the dvr have as input a R0 data file?

maxnoe: The script calculates the pixel masks from the dl1 images, stores it in an hdf5 output file and the actual dvr is performed by a C++ program reading the pixel masks and applying them to the zfits files.

The C++ program is here:
https://gitlab.cta-observatory.org/cta-array-elements/lst/subsystems/daq/offline_gain_selection

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: No, it is just poor wording. picture_threshold is not modified, it is simply the starting value for the actual threshold used, "min_charge_for certain_selection". I will explain it better.

rlopezcoto (approved):

PR #1092, opened 2023-05-04
Fix a failing test on DL3 Tool

chaimain:
A test on DL3 Tools when using IRF interpolation methods was failing as it required a list of IRFs which are not present as fixtures for the scope of the test session. These IRFs are created and tested in `lstchain/high_level/tests/test_interpolate` (through some "lengthy" code) for almost all combinations of IRF interpolations. 
As such, it is not essential to have separate and similar tests on DL3 Tools, for testing the same functionality.

This was missed in the earlier PR #711 as I tested in sequential order instead of in parallel (as done in CI).

In case, we require to have the DL3 Tools tested for the full functionality again, we would just need to shift ~150 lines of code from `lstchain/high_level/tests/test_interpolate.py` to `lstchain/conftest.py`

### DIFF
--- a/lstchain/high_level/interpolate.py
+++ b/lstchain/high_level/interpolate.py
@@ -420,7 +420,7 @@ def interpolate_irf(irfs, data_pars, interp_method="linear"):
 
     # Exclude AZ_PNT as target interpolation parameter
     # For the interpolation parameters only
-    params_sel = [d for d in data_pars.keys() if d != "AZ_PNT"]#[*d.keys()]
+    params_sel = [d for d in data_pars.keys() if d != "AZ_PNT"]
     n_grid = len(irfs)
     irf_pars_sel = np.empty((n_grid, len(params_sel)))
     interp_pars_sel = list()
--- a/lstchain/high_level/interpolate.py
+++ b/lstchain/high_level/interpolate.py
@@ -193,13 +193,9 @@ def get_nearest_az_node(
 
     if irf_num[idx] > 1:
         # Only using the overlapping nodes
-        idx_list = np.flatnonzero(
-            (irf_params_short == irf_params_short[index]).all(1)
-        )
+        idx_list = np.flatnonzero((irf_params_short == irf_params_short[index]).all(1))
         # Finding the shortest distance with respect to target azimuth
-        diff = np.abs(
-            irf_params_full[idx_list, az_idx] - target_params_full[az_idx]
-        )
+        diff = np.abs(irf_params_full[idx_list, az_idx] - target_params_full[az_idx])
         idx = idx_list[np.where(diff == diff.min())[0]][0]
     else:
         # if there are no overlapping nodes
--- a/lstchain/high_level/tests/test_interpolate.py
+++ b/lstchain/high_level/tests/test_interpolate.py
@@ -84,12 +84,8 @@ def test_interp_irf(simulated_irf_file, simulated_dl2_file):
         aeff_1_meta["AZ_PNT"] = (az_2 * 180 / np.pi, "deg")
         aeff_2_meta["B_DELTA"] = (del_2 * 180 / np.pi, "deg")
 
-        aeff_hdu_1 = fits.BinTableHDU(
-            aeff_1, header=aeff_1_meta, name="EFFECTIVE AREA"
-        )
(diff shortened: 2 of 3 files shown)


moralejo:
Hi, 
probably unrelated, but I realized that the tests in high_level do not pass in my Mac (while they work on linux). 
This is the same for the master and for the interp_irfs branch

```
lstchain/data/tests/test_normalised_pulse_template.py::test_load_from_file_and_miscellaneous PASSED [  4%]
lstchain/high_level/tests/test_hdus.py::test_create_event_list ERROR     [  5%]
lstchain/high_level/tests/test_hdus.py::test_create_obs_hdu_index ERROR  [  5%]
lstchain/high_level/tests/test_hdus.py::test_get_timing_params PASSED    [  6%]
lstchain/high_level/tests/test_interpolate.py::test_load_irf_grid PASSED [  6%]
lstchain/high_level/tests/test_interpolate.py::test_interp_irf FAILED    [  7%]
lstchain/high_level/tests/test_interpolate.py::test_compare_irfs PASSED  [  7%]
```

When looking in more detail, I don't understand anything, the error seems to be in the r0 to dl1 step

```
==================================== ERRORS ====================================
___________________ ERROR at setup of test_create_event_list ___________________

temp_dir_observed_files = PosixPath('/private/var/folders/cn/jhp87yg56455zyhkzz36mhl40000gn/T/pytest-of-moralejo/pytest-18/observed_files0')
run_summary_path = PosixPath('/private/var/folders/cn/jhp87yg56455zyhkzz36mhl40000gn/T/pytest-of-moralejo/pytest-18/observed_files0/RunSummary_20200218.ecsv')

    @pytest.mark.private_data
    @pytest.fixture(scope="session")
    def observed_dl1_files(temp_dir_observed_files, run_summary_path):
...
...
E             File "/Users/moralejo/opt/miniconda3/envs/lst-dev/lib/python3.8/si
te-packages/astropy/io/ascii/core.py", line 743, in check_column_names
E               raise InconsistentTableError(
E           astropy.io.ascii.core.InconsistentTableError: Length of names argume
nt (10) does not match number of table columns (3)
```
Anyone has an idea of what may be going on?

gabemery (approved):
Ok with removing the test if it leads to failure. (also for some reason it doesn't reduce coverage?)

gabemery:
@moralejo Is this the full error message you get? It doesn't include the line where you get the error.

chaimain:
> Ok with removing the test if it leads to failure. (also for some reason it doesn't reduce coverage?)

It does reduce the coverage, but of the DL3 Tool file, https://app.codecov.io/gh/cta-observatory/cta-lstchain/pull/1092/blob/lstchain/tools/lstchain_create_dl3_file.py which is an indirect change with respect to this PR.

rlopezcoto (changes requested):
before removing the test because it is not passing, let's try to figure out what is the problem.

moralejo:
> @moralejo Is this the full error message you get? It doesn't include the line where you get the error.

The error itself seems to happen when running  lstchain_data_r0_to_dl1, and more specifically, in ctapipe_io_lst, when it tries to read in the drive report of the test data
https://github.com/cta-observatory/ctapipe_io_lst/blob/28b4a1b55e58fbacae6fe52c90b0e6e28433b984/ctapipe_io_lst/pointing.py#L96-L121

```
test_data/real/monitoring/DrivePositioning/drive_log_20200218.txt

astropy.io.ascii.core.InconsistentTableError: 
ERROR: Unable to guess table format with the guesses listed below:
[...]
```
And indeed, the format does not seem to be what it tries to read. I can reproduce it outside of the tests (both running the lstchain_data_r0_to_dl1 script and directly reading in the file)

`astropy.io.ascii.core.InconsistentTableError: Length of names argument (10) does not match number of table columns (20)
`

gabemery:
@moralejo It seems like a dedicated issue will be needed to discuss this.

moralejo:
> @moralejo It seems like a dedicated issue will be needed to discuss this.

Let's discuss here: #1094

rlopezcoto:
Now that we have worked out the problems with the tests in #1094 and that it was unrelated to this PR, @moralejo can you confirm that tests pass on this branch and that everything is alright for it to be merged?

rlopezcoto (approved):

PR #1096, opened 2023-05-12
Interpolation of alpha cuts and change of default maximum alpha cut

SeiyaNozaki:
Here I implemented the interpolation of `alpha` cuts as `gammaness` interpolation. In addition, I changed the default maximum alpha cut from 45 deg to 20 deg since it was too loose.


lstchain/high_level/tests/test_interpolate.py
```diff
@@ -132,50 +195,103 @@ def test_interp_irf(simulated_irf_file, simulated_dl2_file):
...
+                al_2["cut"][mask] *= factor_zd * factor_del
+
+                al_1_meta["ZEN_PNT"] = (zen_2 * 180 / np.pi, "deg")
+                al_1_meta["AZ_PNT"] = (az_1 * 180 / np.pi, "deg")
+                al_1_meta["B_DELTA"] = (del_1 * 180 / np.pi, "deg")
+                al_2_meta["ZEN_PNT"] = (zen_2 * 180 / np.pi, "deg")
+                al_2_meta["AZ_PNT"] = (az_2 * 180 / np.pi, "deg")
+                al_2_meta["B_DELTA"] = (del_2 * 180 / np.pi, "deg")
+
+                al_hdu_1 = fits.BinTableHDU(al_1, header=al_1_meta, name="GH_CUTS")
```
chaimain, line 249: name as `AL_CUTS`

chaimain (approved):

rlopezcoto:
this PR is going to change the definition of the default ranges assumed when creating DL3 files, so it should also go into the next release and the target branch should be `ctapipe_v0.17` instead of `master`

chaimain (changes requested):
@SeiyaNozaki Can you update the `interpolate_al_cuts` as per #1099 and we can rebase and then merge #1099 before this to have a good follow-up.

docs/examples/irf_tool_config.json
```diff
@@ -24,8 +24,8 @@
     "max_theta_cut": 0.32,
     "fill_theta_cut": 0.32,
     "min_alpha_cut": 1,
-    "max_alpha_cut": 45,
-    "fill_alpha_cut": 45,
+    "max_alpha_cut": 20,
```
chaimain, line 27: These default values needs to be updated in `lstchain/io/event_selection` as well.

chaimain (approved):

PR #1097, opened 2023-05-12
Add code snippet to documentation to show how to print config and calibration files used in the analysis

morcuended:
I also changed the link to ctapipe documentation to readthedocs link (for intersphinx mapping) and add some py class references from ctapipe to nitpick ignore list to avoid showing so many warnings that were preventing automatic deployment of the documentation.

follow-up of #1086


vuillaut (approved):
Very useful, thanks.

PR #1099, opened 2023-05-15
Update IRF interpolation methods with pyirf v0.8.1

chaimain:
Update IRF interpolation methods with the updated functionality from pyirf v0.8.1, and more specifically,

- Update the interpolation method of `GH_CUTS`, similar to `interpolate_rad_max` function of pyirf, 
- add a temp fix of `nan_to_num` on creating Effective Area (missing in pyirf) to avoid having `astropy's Masked` values for `EFFAREA` data
- update interpolation tests to work on the interpolation parameters of the fixtured MC test data sample and not on hard-coded values.

### DIFF
--- a/lstchain/high_level/interpolate.py
+++ b/lstchain/high_level/interpolate.py
@@ -13,13 +13,13 @@ from pyirf.io.gadf import (
     create_psf_table_hdu,
 )
 from pyirf.interpolation import (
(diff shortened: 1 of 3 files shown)


gabemery (approved):
Looks good. Failing tests are already failing in the target branch as far as I see so I think it shouldn't prevent merging.

PR #1072, opened 2023-02-08
PR for ctapipe 0.17 : write pedestal and ff events to a h5 files

FrancaCassol:
The idea is to process the interleaved data after the R1 to dl1b process and to apply the Cat-B calibration in the dl1a to dl1b process

PR #1100, opened 2023-05-15
PR for ctapipe 0.17 : write pedestal and ff events to a h5 files

FrancaCassol:
New commit for the code in PR [https://github.com/cta-observatory/cta-lstchain/pull/1072](https://github.com/cta-observatory/cta-lstchain/pull/1072) with correct rebase to ctapipe_0.17. It substitutes PR 1072

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -286,7 +286,7 @@
       "transform_waveform": true,
       "waveform_dtype": "uint16",
       "waveform_offset": 400,
-      "waveform_scale": 1  
+      "waveform_scale": 80  
     }
   }
 }
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -276,5 +276,17 @@
     "nsb_tuning": false,
     "nsb_tuning_ratio": 0.52,
     "spe_location": "lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat"
+  },
+  "write_interleaved_events":{
+    "DataWriter": {
+      "overwrite":  true,
+      "write_images": false,
+      "write_parameters": false,
+      "write_waveforms": true,
+      "transform_waveform": true,
+      "waveform_dtype": "uint16",
+      "waveform_offset": 400,
+      "waveform_scale": 1  
+    }
   }
 }
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -26,6 +26,9 @@ from ctapipe.image import number_of_islands, apply_time_delta_cleaning
 from ctapipe.io import EventSource, HDF5TableWriter, DataWriter 
 from ctapipe.utils import get_dataset_path
 from traitlets.config import Config
+from ctapipe_io_lst.constants import (
+    PIXEL_INDEX
+)
 
 from . import disp
 from .utils import sky_to_camera
@@ -455,7 +458,7 @@ def r0_to_dl1(
             name = f"interleaved_{name}"
         interleaved_output_file = Path(dir, name)
         interleaved_writer = DataWriter(event_source=source,output_path=interleaved_output_file,config=interleaved_writer_config)
-       
+        interleaved_writer._writer.exclude("/r1/event/telescope/.*", "selected_gain_channel")
 
     with HDF5TableWriter( 
         filename=output_filename,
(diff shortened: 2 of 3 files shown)


lstchain/reco/r0_to_dl1.py
```diff
@@ -444,7 +444,17 @@ def r0_to_dl1(
         lhfit_fitter_config = {'TimeWaveformFitter': config['lh_fit_config']}
         lhfit_fitter = TimeWaveformFitter(subarray=subarray, config=Config(lhfit_fitter_config))
 
-    with HDF5TableWriter(
+    # initialize the writer of the interleaved events 
+    interleaved_writer = None
+    if 'write_interleaved_events' in config.keys():
```
maxnoe, line 449: `.keys()` is not needed.

lstchain/reco/r0_to_dl1.py
```diff
@@ -510,8 +520,14 @@ def r0_to_dl1(
                                            calibration_index,
                                            event.mon.tel[tel_id],
                                            new_ped=new_ped_event, new_ff=new_ff_event)
-
+                    
+                    # write the calibrated R1 waveform without gain selection
+                    source.r0_r1_calibrator.select_gain = False
```
maxnoe, line 525: I really don't like that you modify the r0_r1_calibrator of the source here. Also, isn't it too late for that here? The source already applied the calibration when it reaches lstchain.

lstchain/reco/r0_to_dl1.py
```diff
@@ -746,6 +762,11 @@ def r0_to_dl1(
         muon_output_filename = Path(dir, name)
         table = Table(muon_parameters)
         table.write(muon_output_filename, format='fits', overwrite=True)
+        
+        # close the interleaved output file and write metadata
+        if 'write_interleaved_events' in config.keys():
```
maxnoe, line 767: Better you `if interleaved_writer is not None` since that doesn't need the config again.

lstchain/reco/r0_to_dl1.py
FrancaCassol: Max,  it works because we have the "calibrate_flatfields_and_pedestals": false, in the config file (by the way, this code works well, I could "Cat-B" calibrate the GRB data with it ;-) )
I agree that it is not nice to change the source here, but the source is in any case treating in a special way these events, which one day will be in a dedicated stream hopefully.

FrancaCassol: @maxnoe do you prefer I make a copy of the  source.r0_r1_calibrator  in order to touch an independent object?

lstchain/data/lstchain_standard_config.json
```diff
@@ -276,5 +276,17 @@
     "nsb_tuning": false,
     "nsb_tuning_ratio": 0.52,
     "spe_location": "lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat"
+  },
+  "write_interleaved_events":{
```
rlopezcoto, line 280: what is the defaults value for `write_interleaved_events`? sorry, could not find it.

rlopezcoto (commented):
this PR looks ok, if @maxnoe does not have additional comments, we could move on

lstchain/data/lstchain_standard_config.json
FrancaCassol: They are defined here: https://github.com/cta-observatory/ctapipe/blob/0834ca5c0b852ceccd444917d76d8c812d5d5453/ctapipe/io/datawriter.py#L147

lstchain/reco/r0_to_dl1.py
```diff
@@ -510,10 +526,27 @@ def r0_to_dl1(
...
+                    # write the calibrated R1 waveform without gain selection
+                    source.r0_r1_calibrator.select_gain = False
                     source.r0_r1_calibrator.calibrate(event)
-
+                
+                    if interleaved_writer is not None:
+                        interleaved_writer(event)
+
+                    # gain select the events
+                    source.r0_r1_calibrator.select_gain = True
```
maxnoe, line 538: This is not needed any more

lstchain/reco/r0_to_dl1.py
maxnoe: With the fixed code, this line should not be needed, right?

lstchain/reco/r0_to_dl1.py
FrancaCassol: yes, we need it, because I above I had to put it to False...

lstchain/reco/r0_to_dl1.py
FrancaCassol: the fix was to avoid the re-calibration and just to perform the gain selection

lstchain/data/lstchain_standard_config.json
rlopezcoto: sorry, I am maybe missing something, but in the link you are giving I do not find the `write_interleaved_events` defined

FrancaCassol: The default values are refined in the DataWriter trailets definition (the target of the above link), which is the Component to be configured by the write_interleaved_events json parameters

rlopezcoto: ok thx!

rlopezcoto (approved):

lstchain/reco/r0_to_dl1.py
FrancaCassol: > But calibration events are not gain-selected, are they?
After they are used for creating the calibration coefficients, they are gain selected

FrancaCassol: > Or should we add a new flag to the source to calibrate but not gain select the calibration events?

a new flag would be cleaner but we in any case the present r0_to_dl1 is something we want to improve as soon as possible, I let you decide if you want add this flag and change the source version now

rlopezcoto: I would say that if @maxnoe has time to implement this right away, we could make a minor release of `ctapipe_io_lst`, otherwise @FrancaCassol could open an issue to remember to modify this in the future

FrancaCassol: Hi @rlopezcoto, I can do it but I would like to recall that  in a (hopefully) short future the cat-B calibration will be no more performed in the r0_to_dl1 script and also the interleaved images will be no more written in the dl1 file but will follow a parallel path treated by the calibration pipeline. If @maxnoe has not time to implement it just now I would go ahead with this PR as it is

rlopezcoto: sure, this was my suggestion, if it cannot be implemented right now, let's open an issue to remember this and move on with the PR

rlopezcoto: Opened here:
https://github.com/cta-observatory/ctapipe_io_lst/issues/186

thx!

PR #1102, opened 2023-05-16
Update effective time calculation for observations suffered from busy stuck problem

SeiyaNozaki:
To calculate the elapsed time, a cut of `time_diff`<0.1 s is applied to exclude events suffered from DAQ break. However, such kind of filter is not applied to calculate the average of `delta_t` for the estimation of the trigger rate.

For example, when we have 'break' because of DAQ problem like below, the estimated trigger rate is 4.8 kHz, but the updated codes (excluding large `delta_t` events) estimate 7.4 kHz.
eff_time_calculation

Here I set 0.01 s (100 Hz) as the limit for both elapsed/effective time calculation for the moment. Any comments are welcome.

### DIFF
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -669,12 +669,15 @@ def get_effective_time(events):
 
     # elapsed time: sum of those time differences, excluding large ones which
     # might indicate the DAQ was stopped (e.g. if the table contains more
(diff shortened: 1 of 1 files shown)


moralejo (approved):
Ok! Note: this will have no effect for observations without problems (the cut delta_t<0.01 will have no effect if the 100+100 Hz of interleaved are working, and there is no busy stuck).

Issue #806, opened 2021-12-10
Run black on lstchain scripts for proper formatting


rlopezcoto:
Suggested in:
https://github.com/cta-observatory/cta-lstchain/pull/776#pullrequestreview-827523003

For reference of lstchain code formatting, see also discussion in:
https://github.com/cta-observatory/cta-lstchain/pull/664
in which the auto-formatting of PRs using `black` was proposed, but we never found an agreement due to the possibility of conflicts formatting the code automatically for every PR

PR #1103, opened 2023-05-17
Fix fov offset calculation in the irf tool, fixes #1098

### LINKED ISSUE #1098: IRF metadata floating point precision
Apparently the `create_irf_file` tool uses the mean offset position of the gammas to determine the offset used in the simulation:

https://github.com/cta-observatory/cta-lstchain/blob/4cc8bce8bd1efd4bb9860b6e6341c165ebe03b0b/lstchain/tools/lstchain_create_irf_files.py#L379-L381

https://github.com/cta-observatory/cta-lstchain/blob/4cc8bce8bd1efd4bb9860b6e6341c165ebe03b0b/lstchain/tools/lstchain_create_irf_files.py#L522-L525

This creates an issue in my case where most produced irfs have offset 0.4 deg, but one has 0.3999 making them "incompatible" for interpolation. There are 2 questions arising:
- This used to round to 3 digits, but was changed a year ago by @chaimain in this commit: https://github.com/cta-observatory/cta-lstchain/commit/d75ee4554927455018de9b8d8d8dfd6bcfedac6c
Was there any deeper reason for that? 3 digits already seems excessive to me in this case
- Is this needed at all? Clearly this information should be stored in the simtel files.

LukasNickel:
- Uses the fov_offset that is available in the simulation information as opposed to the individual gamma positions. This should avoid rounding issues in the future
- The individual offset is still used for source-dependent analyses. As I am not quite sure how that works, I left it as is
- I ran black on the files I touched, which created a pretty large diff. Since it all looked reasonable to me, I left it in.


lstchain/io/io.py
```diff
@@ -1188,16 +1223,37 @@ def check_mc_type(filename):
...
 
     else:
-        raise ValueError('mc type cannot be identified')
+        raise ValueError("mc type cannot be identified")
 
     return mc_type
-            
+
+
+def check_mc_fov_offset(filename):
```
maxnoe, line 1240: This doesn't "check" anything. It "computes" or "gets"

lstchain/io/io.py
```diff
@@ -1188,16 +1223,37 @@ def check_mc_type(filename):
...
+    string
+    """
+
+    simu_info = read_simu_info_merged_hdf5(filename)
+
+    min_viewcone = simu_info.min_viewcone_radius.value
+    max_viewcone = simu_info.max_viewcone_radius.value
+
+    # This calculation is slightly more stable
+    mean_offset = 0.5 * min_viewcone - 0.5 * (max_viewcone - min_viewcone)
```
maxnoe, line 1257: You should convert to float64 for the computation as well

lstchain/io/io.py
```diff
@@ -1188,16 +1223,38 @@ def check_mc_type(filename):
...
+def get_mc_fov_offset(filename):
+    """
+    Check MC type ('point_like', 'diffuse', 'ring_wobble') based on the viewcone setting
+    Parameters
+    ----------
+    filename:path (DL1/DL2 hdf file)
+    Returns
+    -------
+    string
+    """
```
chaimain, line 1249: Update with an appropriate description for this function.

lstchain/io/io.py
```diff
@@ -1188,16 +1223,38 @@ def check_mc_type(filename):
...
+
+    simu_info = read_simu_info_merged_hdf5(filename)
+
+    # Make sure we have full precision here
+    min_viewcone = simu_info.min_viewcone_radius.value.astype(float)
+    max_viewcone = simu_info.max_viewcone_radius.value.astype(float)
+
+    # This calculation is slightly more stable
+    mean_offset = 0.5 * min_viewcone - 0.5 * (max_viewcone - min_viewcone)
```
chaimain, line 1258: Rather than calling it `mean_offset`, better to just call it `fov_offset`. 
Have you checked for diffuse MC files (not used for IRFs), as to what values you get for this parameter?

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -364,21 +358,22 @@ def start(self):
...
-                p["events"]["reco_source_fov_offset"] = p["events"]["true_source_fov_offset"]
+                p["events"]["reco_source_fov_offset"] = p["events"][
+                    "true_source_fov_offset"
+                ]
 
         self.log.debug(p["simulation_info"])
         gammas = self.mc_particle["gamma"]["events"]
         geomag_params = self.mc_particle["gamma"]["geomag_params"]
         self.log.info(geomag_params)
+        mean_fov_offset = get_mc_fov_offset(self.mc_particle["gamma"]["file"])
```
chaimain, line 369: `mean_fov_offset` -> `mc_fov_offset` or maybe something similar

chaimain (changes requested):

chaimain:
> * The individual offset is still used for source-dependent analyses. As I am not quite sure how that works, I left it as is

Maybe @SeiyaNozaki can comment on this.

> * I ran black on the files I touched, which created a pretty large diff. Since it all looked reasonable to me, I left it in.

As you can see running `black` gave some strange modifications in some places. and so, I probably missed checking/"correcting" in various places.

lstchain/io/io.py
LukasNickel: I just checked one file and apparently the maximum is 6 deg. The minimum is 0 deg, so that would be 3 deg. How does that help? :sweat_smile: It is only calculated for point like or ring-wobble files.

lstchain/tools/lstchain_create_irf_files.py
LukasNickel: I prefer mean, because of ring wobbles tbh.
But either way is fine

lstchain/io/io.py
chaimain: Not exactly useful here currently, but as the function is in lstchain.io.io, just wanted to be sure that this is applicable for all possible inputs here.

chaimain: And speaking of general inputs, we do have diffuse gamma MC for some special studies and for that case, the FOV offset evaluation as a mean value might not make much sense. Maybe there should be dependence with the check_mc_type function for this evaluation.

LukasNickel: Does it even make sense to compute an offset for diffuse monte carlo at all?
Like what do you even expect to be the result?

chaimain: Indeed, I dont think it is a usable parameter for diffuse MC, but we should have the MC type used for the creation of the IRFs, continued as a general metadata information. Maybe once the CTA IRF/DL3 Model requirement is complete, and we follow it exactly, then we will have a different way of saving such a metadata. I dont know if ctapipe has this right now (and in case it is available in > v0.12, then we can push these changes for ctapipe_0.17 branch instead), and if it is we might have to adapt the code accordingly.

In any case, having the G_OFFSET value as 6 deg does indicate that the simulation used is of diffuse type, and covering the full FoV.

chaimain: @maxnoe what do you suggest for the metadata `G_OFFSET` for diffuse MC-based IRF? Or in general, how should this information be saved as, as per CTAO?

chaimain: Sounds good! It will be clearer when reading the IRFs! 

Then, we would just have to update the same in https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/high_level/interpolate.py#L226, https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/high_level/interpolate.py#L429 and https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/tools/lstchain_create_dl3_file.py#L505-L510. @LukasNickel Can you try some options for these names?

LukasNickel: Proposal:
We get rid of it altogether.

For interpolation you need to know if the axes are compatible and you check that. Why would the axes be compatible if the IRF were computed from different offsets? If you insist, that the MC type needs to be known, we can just put that into the header directly instead of G_OFFSET. But I would argue that is not even necessary: If different simulation types / offsets produce the same irf binning, they are either compatible or incorrect.

chaimain: The logic seems correct to me. Can you test with the test sample MC (diffuse) and any particular node of a recent MC production (ring-wobble) to be certain of this?

I would still prefer to have readable metadata information telling the MC type used for a given IRF production. It will help analyzers be clear of the IRFs they work with, more easily than accessing logging data at least.

LukasNickel: I dont think analyzers should have to deal with checking the metadata of their IRFs, no?

rlopezcoto:
@chaimain I see you have still marked the change request, are you ok with the changes implemented? if so, could you approve please?

chaimain:
> @chaimain I see you have still marked the change request, are you ok with the changes implemented? if so, could you approve please?

There have been no changes made to the PR after my and Seiya's comments, except for merging with `master` branch.

chaimain:
> Don't we already have black running on the code? Also are the settings you used personal ones or ones defined for lstchain?
> 
> Current change will lead to quite non uniform code in lstchain (e.g. usage of " instead of ' in the`__all__` of io.py)

No, there was no consensus on it #806. Can you clarify what settings are you referring to?

LukasNickel:
Regarding black:
I can clean up the PR to not include any black changes if you prefer. There are no special settings here, all default as far as I am aware. Maybe that is the way to go in order to make the review easier?

Regarding renaming variables to not include `mean`:
Easily done. Since it is pretty much one vote for both schemes, it would be great if someone else could give an opinion.

Regarding the larger metadata discussion:
Not sure what the conclusion is here:
- Remove the G_OFFSET?
- fix the calculation and keep it as implemented in this PR
- Do something else / run some checks on diffuse MC? To be honest I am a bit confused what the proposal is there.

LukasNickel:
I force pushed to not include the black changes.
That should make any discussions easier.

Leaves us with two points:
- `mean_fov_offset` or `mc_fov_offset`?
- More importantly: There seems to be no requirement to keep the `G_OFFSET` metadata for interpolation as the compatibility should be ensured by checking for equal bins. Should we drop it or does someone rely on it?

chaimain:
As discussed privately, it will be better to get the input from analyzers who would use diffuse gamma MC (not ring-wobble) for their analysis using IRF interpolation, to understand whether we need a parameter stating the offset (mean or something else) of the MC used. 
As far as I know, only the teams analyzing the Galactic Center and the LHAASO sources use such MC.

rlopezcoto:
> I force pushed to not include the black changes. That should make any discussions easier.
> 
> Leaves us with two points:
> 
> * `mean_fov_offset` or `mc_fov_offset`?

`mean_fov_offset` describes better what that function does

> * More importantly: There seems to be no requirement to keep the `G_OFFSET` metadata for interpolation as the compatibility should be ensured by checking for equal bins. Should we drop it or does someone rely on it?

Do you mean to remove its filling from:
https://github.com/cta-observatory/cta-lstchain/blob/ced74aca7110b68c7749a3a5df7aae5a46ce5326/lstchain/tools/lstchain_create_irf_files.py#L522
and
https://github.com/cta-observatory/cta-lstchain/blob/ced74aca7110b68c7749a3a5df7aae5a46ce5326/lstchain/tools/lstchain_create_dl3_file.py#L531
?
I would still keep those

chaimain:
> > I force pushed to not include the black changes. That should make any discussions easier.
> > Leaves us with two points:
> > 
> > * `mean_fov_offset` or `mc_fov_offset`?
> 
> `mean_fov_offset` describes better what that function does

Agreed.

> 
> > * More importantly: There seems to be no requirement to keep the `G_OFFSET` metadata for interpolation as the compatibility should be ensured by checking for equal bins. Should we drop it or does someone rely on it?
> 
> Do you mean to remove its filling from:
> 
> https://github.com/cta-observatory/cta-lstchain/blob/ced74aca7110b68c7749a3a5df7aae5a46ce5326/lstchain/tools/lstchain_create_irf_files.py#L522
> 
> 
> and
> https://github.com/cta-observatory/cta-lstchain/blob/ced74aca7110b68c7749a3a5df7aae5a46ce5326/lstchain/tools/lstchain_create_dl3_file.py#L531
> 
> 
> ?

Yes, and in general, not to evaluate it altogether.

> I would still keep those

I concur and so far I have not received any requests otherwise. Maybe in the future, if this changes, then we can look into a resolution then.

chaimain:
Discussing with Yoshiki and Shotaro, the metadata information `G_OFFSET` seems to be not necessary as we do compare the `fov_offset_bins` 
- which we get from this `mean_fov_offset` for point-like and ring-wobble gamma MC, and for diffuse gamma MC, we get the range of `fov_offset_bins` separately from either `DataBinning`'s `fov_offset_bins` function, 

in `compare_irfs` function as IRF OFFSET columns. With this, we should not need this explicit metadata information, at least not in the intermediate future.

chaimain:
So, @LukasNickel I would suggest that you remove this header information from the 3 files.

PR #1105, opened 2023-05-18
Update DL3 cuts and some clean-up

chaimain:
Instead of a `max_theta_cut` for energy-dependent theta (source-independent) cuts, `RAD_MAX`, evaluate the maximum `RAD_MAX` as per the wobble offset angle and the number of OFF regions to be used in ON/OFF normalization, in post DL3 analysis, when using Wobble OFF regions to evaluate the background counts. This replaces `max_theta_cut` and `fill_theta_cut` with `n_off_wobbles` in this case.

Also, add a function `update_fill_cuts` to replace the cut values for the energy-dependent cuts evaluated with pyirf, for the energy bins with fewer events than `min_event_p_en_bin` (default = 100), with the cut value of the nearest energy bin, evaluated properly. This is usually relevant for lower and higher energy bins, where there are fewer or no MC events, and will select a cut evaluated for the first or last energy bin with a statistically significant number of events, instead of a constant `fill_value`. This is applied for all such parameters - `gammaness`, `theta` and `alpha` (source-dependent).

**Note:** These changes will make the IRF/DL3 files **incompatible** with older versions

Some general clean-up of code in `lstchain/high_level/hdu_table` and updated docstrings.


chaimain:
Testing with an older MC production using the current master dependencies, and using the new (pyirf v0.8) `n_events` Column data for the cuts table, to be consistent with the functionality of `update_fill_cuts`, to check for energy-dependent gammaness and theta cuts.
fill_dl3_cuts

For testing with the updated dependencies, we would need a new MC DL2 (at least) production, due to the new methods of ctapipe in reading the simulation info (eg `num_showers` -> `n_showers`)

moralejo:
fill_dl3_cuts

It is a bit weird though that the gammaness cut decreases after 8 TeV. Hard to be sure on a color scale, but it does not seem like the gammaness distributions shift down as energy increases... What efficiency is that? Do the cuts seem right (e.g. for the first bin above 10 TeV) if the vertical slices of the histogram are plotted?

chaimain:
> fill_dl3_cuts
> 
> It is a bit weird though that the gammaness cut decreases after 8 TeV. Hard to be sure on a color scale, but it does not seem like the gammaness distributions shift down as energy increases... What efficiency is that? Do the cuts seem right (e.g. for the first bin above 10 TeV) if the vertical slices of the histogram are plotted?

The cut efficiency is default value of 0.95. In a coarser binning (10 instead of earlier 40) for gammaness and with plotting bins with minimum 30 events, we do see a collection of events with lower gammaness value.
fill_dl3_cuts_2

The majority of events do have gammaness value close to the maximum, but because of these group of events, the percentile evaluation yields a lower cut value. The cuts of 95% efficiency holds true for these bins.

lstchain/high_level/hdu_table.py
```diff
@@ -323,30 +311,38 @@ def set_expected_pos_to_reco_altaz(data):
     """
     Set expected source positions to reconstructed alt, az positions for
```
moralejo, line 312: This formulation is a bit misleading. Seems to state the opposite (i.e. take the reco values and put them in the "expected" variables) that what is actually done.
Be more specific, e.g. "Fill the reconstructed alt az with the expected source position. This is just a trick..."

lstchain/io/event_selection.py
```diff
@@ -141,6 +145,35 @@ def apply_global_gh_cut(self, data):
...
+        minimum.
+        """
+        cut_table_new = cut_table.copy()
+
+        low_event_bins = np.argwhere(
+            cut_table["n_events"] < self.min_event_p_en_bin
+        ).ravel()
+        high_event_bins = np.where(
+            cut_table["n_events"] >= self.min_event_p_en_bin
+        )[0]
```
moralejo, line 163: Why use np.argwhere first, then np.where? If I am not mistaken, the way you use them, the effect is the same.
The second one could be 
np.argwhere(cut_table["n_events"] >= self.min_event_p_en_bin).ravel()

lstchain/io/event_selection.py
```diff
@@ -141,6 +145,35 @@ def apply_global_gh_cut(self, data):
...
+
+        for low_ in low_event_bins:
+            if low_ < high_event_bins[0]:
+                cut_table_new["cut"].value[low_] = cut_table["cut"].value[
+                    high_event_bins[0]
+                ]
+            elif low_ > high_event_bins[-1]:
+                cut_table_new["cut"].value[low_] = cut_table["cut"].value[
+                    high_event_bins[-1]
+                ]
```
moralejo, line 173: This does not consider the possibility (perhaps unlikely) that you have a bin without enough statistics which has higher energy than a bin with enough statistics (or the corresponding case in the high-E end of the range). In that case an interpolation could be used.

moralejo (changes requested):

lstchain/io/event_selection.py
maxnoe: Also, using `np.where` just to get the indices is deprecated I think (at least it's slower than just using `np.nonzero`)

maxnoe: > np.argwhere(a) is almost the same as np.transpose(np.nonzero(a)), but produces a result of the correct shape for a 0D array.

> The output of argwhere is not suitable for indexing arrays. For this purpose use nonzero(a) instead.

chaimain: Thanks for the suggestions! Changed to using `np.nonzero`

lstchain/io/event_selection.py
chaimain: Added coverage for this possibility and updated the test accordingly.

lstchain/high_level/hdu_table.py
chaimain: Fixed the docstring. Maybe we can update the function name as well, to something like `use_expected_pos_for_reco_altaz` or `fill_reco_altaz_w_expected_pos`. As this function is only used for source-dependent analysis @SeiyaNozaki do you have a preference?

moralejo: I think the function name is fine

lstchain/io/event_selection.py
chaimain: I think, in that even more unlikely case, we/analyzer have provided a very large value for `min_event_p_en_bin` with respect to the simulation data that we have, and we can tune that instead of covering for this computational possibility.

SeiyaNozaki (approved):

rlopezcoto:
@moralejo changes were made, anything else missing?

moralejo (approved):

PR #1110, opened 2023-05-24
Bump to ctapipe v0.19

rlopezcoto:


maxnoe:
I started the conda-forge build for ctapipe_io_lst 0.20.1, it should be online in an hour or so. Then the tests should pass here.

PR #1111, opened 2023-05-24
dl1ab writing using ctapipe write_table

vuillaut:
Follow up of #1031 using ctapipe read and write_table (ctapipe > v0.12 was required)


lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -218,132 +219,133 @@ def main():
...
 
-            outfile.root[dl1_params_lstcam_key][:] = params
-            if image_mask_save:
-                outfile.root[dl1_images_lstcam_key].modify_column(colname='image_mask', column=image_mask)
+        # reset added parameters types
+        for p in new_params:
+            params[p] = params[p].astype(type(dl1_container[p][0]))
 
         write_metadata(metadata, args.output_file)
+        write_table(params, args.output_file, dl1_params_lstcam_key, overwrite=True, filters=HDF5_ZSTD_FILTERS)
```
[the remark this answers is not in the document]
maxnoe: Edit, sorry, no... that applies only to the table itself, never mind.

However, you could keep the file open with tables, and directly pass the open `tables.File` to `write_table`

gabemery:
Are these methods usable for the first dl1 file writing in r0_to_dl1?

vuillaut:
> Are these methods usable for the first dl1 file writing in r0_to_dl1?

It could be, but I don't want to consider this here and _fix_ a working process (that should already be modified to use `ctapipe-process`).
The goal here was to solve an issue explained in #1031 while simplifying parameters writing in the dl1ab process.

maxnoe:
> Are these methods usable for the first dl1 file writing in r0_to_dl1?

r0_to_dl1 uses the ctapipe TableWriter, which produces the same output format as `write_table`, from containers instead of astropy tables. that is appropriate for r0_to_dl1, since it iterates over the events from the source.

gabemery:
Ok, thanks for the clarifications

rlopezcoto:
is there anything missing with this PR or are you happy with the solutions implemented/answers @maxnoe, @gabemery ?

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -211,29 +212,24 @@ def main():
...
             set_numba_seed(infile.root.dl1.event.subarray.trigger.col('obs_id')[0])
 
         image_mask_save = not args.no_image and 'image_mask' in infile.root[dl1_images_lstcam_key].colnames
+        if image_mask_save:
+            image_mask = image_table['image_mask']
+
+        params = read_table(infile, dl1_params_lstcam_key)
+        new_params = set(parameters_to_update) - set(params.colnames)
+        for p in new_params:
+            params[p] = np.empty(len(params), dtype=np.float32)
```
maxnoe, line 227: Looking at the list of `parameters_to_update`, not all of them are float32, there is for example `n_pixels`.

maxnoe: Well, how about something like this:
```
parameter_dtypes = {
    "n_pixels":  np.int32,
    "n_islands":  np.int32,
}


...


params[p] = np.empty(len(params), dtype=parameter_dtypes.get(p, np.float32)`
```

vuillaut: that's a possibility but not so flexible if new parameters are added.
what about guessing types from the dl1_container?

maxnoe: You could look at `type(d1_container.<field>.default)`, yes.

maxnoe: However, it seems these are using python values / float64 at the moment, which should probably be corrected...

vuillaut: Thanks @maxnoe .
Following this, I made the corresponding changes, setting all types in `DL1ParametersContainer`

lstchain/io/lstcontainers.py
```diff
@@ -35,96 +35,81 @@ class DL1ParametersContainer(Container):
...
-    tel_pos_z = Field(None, "Telescope z position in the ground")
-
-    trigger_type = Field(None, "Trigger type")
-    ucts_trigger_type = Field(None, "UCTS trigger type")
-    trigger_time = Field(None, "Trigger time")
-    event_type = Field(None, "Event type")
+    intensity = Field(np.nan, 'Total intensity (size)', dtype=np.float32)
+    log_intensity = Field(np.nan, 'log of total intensity (size)', dtype=np.float32)
+
+    x = Field(np.nan, 'Centroid x coordinate', dtype=np.float32, unit=u.m)
```
maxnoe, line 41: Default must be the correct dtype and unit, so `u.Quantity(np.float32(np.nan), u.m)` in this case.

We should probably validate this in ctapipe...

vuillaut: I would have thought that the `Field` would set the type and unit to the default as well. Making changes then.

lstchain/scripts/lstchain_dl1ab.py
vuillaut: Was `float64` really necessary for the `intensity` and `log_intensity` ?

rlopezcoto:
tests are not passing, any idea @vuillaut ?

vuillaut:
> tests are not passing, any idea @vuillaut ?

yep, working on it

vuillaut:
Hi.
I integrated @maxnoe change request and merge `ctapipe_0.17` branch. If you could have another look 🙏

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -243,46 +242,46 @@ def main():
...
-
-            new_params = set(parameters_to_update) - set(params_node.colnames)
-            if new_params:
-                log.warning(f"Parameters not in original DL1 file {args.input_file} that can't be recomputed:"
-                            f"{new_params}")
-            parameters_to_update = list(set(parameters_to_update) & set(params_node.colnames))
-            if image_mask_save:
-                image_mask = outfile.root[dl1_images_lstcam_key].col('image_mask')
-
-            if catB_calib:
```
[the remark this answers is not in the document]
vuillaut: - the condition on `image_mask_save` is moved line 274

- the condition on `catB_calib` was useless, the image table is already loaded, as we need the image in any case (line  245). Modifying the images and saving them is under `catB_calib` condition line 298 and line 426)

rlopezcoto (approved):
I think this is good to go, if @maxnoe has time to have a final look, it would be great

PR #1112, opened 2023-05-24
Update muon_analysis.py

ferxinii:
With these changes, the fit_muon function now fits a ring correctly even under high noise conditions. This was previously not the case. The changes made are:

-the min. number of pixel neighbors to survive the tailcuts cleaning is now set to 2 (previously was set to default, which is 0). -If no tailcuts are indicated as an input, the function will select appropiate tailcuts based on the noise of the image. 

Also, in the analyze_muon_event function, when calling the fit_muon function in line 231, the 'tailcuts' input has been removed. This way, when using analyze_muon_event, the tailcuts will be adapted to the image's noise.


lstchain/image/muon/muon_analysis.py
```diff
@@ -125,13 +127,34 @@ def fit_muon(x, y, image, geom, tailcuts):
...
         Image after cleaning
     """
-
+    
+    if tailcuts == None:
+        #we want to quantize the noise of the image. To do so, we will use the negative Q cumulative distribution.
+        negative_Q = np.sort(image[image <= 0])
+        
+        hist, bins = np.histogram(negative_Q,range=(-15,0),bins=30)
+        bins=bins[:-1]
```
rlopezcoto, line 136: comment from https://github.com/cta-observatory/cta-lstchain/pull/1012#discussion_r964947784

lstchain/image/muon/muon_analysis.py
```diff
@@ -125,13 +127,34 @@ def fit_muon(x, y, image, geom, tailcuts):
...
+        
+        normalised_cumulative_hist = np.asarray(np.cumsum(hist)/np.sum(hist))
+
+        idx = (np.abs(normalised_cumulative_hist - 0.318)).argmin() #Find q closest to standard deviation
+        dev = np.abs(bins[idx])
+        #MORALEJO: we want to get, from a single image, a quantity related to the width of the noise distribution,
```
rlopezcoto, line 142: this comment can probably be edited rather than copy/paste that of:
https://github.com/cta-observatory/cta-lstchain/pull/1012#discussion_r964953348

rlopezcoto (commented):
Did you check that the fits give similar results to those we are getting at the moment?
Also, I am not sure that
> -If no tailcuts are indicated as an input, the function will select appropiate tailcuts based on the noise of the image.

is exactly the way to go since we may be removing too many pixels for noisy images, which could affect the total charge extracted. @moralejo did you take that into account when you reviewed this PR?

morcuended:
Are these changes still valid? If so, we should try to finish this PR.

lstchain/image/muon/muon_analysis.py
```diff
@@ -125,13 +127,34 @@ def fit_muon(x, y, image, geom, tailcuts):
     image_clean: `np.ndarray`
         Image after cleaning
     """
-
+    
+    if tailcuts == None:
+        #we want to quantize the noise of the image. To do so, we will use the negative Q cumulative distribution.
```
maxnoe, line 132: Why do this instead of using the pedestal std deviation?

lstchain/image/muon/muon_analysis.py
```diff
@@ -135,7 +135,7 @@ def fit_muon(x, y, image, geom, tailcuts=None):
         hist, bins = np.histogram(negative_Q,range=(-15,0),bins=30)
         bins=bins[:-1]
         
-        normalised_cumulative_hist = np.asarray(np.cumsum(hist)/np.sum(hist))
+        normalised_cumulative_hist = np.cumsum(hist)/np.sum(hist)
```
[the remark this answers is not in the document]
maxnoe: or just look for 0.318 * hist[-1] to not even divide everything again

moralejo:
To add some context: this was done as a summer student project, but time was too short to finalize it.

> Did you check that the fits give similar results to those we are getting at the moment? Also, I am not sure that
> 
> > -If no tailcuts are indicated as an input, the function will select appropiate tailcuts based on the noise of the image.
> 
> is exactly the way to go since we may be removing too many pixels for noisy images, which could affect the total charge extracted. @moralejo did you take that into account when you reviewed this PR?

This will affect only the geometrical ring fit. Only through the ring parameters can the charge-related parameters (computed later) be modified. And they are computed in a way (i.e. rather wide band around the ring) that they should be rather robust against small changes of ring parameters. 

if 4-sigma, 2-sigma tail cuts (which should correspond to ~7.0, 3.5 pe for very dark data) remove a significant part of the ring charge then the ring is too dim for proper analysis anyway.  It may be the case for 10 * darkNSB, which is our limit of operation... On the other hand, muon calibration under such conditions may not be absolutely needed (one can check the calibration from close-in-time data with less NSB)

moralejo:
The original idea of this was, "if we can clearly see the rings on a single image with high NSB (for which nevertheless the standard ring analysis failed completely), and distinguish it clearly from the noise, then it should also be possible to obtain the ring parameters using only the info contained in the event".

moralejo:
BTW what would be the correct way to merge the (many) change in the main branch into this one? I tried, but there are lots of conflicts which I think should not be there (seem totally unrelated to the only file that was changed in this branch)

maxnoe:
This branch doesn't have any conflicts with main and can be merged as is. If you want to update this branch before merging, you can merge or rebase

This just works and doesn't create conflicts:

```
❯ git fetch
❯ git switch muon_analysis
Switched to branch 'muon_analysis'
Your branch is up to date with 'origin/muon_analysis'.
❯ git merge origin/main
```


As does rebasing:

```
❯ git fetch
❯ git switch muon_analysis
❯ git rebase origin/main
```

lstchain/image/muon/muon_analysis.py
```diff
@@ -122,13 +124,36 @@ def fit_muon(x, y, image, geom, tailcuts):
...
         Image after cleaning
     """
-
+    
+    if tailcuts == None:
+        # We want to quantify the noise of the image. To do so, we will use the 
+        # negative Q cumulative distribution.
+        negative_Q = np.sort(image[image <= 0])
+        
+        hist, bins = np.histogram(negative_Q,range=(-15,0),bins=30)
```
morcuended, line 133: why this range from -15?

moralejo: This is in p.e. , and was considered a reasonable value to include the negative fluctuations, based on inspection of moon data. Could of course be made wider.

lstchain/image/muon/muon_analysis.py
```diff
@@ -122,13 +124,36 @@ def fit_muon(x, y, image, geom, tailcuts):
     image_clean: `np.ndarray`
         Image after cleaning
     """
-
+    
+    if tailcuts == None:
```
morcuended, line 128:
```suggestion
    if tailcuts is None:
```

morcuended (approved):
It looks good to me

morcuended (approved):
looks good to me

PR #1119, opened 2023-06-13
Use numba njit in all loops to inject NSB in R1 waveforms

aaguasca:
This PR aims to optimize the NSB tuning at the waveform level grouping all loops in `tune_nsb_on_waveform` and `template_interpolation` in a single function (`template_interpolation_withNSB`). Then, applying njit to speed up the loop computation. `template_interpolation_withNSB` overrules the call of `NormalizedPulseTemplate` as `pulse_templates`

I adapted the function log_pdf to use `template_interpolation_withNSB` despite using only the interpolation part without adding NSB.

Below you can see the comparison between the pixel charge distribution for all MC gammas using the commit 66ef7f9 and using the modifications in this PR in orange and black, respectively. The legend also shows the time required to execute the r0_to_dl1 step for each one considering a `nsb_tuning_ratio=24.5`. The approach in this PR is ~16 times faster than the one in 66ef7f9. The higher the `nsb_tuning_ratio` value, the higher the improvement.

comparison_allnjt_vs_PR1066

### DIFF
--- a/lstchain/reco/reconstructorCC.py
+++ b/lstchain/reco/reconstructorCC.py
@@ -228,6 +228,71 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
 
 
 @njit(cache=True)
+def template_interpolation_withNSB(empty_waveform, time, is_high_gain, additional_nsb, amplitude, t_0,
+                           t0_template, dt_template, a_hg_template, a_lg_template):
+    """
+    Fast template interpolator using uniformly sampled base with known origin and step to add additional_nsb photons
+    as NSB with a certain amplitude at t_0. The code works for all pixel and selected gains.
+
+    Code interpolation duplicated from function template_interpolation.
+
+    Parameters
+    ----------
+    empty_waveform: empty charge (p.e. / ns) in each pixel and sampled time.
+    time: float64 1D array
+        Times of each waveform samples
+    is_high_gain: boolean 1D array
+        Gain channel used per pixel: True=hg, False=lg
+    additional_nsb: float64 1D array
+        Number of NSB photons to add per pixel
+    amplitude: float 2D array
+        Normalisation factor to apply to the template per photon in each pixel
+    t_0: float 2D array
+        Shift in the origin of time per photon in each pixel
+    t0_template: float64
+        Time of the first value of the pulse templates
+    dt_template: float 64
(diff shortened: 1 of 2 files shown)


gabemery:
@aaguasca You will need to redo the PR using a branch of the lstchain repository instead of using a branch from a fork. If you don't tests will fail. I will still look a t the code now.

lstchain/reco/reconstructorCC.py
```diff
@@ -293,12 +358,26 @@ def log_pdf(charge, t_cm, x_cm, y_cm, length, wl, psi, v, rl,
...
+        is_high_gain=is_high_gain,
+        additional_nsb=np.full(n_pixels, n_samples),
+        amplitude=np.ones(data.shape),
+        t_0=t_0,
+        t0_template=template_t0,
+        dt_template=template_dt,
+        a_hg_template=template_hg,
+        a_lg_template=template_lg
+    )
+
```
gabemery, line 380: I am not convinced by changing the call in log_pdf.
The loop before the function is already in njit code. You add multiple operations before the call to the interpolation. And the code inside is in my opinion not faster. (Would need to test).

gabemery (changes requested):
Very nice processing duration improvements for the NSB injection. But I am not convinced by the change to the log_pdf function for 2 main reasons. I don't see any reason this more complex code would be faster and the current version may be wrong.

Changes for the NSB injection require minor corrections (and proof tests are passing). 
Changes to the log_pdf method need proof that it is faster, one pulse is produced for each pixel and pulse is the same as with the previous interpolation.

lstchain/reco/reconstructorCC.py
aaguasca: You are right, even though if I use ` t_0 = (time_shift + t_model)[:, None] * (np.ones((n_pixels, n_samples), dtype=np.float64))` the loop format in njit code is faster than the numpy product. t=4.8e-05 s and 3.3e-05 s, respectively. 
Using `template_interpolation` in `log_pdf` is fast, clear and tested. I will undo the changes in `log_pdf`

aaguasca:
Thanks for the comments, @gabemery! I opened a new PR (PR #1120) to continue this PR in a branch of the lstchain repository. I implemented your comments in PR #1120.

PR #1120, opened 2023-06-14
Create pure NSB R1 waveforms

aaguasca:
This PR is the continuation of PR #1119 using a branch of the lstchain repository. I have already applied @gabemery's suggestions in PR #1119.


lstchain/reco/reconstructorCC.py
```diff
@@ -227,6 +227,69 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
     return out
 
 
+@njit(cache=True)
+def template_interpolation_withNSB(empty_waveform, time, is_high_gain, additional_nsb, amplitude, t_0,
```
gabemery, line 231: The method name is not really representative of what this code actually does. It is not the template interpolation method with NSB. This method creates NSB only waveforms.

lstchain/reco/reconstructorCC.py
```diff
@@ -227,6 +227,69 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
...
+    dt_template: float 64
+        Time step between templates values
+    a_hg_template: float64 1D array
+        Template values for the high gain channel
+    a_lg_template: float64 1D array
+        Template values for the low gain channel
+
+    Returns
+    -------
+    dummy_waveform: float64 2D array
```
gabemery, line 263: Either replace dummy_waveform by "nsb_waveform", "empty_waveform" or don't name it here.

lstchain/reco/reconstructorCC.py
```diff
@@ -227,6 +227,69 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
...
+    n_pixels, _ = empty_waveform.shape
+    m = time.shape[0]
+    size = a_hg_template.shape[0]
+
+    for i in range(n_pixels):
+
+        for j in range(additional_nsb[i]):
+            single_spe_waveform = np.empty(m)
+            # Find the index before the requested time
+            a = (time - t_0[i][j] - t0_template) / dt_template
```
gabemery, line 275: Copy from the previous PR since it was not answered : 
"


Since you are creating a new times object for each additionnal_nsb, you do time-t0_template n_pixels*mean(additionnal_nsb)*m times. If you did time-t0 _template early it would only be done m times.

Probably a minor optimisation but it should still be a bit faster. (It is not the case in the original interpolation function since times is 2D there)
"

gabemery (changes requested):
Minor leftover comments.

lstchain/reco/reconstructorCC.py
aaguasca: Thanks, @gabemery, for your review! I like `nsb_only_waveforms`. What do you think?

lstchain/reco/reconstructorCC.py
aaguasca: I renamed `empty_waveform`  to `nsb_waveform`. I adapted the returns docs accordingly.

lstchain/reco/reconstructorCC.py
aaguasca: Ups, I forgot to move `single_spe_waveform` outside the loops... Putting it outside the loops makes the computation ~30% faster

lstchain/reco/reconstructorCC.py
gabemery: I think it works

lstchain/reco/reconstructorCC.py
```diff
@@ -227,6 +227,69 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
     return out
 
 
+@njit(cache=True)
+def nsb_only_waveforms(nsb_waveform, time, is_high_gain, additional_nsb, amplitude, t_0,
+                           t0_template, dt_template, a_hg_template, a_lg_template):
+    """
+    Generate waveforms of pure NSB. NSB photons injected through a fast interpolator using a
+    uniformly sampled base with known origin and step with a certain amplitude at t_0.
```
gabemery, line 235: Generate waveforms of pure NSB -> technically add pure NSB to a waveform since you take a waveform as input. 

I see two way to change the code (current version is also good, simply ideas). 
1: Keep the function the same but pass the waveform to be modified instead of creating an intermediate zeros waveform. (function could then be "add_nsb_only_waveform") -> advantage : no intermediate waveform object and removes the final sum out of the function
2: Create the zeros waveform in the function -> advantage : removes input and guaranties that the output is indeed an NSB only waveform. (get n_pixels from "additional_nsb") -> possible problem : call to numpy possible and as fast in this case?

gabemery, line 235: "uniformly sampled base with known origin and step with a certain amplitude at t_0" -> uniformly sampled normalised template with known time of first value and time step. The method requires as input the number of NSB events to inject per pixel, the times of injection and the events normalisations.

lstchain/reco/reconstructorCC.py
```diff
@@ -227,6 +227,69 @@ def template_interpolation(gain, times, t0, dt, a_hg, a_lg):
...
+
+    Code interpolation duplicated from function template_interpolation.
```
gabemery, line 237: Interpolation code ...

lstchain/reco/reconstructorCC.py
gabemery: You missed my point but good (unrelated) change

lstchain/reco/reconstructorCC.py
aaguasca: I would prefer option 2. Maybe it is more clean than option 1 and the the overall time of the function `tune_nsb_on_waveform` should be the same.

PR #1121, opened 2023-06-15
Update dependencies to current versions, add python 3.11 CI

rlopezcoto:
There was a [gammapy release](https://github.com/gammapy/gammapy/releases/tag/v1.1) very recently and we can also follow the latest [ctapipe](https://github.com/cta-observatory/ctapipe/releases/tag/v0.19.2) one

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.9
 
       - name: Install doc dependencies
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,8 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
-        ctapipe-version: ["v0.19.0"]
+        python-version: ["3.9", "3.10", "3.11"]
+        ctapipe-version: ["v0.19.2"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


rlopezcoto:
it seems that the problem is still the support of python 3.8 in the ci due to the new version of gammapy, shall we drop it completely in the ci?

maxnoe:
@rlopezcoto Yes, dropping 3.8 is fine.

Let's drop 3.8 and add a 3.11 test

maxnoe:
Update scikit-learn to 1.2 for 3.11 support

maxnoe:
aaaand matplotlib... current one is 3.7

maxnoe:
The docs build also needs to run on at least 3.9 now

rlopezcoto:
> aaaand matplotlib... current one is 3.7

just saw :D
the next one we downgrade to gammapy 1.0 and python 3.8 :P

rlopezcoto:
> The docs build also needs to run on at least 3.9 now

what about pyflakes?

maxnoe:
> what about pyflakes?

It should run on the lowest version we still support

maxnoe:
Now it's numpy, which we for some reason pin to excatly 1.21, I think I'd just remove the pin

maxnoe:
The setup.cfg still has 

>         'numpy<1.22.0a0',

rlopezcoto:
I do not see that one, the only one to be modified should be:
https://github.com/cta-observatory/cta-lstchain/blob/7e0f1646cd682917a579407a98e21a2889a9b4c0/setup.cfg#L13
right?

maxnoe:
Ah, I looked on Master, I probably already removed it in the 0.17 branch

rlopezcoto:
uff, it was tougher than it seemed, thanks for the help @maxnoe !

maxnoe (approved):

PR #1123, opened 2023-06-16
Catch all exceptions in create_run_summary

maxnoe:
The EVBv6 test files created an exception that was not handled, preventing the creating of the run summary.

I think it does not really make sense here to check for specific exceptions, anything we can't identify is probably not analyzable at the moment.

### DIFF
--- a/lstchain/scripts/lstchain_create_run_summary.py
+++ b/lstchain/scripts/lstchain_create_run_summary.py
@@ -182,7 +182,7 @@ def type_of_run(date_path, run_number, counters, n_events=500):
         else:
             run_type = "ERROR"
 
-    except (AttributeError, ValueError, IOError, IndexError) as err:
+    except Exception as err:
         log.error(f"File {filename} has error: {err!r}")
 
         run_type = "ERROR"


morcuended (dismissed):

rlopezcoto:
> @rlopezcoto This was meant to go in **before** the update to all the dependencies.

why? it needs to work with the new ones as well

maxnoe:
I guess it's ok to create the run summary with the new version and then run processing with 0.9.x for now @morcuended ?

Or do we need to revert?

rlopezcoto:
ok, you wanted to have a `master` with the run summary with this bug corrected... fine, we can revert, but I would just get everything merged and create the run summary with lstchain v0.10 that should be released as soon as we get this PR and #1120 merged

morcuended:
Yes, it is fine. Actually, I asked also to move new test files elsewhere, so I guess we could even live without the fix (but better to have it)

morcuended:
> ok, you wanted to have a master with the run summary with this bug corrected... fine, we can revert, but I would just get everything merged and create the run summary with lstchain v0.10 that should be released as soon as we get this PR and https://github.com/cta-observatory/cta-lstchain/pull/1120 merged

No need to revert

morcuended (approved):

PR #1124, opened 2023-06-16
Use interleaved events to get the pedestal std deviation in the likelihood reco

gabemery:
Add a small method to estimate the pedestal deviation pixel-wise from interleaved events instead of event-wise from signal-less pixels in the likelihood reconstruction method.
Also remove obviously dimmed pixels for the subrun.

Will likely become mandatory to analyse DVRed data.


maxnoe:
@FrancaCassol ?

gabemery:
@maxnoe Right, I wrote this before the implementation of the CAT B files. If I don't try to get it in today's release and target the next minor one I can check if using CAT B calibration files works.

PR #1125, opened 2023-06-19
Require at least ctapipe_io_lst 0.21.1

maxnoe:


rlopezcoto (commented):
can you also pin python 3.11 in the environment?

PR #1129, opened 2023-06-23
Make lstchain_dvr_pixselector.py back-compatible with v0.9 DL1

moralejo:
Make lstchain_dvr_pixselector.py back-compatible with DL1 files containing obsolete instrument description: now the camera geometry is obtained from the LST event source.

Also: do not skip files in which no events are tagged as cosmics. In some files the tagging failed, and cosmics are actually tagged as "unknown" (whereas FF and pedestal are heuristically recognized in the DL1 file). We now assume those are cosmics.

### DIFF
--- a/lstchain/scripts/lstchain_dvr_pixselector.py
+++ b/lstchain/scripts/lstchain_dvr_pixselector.py
@@ -233,8 +233,8 @@ def main():
         data_parameters = read_table(dl1_file, dl1_params_lstcam_key)
 
         # Get the LST camera geometry:
-        sa = LSTEventSource.create_subarray(tel_id=1)
-        camera_geom = sa.tel[1].camera.geometry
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -232,10 +232,10 @@ def main():
 
         data_parameters = read_table(dl1_file, dl1_params_lstcam_key)
 
-        # Read some useful extra info from the file:
-        subarray_info = SubarrayDescription.from_hdf(dl1_file)
-        camera_geom = subarray_info.tel[1].camera.geometry
-
+        # Get the LST camera geometry:
+        sa = LSTEventSource.create_subarray(tel_id=1)
```
rlopezcoto, line 236: I would be more descriptive with variable names:
sa -> subarray

rlopezcoto (approved):

Issue #1048, opened 2022-11-22
Use correction for aberration in source-dependent analysis


moralejo:
https://github.com/cta-observatory/cta-lstchain/blob/3c75e08cbf5035f701408562f97e217880abe62d/lstchain/reco/dl1_to_dl2.py#L355-L356

Hi @SeiyaNozaki, above we are using the nominal focal length of 28 m (this is what is inside the files, as far as I know) for the source-dependent analysis. There could be some (marginal) gain if we use a correction for aberration (sharper Alpha distributions and perhaps a bit better performance of gammaness). Most likely this would be visible only at high energies, where the small difference may be significant considering the "sharpness" of the alpha distributions.

SeiyaNozaki:
Following the suggestion, now I want to create `OpticsDescription` in case that subarray table is not readable (because of difference of ctapipe version). 

But I got the following error... Am I doing wrongly or is this ctapipe bug...?
https://github.com/cta-observatory/ctapipe/blob/main/ctapipe/instrument/optics.py#L199-L207
```
>>> from ctapipe.instrument import OpticsDescription
>>> 
>>> OpticsDescription.from_name("LST")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/fefs/aswg/software/conda/envs/lstchain-v0.10.0/lib/python3.10/site-packages/ctapipe/instrument/optics.py", line 207, in from_name
    raise ValueError(f"Unsupported version of optics table: {version}")
ValueError: Unsupported version of optics table: None
>>> 
>>> from ctapipe.utils import get_table_dataset
>>> table = get_table_dataset("optics", role="OpticsDescription.from_name")
>>> table
<Table length=9>
tel_description tel_type tel_subtype  mirror_area  mirror_type num_mirror_tiles equivalent_focal_length
                                           m2                                              m           
     str14        str5      str10       float64        str2         int64               float64        
--------------- -------- ----------- ------------- ----------- ---------------- -----------------------
        MST-SCT      MST         SCT 73.3475723267          SC                2           5.58629989624
        SST-GCT      SST         GCT 12.5663709641          SC                2           2.28299999237
            LST      LST          --  386.73324585          DC              198                    28.0
          MAGIC    MAGIC          --         239.0          DC              964                   16.97
      SST-ASTRI      SST       ASTRI 14.5625667572          SC                2           2.15000009537
            MST      MST          -- 103.830558777          DC               84                    16.0
[...]

PR #1131, opened 2023-07-03
Use effective focal length to compute src-dep parameters

SeiyaNozaki:
Related to #1048 
This PR allows to using the effective focal length (instead of the equivalent focal length:28 m) to compute source-dep parameters. Due to the version incompatibility, it raises an error when reading subarray table using lstchain v0.10 + DL1 data processed by lstchain v0.9. So, if it fails to read subarray table to get the effective focal length, it just uses the standard LST effective focal length.


lstchain/reco/dl1_to_dl2.py
```diff
@@ -349,10 +350,18 @@ def build_models(filegammas, fileprotons,
...
-            subarray_info = SubarrayDescription.from_hdf(filegammas)
-            tel_id = config["allowed_tels"][0] if "allowed_tels" in config else 1
-            focal_length = subarray_info.tel[tel_id].optics.equivalent_focal_length
-            src_dep_df_gamma = get_source_dependent_parameters(df_gamma, config, focal_length=focal_length)
+            try:
+                subarray_info = SubarrayDescription.from_hdf(filegammas)
+                tel_id = config["allowed_tels"][0] if "allowed_tels" in config else 1
+                effective_focal_length = subarray_info.tel[tel_id].optics.effective_focal_length
+            except OSError:
+                print("subarray table is not readable because of the version inompatibility.")
```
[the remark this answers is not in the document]

moralejo:
@SeiyaNozaki I am not 100% sure this is correct.
I see the effective focal length is used to compute the expected source position (in the call to get_expected_source_pos).

However, if you use the effective focal length for computing the image parameters, then it is like having no aberration in the reco params (e.g. reco direction).  Hence the expected source position (to be used as reference for the src-dep parameters computations) should be calculated with the nominal focal length, right?

moralejo:
Could you have a look at this, @SeiyaNozaki?

SeiyaNozaki:
Just let me summarize the current parameterization and how we should correct aberration effect.

If I understand correctly, hillas parameters are computed in camera frame (unit: m) in ctapipe. 

When computing parameters related MC true gamma-ray position (including `src_x`, `src_y`) inside `add_disp_to_parameters_table`, equivalent focal length is used for the conversion of MC gamma-ray arrival direction from sky coordinate to camera frame. So those values are not aberration-corrected.
https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/r0_to_dl1.py#L583
https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/r0_to_dl1.py#L805-L810


`alpha` and `dist` are computed using `x`, `y`, `expected_src_x`, `expected_src_y` (all in camera frame), and `psi`. So those expected source position should be computed after aberration correction (by using effective focal length). It means `src_x` and `src_y` (for MC) should also computed after aberration correction 
https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/dl1_to_dl2.py#L769-L771

Am I correct...?

maxnoe:
I think the preferred solution here would but to just switch to computation in telescope frame for everything, not piece-by-piece applying effective focal length corrections to different parts of the code.

SeiyaNozaki:
what do you think @moralejo ? It could be a big change, but we may need to apply this change at some point.

moralejo:
> 
> If I understand correctly, hillas parameters are computed in camera frame (unit: m) in ~ctapipe~ lstchain.
> 

Ok, so they are in meters, and measured "on the camera" (i.e. the effect of aberration is present, e.g. centroids are slightly further out than they would for ideal optics with the "equivalent focal length" of 28 m) 

> When computing parameters related MC true gamma-ray position (including `src_x`, `src_y`) inside `add_disp_to_parameters_table`, equivalent focal length is used for the conversion of MC gamma-ray arrival direction from sky coordinate to camera frame. So those values are not aberration-corrected.

https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/r0_to_dl1.py#L788

I think there is a problem there. You pass to the function the equivalent focal length. This is used to calculate the true source position on the camera (in m). That is the position for ideal optics (28 m focal), but the computed image parameters are affected by aberration.  Either we calculate the true source position modified by aberration, OR we correct the aberration in the image parameters (in this case just the centroid position).

The way it is now, the disp_dx, disp_dy, disp_norm will come from aberration-affected (centroid_x, centroid_y) and aberration_free (source_position_x, source_position_y). Probably the effect is not large for the source-INdependent analysis, because the RF would "learn" to estimate the source position (for ideal optics) from the aberration-affected images.

moralejo:
> https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/r0_to_dl1.py#L583
> 
> 
> https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/r0_to_dl1.py#L805-L810
> 
> `alpha` and `dist` are computed using `x`, `y`, `expected_src_x`, `expected_src_y` (all in camera frame), and `psi`. So those expected source position should be computed after aberration correction (by using effective focal length). It means `src_x` and `src_y` (for MC) should also computed after aberration correction

If you want to use the parameters in camera frame (=aberration-affected) then, for calculating src-dep params, one should use the "true source position" also affected by aberration (=computed using the effective focal length).  But this is not what we do now.

maxnoe:
> Seems the cleanest solution, yes, if the aberration correction is applied (i.e. effective focal length is always used) when transforming the image parameters from camera to telescope frame.

Technically, we don't transform the image parameters. We transform the pixel coordinates and then compute image parameters in the TelescopeFrame.

moralejo:
> Better than _correcting_ the image parameters after the fact would be already computing them so they are correct, which is what computing them in the `TelescopeFrame` does, if the effective focal length is used to transform the pixel coordinates from `CameraFrame` to `TelescopeFrame`.

Yes, I agree.

> 
> In `r0_to_dl1` and `dl1ab`, we would need to transform the camera geometry:
> 
> ```
> geometry_tel_frame = geometry.transform_to(TelescopeFrame())
> ```
> 
> and then just use that geometry for the computation of all image parameters.
> 
> `ctapipe_io_lst` always fills the `effective_focal_length` into the `frame` of `CameraGeometry`.

Seems simple enough... Would it then be ok to make that part of this PR?

BTW, I think all this still allows running dl1_to_dl2 on existing DL1 files (from v0.9), right? (eff. focal length is hard-coded)

maxnoe:
> BTW, I think all this still allows running dl1_to_dl2 on existing DL1 files (from v0.9), right? (eff. focal length is hard-coded)

Starting from the images yes, the parameters will change though.

moralejo:
Ah, right, no problem **if** we re-do the DL1ab part...

Current DL1b:  

- image parameters in camera frame (in m, obviously "aberration-affected" - apologies for the jargon)
- for MC: true source position in camera frame, src_x, src_y in m, but computed for "ideal f=28 m optics" (hence should not be "mixed" with the image parameters without correcting one or the other!)

If we go for the change, DL1b would be like:

- image parameters in telescope frame (in deg, and "corrected for aberration")
- for MC: true source position in deg in telescope frame src_x, src_y  would be  (src_fov_lon, src_fov_lat). This is fully consistent with the image parameters.

I think we should go for this change before the next reprocessing (from raw) of the full sample, but I hesitate to go for it straightaway. Perhaps it is better to make sure, with as few changes as possible, that the source-dependent analysis is properly done, to check if the discrepancies we see with source-independent may be related to this aberration issue.

maxnoe:
> image parameters in camera frame (in m, obviously "aberration-affected" - apologies for the jargon)

I think its even worse... length and width are transformed by hand into degrees using the equivalent focal length, the rest of the image parameters is left untouched and is in meters or the related units.

moralejo:
Let me summarize my understanding of the current code in this PR:

In [get_expected_source_pos:](https://github.com/cta-observatory/cta-lstchain/blob/8543db816fd528f5d7cc073b10dd3c0b2645b62b/lstchain/reco/dl1_to_dl2.py#L779)

https://github.com/cta-observatory/cta-lstchain/blob/8543db816fd528f5d7cc073b10dd3c0b2645b62b/lstchain/reco/dl1_to_dl2.py#L789-L792

Note that the focal length passed as an argument to  get_expected_source_pos is **not** used at all for MC gammas (only for MC protons). The pre-calculated values data['src_x'] (and src_y) are used for gammas - and they were computed with the equivalent focal length (i.e. 28 m), in r0_to_dl1:  
https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/r0_to_dl1.py#L805-L810

Now, this source position computed with f=28 m, i.e. assuming perfect optics, is used to calculate the source-dependent parameters (from image parameters in camera frame, i.e. "affected by aberration"): 
https://github.com/cta-observatory/cta-lstchain/blob/8543db816fd528f5d7cc073b10dd3c0b2645b62b/lstchain/reco/dl1_to_dl2.py#L731

This is what is not correct in my opinion. 
I think the solution (for the source-dependent analysis) is to modify get_expected_source_pos to recalculate the "true gamma direction" using the effective focal length (i.e. introducing the aberration). With a call to sky_to_camera using the effective focal length (i.e. the same as shown above in r0_to_dl1, but with the effective focal)

moralejo:
That is for source-dependent analysis. For source-independent the target disp values have the same problem, so that should be corrected as well.

SeiyaNozaki:
Okay, I agree with @moralejo 

Strictly speaking, `reco_disp_norm_diff` is computed as the difference between `dist` and `reco_disp_norm`.
https://github.com/cta-observatory/cta-lstchain/blob/8543db816fd528f5d7cc073b10dd3c0b2645b62b/lstchain/reco/dl1_to_dl2.py#L534

So I want just to update all disp parameters (including `src_x`, `src_y`) for MC simulation events using effective focal length for source-dependent analysis. Does it make sense?

moralejo:
> 
> Strictly speaking, `reco_disp_norm_diff` is computed as the difference between `dist` and `reco_disp_norm`.
> 
> https://github.com/cta-observatory/cta-lstchain/blob/8543db816fd528f5d7cc073b10dd3c0b2645b62b/lstchain/reco/dl1_to_dl2.py#L534
> 
> So I want just to update all disp parameters (including `src_x`, `src_y`) for MC simulation events using effective focal length for source-dependent analysis. Does it make sense?

Sorry, I don't understand what you mean. If we recalculate (src_x, src_y) with effective focal length then everything is in camera frame, and all source-dep parameters will be correct, right?

moralejo:
As far as I could see, all calls to apply_models use the equivalent focal length, 28 m:
https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/reco/dl1_to_dl2.py#L533

This means that the conversion from reco event direction in camera frame to Alt Az is done ignoring the effect of aberration. This is not correct, but  may be partly compensated by the fact that the RF was trained with the target direction (on camera) computed also with the equivalent focal length - so the RF could somehow "learn" to go from images with aberration to reco directions without it. 

We should however fix it everywhere, probably it is more clear if done in this PR.

For consistency we should also change the stored values of src_x, src_y in the MC DL1 files. As of now they are also computed with equivalent f. They will anyway not be used in the pipeline if we go for the changes above - so that we can run dl1-to-d2 on existing DL1 files.

SeiyaNozaki:
> > Strictly speaking, `reco_disp_norm_diff` is computed as the difference between `dist` and `reco_disp_norm`.
> > https://github.com/cta-observatory/cta-lstchain/blob/8543db816fd528f5d7cc073b10dd3c0b2645b62b/lstchain/reco/dl1_to_dl2.py#L534
> > 
> > So I want just to update all disp parameters (including `src_x`, `src_y`) for MC simulation events using effective focal length for source-dependent analysis. Does it make sense?
> 
> Sorry, I don't understand what you mean. If we recalculate (src_x, src_y) with effective focal length then everything is in camera frame, and all source-dep parameters will be correct, right?

Sorry please forget it... I just wanted to say I want to recalculate not only src_x, src_y, but also all `disp` parameters (disp_norm, disp_sign etc) which is used for RF training.

lstchain/reco/dl1_to_dl2.py
```diff
@@ -794,11 +806,11 @@ def get_expected_source_pos(data, data_type, config, effective_focal_length=29.3
     # For proton MC, nominal source position is one written in config file
     if data_type == 'mc_proton':
         expected_src_pos = utils.sky_to_camera(
-            u.Quantity(data['mc_alt_tel'].values + config['mc_nominal_source_x_deg'], u.deg, copy=False),
-            u.Quantity(data['mc_az_tel'].values + config['mc_nominal_source_y_deg'], u.deg, copy=False),
+            u.Quantity(data['mc_alt_tel'].values + np.deg2rad(config['mc_nominal_source_x_deg']), u.rad, copy=False),
```
maxnoe, line 809: I am confused about this.... you add something called "nominal source_{x,y}" to something called az/alt?

This seems wrong, at least if the operation is correct, than the names.

SeiyaNozaki: This parameter is already there, it specifies the relative position (0.4 deg on x-axis by default) to compute expected source position for proton MC. 
https://github.com/cta-observatory/cta-lstchain/blob/c9e4c4f2d1d44ba1066be86865b770613d6cca1a/lstchain/data/lstchain_src_dep_config.json#L62-L63

Just I found the bug, though it didn't affect the final result much

maxnoe: > (0.4 deg on x-axis by default)

What x-axis? In which coordinate system?

SeiyaNozaki: yep, indeed it may be confusing... It assumes in AltAz coordinates (0.4 deg).. This parameter was introduced here:
https://github.com/cta-observatory/cta-lstchain/pull/247#issuecomment-605840586

moralejo: > 
> yep, indeed it may be confusing... It assumes in AltAz coordinates (0.4 deg).. This parameter was introduced here: [#247 (comment)](https://github.com/cta-observatory/cta-lstchain/pull/247#issuecomment-605840586)
            
u.Quantity(data['mc_az_tel'].values + np.deg2rad(config['mc_nominal_source_y_deg']), u.rad, copy=False),

It is not correct, camera frame's x is "along the altitude axis", but  y  is definitely not the azimuth axis.... 

Also, why do you need this? mc_nominal_source_x_deg and mc_nominal_source_y_deg are already the position on the camera (well, in degrees and without aberration) w.r.t. which you want the src-dep reco for protons.
You just have to transform it to meters

expected_src_pos_y_m = np.tan(np.deg2rad(mc_nominal_source_y_deg)) * effective_focal 
expected_src_pos_x_m = np.tan(np.deg2rad(mc_nominal_source_x_deg)) * effective_focal

SeiyaNozaki: yep, actually I don't remember why I implemented like this...
I implemented it as Abelardo suggested.

moralejo: I just resolved the conflict resulting from the azimuth range change

maxnoe: > expected_src_pos_y_m = np.tan(np.deg2rad(mc_nominal_source_y_deg)) * effective_focal_length
> expected_src_pos_x_m = np.tan(np.deg2rad(mc_nominal_source_x_deg)) * effective_focal_length

We have the ctapipe coordinate systems for this.

The source position is given in AltAz, correct? or where does this `mc_nominal_source_{x,y}_deg` come from?

Then:

```
source_pos = SkyCoord(alt=..., az=.., frame=AltAz())
pointing = SkyCoord(alt=..., az=..., frame=AltAz())
camera_frame = CameraFrame(telescope_pointing=pointing, focal_length=effective_focal_length)

source_camera = source_pos.transform_to(camera_frame)
```

moralejo: No, the "fake position" for the proton analysis is provided in "camera coordinates" x, y, but in degrees and without aberration. So this converts it to true camera coordinates.

maxnoe: Ok, then  it's:

```
source_pos = NominalFrame(fov_lon=..., fov_lat=.., frame=TelescopeFrame())
camera_frame = CameraFrame(focal_length=effective_focal_length)
source_camera = source_pos.transform_to(camera_frame)
```

moralejo: > Ok, then it's:
> 
> ```
> source_pos = SkyCoord(fov_lon=..., fov_lat=.., frame=TelescopeFrame())
> camera_frame = CameraFrame(focal_length=effective_focal_length)
> source_camera = source_pos.transform_to(camera_frame)
> ```

Yes, and I think  fov_lon = -mc_nominal_source_y_deg  and  fov_lat = mc_nominal_source_x_deg   

It does not matter much, however, because the direction of the "nominal" position chosen for protons is not critical I think.

lstchain/scripts/lstchain_add_source_dependent_parameters.py
```diff
@@ -61,11 +62,17 @@ def main():
...
-    focal_length = subarray_info.tel[tel_id].optics.equivalent_focal_length
-
-    src_dep_df = pd.concat(get_source_dependent_parameters(dl1_params, config, focal_length=focal_length), axis=1)
+    try:
+        subarray_info = SubarrayDescription.from_hdf(dl1_filename)
+        tel_id = config["allowed_tels"][0] if "allowed_tels" in config else 1
+        effective_focal_length = subarray_info.tel[tel_id].optics.effective_focal_length
+    except OSError:
+        print("subarray table is not readable because of the version inompatibility.")
+        print("Use the effective focal lentgh for the standard LST optics")
```
moralejo, line 71: typos: inompatibility, lentgh

Also, better "I will use the effective..."   (otherwise it seems it is asking the user for some change )

maxnoe: "The effective focal length for the standard LST optics will be used"

lstchain/scripts/lstchain_add_source_dependent_parameters.py
```diff
@@ -61,11 +62,17 @@ def main():
...
+        effective_focal_length = OPTICS.effective_focal_length
+        
+    src_dep_df = pd.concat(get_source_dependent_parameters(
+        dl1_params, config, effective_focal_length=effective_focal_length), axis=1)
```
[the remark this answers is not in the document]
SeiyaNozaki: `src_x` and `src_y` is updated using effective focal length here (I thought this was done before `get_expected_source_pos` function, but it is not the case... I will update the code) 
https://github.com/cta-observatory/cta-lstchain/blob/500d7b2e15596c532a67535c369e10d379d0e097/lstchain/reco/dl1_to_dl2.py#L629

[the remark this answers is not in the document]
moralejo: As I said, I think you have to update get_expected_source_pos so that it recalculates the position, not just propagates...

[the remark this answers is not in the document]
SeiyaNozaki: OK, it is true. Then I will modify codes to always recalculate the source position inside this function.
This function is only used for source-dep analysis. But, it would be better to recalculate the source position somewhere else also for the source-indep analysis addressing this comment, correct?
https://github.com/cta-observatory/cta-lstchain/pull/1131#issuecomment-1633677178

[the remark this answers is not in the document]
SeiyaNozaki: Ok, I see.

For `lstchain_add_source_dependent_parameters.py`, I'd even remove this script since source-dep parameters are also computed during dl1_to_dl2 step if src-dep parameters don't exist.
This script overwrites the existing DL1 data to add source-dependent data frame, so it doesn't make sense for the current data analysis scheme since we analyze the common DL1 data produced by OSA.

moralejo:
Looks fine to me. If I understood correctly, the source-independent analysis is also corrected, by using as target in build_models the true gamma direction in camera coordinates computed with effective focal length. Then, after applying the disp RF, the direction is converted to Alt Az using again effective focal length.

moralejo (approved):

Issue #1079, opened 2023-02-24
Defined range of az_tel for real data and simulation data


SeiyaNozaki:
I realized the defined range of az_tel is different between real data (-90 < az_tel < 270) and simulation data (0 < az_tel < 360). Is it a known feature? I think it is not good when using `az_tel` (instead of `sin_az_tel`) for random forest training.

rlopezcoto:
@SeiyaNozaki did you (or anybody) seen anything worrying back when we were using `az_tel` instead of `sin_az_tel` apart from the discontinuity in 0,360 that you reported some time ago? I think we are safe now that we are using `sin_az_tel`, but to be on the safe side we could wrap the angles to 360 deg as @maxnoe is suggesting

moralejo:
@SeiyaNozaki shall we give a try at the wrapping? Or perhaps better, we could make the calculation of the sin(az) inside the DL1 to DL2 stage (for DL1 files where it does not exist, i.e. all our standard ones now), and stop using az in the RFs.

SeiyaNozaki:
yeah, at least it would be better to wrap azimuth to avoid any confusion. It is also good to calculate sin(az) inside the DL1 to DL2 stage, but if you want to know if sin(az) is really needed, I can check.

moralejo:
Well, depends on what you mean by "really needed". I am pretty sure the difference will be small, but considering how RF works, using an azimuth with a discontinuity at 0/360 is not the best you can do. It will treat events at, say, 359.9 and 0.1 as "maximally different" in that variable, while they are in reality the same.

PR #1132, opened 2023-07-03
Normalize `az_tel` to the range of [0, 360) and compute sin_az_tel in dl1_to_dl2 step

SeiyaNozaki:
Related to #1079 

This PR normalize `az_tel` values to the range of [0, 360) degrees which is same as the distribution of MC data points. In addition, if there is no `sin_az_tel` in DL1 file because the DL1 was created using former version of lstchain, it computes `sin_az_tel` in dl1_to_dl2 step.

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -399,9 +399,9 @@ def build_models(filegammas, fileprotons,
 
     # Dealing with `sin_az_tel` missing data because of the former version of lstchain
     if 'sin_az_tel' not in df_gamma.columns:
-        df_gamma.sin_az_tel = np.sin(df_gamma.az_tel)
+        df_gamma['sin_az_tel'] = np.sin(df_gamma.az_tel)
     if 'sin_az_tel' not in df_proton.columns:
-        df_proton.sin_az_tel = np.sin(df_proton.az_tel)
+        df_proton['sin_az_tel'] = np.sin(df_proton.az_tel)
 
     # Training MC gammas in reduced viewcone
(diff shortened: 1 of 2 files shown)


rlopezcoto (approved):

lstchain/reco/dl1_to_dl2.py
```diff
@@ -14,6 +14,7 @@
 import numpy as np
 import pandas as pd
 from astropy.coordinates import SkyCoord
+from astropy.coordinates import Angle
```
chaimain, line 17: Combine these 2 imports

chaimain (approved):

SeiyaNozaki:
Hi @vuillaut
Do you know which altitude value raises an error for the conversion? I think the range of altitude is within what astropy requires, no?

maxnoe:
> and I sometimes get errors using the astropy converter on alt.

From simtel array files? Or from LST R0 files / pointing files?

I don't think we have ever observed at values that would trigger this errors.

The problem with simtelarray should be fixed, both in ctapipe:
https://github.com/cta-observatory/ctapipe/pull/2077

and astropy:
https://github.com/astropy/astropy/issues/13708

vuillaut:
From what I can remember, that was on reconstructed alt, angles > 90 deg that I ended up filtering or clipping.
But I don't have a specific example at hand.

SeiyaNozaki:
I have DL2 data with telescope zenith of 1 degree. I feel `reco_alt` was properly wrapped below 90 degrees. (maximum `reco_alt` was 89.9967188877843 degrees)
download (19)

PR #1133, opened 2023-07-04
Change way to access monitoring info in dl1 file

moralejo:
The former approach np.array(f.root[].cols.charge_mean) results in some cases  in an attempt to allocate a huge array (instead of just 2x2x1855), and hence a crash.  So far I only had this problem at PIC, using lstchain 0.9.13. No idea why...
With current main branch the same code works well at the IT cluster. 

Since the new lines work in both cases, I propose to merge the change.

### DIFF
--- a/lstchain/calib/camera/pixel_threshold_estimation.py
+++ b/lstchain/calib/camera/pixel_threshold_estimation.py
@@ -24,10 +24,10 @@ def get_bias_and_std(dl1_file):
     """
     with tables.open_file(dl1_file) as f:
         ped = f.root[dl1_params_tel_mon_ped_key]
-        ped_charge_mean = np.array(ped.cols.charge_mean)
-        ped_charge_std = np.array(ped.cols.charge_std)
+        ped_charge_mean = ped.col('charge_mean')
+        ped_charge_std = ped.col('charge_std')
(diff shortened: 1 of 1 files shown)


moralejo:
> Is this by chance the same issue as this here?
> 
> [cta-observatory/dl1-data-handler#121](https://github.com/cta-observatory/dl1-data-handler/pull/121)

Yes, seems to be the same!

moralejo:
In my case this is completely deterministic, but I only get it with a v0.9.13 installation at PIC.

maxnoe:
What is the pytables version there and how was it installed?

maxnoe:
Since it's when calling `np.array` on a pytables data structure, it might also depend on the numpy version.

Anyways, the change looks good and is what is actually documented in the pytables docs

moralejo:
> What is the pytables version there and how was it installed?

# Name                    Version                   Build  Channel
pytables                  3.8.0            py38h5dedbee_2    conda-forge

# Name                    Version                   Build  Channel
h5py                      3.9.0           nompi_py38hc4a6f91_101    conda-forge

maxnoe:
H5py should not be in the play here at all, numpy would be more interesting to know

rlopezcoto (approved):

PR #1134, opened 2023-07-06
Bump ctapipe_io_lst version

rlopezcoto:
The new `ctapipe_io_lst` version will allow to read EvB6 data

(tests will pass whenever [the new version of ctapipe_io_lst](https://github.com/cta-observatory/ctapipe_io_lst/releases/tag/v0.22.0) is uploaded to pypi

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -30,7 +30,7 @@ dependencies:
   - pandas>=2
   - pymongo
   - seaborn
-  - ctapipe_io_lst=0.22.0
+  - ctapipe_io_lst=0.22
   - pytest
   - pip:
       - ctaplot~=0.6.2
--- a/environment.yml
+++ b/environment.yml
@@ -31,7 +31,7 @@ dependencies:
   - pandas>=2
   - pymongo
   - seaborn
-  - ctapipe_io_lst=0.21.0
+  - ctapipe_io_lst=0.22.0
   - pytest
   - pip:
       - ctaplot~=0.6.2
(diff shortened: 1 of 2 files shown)


maxnoe:
PyPI is already there, conda-forge is what is missing.

Actually, could you also remove `protozfits` from the conda-env? It is not used directly by lstchain and now that we are using a conda package for ctapipe_io_lst, it will be installed from there. Having it in the environment yaml might lead to conflicts if we forget to update it here.

environment.yml
```diff
@@ -30,7 +30,7 @@ dependencies:
   - pandas>=2
   - pymongo
   - seaborn
-  - ctapipe_io_lst=0.21.1
+  - ctapipe_io_lst=0.22.0
```
maxnoe, line 33: I'd suggest to match the definition in setup.cfg. Allow any 0.22.x version.

So instead of `=0.22.0`, I'd put `= 0.22`, which allows newer versions once we release them

rlopezcoto:
> PyPI is already there, conda-forge is what is missing.

Right, sorry!

> Actually, could you also remove `protozfits` from the conda-env? It is not used directly by lstchain and now that we are using a conda package for ctapipe_io_lst, it will be installed from there. Having it in the environment yaml might lead to conflicts if we forget to update it here.

But we do not have it defined in lstchain, right?

maxnoe:
> But we do not have it defined in lstchain, right?

Ah, sorry, right, so all good

maxnoe (approved):

PR #1139, opened 2023-07-12
Add detection QE spread for outlier estimation

FrancaCassol:
Add the detection QE dispersion  over the camera (~7%) to estimate the cuts for selecting pixels outliers on the base of the calculated number of pe by the F factor method. In a first step we neglected the ~6-7% dispersion, which is ok for the standard FF intensity but not for filters delivering to very high FF intensities (> 1000 pe/pixel). This ended up with many pixels wrongly set as outliers for the last high intensity points of the filter scan.

### DIFF
--- a/lstchain/calib/camera/calibration_calculator.py
+++ b/lstchain/calib/camera/calibration_calculator.py
@@ -210,10 +210,12 @@ class LSTCalibrationCalculator(CalibrationCalculator):
         # define unusables on number of estimated pe
         npe_deviation =  calib_data.n_pe - npe_median[:,np.newaxis]
 
-        # cut on the base of pe statistical uncertanty (neglect the 7% spread due to detection QE)
+        # cut on the base of pe statistical uncertainty (adding a 7% spread due to different detection QE among PMs) 
+        tot_std = np.sqrt(npe_median + (0.07 * npe_median)**2)
+
         npe_outliers = (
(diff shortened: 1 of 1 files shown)


lstchain/calib/camera/calibration_calculator.py
```diff
@@ -210,10 +210,12 @@ def calculate_calibration_coefficients(self, event):
         # define unusables on number of estimated pe
         npe_deviation =  calib_data.n_pe - npe_median[:,np.newaxis]
 
-        # cut on the base of pe statistical uncertanty (neglect the 7% spread due to detection QE)
+        # cut on the base of pe statistical uncertainty (adding a 7% spread due to different detection QE among PMs) 
+        tot_std = np.sqrt(npe_median + (0.07 * npe_median)**2)
```
maxnoe, line 214: Should this be configurable? i.e. `camera_qe_std_percent = FloatTelescopeParameter(0.07).tag(config=True)`

Alternatively, use a module constant.

FrancaCassol: mah, I was also wondering... ok, let's do it configurable

FrancaCassol: done

maxnoe (approved):

PR #1140, opened 2023-07-14
Improve calibration "info" and "warning" messages 

FrancaCassol:
This PR add the option to read more than a sub_run file if necessary and improves the calibration out messages. In particular, add a warning in case there is not enough events statistics in the run (as it happened for run 13653)

### DIFF
--- a/lstchain/tools/lstchain_create_calibration_file.py
+++ b/lstchain/tools/lstchain_create_calibration_file.py
@@ -238,8 +238,8 @@ class CalibrationHDF5Writer(Tool):
 
         self.log.info(f"Written {self.n_calib} calibration events")
         if self.n_calib == 0:
-             self.log.critical(f"!!! No calibration events in the output file !!! : ")
-             self.log.critical(f"flatfiels collected statistics = {self.processor.flatfield.num_events_seen} events")
+             self.log.critical("!!! No calibration events in the output file !!! : ")
+             self.log.critical(f"flatfield collected statistics = {self.processor.flatfield.num_events_seen} events")
              self.log.critical(f"pedestal collected statistics = {self.processor.pedestal.num_events_seen} events")
              self.exit(1)
 
--- a/lstchain/tools/lstchain_create_calibration_file.py
+++ b/lstchain/tools/lstchain_create_calibration_file.py
@@ -103,10 +103,11 @@ class CalibrationHDF5Writer(Tool):
         self.processor = None
         self.writer = None
         self.simulation = False
+        self.n_calib = 0
 
     def setup(self):
 
-        self.log.debug("Opening file")
+        self.log.info("Opening file")
         self.eventsource = EventSource.from_config(parent=self)
 
         self.processor = CalibrationCalculator.from_name(
@@ -130,7 +131,7 @@ class CalibrationHDF5Writer(Tool):
 
         group_name = 'tel_' + str(tel_id)
 
-        self.log.debug(f"Open output file {self.output_file}")
+        self.log.info(f"Open output file {self.output_file}")
 
         self.writer = HDF5TableWriter(
             filename=self.output_file, group_name=group_name, overwrite=True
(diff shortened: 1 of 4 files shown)


lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -223,15 +224,24 @@ def start(self):
...
+                self.n_calib += 1
+
                 if self.one_event:
                     break
 
     def finish(self):
+
+        self.log.info(f"Written {self.n_calib} calibration events")
+        if self.n_calib == 0:
+             self.log.warning(f"!!! No calibration events in the output file !!! : ")
```
maxnoe, line 241: Wouldn't it make sense to exit with an error code in this case? I.e. even use `error` or `critical` and then do `self.exit(1)`?

FrancaCassol: yes, indeed. Perhaps better "critical" (strictly speaking this is not an error)

FrancaCassol: But I would not exit before the file is closed and provenance is written, better to keep track of this type of problems

maxnoe: Provenance will still be written and files closed  (if self.enter_context is used to open the file / writer), even if you call exit, because writing provenance is in a "finally" block:

https://github.com/cta-observatory/ctapipe/blob/c44226a5ee6b28ea5cc3de6d82e828aedf319b30/ctapipe/core/tool.py#L448-L450

maxnoe (approved):

lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -223,15 +224,25 @@ def start(self):
...
+                self.n_calib += 1
+
                 if self.one_event:
                     break
 
     def finish(self):
+
+        self.log.info(f"Written {self.n_calib} calibration events")
+        if self.n_calib == 0:
+             self.log.critical(f"!!! No calibration events in the output file !!! : ")
```
morcuended, line 241: remove f

lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -223,15 +224,25 @@ def start(self):
...
+             self.log.critical(f"flatfiels collected statistics = {self.processor.flatfield.num_events_seen} events")
```
morcuended, line 242: typo: flatfields

morcuended (changes requested):

lstchain/tools/lstchain_create_calibration_file.py
maxnoe: Also: CRITICAL is enough, don't need all those exclamation marks

lstchain/tools/lstchain_create_calibration_file.py
```diff
@@ -238,8 +238,8 @@ def finish(self):
 
         self.log.info(f"Written {self.n_calib} calibration events")
         if self.n_calib == 0:
-             self.log.critical(f"!!! No calibration events in the output file !!! : ")
-             self.log.critical(f"flatfiels collected statistics = {self.processor.flatfield.num_events_seen} events")
+             self.log.critical("!!! No calibration events in the output file !!! : ")
+             self.log.critical(f"flatfield collected statistics = {self.processor.flatfield.num_events_seen} events")
```
maxnoe, line 242: It doesn't really matter here, since critical messages will probably always be printed, but for `info` and `debug` messages there is a big difference:

For logging messages, you shouldn't use f-strings, but the formatting supported by the logging messages itself, since these are "lazy": if the message is not printed, the potentially expensive formatting is not done:

```python
self.log.critical("Flatfield collected statistics = %d", self.processor.flatfield.num_events_seen")
```
```

morcuended (approved):

lstchain/tools/lstchain_create_calibration_file.py
FrancaCassol: I see, goof to know. I will take this into account in the future. Since these messages are pretty short, if you agree, I prefer to go ahead with other developments.

PR #1142, opened 2023-07-19
Patchs for Cat B calibration tool

FrancaCassol:


lstchain/data/catB_camera_calibration_param.json
```diff
@@ -3,13 +3,13 @@
   "CatBCalibrationHDF5Writer": {
     "one_event": false,
     "calibration_product": "LSTCalibrationCalculator",
-    "log_level": "DEBUG"
+    "log_level": "INFO"
   },
   "LSTCalibrationCalculator": {
     "squared_excess_noise_factor": 1.222,
     "use_scaled_low_gain": true,
+    "hg_lg_ratio": 17.4,
```
morcuended, line 11: Can you say a few words about the change in this value?

lstchain/tools/lstchain_create_cat_B_calibration_file.py
```diff
@@ -123,7 +124,7 @@ def setup(self):
 
         group_name = 'tel_' + str(tel_id)
 
-        self.log.debug(f"Open output file {self.output_file}")
+        self.log.info(f"Open output file {self.output_file}")
```
morcuended, line 127: Maybe a good opportunity to use the lazy formatting in logging instead of fstrings

lstchain/data/catB_camera_calibration_param.json
FrancaCassol: this is the correct ratio high gain/low gain (which is also in the default of the trailet). I honestly do not remember why the value in the json was 1 ...

lstchain/tools/lstchain_create_cat_B_calibration_file.py
FrancaCassol: Actually, I prefer to keep a correspondence between the two calibration scripts (Cat-A and Cat-B), these messages are mainly one per job, do not imply calculations and they are in the main function, is this lazy stuff so important here ? (Also I find more readable this format ;-) )

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -241,6 +241,7 @@ def calculate_calibration_coefficients(self, event):
         if self.use_scaled_low_gain:
             calib_data.unusable_pixels[constants.LOW_GAIN] = calib_data.unusable_pixels[constants.HIGH_GAIN]
             calib_data.dc_to_pe[constants.LOW_GAIN] = calib_data.dc_to_pe[constants.HIGH_GAIN] * self.hg_lg_ratio
+            calib_data.time_correction[constants.LOW_GAIN] = calib_data.time_correction[constants.HIGH_GAIN]
```
maxnoe, line 244: Discussed offline: I asked whether we can use HG for the time correction of LG given that the two use slightly different signal paths and different readout chips, answer is yes:
comparison_timecorr

maxnoe (approved):

PR #1145, opened 2023-07-24
Write modified images from dl1ab in output DL1 files and fix `add_config_metadata`

### LINKED ISSUE #1127: discussion: write modified images after dl1ab
Hi,

I think modified images should be written (unless `--no-image` is passed of course) when running dl1ab with an image modifier. At the moment, the original images (unmodified) are written.

For two reasons:
- files should be consistent: I would expect that the parameters in a file have been computed on the images in the same file
- it would very much ease other analyses starting at DL1a level

I remember we quickly discussed this when creating the dl1ab script, but I can't remember if (and which one) there was a good/specific reason for the current behaviour?

@rlopezcoto @moralejo @maxnoe

vuillaut:
This PR allows to write the images in all cases, unless specified not to with `no-image`.

It also modifies the function `add_config_metadata` to fix corner cases (especially concerning `None`) and make it more robust to writing, and re-reading of the config.

Fixes #1127.


moralejo:
This is certainly useful - the original reason for not enabling this option from the start was that this would allow to run again dl1ab (possible with another addition of noise) over the dl1ab outputs. That is a bit dangerous. To be safe, we could write in the file a flag indicating the images are modified (and perhaps also with which parameters), and then prevent the files to be dl1ab'ed again with noise tuning (I think we should start from the original ones if another noise level is needed).

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -92,6 +93,27 @@
 )
 
 
+def image_modifier_checker(config):
```
maxnoe, line 96: "image_modifier_checker" is a noun, functions should be verbs or if boolean, sound like the condition, so maybe better
"
```
def includes_image_modification(config):
    ...

if includes_image_modification(config)

```
?

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -243,6 +265,13 @@ def main():
 
     with tables.open_file(args.input_file, mode='r') as infile:
         image_table = read_table(infile, dl1_images_lstcam_key)
+        # if the image modifier has been used to produce these images, stop here
+        config_from_image_table = json.loads(image_table.meta['config'])
+        if image_modifier_checker(config_from_image_table) and image_modifier_checker(config):
+            log.error(f"The image modifier has already been used to produce the images in file {args.input_file}."
+                        "Re-applying the image modifier is not a good practice, start again from unmodified images please.")
+            sys.exit(0)
```
maxnoe, line 273: This is an error condition. Exit with something other than 0, e.g. 1

vuillaut: thanks for spotting this one

vuillaut:
> This is certainly useful - the original reason for not enabling this option from the start was that this would allow to run again dl1ab (possible with another addition of noise) over the dl1ab outputs. That is a bit dangerous. To be safe, we could write in the file a flag indicating the images are modified (and perhaps also with which parameters), and then prevent the files to be dl1ab'ed again with noise tuning (I think we should start from the original ones if another noise level is needed).

Thank you for the feedback.
I have modified the dl1ab script to write the config parameters in the metadata as it was already done in the r0_to_dl1 workflow.
Then we can simply read the config to know if the images have been produced using the `image_modifier`, and in this case not allow re-modification.

vuillaut:
Note: I had to modify the `add_config_metadata` function as well that was unnecessarily complicated and not allowing simple reading of the dumped config. It now uses the more robust `json.dumps` function while keeping the original intended functionality of dumping some unsupported types such as `LazyConfigValue` or `PosixPath`. I have added a unit test.
Since this has nothing to do with the original PR, I can make a separate one if you prefer (and if there is something to discuss about these changes).

moralejo:
> [...] Since this has nothing to do with the original PR, I can make a separate one if you prefer (and if there is something to discuss about these changes).

May be just change the PR title adding a reference to the modifications about the config reading?

moralejo (dismissed):

maxnoe:
One test failing:

````
FAILED lstchain/scripts/tests/test_lstchain_scripts.py::test_dl1ab_on_modified_images - ValueError: Running lstchain_dl1ab failed with return code 1, output: 
 image_modifier configuration: {'increase_psf': True, 'increase_nsb': True}
Modified images are saved in the output file.
  File "/home/runner/micromamba-root/envs/lst-dev/bin/lstchain_dl1ab", line 8, in <module>
    sys.exit(main())
  File "/home/runner/work/cta-lstchain/cta-lstchain/lstchain/scripts/lstchain_dl1ab.py", line 177, in main
    extra_noise_in_dim_pixels = imconfig["extra_noise_in_dim_pixels"]
KeyError: 'extra_noise_in_dim_pixels'
```

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -430,6 +431,29 @@ def test_dl1ab_no_images(simulated_dl1_file, tmp_path):
             assert (new_parameters['length'] != old_parameters['length']).any()
 
 
+def test_dl1ab_on_modified_images(simulated_dl1ab, tmp_path):
+    config_path = tmp_path / 'config_image_modifier.json'
+    output_file = tmp_path / 'dl1ab_on_modified_images.h5'
+    reprocess_output_file = tmp_path / 'dl1ab_on_modified_images_reprocess.h5'
+    config_path = Path(__file__).parent.joinpath("../../data/lstchain_dl1ab_tune_MC_to_Crab_config.json")
```
morcuended, line 438: Maybe better 
```python
from importlib.resources import files
config_path = files("lstchain.data") / "lstchain_dl1ab_tune_MC_to_Crab_config.json"
```

morcuended:
Failing test is unrelated to changes in this PR. I produced again the run summary file with test data and got run types: DATA, PEDCALIB and DATA (it was expecting DATA, ERROR, DATA types). Did the R0 files in the test data folders change?

maxnoe:
> Failing test is unrelated to changes in this PR. I produced again the run summary file with test data and got run types: DATA, PEDCALIB and DATA (it was expecting DATA, ERROR, DATA types for runs 2005, 2006 and 2008, respectively). Did the R0 files in the test data folders change for run 2006?


Yes, I added more  files for run 2006 for the tests in #1147 , so that's why it now recognizes 2006 (correctly) as a PEDCALIB run

lstchain/scripts/tests/test_lstchain_scripts.py
vuillaut: thank you this is indeed better, I made the change

vuillaut:
Tests are finally passing !

lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -430,6 +431,29 @@ def test_dl1ab_no_images(simulated_dl1_file, tmp_path):
...
+    output_file = tmp_path / 'dl1ab_on_modified_images.h5'
+    reprocess_output_file = tmp_path / 'dl1ab_on_modified_images_reprocess.h5'
+    config_path = files("lstchain.data").joinpath("lstchain_dl1ab_tune_MC_to_Crab_config.json")
+
+    run_program(
+        'lstchain_dl1ab',
+        '-f', simulated_dl1ab,
+        '-o', output_file,
+        '-c', config_path,
+    )
```
morcuended, line 445: is there something here you can check in the output file?

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -92,6 +93,27 @@
 )
 
 
+def includes_image_modification(config):
```
morcuended, line 96: maybe this function can be moved to `image.modifier` (?)

lstchain/scripts/tests/test_lstchain_scripts.py
vuillaut: well, the presence of the config but it is indirectly tested after as it's read correctly and the second run fails because of it

lstchain/scripts/lstchain_dl1ab.py
vuillaut: it's more a config-related function than an image modifier one, no?
would it  be better in `io.config`?

morcuended: Sure, agreed

vuillaut: done

lstchain/io/tests/test_config.py
```diff
@@ -1,6 +1,7 @@
+import tempfile
 from lstchain.io import config
 from lstchain.io.config import get_cleaning_parameters
-import tempfile
+from lstchain.scripts.lstchain_dl1ab import includes_image_modification
```
morcuended, line 4: shouldn't be from `lstchain.io.config`. How does this even work?

maxnoe: Why shouldn't it?

morcuended: I was expecting that you could not import a function defined somewhere else. But, true, it is not the case

morcuended (dismissed):

lstchain/io/tests/test_config.py
vuillaut: @morcuended thank you Daniel, you are right, it should correctly be imported from config, I fixed this.
It worked only because the function was imported by `lstchain_dl1ab` in the first place.

morcuended (approved):

morcuended:
my approval keeps being dismissed automatically

vuillaut:
> my approval keeps being dismissed automatically

Hu? Why is that? Were the tests not over?

morcuended (approved):

morcuended:
> > my approval keeps being dismissed automatically
> 
> Hu? Why is that? Were the tests not over?

I approved again after test completion and also happened the same

morcuended (approved):

Issue #1150, opened 2023-08-28
Problems with some subruns from 2023-08-19 in muon analysis


morcuended:
There were some subruns from 2023-08-19 runs: 14034.0021 (around the middle of the run) and other 6 subruns of run 14029 (25, 26, 27, 29, 30, 31 towards the end of the run) that failed the R0 to DL1 step. Some problems (related to the rates) were reported in the ELOG during those runs.

Example of R0 to DL1 command to reproduce the error:
```
lstchain_data_r0_to_dl1 --input-file=/fefs/aswg/data/real/R0/20230819/LST-1.1.Run14034.0021.fits.fz --output-dir=/fefs/aswg/data/real/running_analysis/20230819/v0.10 --pedestal-file=/fefs/aswg/data/real/monitoring/PixelCalibration/Cat-A/drs4_baseline/20230819/pr
o/drs4_pedestal.Run14024.0000.h5 --calibration-file=/fefs/aswg/data/real/monitoring/PixelCalibration/Cat-A/calibration/20230819/v0.10.2/calibration_filters_52.Run14031.0000.h5 --time-calibration-file=/fefs/aswg/data/real/monitoring/PixelCalibration/Cat-A/drs4_time_samplin
g_from_FF/20220518/v0.10.0/time_calibration.Run08349.0000.h5 --systematic-correction-file=/fefs/aswg/data/real/monitoring/PixelCalibration/Cat-A/ffactor_systematics/20230410/v0.10.3/calibration_scan_fit_20230410.0000.h5 --config=/fefs/aswg/lstosa/config_files_lstchain/lst
chain_standard_v0.10_heuristic_ff.json --pointing-file=/fefs/onsite/monitoring/driveLST1/DrivePositioning/DrivePosition_log_20230819.txt --run-summary-path=/fefs/aswg/data/real/monitoring/RunSummary/RunSummary_20230819.ecsv
```

Here is the error traceback:

```python
...
Traceback (most recent call last):
  File "/fefs/aswg/software/conda/envs/osa-v0.10/lib/python3.11/site-packages/lstchain/scripts/lstchain_data_r0_to_dl1.py", line 227, in main
    r0_to_dl1.r0_to_dl1(
  File "/fefs/aswg/software/conda/envs/osa-v0.10/lib/python3.11/site-packages/lstchain/reco/r0_to_dl1.py", line 701, in r0_to_dl1
    analyze_muon_event(subarray,
[...]

PR #1151, opened 2023-08-29
Fix unusable pixels usage in muon analysis

FrancaCassol:
solves #1150

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -376,7 +376,6 @@ def r0_to_dl1(
     )
 
     if not is_simu:
-        calibration_mon = source.r0_r1_calibrator.mon_data.tel[telescope_id].calibration
 
         # Pulse extractor for muon ring analysis. Same parameters (window_width and _shift) as the one for showers, but
         # using GlobalPeakWindowSum, since the signal for the rings is expected to be very isochronous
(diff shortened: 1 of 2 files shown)


lstchain/reco/r0_to_dl1.py
```diff
@@ -661,7 +661,8 @@ def r0_to_dl1(
 
                 # Muon ring analysis, for real data only (MC is done starting from DL1 files)
                 if not is_simu:
-                    bad_pixels = event.mon.tel[telescope_id].calibration.unusable_pixels[0]
+                    bad_pixels = source.r0_r1_calibrator.mon_data.tel[telescope_id].calibration.unusable_pixels[0]
```
maxnoe, line 664: Maybe add a comment that we want to use the fixed pixel list from the loaded calibration file instead of the new calibration maybe filled into `event.mon`? 

Otherwise looks good.

maxnoe, line 664: Code would be much shorter and reable if you store this once into a variable and then use it:

```
calibration_mon = source.r0_r1_calibrator.mon_data.tel[telescope_id].calibration
```

maxnoe (approved):

morcuended (approved):

morcuended:
offtopic: why a pre-commit action is running in this PR?

maxnoe (approved):

PR #1157, opened 2023-09-07
Transform timestamp to ns before converting it to int in summary script

morcuended:
Right now `run_start` timestamp is transformed to int and then from sec to nsec, hence losing subsecond information. This changes the order of the operations, first transforming it to nsec and then to int


maxnoe:
Just to note that currently, the input data only seems to contain whole second timestamp, so while this is more correct, it won't directly change anything

morcuended:
closing this since it won't make a difference

PR #1165, opened 2023-09-18
add default compression to write_dataframe function to compress dl2

### LINKED ISSUE #1163: compress dl2 data
It appears DL2 data is not compressed.

see
https://github.com/cta-observatory/cta-lstchain/blob/1280e47950726f92200e1624ca38c672760e9d77/lstchain/io/io.py#L771

I can have a look to improve this.

vuillaut:
fixes #1163


moralejo:
Well spotted... indeed DL2 are really bulkier than the corresponding DL1b, funny we have not noticed.
Do you know (or can test) how much the default compression level impacts reading speed?

vuillaut:
> Well spotted... indeed DL2 are really bulkier than the corresponding DL1b, funny we have not noticed. Do you know (or can test) how much the default compression level impacts reading speed?

Hi @moralejo 

I just did the test and here are the results.
From these, I would actually advocate for a default compression level = 1, the impact on file size and reading time being marginal, but the writing time increases gradually.
What do you think?

| Compression level | File size (MB) | Write time (s)      | Read time (s)       |
|-------------------|----------------|---------------------|---------------------|
| 0                 | 1107.80        | 14.58               | 6.27                |
| 1                 | 564.23         | 37.44               | 14.23               |
| 2                 | 561.22         | 43.88               | 15.54               |
| 3                 | 550.47         | 42.75               | 13.66               |
| 4                 | 550.42         | 46.70               | 14.65               |
| 5                 | 548.46         | 64.54               | 16.43               |
| 6                 | 539.72         | 77.75               | 11.71               |
| 7                 | 537.05         | 90.09               | 11.43               |
| 8                 | 534.16         | 139.93              | 17.78               |
| 9                 | 533.99         | 207.13              | 16.41               |

vuillaut:
> @vuillaut Did you test that locally or on the cluster? Because I'd suspect that write and read speed actually would go up when writing much less data to the slowish network file system on the cluster.

Indeed, I tested locally on my laptop, let me run this on the cluster.

morcuended:
+1 to setting the compression level to 1 by default

vuillaut:
I have set default `complevel=1` and redone the test with `bloc:zstd` as complib.

| Compression level | File size (MB) | Write time (s) | Read time (s) |
|------------------|----------------|----------------|---------------|
| 0                | 406.70         | 4.27           | 1.38          |
| 1                | 268.55         | 4.27           | 1.90          |
| 2                | 263.42         | 6.13           | 1.70          |
| 3                | 262.20         | 7.56           | 1.73          |
| 4                | 260.82         | 9.55           | 1.70          |
| 5                | 260.27         | 12.03          | 1.67          |
| 6                | 260.08         | 25.49          | 1.81          |
| 7                | 259.62         | 38.49          | 1.72          |
| 8                | 257.56         | 51.92          | 1.71          |
| 9                | 257.47         | 90.88          | 1.77          |

(I don't think the read time are too relevant, they probably fluctuate too much with cluster usage)

I think this is ready for review.

lstchain/io/io.py
```diff
@@ -750,7 +750,8 @@ def write_dataframe(dataframe, outfile, table_path, mode="a", index=False, confi
     meta: global metadata
     """
 
-    filters = tables.Filters(complevel=complevel, complib=complib)
+    filters = HDF5_ZSTD_FILTERS.copy()
```
morcuended, line 753: `HDF5_ZSTD_FILTERS` still sets level 5 by default. Should the value be updated there as well?

morcuended (changes requested):
thanks. I left some comments

lstchain/io/io.py
vuillaut: That was probably a good choice for other data, no?
Wasn't this tested at the time?

lstchain/io/io.py
```diff
@@ -750,7 +750,8 @@ def write_dataframe(dataframe, outfile, table_path, mode="a", index=False, confi
     meta: global metadata
```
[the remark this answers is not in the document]
vuillaut: Done

lstchain/io/io.py
morcuended: you're right. Then since this complevel is a new argument, it has to be updated in all its usages. Otherwise, it will be always level 1 by default

vuillaut: Could you clarify?

morcuended: Sorry, my bad. `write_dataframe` is only used to write DL2 files as far as I can tell from the usage of the function. `complevel` defaults to 1 in this function. It reads the `HDF5_ZSTD_FILTERS` where `complevel` is defined as 5.
None of the usages of `write_dataframe` set `complevel` differently than 1. So IMO, leaving the 5 by default there, where right now is not being used at all, is going to be confusing.

morcuended (approved):
looks good to me

morcuended (approved):

vuillaut:
Thanks Daniel!

moralejo:
Apologies for missed review and thanks for this @vuillaut !

Issue #1168, opened 2023-09-20
Fetch run type information from TCU database instead of guessing from the data files


morcuended:
https://github.com/cta-observatory/cta-lstchain/blob/e72c0d41ec5f9bf3d97d4c91c6d63d1e62b2ce61/lstchain/scripts/lstchain_create_run_summary.py#L132C2-L132C2

This mostly affects onsite data processing, as it is essential to what runs are to be analysed.

Currently, we guess the run type based on the % of trigger types in the first 500 events of each run.

It should be good to look for this information directly in the TCU mongo database.

Pros:
 - Some special runs (tests, calibrations, etc.) are properly tagged in the TCU differently than those calibrations runs that should be used for data processing. With the current scheme, we cannot distinguish test/filter-wheel scan runs (tagged as PEDCALIB) from those actual PEDCALIB runs for data processing.
 
Cons:
 - This information is only available starting from around mid-2021.
 - There are some run numbers duplicated in the database, so the information is missing for some of them. TCU team are aware, I do not know if they already fixed it.
 - Sometimes, if something goes wrong during data taking and TCU does not write the corresponding entry point in the database, the information is not available. While it could be guessed with the current heuristic way.

A robust run summary script should probably combine both approaches.

See also https://github.com/cta-observatory/lstosa/issues/191

morcuended:
I propose to first check the run type in TCU, if it is not available (which happens now and then), guess it as it is done now as a fallback.

The most straightforward change would be to simply change it in `lstchain_create_run_summary`. However, this information is really not used anywhere in lstchain, but only in lstosa. So perhaps at some point, we could detach the use of the summary for timestamp calculation and for handling the onsite processing.

PR #1173, opened 2023-10-11
Get run type from TCU if it's available

marialainez:
Issue #1168

### DIFF
--- a/lstchain/scripts/lstchain_create_run_summary.py
+++ b/lstchain/scripts/lstchain_create_run_summary.py
@@ -63,6 +63,14 @@ parser.add_argument(
     default=False,
 )
 
+parser.add_argument(
+    "-t",
+    "--tcu-server",
+    type=str,
+    help="TCU database server",
+    default="lst101-int",
+)
+
 dtypes = {
     "ucts_timestamp": np.int64,
     "run_start": np.int64,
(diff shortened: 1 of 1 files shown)


morcuended (commented):
Thanks, @marialainez! This works nicely in the LST IT centre where DB is available. But we have to make tests happy when DB access is not possible. Either catching `ServerSelectionTimeoutError` in a try block or checking first if DB is accessible. something like:

```
def is_db_available(tcu_db_server):
    """Check the connection to the TCU database."""
    tcu_client = MongoClient(tcu_db_server, serverSelectionTimeoutMS=3000)
    try:
        tcu_client.server_info()
    except ConnectionFailure:
        log.warning("Database is not available.")
        return False
    else:
        log.debug("Database is available.")
        return True
```

lstchain/scripts/lstchain_create_run_summary.py
```diff
@@ -296,10 +371,17 @@ def main():
 
     reference_counters = [read_counters(date_path, run) for run in run_numbers]
 
-    run_types = [
-        type_of_run(date_path, run, counters)
-        for run, counters in zip(run_numbers, reference_counters)
-    ]
+    if is_db_available("lst101-int"):
```
morcuended, line 374: I would probably add tcu_server (set by default as `'lst101-int'` ) as input argument to make it more flexible and change it accordingly across the script. Otherwise it looks good to me

morcuended (approved):
thanks @marialainez, looks good to me

PR #1174, opened 2023-10-13
Add possibility to copy and not merge some keys  by lstchain_merge_hdf5_files 

FrancaCassol:
This is in particular necessary for the calibration keys in DL1 and DL2 files, which are the same for all subruns and should not be merged but simply copied from the first file of the list.
I wonder if we can already add them as default in the script...


vuillaut:
Hi
Should these keys be added to `merging_checks` to make sure they are indeed the same in all files? (open question)

lstchain/scripts/lstchain_merge_hdf5_files.py
```diff
@@ -72,6 +72,11 @@
     help='Skip checks when merging files'
 )
 
+parser.add_argument(
+    '--keys-to-copy',
+    nargs="+", default=[''],
+    help='List of duplicated keys to be copied and not to be merged'
```
maxnoe, line 78: I don't think this should be a command line argument. We know which keys are event-wise and should be merged and which are global so shouldn't be merged.

Let's define this in the code.

lstchain/scripts/lstchain_merge_hdf5_files.py
```diff
@@ -72,6 +72,11 @@
     help='Skip checks when merging files'
 )
 
+parser.add_argument(
+    '--keys-to-copy',
+    nargs="+", default=[''],
```
maxnoe, line 77: This should be `nargs="*"` (allow it to be empty) and the default is then an empty list.

The default should just be an empty list, not a list with one element that is an empty string.

lstchain/scripts/lstchain_merge_hdf5_files.py
FrancaCassol: Indeed, this would be the best, but since this seems a general merging function I was not sure we wanted to specify it inside the function. If we agree on that I prefer it also.

maxnoe: It's a general merging function for our data, we do know all possible keys.

lstchain/scripts/lstchain_merge_hdf5_files.py
```diff
@@ -22,6 +22,22 @@
...
+from lstchain.io.io import (
+    dl1_params_tel_mon_ped_key,
+    dl1_params_tel_mon_cal_key,
+    dl1_params_tel_mon_flat_key,
+    dl1_mon_tel_CatB_cal_key,
+    dl1_mon_tel_CatB_ped_key,
+    dl1_mon_tel_CatB_flat_key
+)
+
+default_keys_to_copy = [dl1_params_tel_mon_ped_key,
```
morcuended, line 34: The test is currently failing because the merging function expects `CatB` keys that are not present in the test h5 files. The function should check first whether they exit

lstchain/scripts/lstchain_merge_hdf5_files.py
morcuended: I'm reopening this conversation because I think it is not resolved yet. Having this command line argument makes it a bit confusing at the same time as having a set of default keys to be copied. Do we really need this additional option? I think they can be simply defined in lstchain/io/io.py

morcuended (changes requested):

lstchain/io/io.py
```diff
@@ -332,15 +337,33 @@ def auto_merge_h5files(
     else:
         keys = set(nodes_keys)
 
+    copy_keys = {}
```
maxnoe, line 340: `{}` creates an empty dict, you want a set: `copy_keys = set()`

lstchain/io/io.py
```diff
@@ -332,15 +337,33 @@ def auto_merge_h5files(
...
+    if keys_to_copy:
+        keys_to_copy = {k for k in keys_to_copy if k in keys}
```
maxnoe, line 342: That's `keys_to_copy.intersection(keys)`

lstchain/io/io.py
maxnoe, line 340: As commented below, I think the default should be the keys we know should be copied instead of merged (monitoring and instrument for starters)

FrancaCassol: I added it in the tool

FrancaCassol:
Hi @maxnoe , @morcuended , I added the default keys, but I prefer to keep the trailet in case of tests on new implementations. Do not hesitate to add changes if you have strong opinions on this. We need to merge this PR before next release, which should be done soon.

rlopezcoto:
@maxnoe , @morcuended  are you ok with the changes so we can move on with this PR as well?

Issue #1160, opened 2023-09-14
How to deal with the correction of EDISP calculation


morcuended:
We have to think about how to deal with the correction of EDISP calculation for those analyses using v0.9 (and even v0.10) that are in an advanced state. Fix was introduced in pyirf v0.10.0 (https://github.com/cta-observatory/pyirf/pull/250).

lstchain v0.9 depends on pyirf v0.6
lstchain v0.10 depends on pyirf v0.8

One possibility is to include the correction directly in lstchain and make bugfix releases for lstchain v0.9 y v0.10. For lstchain v0.11, directly use pyirf v0.10 with the fix included, and drop the hack included in lstchain.

moralejo:
Other opinions?

rlopezcoto:
for the second case, I would make lstchain compatible with pyirf v0.10 and make a bugfix release (lstchain v0.10.5)
for DL1 files produced with lstchain v0.9, we need to decide if it is worth it to introduce that bugfix

morcuended:
> The solution of analyzing everything with lstchain >= v0.10.5 will be clearner, otherwise I am predicting a mess for which only very expert analyzers will know which version to use

If we can require pyirf v0.10 in lstchain >= v0.10.5 I think this is the best option. Then advocate using lstchain >= v0.10.5 for the ongoing projects.

maxnoe:
> If we can require pyirf v0.10 in lstchain >= v0.10.5 I think this is the best option. Then advocate using lstchain >= v0.10.5 for the ongoing projects.

This would apply updates to the IRF interpolation code, which would be very good to have, but the pyirf API has changed here with the introduction of the more advanced interpolation algorithms.

@RuneDominik can comment what might be needed. I think we can make this API compatible at the lstchain level, so it would be possible.

For lstchain 0.9, I'd indeed propose to post-process the pyirf computed EDISP in lstchain to fix them, I'd be very glad If I don't have to do a whole bug release for an old version of pyirf if the fix here is two lines of code.

RuneDominik:
[...]

PR #1184, opened 2023-11-27
Update interpolation code to pyirf v0.10 API

morcuended:
See #1160

I still need to add some optional arguments to the estimators. definitions.

### DIFF
--- a/lstchain/high_level/interpolate.py
+++ b/lstchain/high_level/interpolate.py
@@ -14,10 +14,10 @@ from pyirf.io.gadf import (
 )
 from pyirf.interpolation import (
     GridDataInterpolator,
-    interpolate_effective_area_per_energy_and_fov,
-    interpolate_energy_dispersion,
-    interpolate_psf_table,
-    interpolate_rad_max,
+    EnergyDispersionEstimator,
+    EffectiveAreaEstimator,
+    PSFTableEstimator,
+    RadMaxEstimator,
 )
 from scipy.spatial import Delaunay, distance, QhullError
 
@@ -361,8 +361,8 @@ def interpolate_gh_cuts(
     ----------
     .. [1] https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html
     """
-    interp = GridDataInterpolator(grid_points=grid_points, params=gh_cuts)
-    gh_cuts_interp = interp(target_point, method=method)
+    interp = GridDataInterpolator(grid_points=grid_points, params=gh_cuts, method=method)
+    gh_cuts_interp = interp(target_point)
 
     return gh_cuts_interp
 
@@ -402,8 +402,8 @@ def interpolate_al_cuts(
     ----------
     .. [1] https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html
     """
-    interp = GridDataInterpolator(grid_points=grid_points, params=al_cuts)
-    al_cuts_interp = interp(target_point, method=method)
+    interp = GridDataInterpolator(grid_points=grid_points, params=al_cuts, method=method)
+    al_cuts_interp = interp(target_point)
 
     return al_cuts_interp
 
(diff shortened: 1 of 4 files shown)


lstchain/high_level/interpolate.py
```diff
@@ -361,8 +362,10 @@ def interpolate_gh_cuts(
     ----------
     .. [1] https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html
     """
-    interp = GridDataInterpolator(grid_points=grid_points, params=gh_cuts)
-    gh_cuts_interp = interp(target_point, method=method)
+    interp = GridDataInterpolator(
+        grid_points=grid_points, params=gh_cuts, method=method
+    )
+    gh_cuts_interp = interp(target_point)
```
RuneDominik, line 367: If extrapolation is needed I'd suggest to create a Estimator object (same for the AL_CUTS), just like it is done for RAD_MAX tables in pyirf. If you don't want extrapolation this should be fine.

morcuended:
@RuneDominik, ready for review. Please have a look at the changes.

Although you suggested leaving all the optional arguments by default, I have included the interpolation method in the effective area and RADMAX estimators (via `interpolator_kwargs`) to keep this possibility as it was before.

RuneDominik:
LGTM, if you want this to work just as the old version. One could argue that this would be a good opportunity to support the whole API, including extrapolation.

morcuended:
> LGTM, if you want this to work just as the old version. One could argue that this would be a good opportunity to support the whole API, including extrapolation and support for different inter-/extrapolator classes.

Many thanks for the quick review. I agree with you. However, this change has been pending for quite some time and I'd say it's somewhat urgent to make a new minor release (with the interpolation code at least working as before). I suggest merging this PR as it is. Make the release. And extend the support for extrapolation in another PR. What do you think, @rlopezcoto?

rlopezcoto:
thanks @RuneDominik for the review and the comment. I support @morcuended's point of view, since this change is urgently needed to release a bugfix, I would just stick to solve this issue.

morcuended:
I tested these changes (using pyirf v0.10) on a Crab sample and compared them with results from lstchain v0.10.4 (pyirf v0.8 + applying the edisp fix script). Both use interpolation. The resulting SED is not the same. 

pyirf v0.8 + applying the edisp fix script

flux_points_LP_8bin_gh_eff_0 7_th_cont_0 7_pyirf08

pyirf v0.10 (this PR)

flux_points_LP_8bin_gh_eff_0 7_th_cont_0 7_pyirf010

@RuneDominik was there any change in the interpolation code of pyirf (from v0.8 to v0.10) that can explain this change in the spectrum?

RuneDominik:
The was the edisp fix which was also applied to the interpolation code. As far as I see when going though the `QuantileInterpolator` history, this was the only one actually changing the code. I assume the IRFs had the applicable edisp corrections, so that

- **pyirf v0.8 + applying the edisp fix script** used uncorrected IRFs (thus with edisp summed to 1) as input and the output was corrected to an integral of 1
- **pyirf v0.10 (this PR)** used correct IRFs as input and no re-normalization was applied to the output?

morcuended:
> pyirf v0.8 + applying the edisp fix script used uncorrected IRFs (thus with edisp summed to 1) as input and the output was corrected to an integral of 1

Correct. I applied the script correcting the edisp normalization to the DL3 files produced with pyirf v0.8

> pyirf v0.10 (this PR) used correct IRFs as input and no re-normalization was applied to the output?

No re-normalization was applied to the output in this case.

RuneDominik:
I tested a bit and found that for data with a logarithmic binning (hence edisps) the results have changed when switching between pyirf 0.9 (pre normalization fix) and pyirf 0.10 (post normalization fix), as you can see here in the residuals between truth and estimated (and re-normalized if needed) Gaussian histograms. 

Residuals

For linear binning I found no difference. Thinking of it this seems plausible, as the bin-width quite drastically changes for logarithmic binning and is applied for normalization in the pyirf 0.10 version. As of now I would not call each of them better or worse, just different.

morcuended:
Thanks a lot for the check @RuneDominik. In view of this, I think we can proceed with the merging.

rlopezcoto:
thanks a lot for the fix and the checks @morcuended and @RuneDominik! if you approve this, I think we can move on with this PR

PR #1190, opened 2023-12-03
New script and notebook for the selection of good quality runs

moralejo:
This will eventually replace the data_selection notebook for find good quality runs


lstchain/scripts/lstchain_cherenkov_transparency.py
```diff
@@ -0,0 +1,680 @@
...
+def expfunc(x, a, b):
+    return a * np.exp(b * x)
+
+
+def time_to_yyyymmdd(t):
+    hour = np.array([datetime.fromtimestamp(x).hour for x in t])
+    t2 = np.array([x if h > 12 else x - 43200 for x, h in zip(t, hour)])
+    dt = [datetime.fromtimestamp(x) for x in t2]
+
+    return np.array([(x.year * 10000 + x.month * 100 + x.day) for x in dt])
```
morcuended, line 676: I think you don't need this function, you already know the date in `YYYYMMDD` format from the beginning `date = int(output_file[-11:-3])`

lstchain/scripts/lstchain_cherenkov_transparency.py
```diff
@@ -0,0 +1,680 @@
...
+#
+# The night's datacheck summary file is the output_file, because it will be
+# updated by this script
+#
+parser.add_argument('-u', '--update_datacheck_file', dest='output_file',
+                    type=str, default='',
+                    help='DL1_datacheck_YYYYMMDD.h5 file')
+
+parser.add_argument('-d', '--input_dir', dest='input_dir',
+                    type=str, default='v0.10/tailcut84/datacheck',
```
morcuended, line 38: This input directory argument could be named `prod_id` or `prod_version` instead. As an input directory, I would rather give the whole path.

lstchain/scripts/lstchain_cherenkov_transparency.py
```diff
@@ -0,0 +1,680 @@
...
+                    help='relative path of run-wise datacheck directory under '
+                         'DL1/YYYYMMDD')
+
```
morcuended, line 41: I would separate the `update_datacheck_file` option in two, one is the date in YYYYMMDD, and the other is the output directory where you want to store the recalculated file (right now stored in the cwd).

lstchain/scripts/lstchain_cherenkov_transparency.py
moralejo: With the latest changes we do need it - and I prefer not to depend on a file name for that.

lstchain/scripts/lstchain_cherenkov_transparency.py
moralejo: I really want to update the file, just add a table to it, not have a different file.

notebooks/data_quality.ipynb
```diff
@@ -0,0 +1,1833 @@
+{
```
morcuended, line 1: <p>Maybe also comment here that the DL1 datacheck files should have been "reprocess" using the cherenkov transparency script</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

morcuended, line 1: <pre class="ql-syntax" spellcheck="false">DL1_datacheck_2023*
Also, at least for me, the ? symbol does not work as a wildcard for globbing as if it were the shell
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/1190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

moralejo: I plan to copy the (updated) older files to the standard location, and then we will have the cherenov transparency script as part of the onsite automatic processing, so I think it is not necessary.

moralejo: Doesn't it? (and if you just try in a python session?)
As for the specific path in the comments in the notebook, there is an error of mine (asked for 2022 directories and then for 2023 files). Will fix.

morcuended: > Doesn't it? (and if you just try in a python session?) It works for me.

Sorry, yes, it works in both notebooks and python interactive sessions. My bad

morcuended (approved):
Looks fine to me. I tested it for several sources and it works fine.

rlopezcoto (approved):
Already reviewed by @morcuended , just approving after a small comment implementation

PR #1198, opened 2023-12-19
Script for gain selection (R0 to R0G), for data taken in ADHF format …

moralejo:
(i.e. for data taken from July 2023).
Still being tested.


morcuended:
Will this gain selector also work with EVB6 data?

moralejo:
> Will this gain selector also work with EVB6 data?
It should, because the format is the same (ADHF). Will try it.

moralejo:
> > Will this gain selector also work with EVB6 data?
> 
> It should, because the format is the same (ADHF). Will try it.

No, it does not work...  `'File' object has no attribute 'CameraConfig'`

maxnoe:
What is "ADHF" and where does this acronym come from?

moralejo:
> What is "ADHF" and where does this acronym come from?

I am afraid I made it up, sorry. The original wording (in the ELOGs) is "ADH-ZFW", butI am not sure either whether that is official or not.

maxnoe:
Ah ok. There are three major combinations of software versions I think.

They can be distinguished by looking at the `EXTNAME` and the `PBFHEAD` header keywords in the corresponding FITS HDUs.

* EVBv5 + "old" ZFW  (ZFitsWriter compiled from svn `CamerasToACTL` repository).
    
    This is the oldest combination and it uses the `ProtoR1` protocol buffer definitions with the `PBFHEAD` keyword set to `R1.Event` etc.
    
```
In [6]: f = fits.open("/fefs/aswg/data/real/R0/20201120/LST-1.1.Run02962.0000.fits.fz")

In [7]: [hdu.name for hdu in f]
Out[7]: ['PRIMARY', 'CameraConfig', 'Events']

In [8]: f['CameraConfig'].header["PBFHEAD"]
Out[8]: 'R1.CameraConfiguration'

In [9]: f['Events'].header["PBFHEAD"]
Out[9]: 'R1.CameraEvent'
```
 
* EVBv5 + ADH-ZFW ("new" ZFitsWriter compiled from git `array-data-handler` repository).
    
    This combination was used since mid of last year. It uses the same data model, but with a slightly different `PBFHEAD` keyword: `ProtoR1.Event` etc.

```
In [10]: f = fits.open("/fefs/aswg/data/real/R0/20230811/LST-1.1.Run13880.0000.fits.fz")

In [11]: [hdu.name for hdu in f]
Out[11]: ['PRIMARY', 'CameraConfig', 'Events']

In [12]: f['CameraConfig'].header["PBFHEAD"]
Out[12]: 'ProtoR1.CameraConfiguration'

In [13]: f['Events'].header["PBFHEAD"]
Out[13]: 'ProtoR1.CameraEvent'
```

* EVBv6 + ADH-ZFW
    
    This is using the new official CTAO R1v1 format. The extname of the camera configuration changed, the `TelescopeDataStream` is new etc.:
    
```
In [14]: f = fits.open("/fefs/aswg/data/real/R0/20231215/LST-1.1.Run16120.0000.fits.fz")

In [15]: [hdu.name for hdu in f]
Out[15]: ['PRIMARY', 'DataStream', 'CameraConfiguration', 'Events']

In [16]: f['DataStream'].header["PBFHEAD"]
Out[16]: 'R1v1.TelescopeDataStream'

In [17]: f['CameraConfiguration'].header["PBFHEAD"]
Out[17]: 'CTAR1.CameraConfiguration'

In [18]: f['Events'].header["PBFHEAD"]
Out[18]: 'CTAR1.Event'
```


I notice the PBFHEAD is inconsistent... will report to dirk and julien

maxnoe:
The best solution is to treat data with `R1.Event` and `ProtoR1.Event` identically to the existing tool and add new code for dealing with the `EVBv6` data.

moralejo:
> The best solution is to treat data with `R1.Event` and `ProtoR1.Event` identically to the existing tool and add new code for dealing with the `EVBv6` data.

I don't understand - what I do now is to check empirically whether I find "CameraConfig" or "CameraConfiguration" in the file - the latter seems to be the case for EvB6 data. Isn't that enough to distinguish the two cases we find since mid-2023?
There is no need for the program to work for EvB5 data, which will were already gain-selected with your C++ code.

maxnoe:
> There is no need for the program to work for EvB5 data, which will were already gain-selected with your C++ code.

What identifies the new data format most clearly is the content of `PBFHEAD`, if it starts with `R1v1` or `CTAR1`, it's the new official data format written by EVBv6.

That the extname of the `CameraConfiguration` HDU changed is more accidental...

maxnoe:
> There is no need for the program to work for EvB5 data, which will were already gain-selected with your C++ code.

I thought there is, as the C++ code also doesn't work with case 2: EVBv5 plus ADH-ZFW. Isn't that the case?

maxnoe:
However, as said above, that would be relatively easy to fix, so maybe the easiest is to fix the C++ code for the `EVBv5 + ADH-ZFW` case and let the code here only deal with EVBv6 aka CTAR1 aka R1v1?

moralejo:
> However, as said above, that would be relatively easy to fix, so maybe the easiest is to fix the C++ code for the `EVBv5 + ADH-ZFW` case and let the code here only deal with EVBv6 aka CTAR1 aka R1v1?

This script already works with both, using the "CameraConfig" vs. "CameraConfiguration" approach to identify what type of input data is being read in.

moralejo:
About this warning:
`WARNING: Unexpected extra padding at the end of the file.  This padding may not be preserved when saving changes. [astropy.io.fits.header]`

Is there any way of getting rid of it?

maxnoe:
Is this happening for data where no event was written maybe? protozfits doesn't properly create / close the files if no event was ever written.

moralejo:
> Is this happening for data where no event was written maybe? protozfits doesn't properly create / close the files if no event was ever written.

No, there were events written out. But this only happened with the EvB v5 file I tested (not for EvB6 data, which are the ones for which we will use this script)

moralejo:
@maxnoe  can we approve and merge this? If we get it in the next release we can start gain-selecting the data as part of the onsite automatic processing (hence reduce the disk-filling rate)

maxnoe:
@moralejo We can remove the EVBv5 parts here, right? We will handle all the EVBv5 data with the old, c++ based tool, correct?

It also would be good to test this script (all lines are untested as of now). I can create a small sample with the zfitsrecompressor. Do you have an suitable run number for this?

environment.yml
```diff
@@ -18,6 +18,7 @@ dependencies:
   - joblib~=1.2.0
   - toml
   - protobuf=3.20
+  - protozfits=2.2
```
morcuended, line 21: add also to `setup.py`

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,212 @@
...
+                if ff_pix.sum() / event.r0.tel[1].waveform.shape[1] > \
+                        MIN_FLATFIELD_PIXEL_FRACTION:
+                    # Looks like a FF event:
+                    event_type[-1] = EventType.FLATFIELD
+                elif event_type[-1] == EventType.FLATFIELD:
+                    # If originally tagged as FF, but not looking like FF:
+                    event_type[-1] = EventType.UNKNOWN
+
+                event_id.append(event.index.event_id)
+            print('Finished first loop over input files - all ok!')
```
morcuended, line 200: log.info

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,212 @@
...
+
+        except Exception as err:
+            print(err)
```
morcuended, line 203: log.error

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,212 @@
...
+
+    event_type = []
+    event_id = []
+    with EventSource(input_url=input_file,
+                     config=Config(standard_config['source_config'])) as source:
+        source.pointing_information = False
+        source.trigger_information = True
+        source.log.setLevel(logging.WARNING)
+        try:
+            for ievent, event in enumerate(source):
```
morcuended, line 173: ievent not used, remove enumerate

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env python
```
morcuended, line 1: I'd add here a comment explaining which version of data this is expected to be used for (EvB6, as of date, etc)

morcuended (commented):
Apart from having a small test sample, I suggest testing it over an entire night.

moralejo:
> @moralejo We can remove the EVBv5 parts here, right? We will handle all the EVBv5 data with the old, c++ based tool, correct?
> 

Well, apparently it works, and I think it does not harm to have it available in python form, even if we are not now using it for those data.

> It also would be good to test this script (all lines are untested as of now). I can create a small sample with the zfitsrecompressor. Do you have an suitable run number for this?

I used Run13631.0042 (20230708), but any run should do.

moralejo:
@maxnoe @morcuended apologies... FYI: I just realized I had some changes not yet pushed (which were actually necessary to make the script work for EvB6). I will try to check your comments to the previous version and see if they still apply.

moralejo:
> @maxnoe @morcuended apologies... FYI: I just realized I had some changes not yet pushed (which were actually necessary to make the script work for EvB6). I will try to check your comments to the previous version and see if they still apply.

And I may have made something weird when merging the changes in main... should they appear explicitly (as they do) in this PR?

moralejo:
> Apart from having a small test sample, I suggest testing it over an entire night.

I agree, but I prefer to make a first release and try it on a full night.

morcuended:
> > @maxnoe @morcuended apologies... FYI: I just realized I had some changes not yet pushed (which were actually necessary to make the script work for EvB6). I will try to check your comments to the previous version and see if they still apply.
> 
> And I may have made something weird when merging the changes in main... should they appear explicitly (as they do) in this PR?

I think they were merged instead of rebasing your branch with the main. I think it's not a problem anyway

maxnoe:
They shouldn't appear like this in either case

moralejo:
> They shouldn't appear like this in either case

May have to do with me committing the change in December, but not pushing until today. And between the two I merged the changes in main.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,223 @@
...
+                # Get the event type, from the first loop over the files:
+                evtype = event_type_val[event_id==event.event_id][0]
+
+                if ((evtype == EventType.SUBARRAY.value) |
+                    (evtype == EventType.UNKNOWN.value)):
+                    # Find pixels with HG above gain switch threshold:
+                    wf = protozfits.any_array_to_numpy(event.waveform)
+                    num_gains = int(wf.size / num_pixels / num_samples)
+
+                    if num_gains != 2:
```
maxnoe, line 122: What about applying dvr to data that already was gain selected by EVB?

moralejo: > What about applying dvr to data that already was gain selected by EVB?

In principle I wanted to separate DVR from gain selection (unlike in the C++ tool).

maxnoe:
> I used Run13631.0042 (20230708), but any run should do.

That's an EVBv5 run. What would be a suitable EVBv6 run?

moralejo:
> > I used Run13631.0042 (20230708), but any run should do.
> 
> That's an EVBv5 run. What would be a suitable EVBv6 run?

Ah, sorry, right!  I was testing on 20230808 Run13830.0000

moralejo:
I tested on Run16102.0010 from 20231214

maxnoe:
I added now 

```
├── R0
│   └── 20231218
│       ├── LST-1.1.Run16231.0000_first50.fits.fz
│       ├── LST-1.2.Run16231.0000_first50.fits.fz
│       ├── LST-1.3.Run16231.0000_first50.fits.fz
│       └── LST-1.4.Run16231.0000_first50.fits.fz
```

to the test data

maxnoe:
Concerning the EVBv6 data, I think there were some where the flatfield events were not tagged correctly due to a misconfiguration and we need to use the flatfield heuristic to tag those.

We could fix the event type for those here?

morcuended:
> Concerning the EVBv6 data, I think there were some where the flatfield events were not tagged correctly due to a misconfiguration and we need to use the flatfield heuristic to tag those.
> 
> We could fix the event type for those here?

That was the night of 2023-11-10

moralejo:
> Concerning the EVBv6 data, I think there were some where the flatfield events were not tagged correctly due to a misconfiguration and we need to use the flatfield heuristic to tag those.
> 
> We could fix the event type for those here?

I followed the approach "touch as little as possible" that we have followed so far. This would be relatively easy (and anyway, if the heuristic was wrong the gain selection would also be wrong), but it would be confusing to have gain-selected data with fixed trigger tags, and others (all those we reduced in the past, with the C++ tool) in which the original trigger tag is kept.

maxnoe:
Event type doesn't exist in evbv5 data

moralejo:
> Event type doesn't exist in evbv5 data

Ah ok, that would make it more clear (I think back then we discussed modifying the trigger type). 

If we go for that, how would we do it? I guess we do not want to overwrite event_type when it is correct. Should it be a command-line option and then apply it on runs we positively know have issues with the tagging?

moralejo:
> I added now
> 
> ```
> ├── R0
> │   └── 20231218
> │       ├── LST-1.1.Run16231.0000_first50.fits.fz
> │       ├── LST-1.2.Run16231.0000_first50.fits.fz
> │       ├── LST-1.3.Run16231.0000_first50.fits.fz
> │       └── LST-1.4.Run16231.0000_first50.fits.fz
> ```
> 
> to the test data

Thanks, I will do the test, but we need also the Drive file and the RunSummary file for that night, can you please include them, in the same directory?

maxnoe:
> Thanks, I will do the test, but we need also the Drive file and the RunSummary file for that night, can you please include them, in the same directory?

No problem in principle. On the other hand I wonder why they are needed at all..

moralejo:
> > Thanks, I will do the test, but we need also the Drive file and the RunSummary file for that night, can you please include them, in the same directory?
> 
> No problem in principle. On the other hand I wonder why they are needed at all..

I use the LST event source to go over the events and get the right event types (e.g. to be safe against "ucts jumps" - I don't really know if these can ever re-appear). So the Run summary is needed (to get the ref values for computing the dragon_time), and the Drive is not but, I think the event source requires it anyway.

maxnoe:
The jumps part of the code is not used for evbv6 data, so no need for the run summary and you can pass `pointing_information=False` to the source to not load and interpolate pointing.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -76,15 +69,13 @@ def main():
...
     try:
-        num_pixels = input_streams[0].CameraConfig[0].num_pixels
-        num_samples = input_streams[0].CameraConfig[0].num_samples
-    except:
-        is_evb6_data = True
         num_pixels = input_streams[0].CameraConfiguration[0].num_pixels
         num_samples = input_streams[0].CameraConfiguration[
             0].num_samples_nominal
+    except:
+        logging.error('CameraConfiguration not found! Is this CTAR1 data?')
```
maxnoe, line 77: CTAR1 is an old name that will disappear soon, we should use `R1v1` (official CTAO / new EVBv6 protobuf message) vs. `ProtoR1` (old EVBv5 protobuf messages)

moralejo: I think the data you uploaded are the 1-gain test data... (20231218) tests are failing because of that...

maxnoe: My bad, sorry, will upload files from another date... Do you know one by heart that would work?

moralejo: 20231214/LST-1.1.Run16102.0010.fits.fz

moralejo: Thanks a lot!

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+                if event.r0.tel[1].waveform is None:
+                    logging.error('The data seem to contain no R0 waveforms. '
+                                  'Is this already gain-selected data?')
+                    exit(1)
+
+                # Check if this may be a FF event
+                # Subtract baseline offset (convert to signed integer to
+                # allow for negative fluctuations!)
+                wf_hg = event.r0.tel[1].waveform[0][:,
+                        SAMPLE_START:SAMPLE_END].astype('int16') - OFFSET
```
maxnoe, line 176: Subtracting the offset is wrong for EVBv6 data. It will be done automatically by the source since the offset is now stored in the file.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+                # pixel-wise integral:
+                wf_hg_sum = np.sum(wf_hg, axis=1)
+                ff_pix = ((wf_hg_sum > MIN_FLATFIELD_ADC) &
```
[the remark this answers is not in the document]
moralejo: The idea was to keep the same selection we had in the previous (C++) tool. Also, my understanding (perhaps wrong) is that the FF identification in the event source was designed to work on calibrated data - here we apply no calibration.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+    event_type = []
+    event_id = []
+    with EventSource(input_url=input_file,
+                     config=Config(standard_config['source_config'])) as source:
+        source.pointing_information = False
+        source.trigger_information = True
+        source.log.setLevel(logging.WARNING)
+        try:
+            for event in source:
+                if event.r0.tel[1].waveform is None:
```
maxnoe, line 167: EVBv6 data will have always the waveforms in `r1`.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+parser.add_argument('-o', '--output-dir', dest='output_dir',
+                    type=str, default='./',
+                    help='Output directory')
+
+# Range of waveform to be checked (for gain selection & heuristic FF
+# identification)
+SAMPLE_START = 3
+SAMPLE_END = 39
+# Baseline offset:
+OFFSET = 400
```
maxnoe, line 39: In EVBv6 data, the offset is not fixed. It is stored in the file in the `TelescopeDataStream` structure and the `EventSource` automatically applies the conversion of offset / scale when storing the data into `event.r1.tel[...].waveform`.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+                    n_tiles=n_tiles,
+                    rows_per_tile=rows_per_tile,
+                    compression_block_size_kb=64*1024,
+                    defaul_compression="lst"))
+            stream.open(name)
+
+            stream.move_to_new_table("CameraConfiguration")
+            stream.write_message(input_streams[i].CameraConfiguration[0])
+            stream.move_to_new_table("DataStream")
+            stream.write_message(input_streams[i].DataStream[0])
```
maxnoe, line 98: DataStream only exists for EVBv6, yes in other files there are assumptions only valid for EVBv5 data, I am confused about what this tool supports now.

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: > Subtracting the offset is wrong for EVBv6 data. It will be done automatically by the source since the offset is now stored in the file.

Is that also the case if all DRS4 corrections are disabled? Can't we read with the event source the waveforms as they are stored?

moralejo: > > Subtracting the offset is wrong for EVBv6 data. It will be done automatically by the source since the offset is now stored in the file.
> 
> Is that also the case if all DRS4 corrections are disabled? Can't we read with the event source the waveforms as they are stored?

I confirm I had tried to use the FF heuristic identification in the event source, but it only works if the calibration is switched on. And of course, we do not have it at the point at which we want to make the gain selection.

maxnoe: > Is that also the case if all DRS4 corrections are disabled? 

Yes, it's the first step when reading the waveforms. This is completely independent of DRS4 corrections.

The scaled waveform is put into `r1`, if no pe calibration has been performed, the unscaled waveform is also stored, in r0.

But I don't understand why you would want to reimplement this here instead of just using what we already have.

This is now a manual translation of the C++ code back to python which was a manual translation of the code in ctapipe_io_lst to C++,

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: Now it support only EvB6 (R1v1) data.

moralejo: The assumptions you mean are the baseline etc, right? I will look into that.

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: ?? I just tried, and when no drs4 calib is applied the r0 is filled by the source (also r1), so I think it works.  But ok, the baseline might eventually be different, so I will ~~either use r1 or~~ subtract the correct one (should be equivalent, right?). On second thought, using r1 with no calibration is dangerous because it might occasionally take the low gain, and the condition is on adc counts above baseline.

maxnoe: > I just tried, and when no drs4 calib is applied the r0 is filled by the source (also r1), so I think it works. 

If no pe calibration is performed, the original waveform is also filled into r0. 

I didn't implement storing the unscaled waveform in case of pe calibration alrady applied, since that is relatively meaningless... it's not original adc counts but scaled pe.

moralejo: So no issue here, right?

lstchain/scripts/lstchain_r0_to_r0g.py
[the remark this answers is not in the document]
moralejo: Will have a look, thanks!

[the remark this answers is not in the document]
maxnoe: That's a docstring, it's not actually checked. The check is done before outside of that method. 

The "must be done" comes from our previous observations that spikes influence the gain selection and lead to false positives (waveform with spike over threshold, after not).

Doesn't make a big difference though. You can just call the method without having drs4 corrections applied I think.

The main point is: we shouldn't duplicate that code. If the one in ctapipe_io_lst doesn't work here for some reason, we should fix it and not have to very similar versions of the same thing.

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: The baseline is now read from the data.

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: Baseline now read from TelescopeDataStream

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env python
+"""
+Script to apply gain selection to the raw fits.fz produced by EvB6 in CTAR1
+format (aka R1v1).
+
+Gain selection is only applied to shower events (not to interleaved pedestal
+and flatfield)
+
+It uses heuristic identification of the interleaved flatfield (FF) events (
+given the occasional problems we have with the FF event tagging)
```
maxnoe, line 10: Should this be an option? Do we really want to force the detection heuristic, that e.g. doesn't work I think in moonlight conditions, given that it is supposed to be working now?

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+    output_dir = args.output_dir
+
+    # First identify properly interleaved pedestals (also in case there are
+    # ucts jumps) and FF events (heuristically):
+    event_id, event_type = get_event_types(input_file)
+
+    event_id = np.array(event_id)
+    event_type_val = np.array([x.value for x in event_type])
+
+    logging.info('Identified event types and number of events:')
```
maxnoe, line 55: Create a logger 

```
log = logging.getLogger("lstchain_r0_to_r0g")
```

and then use `log.info`. Like this, you are using the root logger.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+    for j in np.unique(event_type_val):
```
maxnoe, line 56: `np.unique` can directly give you the counts:

```
types, counts = np.unique(event_type_val, return_counts=True)
```

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+        logging.info(f'{EventType(j)}:, {np.sum(event_type_val == j)}\n')
```
maxnoe, line 57: logging should use %-formatting with extra arguments:

```
log.info("Counts for event type %s: %d", event_type, count)
```

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # Now loop over the files (4 streams) again to perform the actual gain
+    # selection:
+    k = input_file.find('/LST-1.')
```
maxnoe, line 61: We have functions for dealing with LST raw data file names in `lstchain.paths`

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+    output_stream_names = [output_dir + name[k:] for name in input_stream_names]
+
+    input_streams = []
+    for name in input_stream_names:
+        input_streams.append(protozfits.File(name, pure_protobuf=True))
+
+    try:
+        num_pixels = input_streams[0].CameraConfiguration[0].num_pixels
+        num_samples = input_streams[0].CameraConfiguration[
+            0].num_samples_nominal
```
maxnoe, line 73: This linebreak makes it very weird to read. Also the try/except now spans multiple lines where the error will always happen on line one, maybe better:

```
try:
    camera_config = input_streams[0].CameraConfiguration[0]
except (AttributeError, IndexError):
    log.error("CameraConfiguration not found! Is this R1v1 data?')
    sys.exit(1)

num_pixels = camera_config.num_pixels
num_samples = camera_config.num_samples_nominal
```

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+    except:
```
maxnoe, line 74: Bare except (see above). Should at least be `except Exception` to not catch `KeyboardInterrupt` and `SystemExit`.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+        logging.error('CameraConfiguration not found! Is this R1v1 data?')
+        exit(1)
+
+    with ExitStack() as stack:
+        for i, name in enumerate(output_stream_names):
+
+            n_tiles = input_streams[i].Events.protobuf_i_fits.header[
```
maxnoe, line 81: This formatted a bit awkwardly due to black, the linebreak is at a weird position.
I also think that `protobuf.File` directly exposes the header as parsed by astropy:

```
    header = input_streams[i].Events.header
    n_tiles = header["NAXIS2"]  # is that true?
    rows_per_tile = header["ZTILELEN"]
    
```

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+                    n_tiles=n_tiles,
+                    rows_per_tile=rows_per_tile,
+                    compression_block_size_kb=64*1024,
+                    defaul_compression="lst"))
+            stream.open(name)
+
+            stream.move_to_new_table("CameraConfiguration")
+            stream.write_message(input_streams[i].CameraConfiguration[0])
+            stream.move_to_new_table("DataStream")
+            stream.write_message(input_streams[i].DataStream[0])
```
maxnoe, line 96: Doesn't the datastream come first?

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+
+            stream.move_to_new_table("Events")
+
+            for event in input_streams[i].Events:
+                # skip corrupted events:
+                if event.event_id == 0:
+                    continue
+                # Get the event type, from the first loop over the files:
+                evtype = event_type_val[event_id==event.event_id][0]
```
maxnoe, line 105: This is very inefficient to for each event build a new mask and check the full array. Can use a dict or a table with an index to lookup by the event id instead of doing a full search for each event.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+
+                if ((evtype == EventType.SUBARRAY.value) |
```
maxnoe, line 107: `evtype in {EventType.SUBARRAY.value, EventType.UNKNOWN.value}`

Maybe define "EVENT_TYPES_TO_REDUCE` above?

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+                    (evtype == EventType.UNKNOWN.value)):
+                    # Find pixels with HG above gain switch threshold:
+                    wf = protozfits.any_array_to_numpy(event.waveform)
+                    num_gains = int(wf.size / num_pixels / num_samples)
```
maxnoe, line 111: event has `num_channels`, no need to compute.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+
+                    if num_gains != 2:
+                        logging.error('You are attempting gain selection on '
+                                      'data with only one gain!')
+                        exit(1)
+
+                    wf = wf.reshape((num_gains, num_pixels, num_samples))
+                    hg = wf[0]
+                    lg = wf[1]
+                    use_lg = (np.max(hg[:, SAMPLE_START:SAMPLE_END], axis=1)
```
maxnoe, line 121: Did you test and find that it is faster to first compute the maximum and then check if it is larger than a threshold?

I'd assume that the same logic as in the C++ tool would be more performant:

```
np.any(hg[:, SAMPLE_START:SAMPLE_END] > THRESHOLD, axis=1)
```
```

File: lstchain/scripts/lstchain_r0_to_r0g.py [hunk 26]
```diff
@@ -0,0 +1,206 @@
...
+                                          pixel_status & 0b0111)
+                    event.pixel_status.data = new_status.tobytes()
+
+                stream.write_message(event)
+
+            stream.close()
+            input_streams[i].close()
+
+    logging.info('R0 to R0G conversion finished successfully!')
+    return(0)
```
maxnoe, line 140: Don't return anything. 

(also return is a statement, not a function)

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,206 @@
...
+                    type=str, default='./',
+                    help='Output directory')
+
+# Range of waveform to be checked (for gain selection & heuristic FF
+# identification)
+SAMPLE_START = 3
+SAMPLE_END = 39
+
+def main():
+    args = parser.parse_args()
```
maxnoe, line 40: I think we should add a logging config that at least logs the timestamps by default.

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: I think it works also for moon conditions (the calibration charge is pretty high). But indeed, we should optionally trust the event_type, or perhaps even do a check of how many of the tagged FF fulfill the heuristics or vice versa. Will look into it.

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: As to:   "is that true?"
This was an empirical guess... how can we make sure?

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: yes, thanks - I did not know it was relevant.

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: Can you clarify the thing about the timestamps?

morcuended: that the logging format should contain the timestamp of each message. Default format is: `BASIC_FORMAT = "%(levelname)s:%(name)s:%(message)s"`, then it should include also `%(asctime)s` (timestamp)

lstchain/scripts/lstchain_r0_to_r0g.py
moralejo: Now the heuristic identification can be disabled.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,235 @@
...
+    formatter = logging.Formatter('%(asctime)s - '
+                                  '%(levelname)s - %(message)s',
+                                  '%Y-%m-%d %H:%M:%S')
+    handler = logging.FileHandler(log_file, mode='w')
+    handler.setFormatter(formatter)
+    logging.getLogger().addHandler(handler)
+
+    # Loop over the files (4 streams) to perform the gain selection:
+    
+    input_stream_names = [Path(Path(input_file).parent,
```
maxnoe, line 81: Since each stream is now independent, do we want to have an internal loop here? Why not just call the script for each input file?

moralejo: I see no obvious advantage, and I think it is cleaner to keep the whole subrun as the input here.
Imagine we eventually want to e.g. check the rate of interleaveds, or something else which needs access to go over all the events.

lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -0,0 +1,234 @@
...
+  
+    output_stream_names = [Path(output_dir, Path(inputsn).name) 
+                           for inputsn in input_stream_names]
+
+    input_streams = []
+    for name in input_stream_names:
+        input_streams.append(protozfits.File(str(name), pure_protobuf=True))
+
+    try:
+      camera_config = input_streams[0].CameraConfiguration[0]
```
maxnoe, line 94: Is this a two-space indent?

environment.yml
maxnoe, line 21: In `setup.py` you define `~=2.2` (Which means any 2.x that is greater equal 2.2), and here you define `=2.2`, which means "the latest 2.2.x".

I.e. installing via pip will currently give you 2.4, but setting up the env will give you 2.2

morcuended: yes, I misread the comment. 

> What is better here? Should I be afraid of any existing or future 2.x being incompatible with anything I do with protozfits? I just checked I have been testing with 2.3.0

Are the 2.x minor releases and not supposed to break anything? In any case maybe safer to limit <2.5?

maxnoe: So the equivalent for what is in `setup.py` would be `>=2.2,<3`

maxnoe (approved):

PR #1200, opened 2024-01-22
Likelihood reconstruction update

gabemery:
This pull request covers 4 small changes:

1. Adds the extraction of the pedestal variance used in the likelihood function from interleaved events. This was already covered in https://github.com/cta-observatory/cta-lstchain/pull/1124 but not merged at the time. This is more stable than the current, event-wise extraction and allows to account for pixel differences. It is also needed for data after DVR.
2. Clean up a test and log message from when the code needed to be pre-compiled
3. Add support of source dependent analysis with the likelihood reconstruction
4. Remove the 'weight' argument. It was tested during development but is not used anymore.

All of these changes were already tested extensively in a modified version of lstchain 0.9.8

I also checked that it run successfully up to the creation of DL2 files in the current version on small v0.10 (no DVR) datasets (without in depth check of the file content). The updated config file fails during training due to #1206


lstchain/reco/dl1_to_dl2.py
```diff
@@ -787,14 +785,21 @@ def calc_source_dependent_parameters(data, expected_src_pos_x_m, expected_src_po
...
         expected_src_pos_y_m,
-        data['x'],
-        data['y'],
-        data['psi'])
+        data[x],
+        data[y],
+        data[psi])
 
     src_dep_params['time_gradient_from_source'] = data['time_gradient'] * np.sign(disp) * -1
     src_dep_params['skewness_from_source'] = data['skewness'] * np.sign(disp) * -1
```
SeiyaNozaki, line 805: `time_gradient` and `skewness` (and the sign of those parameters) are computed based on the standard hillas parameter (`x`, `y`, `psi`), then `lhfit_time_gradient` and `lhfit_length_assymmetry` are computed based on lhfit parameters (`lhfit_x`, `lhfit_y`, `lhfit_psi`), right? If so, this sign correction should be done using the base parameters.

gabemery: Right, I will change this

SeiyaNozaki: Now `dist` and `alpha` are computed using LHfit parameters, but the sign of `time_gradient` is not corrected. Is this on purpose?

lstchain/data/lstchain_lhfit_config.json
```diff
@@ -141,39 +141,69 @@
...
     "lhfit_time_gradient",
+    "log_intensity",
+    "width",
+    "length",
+    "x",
+    "y",
+    "wl",
+    "signed_skewness",
+    "kurtosis",
+    "signed_time_gradient",
```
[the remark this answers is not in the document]
gabemery: About your second point : I removed parameters with low contribution/ redundant information when training using the likelihood reconstruction parameters (nearly no effects on results).

Now for the first point, this new example json uses the features I used in the likelihood reconstruction paper. Late in the paper I realised that the classification required the standard analysis parameters to be competitive. So I added all parameters from the LST performance paper, including this signed_time_gradient (which is not the one obtained in the source dependent analysis code).

lstchain/reco/dl1_to_dl2.py
gabemery: The sign of the `skewness` and `time gradient` are corrected line 795-796 of the current version. But this is now in non-edited code as it uses a different disp from Hillas x and y, while `dist` and `alpha` use the one obtained with the lhfit quantities.

lstchain/data/lstchain_lhfit_config.json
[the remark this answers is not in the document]
SeiyaNozaki: okay thanks for the explanation!

SeiyaNozaki (dismissed):
Thanks for the answers @gabemery 
I approved the change of codes related to source-dependent analysis

chaimain (approved):

PR #1201, opened 2024-01-22
Limit scipy to version before 1.12 which introduces a breaking change

gabemery:
The breaking change I encountered is fixed in gammapy main branch but not in a release yet.

Error during lstchain tests with the latest scipy : 

 ```
../../../micromamba-root/envs/lst-dev/lib/python3.10/site-packages/gammapy/utils/roots.py:9: in <module>
    BAD_RES = RootResults(root=np.nan, iterations=0, function_calls=0, flag=0)
E   TypeError: RootResults.__init__() missing 1 required positional argument: 'method'
```
'method' is introduced in scipy 1.12.0

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,6 +19,7 @@ dependencies:
   - toml
   - protobuf=3.20
   - pyparsing
+  - scipy~=1.11.4
   - scikit-learn=1.2
   - sphinx=4
   - sphinx-automodapi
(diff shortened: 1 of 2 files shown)


morcuended (dismissed):

rlopezcoto (dismissed):
Thanks @gabemery ! I opened #1202 to remember that we can follow later scipy releases whenever we do it for gammapy

rlopezcoto (approved):

PR #1208, opened 2024-01-25
Set minimum intensity by default to 50 p.e. in example DL3 and IRF configs

morcuended:
Also, I still find it confusing that these config files are not in lstchain/data, together with the rest of lstchain config files.
Shouldn't they be placed in the same location?

Moreover, these example config files are not included in the package data in `setup.py`:

```
    package_data={
        'lstchain': [
            'data/*',
            'resources/*',
        ],
    },
```

### DIFF
--- a/docs/examples/dl3_tool_config.json
+++ b/docs/examples/dl3_tool_config.json
@@ -1,7 +1,7 @@
 {
   "EventSelector": {
     "filters": {
-      "intensity": [0, Infinity],
+      "intensity": [50, Infinity],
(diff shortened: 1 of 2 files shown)


chaimain:
I believe we had paused attempts to make the software easily usable for non-expert analyzers, and so we continued with the analysis practices that we initiated as is. Not sure if it is too late to make dedicated changes before the start of the upcoming software school...

rlopezcoto (approved):

PR #1209, opened 2024-01-29
Improve non-finite parameters rejection when training models

### LINKED ISSUE #1206: Training failure when using features computed after utils.filter_events
In lstchain.reco.dl1_to_dl2  build_models() there is a set of quantities which are computed after the call to utils.filter_events.
Namely : signed_time_gradient, signed_skewness, sometimes sin_az_tel, and the reconstructed energy and disp.

It means that the training can fail (in particular for the classifier) if any of these quantities are Nan since the filtering code is called before their creation.

How did I encounter the error : The time_gradient often has more Nan than Hillas parameters (when the image is 2 pixels).
Trying to train a models without the time_gradient but with the signed_time_gradient in the classifier, these values are not removed.

Possible fix : re-run the filterbetween the creation of new features and trainings.

gabemery:
Update of the build_models method to ensure more effectively that non-finite values in features are removed.

Main change : additional call to utils.filter_events before the classifier training to account for the new 'signed_time_gradient'. Other new parameters that may be  affected include the 'signed skewness', reconstructed energy and reconstructed disp.

I also moved the 'sin_az_tel' computation before the first call to utils.filter_events. I have never seen a case when 'az_tel' is not finite but this would handle it.

Dedicated tests could be added in a LHfit related PR since this is where a failure case was observed.

closes #1206

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -349,7 +349,7 @@ def build_models(filegammas, fileprotons,
     df_proton = pd.read_hdf(fileprotons, key=dl1_params_lstcam_key)
 
     # Update parameters related to target direction on camera frame for gamma MC
-    # taking into account of the abrration effect using effective focal length
+    # taking into account of the aberration effect using effective focal length
     try:
         subarray_info = SubarrayDescription.from_hdf(filegammas)
         tel_id = config["allowed_tels"][0] if "allowed_tels" in config else 1
@@ -359,7 +359,7 @@ def build_models(filegammas, fileprotons,
         logger.warning("The effective focal length for the standard LST optics will be used.")
         effective_focal_length = OPTICS.effective_focal_length
 
-    df_gamma = update_disp_with_effective_focal_length(df_gamma, effective_focal_length = effective_focal_length)
+    df_gamma = update_disp_with_effective_focal_length(df_gamma, effective_focal_length=effective_focal_length)
 
     if config['source_dependent']:
         # if source-dependent parameters are already in dl1 data, just read those data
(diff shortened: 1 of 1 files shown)


SeiyaNozaki:
Hi @gabemery The code looks fine. Maybe it is possible to remove `particle_classification_features` for the first event filtering since it will be applied later?

gabemery:
@SeiyaNozaki I would keep it there. The idea being that the best would be to filter all problematic events from the start, if possible (i.e. if events have bad features for particle classification it is better to also remove them from the energy and direction training). Then we filter again for the features which are produced later just in case.

SeiyaNozaki:
Okay, thanks for the explanation! Then. it looks already fine!

SeiyaNozaki (approved):

lstchain/reco/dl1_to_dl2.py
```diff
@@ -402,6 +402,16 @@ def build_models(filegammas, fileprotons,
         lhfit_df_proton = pd.read_hdf(fileprotons, key=dl1_likelihood_params_lstcam_key)
         df_proton = pd.concat([df_proton, lhfit_df_proton], axis=1)
 
+    # Normalize all azimuth angles to the range [0, 360) degrees
```
vuillaut, line 405: This transformation should not be applied when training model, but in the DL1 parameters construction.
The risk is to have different parameters at training and inference.

vuillaut (changes requested):

lstchain/reco/dl1_to_dl2.py
vuillaut: Alternatively, one could have a `preprocess` step, including filtering and this type of transformation (including also lines 409-413), that should be applied at training and inference.

gabemery: Here I simply moved this block of code, but I agree it would be better in an earlier step.
Changing it now may lead to breaking changes. Should it be opened as an issue for the next major release?

vuillaut: Indeed I did not realise you just moved it.
Yes please open an issue, I change my review.

vuillaut (approved):

PR #1212, opened 2024-01-30
Script to compute NSB-tuning parameter for MC at the waveforms level

moralejo:


rlopezcoto (commented):
@gabemery I think this was discussed with you, are you ok with the script?

lstchain/scripts/lstchain_tune_nsb_waveform.py
```diff
@@ -0,0 +1,107 @@
...
+    config = read_configuration_file(args.config)
+
+    nsb_correction_ratio, data_ped_variance, mc_ped_variance = \
+        calculate_required_additional_nsb(args.input_mc, args.input_data,
+                                          config=Config(config))
+
+    dict_nsb = {
+        "nsb_tuning": True,
+        "nsb_tuning_ratio": np.round(nsb_correction_ratio, decimals=2),
+        "spe_location": "lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat"
```
gabemery, line 91: The spe_location needs to be filled by an actual path to the file. Since you update the config file with the script it would be good to fill this path with the full path to the installed lstchain.

A line like in io.config.py (line 43) may work:
"spe_location": Path(__file__).parent.joinpath("../data/SinglePhE_ResponseInPhE_expo2Gaus.dat")

You may need to convert to string

gabemery (commented):
Looks good except a small possible improvement

lstchain/scripts/lstchain_tune_nsb_waveform.py
moralejo: But if you run this in a different system, or with a different installation, the resulting path may be unknown. This should be put somewhere in the R0 to DL1 stage, not here.

gabemery (approved):
The answer by @moralejo makes sense.
So I approve the script and will later update things to not need to provide the path to the spe file.

PR #1219, opened 2024-02-07
Update notebooks pointlike IRF and DL3 creation examples

morcuended:
For the DL2 to IRF notebook, I've created a separate updated notebook focusing on the point-like IRF. For the moment I leave the old notebook that was also explaining the full-enclosure and background IRF creation (which needs to be updated as well, but there are no new DL2 MC files for using them at the moment).


rlopezcoto:
I guess this is necessary for your demonstration, right?

morcuended:
> I guess this is necessary for your demonstration, right?

Yes, I will use them

rlopezcoto (approved):

rlopezcoto:
@moralejo can you update the notebooks for the school?

Issue #1159, opened 2023-09-12
Weird default energy binning


LukasNickel:
The default energy bins here:
https://github.com/cta-observatory/cta-lstchain/blob/main/lstchain/io/event_selection.py#L346

do not match what I would expect:
- Its not 5 (or any n) bins per decade. That would require 6*5+1 = 31 bins instead of 25
- 500 TeV seems excessive. Wouldn't 50 TeV be more than enough?

```src python
In [1]: np.geomspace(0.005, 500, 25)
Out[1]: 
array([5.00000000e-03, 8.07799049e-03, 1.30507861e-02, 2.10848252e-02,
       3.40646035e-02, 5.50347086e-02, 8.89139705e-02, 1.43649242e-01,
       2.32079442e-01, 3.74947105e-01, 6.05763829e-01, 9.78670891e-01,
       1.58113883e+00, 2.55448489e+00, 4.12702093e+00, 6.66760716e+00,
       1.07721735e+01, 1.74035029e+01, 2.81170663e+01, 4.54258788e+01,
       7.33899634e+01, 1.18568685e+02, 1.91559342e+02, 3.09482909e+02,
       5.00000000e+02])
```

Is there any deeper meaning behind these values or was there an error in #1036 ?

LukasNickel:
Sorry, I got confused looking at the code.
The missing bin edge is added when creating the axes:
https://github.com/cta-observatory/cta-lstchain/blob/a8058123c2ab44e7d19dbc44e8de83a1158e562b/lstchain/io/event_selection.py#L450

In that case only the question about the upper edge of 500 TeV remains

maxnoe:
> 50 TeV would be too low, right? CTA requirements are defined up to 200+ TeV - a different issue is how often we will have observations which go beyond a few tens of TeV, but no need to limit that from the start.

How do CTA requirements for the whole array affect the value of the energy migration matrix used for LST mono analysis?

moralejo:
> > 50 TeV would be too low, right? CTA requirements are defined up to 200+ TeV - a different issue is how often we will have observations which go beyond a few tens of TeV, but no need to limit that from the start.
> 
> How do CTA requirements for the whole array affect the value of the energy migration matrix used for LST mono analysis?
> 
> We don't even simulate that high energies

I may be missing something here... what is the advantage of setting a low value for the E-axis higher end? 
We may not have data, or MC, until the very end of that range, but does it harm to have a safe limit?

PR #1223, opened 2024-02-09
Add MC DL2 to point-like IRF notebook

morcuended:
With this PR I intend to upload a new notebook (`MC_DL2_to_pointlike_IRF.ipynb`) focusing on the point-like IRF production as opposed to the existing notebook that describes the full-enclosure IRFs (including DL2 diffuse gammas, protons and electrons) `MC_DL2_to_IRF.ipynb`. I accidentally uploaded the new notebook with the same name as the old one, which was not my intention. Here I also revert that commit that overwrote the old notebook. 

The old notebook is no longer working with the current lstchain version anyway. Still, I did not want to modify it because its content on full-enclosure IRFs, which I find interesting to keep, although it should be updated somewhere else.

### DIFF
--- a/notebooks/MC_DL2_to_pointlike_IRF.ipynb
+++ b/notebooks/MC_DL2_to_pointlike_IRF.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "bb906d0a",
+   "id": "0",
    "metadata": {},
    "source": [
     "# Produce IRF from test MC DL2 (all-sky MC production) with `lstchain_create_irf_files`"
@@ -18,7 +18,7 @@
     }
    },
    "cell_type": "markdown",
-   "id": "f209fe18",
+   "id": "1",
    "metadata": {},
    "source": [
     "Create instrument response functions (IRF) in FITS format, from **test MC DL2** files assuming certain **selection cuts and event filtering** using the `lstchain_create_irf_files` Tool, which in turn uses [**pyirf**](https://pyirf.readthedocs.io).\n",
@@ -53,7 +53,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "8d3d03cf",
+   "id": "2",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -73,7 +73,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "da5a555d-ad84-4160-83d8-d41820ed9b4c",
+   "id": "3",
    "metadata": {},
    "source": [
     "# Settings of the tool (defined by the user)"
@@ -81,7 +81,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "bcc7ef69-e499-4f44-a3b8-cb88428ec352",
+   "id": "4",
    "metadata": {},
    "source": [
     "## Input files: standard all-sky DL2 test MC\n",
@@ -93,7 +93,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "70302483-b6bb-4457-8966-be5f83c45480",
+   "id": "5",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -103,7 +103,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "4830afa8-ad43-454f-bf77-887fb4f3a677",
+   "id": "6",
    "metadata": {},
    "source": [
     "For this example, let's use one of the latest all-sky base productions:\n",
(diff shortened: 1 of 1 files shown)


morcuended:
Updated the notebook example on pointlike IRF creation, clearing the outputs. I also leave the old notebook on IRF creation (also describing full-enclosure IRFs) untouched. Ready for review.

chaimain (changes requested):
- If we would like to keep the older notebooks (just for the extra Background IRF plots...?), it would be helpful to mention explicitly that these are produced with older MC production and lstchain/pyirf versions. It would also help (a general/fresh user) to understand or hint at, why we do not have some of the simulations.
- Can you also mention explicitly that these IRFs are for source-independent analysis? We may link to an updated notebook discussing the source-dependent analysis files later.
- Can you also explain why we need to produce IRFs for _all_ nodes of a declination line? The answer lies in providing the full scope for IRF interpolation. But maybe a smarter analyzer can avoid some nodes when using them for analyses of specific sources.
- Maybe it is trivial, but it might be useful to explain empty IRF bins (in energy axes) for some IRFs (zenith-dependence and usage of over-encompassing range in the common config)

morcuended:
Many thanks for the comments @chaimain, will work on the changes.

> If we would like to keep the older notebooks (just for the extra Background IRF plots...?), it would be helpful to mention explicitly that these are produced with older MC production and lstchain/pyirf versions. It would also help (a general/fresh user) to understand or hint at, why we do not have some of the simulations.

Indeed, I can indicate that they were produced with older software versions and MC productions.

> Can you also mention explicitly that these IRFs are for source-independent analysis? We may link to an updated notebook discussing the source-dependent analysis files later.

Sure

> Can you also explain why we need to produce IRFs for all nodes of a declination line? The answer lies in providing the full scope for IRF interpolation. But maybe a smarter analyzer can avoid some nodes when using them for analyses of specific sources.

Yes, I can mention this. The idea of having the IRF produced for all nodes was for simplicity and to have IRFs ready for "any source". Indeed you will not use all nodes for a specific source, but looking for the closest MC nodes, to interpolate later requires an extra step that it's not trivial in my opinion. Side comment: I've noticed sometimes that for some data, no simplex in which IRFs are interpolated can be found and the nearest node is used. I can also mention that this can happen.

> Maybe it is trivial, but it might be useful to explain empty IRF bins (in energy axes) for some IRFs (zenith-dependence and usage of over-encompassing range in the common config)

Ok, is there something that can be done in these cases? Like redefining the energy binning in the irf config? Later in Gammapy, you redefine the energy axes anyway.

chaimain:
> Many thanks for the comments @chaimain, will work on the changes.
> 
> > If we would like to keep the older notebooks (just for the extra Background IRF plots...?), it would be helpful to mention explicitly that these are produced with older MC production and lstchain/pyirf versions. It would also help (a general/fresh user) to understand or hint at, why we do not have some of the simulations.
> 
> Indeed, I can indicate that they were produced with older software versions and MC productions.

Can you clarify again why we need to see the older (obsolete) Background IRF plots? The Full-Enclosure contains PSF, and in that notebook, I had not produced or shown the plots. Maybe we can just link to the Gammapy docs (For example [CTA with Gammapy](https://docs.gammapy.org/1.2/tutorials/data/cta.html#sphx-glr-tutorials-data-cta-py) or the v1.1 version of it) for the users to see what FE IRFs for CTA would look like.

> > Can you also explain why we need to produce IRFs for all nodes of a declination line? The answer lies in providing the full scope for IRF interpolation. But maybe a smarter analyzer can avoid some nodes when using them for analyses of specific sources.
> 
> Yes, I can mention this. The idea of having the IRF produced for all nodes was for simplicity and to have IRFs ready for "any source". Indeed you will not use all nodes for a specific source, but looking for the closest MC nodes, to interpolate later requires an extra step that it's not trivial in my opinion. Side comment: I've noticed sometimes that for some data, no simplex in which IRFs are interpolated can be found and the nearest node is used. I can also mention that this can happen.

I was thinking more in the case, where an analyzer selects a particular declination line of testing MC, for analyses of some source/s, where usually we have some zenith (or maybe azimuth as well) limits. For such case-by-case analysis, the analyzer need not concern with interpolated IRFs "far beyond" the expected/selected range.

> > Maybe it is trivial, but it might be useful to explain empty IRF bins (in energy axes) for some IRFs (zenith-dependence and usage of over-encompassing range in the common config)
> 
> Ok, is there something that can be done in these cases? Like redefining the energy binning in the irf config? Later in Gammapy, you redefine the energy axes anyway.

In principle, yes the analyzer may change the irf_dl3_tool_config file to be as required. However, there is no problem with any (or most) high-level analysis with Gammapy when such empty bins exist. So, for any general analyses, there should be no problem with working with the default setting. See #1159 for possible confusion an analyzer might face when seeing the empty bins.

A concern may be on the interpolations within a grid of IRFs with different energy ranges (of the MC events), around discrete changes (for example at 52.5 deg zenith). This is not a significant problem as the interpolation code considers the empty IRF bins to have value as 0, making it mathematically possible to continue with the interpolation process.

PR #1224, opened 2024-02-09
Update dl1 to dl2 script (Less required memory and less required step)

SeiyaNozaki:
This PR updates the dl1 to dl2 step:

- Load each random forest model just before using it. Now the script loads all models at the beginning of the script and it consumes much memory. Just pass the model path to `apply_models`. This option was already implemented in `apply_models`, but not used for the script before.
- This script applies the particle classification models to the data twice for `gammaness` and `reco_type`. But `reco_type` can be set based on `gammaness` as suggested by @maxnoe here. Even, we can remove `reco_type` which will not be used for the analysis.
https://github.com/cta-observatory/cta-lstchain/pull/1077#discussion_r1104715136

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -728,10 +728,11 @@ def apply_models(dl1,
                          "the predicted probabilty to assign as gammaness is unclear."
                          "Please check training data")
 
-    # gammaness is the prediction probability for the first class (0)
-    dl2['gammaness'] = probs[:, 0]
-    col = list(classifier.classes_).index(0)
-    dl2['reco_type'] = np.where(probs[:, col] > 0.5, 0, 101)
+    # gammaness is the prediction probability for the class 0 (proton: class 101)
+    mc_type_gamma, mc_type_proton = 0, 101
+    col = list(classifier.classes_).index(mc_type_gamma)
+    dl2['gammaness'] = probs[:, col]
+    dl2['reco_type'] = np.where(probs[:, col] > 0.5, mc_type_gamma, mc_type_proton)
     del classifier
 
     return dl2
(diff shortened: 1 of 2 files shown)


lstchain/reco/dl1_to_dl2.py
```diff
@@ -731,6 +730,9 @@ def apply_models(dl1,
 
     # gammaness is the prediction probability for the first class (0)
     dl2['gammaness'] = probs[:, 0]
+    col = list(classifier.classes_).index(0)
```
maxnoe, line 733: Nice that you get the col automtaically.. but the line above still uses the hard-coded index

SeiyaNozaki:
Concerning the first point, it was on purpose for multiple input files...
https://github.com/cta-observatory/cta-lstchain/pull/985

So I will modify codes like

- If there is a single input file, each model will be loaded just before it is used
- If there are multiple input files, all modes will be loaded once at the beginning and there will be no reload of models for each file

SeiyaNozaki:
I realized I didn't read carefully comments from @gabemery during the school... sorry!

gabemery (dismissed):

lstchain/reco/dl1_to_dl2.py
vuillaut: @SeiyaNozaki could you move line 733 up one line and use col in 732?
After that we can merge I think

vuillaut (changes requested):

lstchain/reco/dl1_to_dl2.py
SeiyaNozaki: sorry I forgot to implement @maxnoe's comment.. now implemented!

vuillaut (approved):

PR #1227, opened 2024-02-12
Fix fake errors in DL2 to DL3 & Index creation step

### LINKED ISSUE #1225: Fake ERRORs in DL2 to DL3 step
When running the DL2 to DL3 step for a source-independent analysis, one gets this error if IRF interpolation is used:

2024-02-10 13:32:49,219 ERROR [lstchain.high_level.interpolate] (interpolate.interpolate_irf): AL CUTS not present for IRF interpolation
Obviously, Alpha cuts do not exist for this type of analysis, and the DL3 file is nevertheless produced (as it should).
If one uses explicitly the option 
--source-dep False

it is even worse... it throws the same error plus this one:
2024-02-10 13:28:32,031 ERROR [lstchain.DataReductionFITSWriter] (tool.run): Caught unexpected exception: File does not contain source-dependent parameters

The more you tell it you are doing a source-**in**dependent analysis, the more it wants to do it!

SeiyaNozaki:
Fix #1225  with some code refactoring.
I updated codes adding case handling for IRF interpolation and index creation (point-like IRF: effective area and energy dispersion. full-enclosure: + PSF and background model). I added a warning message for background IRF.
Actually I also found that there was a small typo/bug in PSF interpolation and pytest of the interpolation for full-enclosure IRF, so I also fixed it.


lstchain/high_level/hdu_table.py
```diff
@@ -198,25 +198,35 @@ def create_hdu_index_hdu(file_list, hdu_index_file, overwrite=False):
...
-            "PSF",
-            "RAD_MAX",
         ]
 
-        for irf in hdu_names:
+        # 0:PRIMARY, 1:EVENTS, 2:GTI, 3:POINTING, 4-:IRF 
+        if hdu_list[4].header["HDUCLAS3"] == "FULL-ENCLOSURE":
+            hdu_names.extend(["BACKGROUND", "PSF"])
+        if "TH_CONT" in hdu_list[4].header:
+            hdu_names.extend("RAD_MAX")
```
maxnoe, line 207: rad_max is required for point-like IRFs, so I think it would make more sense to check for `"HDUCLAS3" == "POINT-LIKE``.

It would also make more sense to just check all the hdus present in the file one by one instead of a pre-defined list of hdus.

lstchain/high_level/hdu_table.py
```diff
@@ -195,28 +195,19 @@ def create_hdu_index_hdu(file_list, hdu_index_file, overwrite=False):
...
-                    "_" + irf_hdu.lower().split("_")[-1]
-                )
-                t_irf["HDU_NAME"] = irf
-                hdu_index_tables.append(t_irf)
-            except KeyError:
-                log.error(f"Run {t_events['OBS_ID']} does not contain HDU {irf}")
 
+        # 0:PRIMARY, 1:EVENTS, 2:GTI, 3:POINTING, 4-:IRF 
+        for hdu in hdu_list[4:]:
+            irf_hdu = hdu.header["HDUCLAS4"]
```
chaimain, line 201: This will not work for the [GH_CUTS](https://github.com/cta-observatory/cta-lstchain/blob/main/lstchain/tools/lstchain_create_irf_files.py#L674-L687) and [AL_CUTS](https://github.com/cta-observatory/cta-lstchain/blob/main/lstchain/tools/lstchain_create_irf_files.py#L701-L714) HDU as they are not pyirf created HDUs and do not (and should not) have HDUCLAS4 header.

chaimain (changes requested):
Maybe also update the [DL2_to_DL3](https://github.com/cta-observatory/cta-lstchain/blob/main/notebooks/DL2_to_DL3.ipynb) notebook to remove the fake errors for better documentation

moralejo:
@SeiyaNozaki, this got stuck... can you please address the comments?

lstchain/high_level/hdu_table.py
SeiyaNozaki: It would be better to include GH_CUTS and AL_CUTS as well? The original code didn't include those cuts in `hdu_names`

SeiyaNozaki:
Regarding DL2_to_DL3 notebook, can you re-run the notebook with this branch? @morcuended 
We also need the config file name as well. Then I got the following error when launching `lstchain_create_dl3_file` in this notebook. I'm not much sure what is the reason...
```
The metadata are not comparable
The other parameter axes data are comparable
2024-06-05 16:39:32,544 ERROR [lstchain.DataReductionFITSWriter] (tool.run): IRF files in /fefs/aswg/workspace/analysis-school-2024/DL2_to_DL3/IRF with pattern, *irf*.fits.gz are not similar and cannot be used to interpolate. Use different list of IRFs.
```

lstchain/high_level/hdu_table.py
SeiyaNozaki: okay I see! Maybe we can keep it as it was in this PR. Then let's make another PR if needed!

morcuended:
> *irf*.fits.gz

Sure, I will try it out. 

Regarding the problem with the irfs, I think it is because I produced in the same directory an IRF file with different cuts or something like this. I removed the file I think is problematic. You can try again. I will do so as well.

morcuended:
I will update the notebook as well, in my to-do list

lstchain/high_level/hdu_table.py
SeiyaNozaki: I modified the code to check all the hdus present in the file one by one
```
        # 0:PRIMARY, 1:EVENTS, 2:GTI, 3:POINTING, 4-:IRF 
        for hdu in hdu_list[4:]:
            irf_hdu = hdu.header["HDUCLAS4"]
```

SeiyaNozaki:
Okay thank you @morcuended !

@chaimain @maxnoe Could you please review the code again (apart from DL2_to_DL3 notebook which will be updated by Daniel later)? Thanks!

lstchain/high_level/hdu_table.py
chaimain: This then checks the HDUs `GH_CUTS` and `AL_CUTS` for `HDUCLAS4`, where it is not present

chaimain: The tests pass because there is only a single test for creating the DL3 index files, which uses the source-independent DL3 files with global cuts.
Maybe adding a test for indexing the source-dependent DL3 files with energy-dependent cuts would be a good measure. (Basically replicating the current test for the path `temp_dir_observed_srcdep_files` instead of `temp_dir_observed_files`)

chaimain (approved):

PR #1228, opened 2024-02-12
Update docs for DL2 to DL3 step

moralejo:
It is dangerous to have two different configuration files for the IRF creation and the DL3 creation. Cuts should be the same...


moralejo:
Now done in PR#1229

PR #1229, opened 2024-02-12
Update irf_tool_config.json

moralejo:
Set as standard "efficiency cuts" to a 70% efficiency, which is the "baseline analysis" in the LST1 performance paper.


moralejo:
Removed superfluous (and confusing) config file for DL3 creation. It must be the same as for IRF creation.
Also: 

- Changed default efficiency cuts to 70% efficiency. 
- Raised the maximum allowed gammaness cut from 0.95 to 0.98.  For very high energies gammaness distribution can be very peaky at 1, and this could be limiting performance.

morcuended (dismissed):
Looks good. Maybe the remaining example config file could be named `irf_dl3_tool_config.json`

morcuended (approved):

PR #1234, opened 2024-02-23
Script to perform data volume reduction on data taken with EvB6 (CTAR1 format)

moralejo:
Performs data volume reduction for EvB6 (CTAR1) data


moralejo:
TBD: improve logging.  Modify lstchain_dvr_pixselector to be faster (use just a fraction of the subruns in a run to determine the charge for pixel selection)

lstchain/scripts/lstchain_r0g_to_r0v.py
```diff
@@ -0,0 +1,175 @@
...
+from astropy.io import fits
+
+import lstchain.paths as paths
+import numpy as np
+from contextlib import ExitStack
+
+parser = argparse.ArgumentParser(description="Volume reducer program (R0G to "
+                                             "R0V)")
+parser.add_argument('-f', '--R0G-file', dest='input_file', required=True,
+                    type=str, help='Input R0G file name (of stream 1)')
```
maxnoe, line 31: you can directly use `type=Path` here to get the right type, also for the others

maxnoe: Why not use positional arguments for the required ones?

moralejo: No obvious advantage, right? And it is similar in other scripts, I think (e.g. -f  for mandatory input files)

lstchain/scripts/lstchain_r0g_to_r0v.py
```diff
@@ -139,7 +139,8 @@ def main():
 
                 # Check if this event is known
                 if event.event_id not in pixel_mask:
-                    log.warn('Event id %d not found in pixel selection file!')
+                    log.warn('Event id %d not found in pixel selection file!',
```
maxnoe, line 142: https://docs.python.org/3/library/logging.html#logging.Logger.warning

lstchain/scripts/lstchain_r0g_to_r0v.py
```diff
@@ -28,18 +28,18 @@
 parser = argparse.ArgumentParser(description="Volume reducer program (R0G to "
                                              "R0V)")
 parser.add_argument('-f', '--R0G-file', dest='input_file', required=True,
-                    type=str, help='Input R0G file name (of stream 1)')
+                    type=Path, help='Input R0G file name (of stream 1)')
 
 parser.add_argument('-o', '--output-dir', dest='output_dir',
                     type=str, default='./',
```
morcuended, line 34:
```suggestion
                    type=Path, default='./',
```

moralejo: No advantage in that case, but ok.

lstchain/scripts/lstchain_r0g_to_r0v.py
```diff
@@ -0,0 +1,175 @@
...
+                stream.write_message(event)
+
+            stream.close()
+            input_streams[i].close()
+
+            # Open again output file for setting the LSTDVR flag:
+            outfile = fits.open(name, mode="update")
+            outfile['Events'].header.set("LSTDVR", True, 
+                                         "LST offline DVR applied")
+            outfile.close()
```
morcuended, line 172: Not that it changes much, but can't this be set already above, while you have the file opened?

morcuended (commented):
Looks good to me overall. Just left some minor comments

lstchain/scripts/lstchain_r0g_to_r0v.py
moralejo: No, it can't be done like that... protzfits does not have the functions to do it.

maxnoe: I just released protozfits 2.5 with added `set_{str,bool,int,float}` methods on `ProtobufZOFits` and also the possibility to set the per-column compressions. 

The header keywords need to be set before the first row of the table is written.

maxnoe: If using astropy, you should use a context manager:

```
with fits.open(name, mode="update") as outfile:
    outfile["Events"].header.set("LSTDVR", True, "LST offline DVR applied")
```

lstchain/scripts/lstchain_r0g_to_r0v.py
```diff
@@ -0,0 +1,175 @@
...
+                    defaul_compression="zstd9"))
+            stream.open(str(name))
+
+            stream.move_to_new_table("DataStream")
+            stream.write_message(input_streams[i].DataStream[0])
+          
+            stream.move_to_new_table("CameraConfiguration")
+            stream.write_message(input_streams[i].CameraConfiguration[0])
+
+            stream.move_to_new_table("Events")
```
maxnoe, line 131: You can now do (with protozfits 2.5) at this line:

```
stream.set_bool("LSTDVR", True, "LST offline DVR applied")
```

lstchain/scripts/lstchain_r0g_to_r0v.py
moralejo: I have to change the protozfits version in environment.yml, right?  (not needed in setup.py which has >=2.2)

maxnoe: Since we would be using features only available with 2.5, we need to update the requirements to be `>= 2.5, <3`

setup.py
```diff
@@ -54,14 +54,13 @@ def find_scripts(script_dir, prefix, suffix='.py'):
         'numba',
         'numpy',
         'pandas',
-        'protobuf~=3.20.0',
```
morcuended, line 57: why is protobuf gone? not needed anymore with protozfits >=2.5?

moralejo: No need to explicitly list it at least

morcuended (approved):

setup.py
maxnoe: The pin was only necessary because older versions of `protozfits` were incompatible with newer versions of protobuf but didn't specify that. That is long fixed (protozfits correctly pins what versions of protobuf it is compatible with) and newer versions of protozfits are also compatible with the latest protobuf.

maxnoe: Since lstchain does not directly use protobuf, it should just rely on the transitive dependency specification of `protozfits`. The pin was only necessary because in the past, that specification was wrong.

morcuended: That's what I though. Thanks for the explanation

PR #1237, opened 2024-02-26
Fixed bug & couple of improvements in lstchain_dvr_pixselector

moralejo:
Fixed long-standing BUG which made that the info in the DVR_settings file produced by the script contained only the information of the last subrun.

Improvements in logging

Now by default only 10 files (subruns) of the input set, distributed uniformly along the run, are actually processed to determine the DVR settings. This is to speed up the process (since the settings are determined for the whole run, we do not need to check all subruns).

### DIFF
--- a/lstchain/scripts/lstchain_dvr_pixselector.py
+++ b/lstchain/scripts/lstchain_dvr_pixselector.py
@@ -456,19 +456,14 @@ def main():
                 writer_conf = tables.Filters(complevel=9, fletcher32=True)
                 filemode = 'w'
                 if not write_pixel_masks:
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -409,7 +459,7 @@ def main():
                     if run_id != current_run_number:
                         current_run_number = run_id
                     else:
-                        filemode == 'a'
```
moralejo, line 412: THIS was the blunder!

rlopezcoto (approved):

PR #1238, opened 2024-02-27
Update of the single photoelectron response file

Mitsunari-Takahashi:
The single photoelectron spectrum data `lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat` has an artificial bump in the spectrum due to the analysis method, and we made a new one `lstchain/data/spe_LST1_2022-04-27_AP2.0e-4.dat` to remove it. In addition, the new spectrum is folded with the afterpulsing while the old one is not. 

Please find the details at https://github.com/cta-observatory/lst-sim-config/issues/40.


rlopezcoto:
thanks @Mitsunari-Takahashi !
As I understood from the discussion in https://github.com/cta-observatory/lst-sim-config/issues/40, this will be the file used for Prod6. This change will mainly be affecting the NSB tuning here in lstchain, are there any changes that need to be applied in [ctapipe_io_lst](https://github.com/cta-observatory/ctapipe_io_lst) as well for the calibration @FrancaCassol , @maxnoe ?

maxnoe:
Is this the single pe pulse shape? Or the single pe integration spectrum?


ctapipe_io_lst only uses the single pe pulse shape (to fill it into the `SubarrayDescription` / `CameraReadout` ).

The file used currently is 

```
oversampled_pulse_LST_8dynode_pix6_20200204.dat
```

here: https://github.com/cta-observatory/ctapipe_io_lst/blob/main/src/ctapipe_io_lst/resources/oversampled_pulse_LST_8dynode_pix6_20200204.dat

Mitsunari-Takahashi:
> Is this the single pe pulse shape? Or the single pe integration spectrum?

It's the LST1-PMT output charge spectrum integrated over an extraction time window. 

I missed the format of the new dat file different from the old one. So please reject this pull request. Sorry, I will correct the file. Before that, could anyone confirm whether the NSB tuning needs the AP-folded spectrum?

moralejo:
> > Is this the single pe pulse shape? Or the single pe integration spectrum?
> 
> It's the LST1-PMT output charge spectrum integrated over an extraction time window.
> 
> I missed the format of the new dat file different from the old one. So please reject this pull request. Sorry, I will correct the file. Before that, could anyone confirm whether the NSB tuning needs the AP-folded spectrum?

NSB tuning at the waveform level (which is the preferred method) needs both the pulse template (shape) and the amplitude fluctuations. About the APs, I always find it quite confusing to mix them with the spectrum of amplitude fluctuations. It is ok for the NSB, because it is anyway arriving at random times. But for the p.e.'s from C-photons we should never use such a spectrum of amplitude fluctuations including APs, right? The APs produced by p.e.'s from C-photons should not be part of the same event, but produce a delayed signal.

So my understanding is that for the specific case of simulating NSB is it fine (even convenient) to include the APs, but not for the simulation of the signal (which of course it is not done in lstchain!). But I am not 100% sure.

Mitsunari-Takahashi:
> So my understanding is that for the specific case of simulating NSB is it fine (even convenient) to include the APs, but not for the simulation of the signal (which of course it is not done in lstchain!). But I am not 100% sure.

I see, so we should separate these problems.

For `lstchain`, the single p.e. spectrum should be folded by AP because it is used only for the NSB tuning (See the diff list in the commits). I will make a new pull request.

In addition, to implement the NSB tuning at the waveform level, I understand we need the pulse shape template. I'd like to remark that the AP pulse is shorter than the photon pulse, but probably the difference is negligible. 

For the Cherenkov photon signal simulation, I agree that the AP should not be folded. Is this a matter of `lst-sim-config`? @moralejo

moralejo:
> 
> For `lstchain`, the single p.e. spectrum should be folded by AP because it is used only for the NSB tuning (See the diff list in the commits). I will make a new pull request.
> 

Correct

> In addition, to implement the NSB tuning at the waveform level, I understand we need the pulse shape template. I'd like to remark that the AP pulse is shorter than the photon pulse, but probably the difference is negligible.

I'd say so...  APs are after all just a small part of the noise.

> For the Cherenkov photon signal simulation, I agree that the AP should not be folded. Is this a matter of `lst-sim-config`? @moralejo

I just checked that it is handled properly in the simulation (I had no doubts). From the sim_telarray documentation:
_"The s.p.e. response tables also accounts for afterpulses. Since afterpulses from the Cherenkov light itself arrives much too late to be registered with the shower, afterpulses are only taken into account for night-sky background light, where it doesn’t matter if the original photo-electron was a few hundred nanoseconds earlier"_

moralejo:
@Mitsunari-Takahashi can we merge and close this? 
I see you are trying to merge from a fork. I cannot re-run tests which failed (probably because of that?)

Mitsunari-Takahashi:
@moralejo Sorry, I forgot to do that! Actually, I'd like to replace the single phe file by a new one, which I derived from a much larger number of PMTs. I will do this as soon as possible.

PR #1239, opened 2024-02-27
Compute cleaning thresholds in lstchain_dvr_pixselector

moralejo:
When used for determining the settings for the Data Volume Reduction (DVR) of a given run, the program now outputs also a .json file (adapted to the NSB conditions) for use in the lstchain_dl1ab step

### DIFF
--- a/lstchain/scripts/lstchain_dvr_pixselector.py
+++ b/lstchain/scripts/lstchain_dvr_pixselector.py
@@ -520,15 +520,15 @@ def main():
             picture_threshold = np.floor(meanq)
             if picture_threshold % 2 != 0:
                 picture_threshold += 1  # We change picture_threshold in steps of 2 p.e.
-                boundary_threshold = picture_threshold / 2
-                newconfig = get_standard_config()['tailcuts_clean_with_pedestal_threshold']
-                newconfig['picture_thresh'] = picture_threshold
-                newconfig['boundary_thresh'] = boundary_threshold
-                run = int(file.name[file.name.find('Run')+3:-3])
-                json_filename = Path(output_dir, f'dl1ab_Run{run:05d}.json')
-                dump_config({'tailcuts_clean_with_pedestal_threshold': 
+            boundary_threshold = picture_threshold / 2
+            newconfig = get_standard_config()['tailcuts_clean_with_pedestal_threshold']
+            newconfig['picture_thresh'] = picture_threshold
+            newconfig['boundary_thresh'] = boundary_threshold
+            run = int(file.name[file.name.find('Run')+3:-3])
+            json_filename = Path(output_dir, f'dl1ab_Run{run:05d}.json')
+            dump_config({'tailcuts_clean_with_pedestal_threshold': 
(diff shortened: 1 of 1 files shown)


moralejo:
Also corrected a small error in earlier changes: in case more than one run was processed (in the "compute_dvr_settings" mode) the total number of (DL1) subruns processed was 10, instead of 10 per run, as intended.

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -168,19 +177,12 @@ def main():
         log.error('Unknown option in --action flag!')
         sys.exit(1)
 
-    dl1_files = []
-
-    if (write_pixel_masks or 
-        len(all_dl1_files) <= max_number_of_processed_subruns):
```
rlopezcoto, line 174: this was a safe check, why removing it?

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -362,17 +367,19 @@ def main():
                                                                   args.picture_threshold,
                                                                   camera_geom)
         else:
+            # For the creation of pixel masks, we get the most typical value
+            # of min_charge_for_certain_selection among the subruns of the run
+            # in the DVR_settings_LST*.h5 file
             runmask = dvr_settings['run_id'] == run_id
-            meanq = dvr_settings['min_charge_for_certain_selection'][runmask].mean()
```
rlopezcoto, line 366: not needed the mean charge any more?

lstchain/scripts/lstchain_dvr_pixselector.py
```diff
@@ -496,6 +506,33 @@ def main():
                 time.sleep(300)
```
rlopezcoto, line 506: what was the reason for this `sleep` here?

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: this was only to obtain min_charge_for_certain_selection
Now it is done by get_typical_dvr_min_charge, so we do not need the meanq intermediate value
(not it is no longer the mean, but the mode, safer against carflashes etc)

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: This was to just process all input DL1 if the total number was smaller than the maximum we want to process (10)
But now the whole thing is done inside get_input_files

In case we are writing the pixel masks we do want all files to be processed, that is why the "if"

lstchain/scripts/lstchain_dvr_pixselector.py
moralejo: I introduced this because at some point we had many errors in writing out the pixel masks, related to glitches in the system @ LP. So I introduced this: in case of error it waits for five minutes and tries again to write out.  It does it a maximum number of times.

rlopezcoto (approved):

PR #1245, opened 2024-03-14
LHfit : DVR based pixel selection + new ped variance extraction for MC

gabemery:
Two elements implemented for (MC) processing : 
- DVR based pixel selection for both data and MC before performing the likelihood fit. This is a small change to the pixels selected for data after volume reduction, but it allows to have the same selection in MC.
- A new method to extract the pedestal variance in MC : extract once at the start (like it is done for data using interleaved events). This is required to keep events where the full camera is selected and to avoid bias in the extraction in the case of dvr pixel selection

Draft to extend test coverage

### DIFF
--- a/lstchain/reco/reconstructor.py
+++ b/lstchain/reco/reconstructor.py
@@ -13,7 +13,6 @@ from lstchain.image.modifier import tune_nsb_on_waveform
 from lstchain.io.lstcontainers import DL1LikelihoodParametersContainer
 from lstchain.reco.reconstructorCC import log_pdf as log_pdf
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


chaimain (changes requested):
Some simple changes based on code quality

lstchain/reco/reconstructor.py
```diff
@@ -24,6 +25,16 @@ class TimeWaveformFitter(TelescopeComponent):
                                       allow_none=False).tag(config=True)
     crosstalk = FloatTelescopeParameter(default_value=0, help='Average pixel crosstalk.',
                                         allow_none=False).tag(config=True)
+    spatial_selection = Unicode(default_value='dvr',
+                                help='Method to select pixels to perform the likelihood fit. Can be:\n'
+                                     'hillas : use the hillas length and width times sigma_space as an ellipsis.'
+                                     '\ndvr : use data volume reduction logic with dvr_pic_threshold and'
+                                     ' dvr_pix_for_full_image',
+                                allow_none=False).tag(config=True)
+    dvr_pic_threshold = Int(8, help='Pixel charge threshold for dvr like pixel selection.',
```
chaimain, line 34: Just for code consistency, can you add `default_value=` to justify the first argument for this and some other parameters?

lstchain/reco/reconstructor.py
```diff
@@ -299,7 +342,7 @@ def __call__(self, event, telescope_id, dl1_container):
 
             return self.predict(unit_cam, fit_params)
 
-    def clean_data(self, pix_x, pix_y, pix_radius, times, start_parameters, telescope_id):
+    def clean_data(self, image, geometry, pix_x, pix_y, pix_radius, times, start_parameters, telescope_id):
         """
         Method used to select pixels and time samples used in the fitting procedure.
```
chaimain, line 347: Can you update the docstring to reflect the additional support for DVR parameters?

chaimain (approved):

PR #1254, opened 2024-05-16
fix missing SinglePhE_ResponseInPhE_expo2Gaus.dat

### LINKED ISSUE #1253: FileNotFoundError: lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat not found
Hi

With lstchain >= v0.10.7 I get the following error when running `lstchain_mc_r0_to_dl1`:

```
FileNotFoundError: lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat not found
```

vuillaut:
This PR fixes #1253 by:
- creating a function to get resource data from lstchain
- setting the path to None by default in the config and looking for the default file path in this case


morcuended:
pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources)

lstchain/io/io.py
```diff
@@ -1345,3 +1347,9 @@ def get_mc_fov_offset(filename):
 
     return mean_offset
 
+
+def get_resource_path(filename):
+    """
+    Get a resource data path in lstchain package
+    """
+    return Path(resource_filename("lstchain", filename))
```
maxnoe, line 1355: Use importlib.resources:

```
return files("lstchain") / filename
```

maxnoe: https://docs.python.org/3/library/importlib.resources.html

vuillaut: Any particular reason?
`resource_filename` was used everywhere in the code, I just created a generic function instead of copy/pasting the same code again.

vuillaut: H just say @morcuended comment, thanks

lstchain/image/tests/test_modifier.py
```diff
@@ -77,11 +82,6 @@ def test_calculate_required_additional_nsb(mc_gamma_testfile, observed_dl1_files
 
 
 def test_tune_nsb_on_waveform():
-    import os
-    import astropy.units as u
-    from scipy.interpolate import interp1d
-    from lstchain.image.modifier import tune_nsb_on_waveform
```
maxnoe, line 82: It is good practice to import the internal imports here. That way import errors appear as part of the test and do not break the whole test suite

lstchain/scripts/lstchain_tune_nsb_waveform.py
```diff
@@ -88,7 +89,7 @@ def main():
     dict_nsb = {
         "nsb_tuning": True,
         "nsb_tuning_ratio": np.round(nsb_correction_ratio, decimals=2),
-        "spe_location": "lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat"
+        "spe_location": get_resource_path("data/SinglePhE_ResponseInPhE_expo2Gaus.dat")
```
morcuended, line 92: `get_reource_path` returns a Path object, I do not know if that'd be a problem here

morcuended (commented):
thanks @vuillaut looks, good to me. I just left a comment

lstchain/scripts/lstchain_tune_nsb_waveform.py
vuillaut: No problem there, it's then passed to `loadtxt`

morcuended (approved):

vuillaut:
Thanks for the review @morcuended 
Sorry I made a last-minute change to make sure the default file is indeed used in the tests. Could you review again please?

PR #1257, opened 2024-05-29
Update codecov version in CI

morcuended:
It seems that coverage reports are not being uploaded to codecov, maybe because the old version


morcuended:
Checkout action is already at the latest version, I updated setup python to v5, and changed the deprecated provision mamba by the setup micromamba action

morcuended:
I think this is ready to go @maxnoe

maxnoe (approved):

PR #1259, opened 2024-06-05
Computation of pointing-dependent weights for MC events 

moralejo:
This is to avoid jumps in RF performance in the "mid-points" between training pointing nodes, just because the event statistics occasionally have jumps between neighboring nodes. 

The idea is to  use "sample_weight", i.e. a weight for each event in the training set, calculated such that each of the pointing nodes has the same total weight.

The calculation is done with all events that are passed to the RFs.  Of course, the energy (and impact) range of those events between culmination and high zenith change a lot, so "same number of events" does not mean much when comparing such extremes, but the point here is to avoid steps between _neighboring_ bins (and for those, indeed, the numbers of events are comparable quantities). Only if the impact and energy ranges were poorly chosen (and distributions truncated) would this normalization result in performance jumps like the ones we have without it.

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -65,6 +65,10 @@
     "fraction_cleaning_intensity": 0.03
   },
 
+  "random_forest_weight_settings": {
+    "pointing_wise_weights": true
+  },
+
   "random_forest_energy_regressor_args": {
     "max_depth": 30,
     "min_samples_leaf": 10,
@@ -79,8 +83,7 @@
     "min_weight_fraction_leaf": 0.0,
     "oob_score": false,
     "random_state": 42,
-    "warm_start": false,
-    "pointing_wise_weights": true
+    "warm_start": false
   },
 
   "random_forest_disp_regressor_args": {
@@ -97,8 +100,7 @@
     "min_weight_fraction_leaf": 0.0,
     "oob_score": false,
     "random_state": 42,
-    "warm_start": false,
-    "pointing_wise_weights": true
+    "warm_start": false
   },
 
   "random_forest_disp_classifier_args": {
@@ -116,8 +118,7 @@
     "oob_score": false,
     "random_state": 42,
     "warm_start": false,
-    "class_weight": null,
-    "pointing_wise_weights": true
+    "class_weight": null
   },
 
   "random_forest_particle_classifier_args": {
@@ -135,8 +136,7 @@
     "oob_score": false,
     "random_state": 42,
     "warm_start": false,
-    "class_weight": null,
-    "pointing_wise_weights": true
+    "class_weight": null
   },
 
 
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -79,7 +79,8 @@
     "min_weight_fraction_leaf": 0.0,
     "oob_score": false,
     "random_state": 42,
-    "warm_start": false
+    "warm_start": false,
+    "pointing_wise_weights": true
   },
 
   "random_forest_disp_regressor_args": {
@@ -96,7 +97,8 @@
     "min_weight_fraction_leaf": 0.0,
     "oob_score": false,
     "random_state": 42,
-    "warm_start": false
+    "warm_start": false,
+    "pointing_wise_weights": true
   },
 
   "random_forest_disp_classifier_args": {
@@ -114,7 +116,8 @@
     "oob_score": false,
     "random_state": 42,
     "warm_start": false,
-    "class_weight": null
+    "class_weight": null,
+    "pointing_wise_weights": true
   },
 
   "random_forest_particle_classifier_args": {
@@ -132,7 +135,8 @@
     "oob_score": false,
     "random_state": 42,
     "warm_start": false,
-    "class_weight": null
+    "class_weight": null,
+    "pointing_wise_weights": true
   },
 
 
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -119,7 +119,7 @@ def train_disp_vector(train, custom_config=None, predict_features=None):
     reg = model(**disp_regression_args)
     x = train[features]
     y = np.transpose([train[f] for f in predict_features])
-    reg.fit(x, y)
+    reg.fit(x, y, sample_weight=train['weight'])
 
     logger.info("Model {} trained!".format(model))
 
@@ -153,7 +153,7 @@ def train_disp_norm(train, custom_config=None, predict_feature='disp_norm'):
     reg = model(**disp_regression_args)
     x = train[features]
     y = np.transpose(train[predict_feature])
-    reg.fit(x, y)
+    reg.fit(x, y, sample_weight=train['weight'])
 
     logger.info("Model {} trained!".format(model))
 
@@ -187,7 +187,7 @@ def train_disp_sign(train, custom_config=None, predict_feature='disp_sign'):
     clf = model(**classification_args)
     x = train[features]
     y = np.transpose(train[predict_feature])
-    clf.fit(x, y)
+    clf.fit(x, y, sample_weight=train['weight'])
 
     logger.info("Model {} trained!".format(model))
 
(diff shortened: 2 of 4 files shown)


moralejo:
It seems to work well:
image

Gammaness cut is 0.5 for the standard, and tuned to get the same amount of events at zd~18 deg.

With "standard RFs", aside from the (smooth) physical dependence with zenith, we have jumps which result from the non-smoothness of the evolution of the number of training events with pointing, along declination lines. The jump above (in the blue histogram) is located at the mid-point between two training nodes.

These artifacts can be removed with the changes proposed here. In all likelihood they are partly responsible for the few-percent systematic errors in flux estimation that we have.

moralejo:
@vuillaut  The training with the Crab line took nearly 10 hours in cp03. Is that normal, or perhaps the use of weights makes it slower?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -79,7 +79,8 @@ def train_energy(train, custom_config=None):
 
     reg = model(**energy_regression_args)
     reg.fit(train[features],
-            train['log_mc_energy'])
+            train['log_mc_energy'],
+            sample_weight=train['weight'])
```
rlopezcoto, line 83: sorry, I do not understand how this is done... is this way added as a column to the table? how is it calculated?

moralejo: I updated the description

moralejo: https://github.com/cta-observatory/cta-lstchain/blob/a595cee71afec4a7012e77d0395f84941bc44762/lstchain/reco/utils.py#L867
is what computes the weights

rlopezcoto: shouldn't you return the `events` dataframe in this function?

lstchain/reco/utils.py
```diff
@@ -862,3 +863,51 @@ def get_events_in_GTI(events, CatB_cal_table):
...
+
+    # Add a 'weight' column to the input table
+    weights = np.array(np.ones(len(events)))
+
+    # First identify existing telescope pointings in the sample:
+    # Convert to degrees and round to avoid potential
+    # rounding issues:
+    alt = np.round(events['alt_tel'], decimals=5)
+    az = np.round(events['az_tel'], decimals=5)
+    dummy = np.unique(np.array([alt, az]).T, axis=0)
```
maxnoe, line 884: `unique` already returns sorted values, so if you want to have it sorted by azimuth, you can just do `np.unique([alt, az], axis=1)`

lstchain/reco/dl1_to_dl2.py
moralejo: > shouldn't you return the `events` dataframe in this function?

That was my first intention, but the dataframe is modified, ,so it is not needed.

lstchain/reco/utils.py
moralejo: No, it does not work... behavior changed, I don't know why

vuillaut:
> @vuillaut The training with the Crab line took nearly 10 hours in cp03. Is that normal, or perhaps the use of weights makes it slower?

Hi
Using how many cores ?   
But that sounds right

lstchain/reco/utils.py
```diff
@@ -862,3 +863,49 @@ def get_events_in_GTI(events, CatB_cal_table):
...
+    # Weight in a given node will be mean_events_per_node / n_events_in_node
+
+    for ipointing, tel_alt_az in enumerate(pointings):
+        mask = (np.isclose(tel_alt_az[0], events['alt_tel'],
+                           atol=1e-5, rtol=0) &
+                np.isclose(tel_alt_az[1], events['az_tel'],
+                           atol=1e-5, rtol=0))
+        weights[mask] = weight_per_pointing[ipointing]
+
+    events['weight'] = weights
```
vuillaut, line 908: If the dataframe already has a column called `weights`, it is silently replaced. It should probably raise a ValueError in this case.

vuillaut (changes requested):

lstchain/data/lstchain_standard_config.json
```diff
@@ -65,6 +65,10 @@
     "fraction_cleaning_intensity": 0.03
   },
 
+  "random_forest_weight_settings": {
+    "pointing_wise_weights": true
+  },
```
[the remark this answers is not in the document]
moralejo: > is it expected to have more settings in `random_forest_weight_settings`? Otherwise, it may be simplified to `"random_forests _pointing_weights": true`

I thought we might want to set this independently for each RF - I tend to think otherwise now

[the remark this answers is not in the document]
moralejo: Still, I think this way allows better future enhancements (like class weights) without changing the current way of setting this.

lstchain/reco/utils.py
moralejo: I added a warning, which I think is enough (and we do not write out this column anywhere, so it would have to be a different 'weight')

lstchain/reco/utils.py
```diff
@@ -862,3 +863,70 @@ def get_events_in_GTI(events, CatB_cal_table):
     gti_mask = gti[events['calibration_id']]
 
     return events[gti_mask]
+
+def compute_rf_event_weights(events):
+    """
+    Compute event-wise weights. Can be used for correcting for the different
+    statistics present in each pointing node of the MC training sample,
+    to avoid "jumps" in the performance of the random forests
+
```
morcuended, line 872:
```suggestion
Parameters
----------
```

lstchain/reco/utils.py
```diff
@@ -862,3 +863,70 @@ def get_events_in_GTI(events, CatB_cal_table):
...
+    events: a DL1 parameters dataframe. The table is modified in place by the
+    addition of a column called 'weight' (unless it exists already). The
+    column contains an event-wise weight to be used in the Random Forest
+    training, to give each of the telescope pointings in the training sample
+    the same overall weight in the training.
+
+    Returns: pointings, weight_per_pointing
```
morcuended, line 879:
```suggestion
    Returns
    -------
```

lstchain/reco/utils.py
```diff
@@ -862,3 +863,70 @@ def get_events_in_GTI(events, CatB_cal_table):
...
 
     return events[gti_mask]
+
+def compute_rf_event_weights(events):
+    """
+    Compute event-wise weights. Can be used for correcting for the different
+    statistics present in each pointing node of the MC training sample,
+    to avoid "jumps" in the performance of the random forests
+
+    events: a DL1 parameters dataframe. The table is modified in place by the
```
morcuended, line 873:
```suggestion
    events : `~pd.DataFrame`
        DL1 parameters dataframe. The table is modified in place by the
```

lstchain/reco/utils.py
```diff
@@ -862,3 +863,70 @@ def get_events_in_GTI(events, CatB_cal_table):
...
+    addition of a column called 'weight' (unless it exists already). The
+    column contains an event-wise weight to be used in the Random Forest
+    training, to give each of the telescope pointings in the training sample
+    the same overall weight in the training.
+
+    Returns: pointings, weight_per_pointing
+
+    pointings: ndarray of shape (number_of_pointings, 2) Alt Az (in radians)
+    for each of the identified telescope pointings in the input MC sample
```
morcuended, line 882:
```suggestion
        for each of the identified telescope pointings in the input MC sample
```

lstchain/reco/dl1_to_dl2.py
```diff
@@ -491,6 +508,13 @@ def build_models(filegammas, fileprotons,
                 "The requested number of protons for the classifier training is not valid."
             ) from e
 
+    if pointing_wise_weights:
+        # Give same total weight to all events in every pointing node, by
+        # applying event-wise weights which depend on the statistics per node
+        _, _ = utils.compute_rf_event_weights(df_proton)
```
morcuended, line 514: Maybe add a comment here that `utils.compute_rf_event_weights(df_proton)` is adding the `weight` column to `df_proton`

morcuended (commented):
Looks good. I just left some comments that could be added and some other formatting stuff for docstrings (although we are not keeping consistency across the code, it would be good to enforce it for proper docs rendering)

morcuended (approved):

PR #1261, opened 2024-06-07
Allow for scaling true energy in IRFs to assess energy-scale systematics

deborahDOR:
Modifications of the code for the consideration of systematics in the energy. 
The goal is to generate modified IRFs due to the effect of a lower or higher light efficiency in the system (telescope + atmosphere) with respect to simulations, by scaling the true energy values in MC simulations.

The changes are:

  - A new argument in the DataBinning Component in lstchain/io/event_selection.py : "scale true energy" (value that will be applied in the true energy values of the events of the MC files)
  - Add "scale true energy" in the irf_dl3_config_tool.json in the DataBinning section (only if the value is different from 1, it will be considered in the computation).
  - Changes in lstchain_create_irfs_files.py to modify the values when pyrif commands are used. Added comments to describe how it can be used and its effects.


lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -28,6 +28,9 @@
 If you want to generate source-dependent IRFs, source-dep flag should be activated.
 The global alpha cut used to generate IRFs is stored as AL_CUT in the HDU header.
 
+Modified IRFs can be created in case of considering systematics in the
+light efficiency. This can be done by setting a value different to one for the
+"scale_true_energy" argument present in the DataBinning Component of the configuration file.
```
morcuended, line 33: I suggest something like:
```
Modified IRFs with true energy scaled by a given factor can be created to evaluate the systematic uncertainty in the light collection efficiency. This can be done by setting a value different from one for the "scale_true_energy" argument present in the DataBinning Component of the configuration file of the IRF creation Tool.
```

morcuended (changes requested):
thanks a lot @deborahDOR, it looks good to me, I just left some comments regarding the description of the new energy scaling factor.

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -317,6 +320,14 @@ def start(self):
                 p["geomag_params"],
             ) = read_mc_dl2_to_QTable(p["file"])
 
+            
+            if self.data_bin.scale_true_energy != 1.0:
+                # This will scale the true energy of the MC events just before filling the IRFs histograms when pyirf commands are used.
+                # Effects expected: non-diagonal energy dispersion matrix and a different spectrum.
```
chaimain, line 326: Just for the sake of reference, can you show a comparative plot of the modified IRFs and Crab Nebula spectrum? Or refer to a presentation if already shown in the collaboration.

deborahDOR: They will be presented for the next LST-reco-call (10/06).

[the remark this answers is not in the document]
deborahDOR: The code does not change if we use this method because the only modification for the user is in the configuration file.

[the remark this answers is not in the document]
chaimain: You are correct. I guess I missed the part where we also add extra aliases for changing the config values from this Tool itself. For example  `"gh-efficiency": "DL3Cuts.gh_efficiency"` (L226 in the current file) and we can have an example usage of this Tool as shown in L129:L137 in the current file (`--gh-efficiency 0.95`).

So in this case, you may have something like `"scale-true-energy": "DataBinning.scale_true_energy"` added to the list of aliases. The value given in the command line for running this tool takes precedence over the values in the config and the default value.

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -527,7 +543,10 @@ def start(self):
             geomag_params["GEOMAG_DELTA"].to_value(u.deg),
             "deg",
         )
-
+        extra_headers["ETRUE SCALE"]= (
```
chaimain, line 546: "ETRUE SCALE" -> "ETRUE_SCALE"

deborahDOR: Yes. Sorry for that.

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -145,7 +145,13 @@ class IRFFITSWriter(Tool):
         --global-alpha-cut 10
         --source-dep
 
-    """
+    To build modified IRFs by specifying a scaling factor applying to the true energy (without using a config file):
+    > lstchain_create_irf_files
+        -g /path/to/DL2_MC_gamma_file.h5
+        -o /path/to/irf.fits.gz
+        --point-like (Only for point_like IRFs)
```
morcuended, line 152: if only valid for the point-like case I would make a check in the code, that we are in that case or raise an error otherwise

chaimain: I think it is meant only for a simple example. If so, you may exclude this flag/line

lstchain/tools/tests/test_tools.py
```diff
@@ -415,3 +414,101 @@ def test_index_srcdep_dl3_files(temp_dir_observed_srcdep_files):
...
+    """
+
+    from lstchain.tools.lstchain_create_irf_files import IRFFITSWriter
+    from gammapy.irf import EffectiveAreaTable2D, EnergyDispersion2D
+    import astropy.units as u
+
+    irf_file=temp_dir_observed_files / "fe_irf.fits.gz"
+    irf_file_mod = temp_dir_observed_files / "mod_irf.fits.gz"
+    config_file = os.path.join(os.getcwd(), "docs/examples/irf_dl3_tool_config.json")
+    config_file_mod = os.path.join(os.getcwd(), "docs/examples/irf_dl3_tool_config_mod.json")
```
morcuended, line 430: if the e_scale factor can be parsed through cmd line, I'd do that instead of using the file

docs/examples/irf_dl3_tool_config_mod.json
```diff
@@ -0,0 +1,52 @@
+{
+  "EventSelector": {
```
morcuended, line 2: I would not include this modified file in the repository

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -221,6 +236,7 @@ class IRFFITSWriter(Tool):
         "global-alpha-cut": "DL3Cuts.global_alpha_cut",
         "allowed-tels": "DL3Cuts.allowed_tels",
         "overwrite": "IRFFITSWriter.overwrite",
+        "scale_true_energy": "DataBinning.scale_true_energy"
```
chaimain, line 239: For the command-line argument, use `scale-true-energy`

chaimain: Also, update it in the above-added example

lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -149,7 +149,7 @@ class IRFFITSWriter(Tool):
     > lstchain_create_irf_files
         -g /path/to/DL2_MC_gamma_file.h5
         -o /path/to/irf.fits.gz
-        --scaling_true_energy 1.15
+        --scaling-true-energy 1.15
```
chaimain, line 152: typo

lstchain/tools/tests/test_tools.py
```diff
@@ -415,3 +414,106 @@ def test_index_srcdep_dl3_files(temp_dir_observed_srcdep_files):
...
+
+    aeff_hdu=EffectiveAreaTable2D.read(irf_file, hdu='EFFECTIVE AREA')
+    aeff_mod_hdu=EffectiveAreaTable2D.read(irf_file_mod, hdu='EFFECTIVE AREA')
+
+    edisp_hdu=EnergyDispersion2D.read(irf_file, hdu='ENERGY DISPERSION')
+    edisp_mod_hdu=EnergyDispersion2D.read(irf_file_mod, hdu='ENERGY DISPERSION')
+
+
+    #assert (aeff_mod.data.shape==(25,1))
+    #assert (edisp_mod.data.shape==(25,30,1))#(E,migra,offset)
```
chaimain, line 481: 1 small thing. Please remove the redundant comments. Everything else looks good

chaimain (approved):

lstchain/tools/tests/test_tools.py
```diff
@@ -476,9 +476,6 @@ def test_add_scale_true_energy_in_irfs(temp_dir_observed_files, simulated_dl2_fi
     edisp_hdu=EnergyDispersion2D.read(irf_file, hdu='ENERGY DISPERSION')
     edisp_mod_hdu=EnergyDispersion2D.read(irf_file_mod, hdu='ENERGY DISPERSION')
 
-
-    #assert (aeff_mod.data.shape==(25,1))
-    #assert (edisp_mod.data.shape==(25,30,1))#(E,migra,offset)
     assert(aeff_mod_hdu.data.shape==aeff_hdu.data.shape)
```
maxnoe, line 479: Assert is a statement, not a function in python, the parentheses are superfluous:

```
 assert aeff_mod_hdu.data.shape == aeff_hdu.data.shape
```

lstchain/tools/tests/test_tools.py
```diff
@@ -415,3 +414,103 @@ def test_index_srcdep_dl3_files(temp_dir_observed_srcdep_files):
             'EFFECTIVE AREA', 'ENERGY DISPERSION'
     ]:
         assert hdu_name in data.hdu_table['HDU_NAME']
+
+def test_add_scale_true_energy_in_irfs(temp_dir_observed_files, simulated_dl2_file):
+    """
+    Checking the validy of modified IRFs after scaling the True Energy by a factor.
```
morcuended, line 420:
```suggestion
    Checking the validity of modified IRFs after scaling the True Energy by a factor.
```

morcuended (approved):
Looks good! Thanks @deborahDOR and @chaimain

morcuended (approved):

chaimain (approved):

chaimain (approved):

morcuended (approved):

PR #1268, opened 2024-06-20
Add test for nsb waveform tuning script

morcuended:
Bug #1266 went unnoticed because the script was not covered by tests. Adding a basic test that runs the script.

### DIFF
--- a/lstchain/scripts/tests/test_lstchain_scripts.py
+++ b/lstchain/scripts/tests/test_lstchain_scripts.py
@@ -235,6 +235,11 @@ def test_validity_tune_nsb(tune_nsb):
 
 
 def test_validity_tune_nsb_waveform(tune_nsb_waveform):
+    """
+    The resulting nsb_tuning_ratio value of -1 expected in this test is meaningless
+    because the input data do not allow a full test of the functionality.
+    This test is only a formal check that the script runs.
+    """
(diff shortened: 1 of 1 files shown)


lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -213,6 +234,17 @@ def test_validity_tune_nsb(tune_nsb):
             assert line == '  "extra_noise_in_bright_pixels": 0.0'
 
 
+def test_validity_tune_nsb_waveform(tune_nsb_waveform):
+    output_lines = tune_nsb_waveform.stdout.splitlines()
+    for line in output_lines:
+        if '"nsb_tuning"' in line:
+            assert line == '  "nsb_tuning": true,'
+        if '"nsb_tuning_ratio"' in line:
+            assert line == '  "nsb_tuning_ratio": -1.0,'
```
moralejo, line 243: May be comment this result is meaningless because the input data do not allow a full test of the functionality (hence it is only a "formal" test that it does run)

morcuended: sure

morcuended: done

moralejo (approved):

PR #1270, opened 2024-06-25
Fix the shape of the interpolated Effective Area

### LINKED ISSUE #1269: interpolate_irfs reduces dimentions of EFFAREA IRF
`lstchain.high_level.interpolate_irfs()` reduces the dimensions of the returned effective area array, resulting in inconsistency between the specified energy-offset binning and array shape of the written IRF. To fix this, apparently, it is sufficient to replace

```python
    aeff_hdu_interp = create_aeff2d_hdu(
            effective_area=aeff_interp.T[0],
            ...
```
with

```python
    aeff_hdu_interp = create_aeff2d_hdu(
            effective_area=aeff_interp.T,
            ...
```
in [L521](https://github.com/cta-observatory/cta-lstchain/blob/v0.10.11/lstchain/high_level/interpolate.py#L521) of "interpolate.py".

The issue manifests itself in v0.10.11 and is not fixed in `master` so far.

chaimain:
Fixes #1269

### DIFF
--- a/lstchain/high_level/interpolate.py
+++ b/lstchain/high_level/interpolate.py
@@ -422,7 +422,7 @@ def interpolate_irf(irfs, data_pars, interp_method="linear"):
     main_headers = fits.open(irfs[0])[1].header
 
     point_like = main_headers["HDUCLAS3"] == "POINT-LIKE"
-    
+
     # Update headers to be added to the final IRFs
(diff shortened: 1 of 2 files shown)


morcuended (approved):
Thanks @chaimain. Looks good to me

PR #1272, opened 2024-07-03
Fix original MC nsb reading

### LINKED ISSUE #1271: Wrong base nsb level used in tuning at waveform level
The tuning at waveform level is done by extracting the relative NSB increase required and getting the original simulated NSB from a Header in simtel files.

Original implementation extracted all occurrences of 'NIGHTSKY_BACKGROUND' in yield_all_subobjects(EventIOFile(filename), [History, HistoryConfig]).

When tested on prod5_tran80 files the output was : 
[0.25307, 0.24586, 0.25307, 0.25307, 0.25307, 0.1, 0.12]
where index 1 was the value for LST-1, index 2 for LST-2, ... index 5 for MAGIC 1 and index 6 for MAGIC 2.
Index 0 was ignored and assumed to be some unused default value.

It was then correct to use tel_id to access the correct original NSB level.

This is not the case with the current MC production which has an additional value : [0.24586, 0.217, 0.24586, 0.1, 0.12].
Reported by @jstvdk

The preferred way to solve this issue would be to have a reliable way to find the NSB level effectively used for each telescope.

gabemery:
closes #1271 

Change the behaviour of `extract_simulation_nsb` to identify the nsb entries used with the telescopes and ignore other entries (for prod5 and LSTprod2). Return a dictionary to be accessed by tel_id.

Replace the access to simtel object by hand in favour of `SimTelFile.history`.

### DIFF
--- a/lstchain/io/io.py
+++ b/lstchain/io/io.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import re
 import warnings
 from multiprocessing import Pool
 from contextlib import ExitStack
@@ -22,9 +21,8 @@ from ctapipe.containers import SimulationConfigContainer
 from ctapipe.instrument import SubarrayDescription
 from ctapipe.io import HDF5TableReader, HDF5TableWriter
 
-from eventio import Histograms, EventIOFile
-from eventio.search_utils import yield_toplevel_of_type, yield_all_subobjects
-from eventio.simtel.objects import History, HistoryConfig
+from eventio import Histograms, SimTelFile
+from eventio.search_utils import yield_toplevel_of_type
 
 from pyirf.simulations import SimulatedEventsInfo
 
@@ -59,7 +57,6 @@ __all__ = [
     'global_metadata',
     'merge_dl2_runs',
     'merging_check',
-    'parse_cfg_bytestring',
     'read_data_dl2_to_QTable',
     'read_dl2_params',
     'read_mc_dl2_to_QTable',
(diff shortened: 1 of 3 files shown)


lstchain/io/io.py
```diff
@@ -1264,37 +1261,32 @@ def remove_duplicated_events(data):
...
+    """
+    nsb = {}
+    # In current MC, correct NSB are logged after 'STORE_PHOTOELECTRONS' entries
+    # TODO In any new production, behaviour needs to be verified.
+    # New version of simtel will allow to use better metadata
+    next_nsb = False
+    tel_id = 1
+    with SimTelFile(filename) as f:
+        try:
+            for _, line in f.history:
```
maxnoe, line 1278: This seems very brittle. Don't we have a better way to understand which option is the correct one to use other than the one that follow `STORE_PHOTOELECTRONS`?

gabemery: I don't see one. But I am not an expert of simtel so maybe I am missing something. As far as I know the information is only located here and there is no clear indication of its usage.

What I checked is that, for whatever reason, the nsb entries corresponding to the final config of each telescope follow this pattern for the two latest MC productions (in the files I checked). But not other entries.

I agree it is brittle and the use of the new metadata available in Simtel will be better once we use it for MC productions. Until then, we have to hope that the config used are somewhat consistent...

moralejo: I see no better option as of now. 
Perhaps you can add a clear warning log message stating what value will be used, so that the user can eventually check on the simtel config files that it is what it should be.

gabemery: Good idea, this info is currently given in a info level message without warning the user to check it. I will add a dedicated warning message.

moralejo:
Ok from my side! Unless @maxnoe has further comments, I am fine with this.

maxnoe:
Seems like this is the best we can do for now

lstchain/io/io.py
```diff
@@ -1264,37 +1261,32 @@ def remove_duplicated_events(data):
...
-                        if cfg_element[0] == 'NIGHTSKY_BACKGROUND':
-                            nsb.append(float(cfg_element[1].strip('all:')))
-                except Exception as e:
-                    print('Unexpected end of %s,\n caught exception %s', filename, e)
+    :return dict of `float` by tel_id: NSB rate
+    """
+    nsb = {}
+    # In current MC, correct NSB are logged after 'STORE_PHOTOELECTRONS' entries
+    # TODO In any new production, behaviour needs to be verified.
+    # New version of simtel will allow to use better metadata
```
morcuended, line 1273: Maybe this could be a Note in the function docstring instead of a comment here so it is visible e.g. in the documentation

morcuended (approved):
looks good, thx!

PR #1274, opened 2024-07-09
NSB tunning : pre-generated NSB only waveforms and a new class 

### LINKED ISSUE #1273: Speeding-up the NSB addition at waveform level
The generation of random NSB p.e.'s to be added to waveforms in the R0 to DL1 stage of MC is currently done "on the fly" for every event & pixel. That takes a significant part of the total cpu for the case of nigh NSB - for 16 * dark we saw like a factor 6 slower processing.

One could just generate, at the beginning of the r0_to_dl1 execution, a database to be kept in memory of a few thousand waveforms containing different realizations of the additional NSB, and just pick one at random from there for every pixel. In the proton simulations we have ~800 events, so if we just simulate, say, 10*1855 waveforms and use them shuffled, that would be a factor 80 less effort used in the noise waveform production.

gabemery:
In order to implement the possibility to pre-generate nsb waveforms to add nsb to events instead of generating new nsb waveforms on the fly, a new class was added.

The class `WaveformNsbTunner` take a new config file entry `pre_computed_multiplicity` which can be set to 0 to generate nsb on the fly. Or to any int to pre-generate `pre_computed_multiplicity` X number of pixels X number of gains waveforms.

Closes #1273 

To do before removing draft : 

- [x] Add docstrings to function and class
- [x] Pass CI 
- [x] Quick result consistency check

Some processing duration to illustrate the gain on a single (proton) MC file : 
+50% NSB, no pre-generation : 1min 39s
+50% NSB, 10X pre-generation : 1min 28s
+1000% NSB, no pre-generation : 3min 18s
+1000% NSB, 10X pre-generation : 1min 31s

Results of the tuning are compatible between using pre-generated waveforms or on the fly generation. Results are also the same as the previous implementation. See the signal less pixel charges for a data runs vs MC without and with tuning. 
image

Unrelated to this PR , but the injection of 55% nominal NSB here is a bit higher than automatically extracted using calculate_required_additional_nsb (~40%). This will need to be investigated independently.

### DIFF
--- a/lstchain/image/modifier.py
+++ b/lstchain/image/modifier.py
@@ -407,7 +407,6 @@ def calculate_noise_parameters(simtel_filename, data_dl1_filename,
 
 
 def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config=None):
-    # TODO check if good estimation
     # TODO reduce duplicated code with 'calculate_noise_parameters'
     """
     Calculates the additional NSB needed in the MC waveforms
@@ -538,8 +537,26 @@ def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config
 
 
 class WaveformNsbTunner:
+    """
+    Handles the injection of additional NSB pulses in waveforms.
+    """
     def __init__(self, added_nsb_fraction, original_nsb, pulse_template, charge_spe_cumulative_pdf,
                  pre_computed_multiplicity=10):
+        """
+        Parameters
+        ----------
+        added_nsb_fraction: `float`
+            Fraction of `original_nsb` to be added during NSB injection. Can be more than 1.
+        original_nsb: `dict`[`int`,`astropy.units.Quantity`]
+            NSB frequency used in the original waveforms per telescope
+        pulse_template: `dict`[int,`lstchain.data.NormalizedPulseTemplate`]
+            Single photo-electron pulse template per telescope
+        charge_spe_cumulative_pdf: `scipy.interpolate.interp1d`
+            Cumulative amplitude probability density function for single photo-electrons
+        pre_computed_multiplicity: `int`
+            Multiplicative factor on the number of pixels used to determine the number of pre-generated, nsb-only,
+            waveforms. Later used during event modification. Set to 0 to always compute the correction on the fly.
+        """
         self.added_nsb_fraction = added_nsb_fraction
         self.original_nsb = original_nsb
         self.pulse_template = pulse_template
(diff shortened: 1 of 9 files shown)


vuillaut:
Hi Grabiel, all,

Won't this negatively affect image-based methods that do not use cleaning?

gabemery:
Hi @vuillaut. You mean that using correlated nsb injection could lead to issues? Or are you speaking of something else?

vuillaut:
> Hi @vuillaut. You mean that using correlated nsb injection could lead to issues? Or are you speaking of something else?
> 

Hi.
Yes exactly. I understand that after cleaning, most of the injected nsb will anyway be removed so re-using nsb waveforms should not be a problem but I am afraid it won't be the case for methods using uncleaned images.

gabemery:
It also concerned me, but with the current implementation I think it should not be a massive issue because:

- Assuming 10X pre-generation, I have 18550 simulated waveform of which I use 1855 selected in a random order. So each full injection will be likely unique in terms of 1855 selected waveforms. 
- Injection is done on already existing r1, so the resulting waveform is the sum of the cleanly simulated simtel waveform and my injection. So they all are unique. Of course if the injected NSB is significantly higher than the original, the resulting injection will be dominated by the new waveforms.
- All of this is done file-wise. So a new generation of nsb waveform is done every O(1000) events
- This is adding stochastic noise on top of signal without changing the baseline, so there should be no specific pattern with useful information for machine learning algorithm?

lstchain/image/modifier.py
```diff
@@ -577,3 +537,160 @@ def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config
...
+        waveform: ndarray
+            Waveform used to know the number of pixels and time samples for a given telescope id
+        dt: `astropy.units.Quantity`
+            Time between samples
+        tel_id: `int`
+            Telescope identifier for which nsb waveforms are generated
+
+        """
+        log.info(f"Pre-generating nsb waveforms for nsb tuning and telescope id {tel_id}.")
+        n = 25
```
morcuended, line 593: this n = 25 is defined several times, does it make sense to define it once at the beginning of the file?

morcuended (commented):
thanks @gabemery, I left some comments

lstchain/tests/test_lstchain.py
```diff
@@ -130,6 +130,9 @@ def test_r0_to_dl1_lhfit_mc(tmp_path, mc_gamma_testfile):
     config['lh_fit_config']["use_interleaved"] = True
     config['waveform_nsb_tuning']['nsb_tuning'] = True
     r0_to_dl1(mc_gamma_testfile, custom_config=config, output_filename=tmp_path / "tmp.h5")
+    os.remove(tmp_path / "tmp.h5")
+    config['waveform_nsb_tuning']['pre_computed_multiplicity'] = 0
```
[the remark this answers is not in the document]
gabemery: It defaults to 10 from line 432 in r0_to_dl1.py (and in the class, but this is not used) and is not defined in the standard config file.
I should probably add it there too as being 10 for consistency. (Or make it None in the default config file and change the code to support it)
And in this test it may be clearer if I explicitly select its value as 10 to avoid confusion.

lstchain/image/modifier.py
gabemery: It does, I will change it.

moralejo (approved):

PR #1278, opened 2024-07-16
Update test-data url

maxnoe:
We had technical issues in Dortmund with the server hosting the lst test data.

I moved them to another server, update the URLs here.

It would really be great if we could find an LST solution for this.


morcuended:
I think we should use CNAF or CSCS for this

morcuended (approved):

PR #1282, opened 2024-07-31
Waveform noise tuner

moralejo:
Code to compute the needed NSB tuning in the MC waveforms to match a certain data file.
It does not require to know the level of NSB simulated, or the level of electronic noise. It obtains the necessary noise addition in a purely empirical way.

### DIFF
--- a/lstchain/image/modifier.py
+++ b/lstchain/image/modifier.py
@@ -411,7 +411,26 @@ def calculate_noise_parameters(simtel_filename, data_dl1_filename,
 
 def get_pix_median_charges(data_dl1_filename, event_type):
     """
+    Obtain from a DL1 real data file (containing camera images, i.e. DL1a)
+    the median (accross camera) of mean (and std dev) of pixel charge for
+    events of type event_type. The medians are obtained using only healthy
+    pixels and excluding too bright outliers (we want to get the typical
+    pixel values, unaffected by the few stars in the FoV)
+
+    Parameters
+    ----------
+    data_dl1_filename: DL1 file name
+    event_type (ctapipe.containers.EventType): type of events, e.g. pedestals
+
+    Returns
+    -------
+    median_ped_meanpixq (p.e.)
+    median_ped_stdpixq  (p.e.)  Median values for healthy and not too
+    noisy pixels
+    tel_id : telescope id to which the events in the file correspond (we assume
+    for now that a given DL1 file contains data from just one telescope)
     """
+
     # Real data DL1 tables:
     camera_images = read_table(data_dl1_filename,
                                '/dl1/event/telescope/image/LST_LSTCam')
@@ -421,7 +440,7 @@ def get_pix_median_charges(data_dl1_filename, event_type):
 
     # All events in a DL1 file should correspond to the same telescope,
     # just take it from the first event:
-    # tel_id = image_params['tel_id'][0]
+    tel_id = image_params['tel_id'][0]
 
     data_dl1_calibration = read_table(data_dl1_filename,
                                       '/dl1/event/telescope/monitoring/calibration')
@@ -454,7 +473,7 @@ def get_pix_median_charges(data_dl1_filename, event_type):
     median_ped_meanpixq = np.median(ped_meanpixq[good_pixels])
     median_ped_stdpixq = np.median(ped_stdpixq[good_pixels])
 
-    return median_ped_meanpixq, median_ped_stdpixq
+    return median_ped_meanpixq, median_ped_stdpixq, tel_id
 
 
 
@@ -491,7 +510,10 @@ def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config
     if config is None:
         config = standard_config
 
-    median_ped_meanpixq, median_ped_stdpixq = get_pix_median_charges(
+    # Obtain the median (accross camera) of mean (and satd dev of) pixel charge
+    # taking into account only healthy pixels, and excluding outliers.
+    # We also get the tel_id to which the DL1 file corresponds:
+    median_ped_meanpixq, median_ped_stdpixq, tel_id = get_pix_median_charges(
             data_dl1_filename, EventType.SKY_PEDESTAL)
 
     # Now we process the Monte Carlo:
--- a/lstchain/image/modifier.py
+++ b/lstchain/image/modifier.py
@@ -530,12 +530,12 @@ def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config
 
 
     for nsb_rate in np.diff(total_added_nsb):
-        # Create an array to put the NSB value for tel_id. It is an array for
-        # the future case in which there are more telescopes, but for now we
-        # just create it with a single value for tel_id
-        nsb = np.zeros(tel_id+1)
-        nsb[tel_id] = nsb_rate
-        nsb_tuner.append(WaveformNsbTunner(nsb * u.GHz, pulse_templates,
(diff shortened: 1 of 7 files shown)


lstchain/image/modifier.py
```diff
@@ -405,9 +408,7 @@ def calculate_noise_parameters(simtel_filename, data_dl1_filename,
     return extra_noise_in_dim_pixels, extra_bias_in_dim_pixels, \
         extra_noise_in_bright_pixels
 
-
 def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config=None):
```
morcuended, line 411: this function is quite big, I suggest extracting some parts with a given functionality in additional smaller functions

moralejo: Ok, I separated the real data part, which indeed will be useful to identify which (NSB-tuned) MC is needed to process a given data run

lstchain/scripts/lstchain_tune_nsb_waveform.py
```diff
@@ -82,13 +82,13 @@ def main():
...
 
-    nsb_correction_ratio, data_ped_variance, mc_ped_variance = \
+    extra_nsb_rate, data_ped_variance, mc_ped_variance = \
         calculate_required_additional_nsb(args.input_mc, args.input_data,
                                           config=Config(config))
 
     dict_nsb = {
         "nsb_tuning": True,
-        "nsb_tuning_ratio": np.round(nsb_correction_ratio, decimals=2),
+        "nsb_tuning_rate": np.round(extra_nsb_rate, decimals=3),
```
maxnoe, line 91: didn't you rename to `nsb_tuning_rate_GHz`?

moralejo: Yes, thanks!  I had just realized... 
Don't waste time with it yet, still WIP...

lstchain/reco/r0_to_dl1.py
```diff
@@ -411,7 +411,8 @@ def r0_to_dl1(
                 pulse_templates = {tel_id: NormalizedPulseTemplate.load_from_eventsource(
                     subarray.tel[tel_id].camera.readout, resample=True)
                     for tel_id in config['source_config']['LSTEventSource']['allowed_tels']}
-                if 'nsb_tuning_rate' in config['waveform_nsb_tuning'].keys():
+                if 'nsb_tuning_rate_GHz' in config[
+                    'waveform_nsb_tuning'].keys():
```
maxnoe, line 415: .keys() is superfluous

moralejo: There are also other instances, will take care

moralejo:
An example of tuning performed with this code, between 
`/fefs/aswg/data/real/DL1/20240627/v0.10/tailcut84/dl1_LST-1.Run17881.0030.h5`
and
`/fefs/aswg/data/mc/DL0/LSTProd2/TrainingDataset/GammaDiffuse/dec_2276/sim_telarray/node_corsika_theta_6.000_az_180.000_/output_v1.4/simtel_corsika_theta_6.000_az_180.000_run10.simtel.gz
`
image

moralejo:
The code is not completely deterministic, since it applies the random noise (different levels) to compare it to the one in the data. I do not think this is a problem, but I guess @gabemery that one could fix all seeds to make it 100% reproducible

moralejo:
And this is with the lstchain test files (only one ped event in the DL1 file, which is enough to roughly match the NSB noise):
image

So a meaningful test of the algo can be done with this new approach.
Unfortunately, the result of the tuning in the CI tests is completely different to what I get when I run it myself (hence the CI fails). Instead of ~3.2p.e./ns its gives ~2.5 p.e./ns  (this is well beyond the fluctuations from the randomness of the algorithm). I have no idea why, any suggestions?

morcuended (changes requested):
No major comments, just some minor questions and suggestions

lstchain/image/modifier.py
```diff
@@ -406,8 +409,75 @@ def calculate_noise_parameters(simtel_filename, data_dl1_filename,
...
+    tel_id = image_params['tel_id'][0]
+
+    data_dl1_calibration = read_table(data_dl1_filename,
+                                      '/dl1/event/telescope/monitoring/calibration')
+    unusable = data_dl1_calibration['unusable_pixels']
+
+    # Locate pixels with HG declared unusable either in original calibration or
+    # in interleaved events:
+    bad_pixels = unusable[0][0]  # original calibration
+    for tf in unusable[1:][0]:  # calibrations with interleaved
```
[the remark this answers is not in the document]
moralejo: just "true/false" :D  Since it is only used in the next line I don't think a more meaningful name is needed.

lstchain/image/modifier.py
```diff
@@ -439,119 +510,167 @@ def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config
...
+    # that extra_nsb is negative, but just in case:
+    if extra_nsb < 0:
+        extra_nsb = 0
+
+    # Now open the MC file again and test that the tuning was successful:
+    mc_reader = EventSource(input_url=simtel_filename, config=Config(config))
+    nsb = {tel_id: extra_nsb * u.GHz for tel_id in
+           config['source_config']['LSTEventSource']['allowed_tels']}
+    tuner = WaveformNsbTunner(nsb, pulse_templates,
+                              charge_spe_cumulative_pdf, 10)
```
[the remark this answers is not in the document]
moralejo: I prefer the former, lest the default change

lstchain/image/modifier.py
```diff
@@ -439,119 +510,167 @@ def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config
...
+
+    numevents = 0
+    maxmcevents = 200 # Enough statistics to determine the right NSB level
+
+    # Simulated levels of additional NSB, rate in p.e./ns (a.k.a. GHz):
+    total_added_nsb = np.array([0, 0.125, 0.25, 0.5, 1, 2, 4])
+
+    # Just the product of added_nsb and original_nsb (first two arguments
+    # below) is relevant! Units "GHz" (p.e./ns)
+    nsb_tuner = [None]
```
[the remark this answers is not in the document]
moralejo: The first element, which is what I initialize here, corresponds to no NSB tuning. Could be done in other ways, but this one is ok I think.

lstchain/image/modifier.py
```diff
@@ -439,119 +510,167 @@ def calculate_required_additional_nsb(simtel_filename, data_dl1_filename, config
...
-    mc_ped_charges = []
+    config['LocalPeakWindowSum']['apply_integration_correction'] = True
+    r1_dl1_calibrator = CameraCalibrator(image_extractor_type=config[
+        'image_extractor'], config=Config(config), subarray=subarray)
+
+    numevents = 0
+    maxmcevents = 200 # Enough statistics to determine the right NSB level
+
+    # Simulated levels of additional NSB, rate in p.e./ns (a.k.a. GHz):
+    total_added_nsb = np.array([0, 0.125, 0.25, 0.5, 1, 2, 4])
```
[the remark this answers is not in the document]
moralejo: No, it is "total" as opposed to "incremental" because the added noise accumulates on the waveforms (which is good, it saves computing time here). It does not refer to "all the levels" we are testing.

morcuended (commented):
no more comments from my side, but better let's @gabemery have a look

morcuended (approved):

PR #1285, opened 2024-09-04
Better warning message in R0 to R0G

moralejo:
The previous message suggested an action which is not actually needed in most cases.

### DIFF
--- a/lstchain/scripts/lstchain_r0_to_r0g.py
+++ b/lstchain/scripts/lstchain_r0_to_r0g.py
@@ -188,8 +188,8 @@ def main():
         log.info('FF-like events tagged as FF: %d', 
                  num_FF_like_with_FF_type)
     else:
-        log.warn('FF-like events tagged as FF: %d !!', 
-                 num_FF_like_with_FF_type)
+        log.warning('FF-like events tagged as FF: %d !!', 
+                    num_FF_like_with_FF_type)
 
     log.info('FF-like events not tagged as FF: %d', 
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_r0_to_r0g.py
```diff
@@ -203,13 +203,15 @@ def main():
...
-              log.warn('You should use heuristic identification of FF events!')
-      else:
-          log.info('R0 to R0G conversion finished successfully!')
+        frac_untagged_ff_like = num_FF_like_with_no_FF_type / num_FF_like
+        if frac_untagged_ff_like > max_frac:
+            log.warn('%d percent of FlatField(FF)-like events are not tagged as FF', 
+                     int(100*frac_untagged_ff_like))
+            log.warn('This may be due to anomalous events, like car flashes or LIDAR shots')
+            log.warn('More rarely, it could result from mis-tagging of FF events')
+            log.warn('A thorough inspection of the data check plots is recommended\n')
```
morcuended, line 212:
```suggestion
            log.warning('%d percent of FlatField(FF)-like events are not tagged as FF', 
                     int(100*frac_untagged_ff_like))
            log.warning('This may be due to anomalous events, like car flashes or LIDAR shots')
            log.warning('More rarely, it could result from mis-tagging of FF events')
            log.warning('A thorough inspection of the data check plots is recommended\n')
```

morcuended (changes requested):

morcuended (approved):

PR #1291, opened 2024-09-17
Replace test file

maxnoe:
New file is not yet on the data server, but will be there tomorrow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,8 @@ __pycache__
 */cython_version.py
 htmlcov
 .coverage
+.coverage.*
+coverage.xml
 MANIFEST
 
 # Sphinx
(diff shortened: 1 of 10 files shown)


morcuended:
@maxnoe is the new test file already in the data server?

maxnoe:
No, not yet. @kosack had some issues connecting to the server, but I think they are solved now.

But he's on vacation until the 4th

morcuended:
@maxnoe, @kosack, a friendly reminder for the replacement of the test file.

maxnoe:
Test file is now there, let's see

morcuended:
Maybe add a test for r0 to dl1 script running without input argument to make sure it works in the future.

morcuended:
Added a test to check that running the script without input argument also works. I think this is good to go.

morcuended (approved):

PR #1294, opened 2024-09-19
Reduce number of trees in RF regressors

moralejo:
Reduced to 150 to 50 trees, physics performance is not impacted, while memory needs will decrease (and speed increase)


lstchain/data/lstchain_standard_config.json
```diff
@@ -73,7 +73,7 @@
     "max_depth": 30,
     "min_samples_leaf": 10,
     "n_jobs": -1,
-    "n_estimators": 150,
+    "n_estimators": 50,
```
morcuended, line 76: to be changed also in `lstchain_lhfit_config.json`?

morcuended (commented):
I think it would be good to upload the comparison in the wiki for bookkeeping

lstchain/data/lstchain_standard_config.json
moralejo: > to be changed also in `lstchain_lhfit_config.json`?

I guess so, but it is a question to @gabemery I guess

moralejo:
Some comparisons here: [RF_tree_reduction.pdf](https://github.com/user-attachments/files/17056563/RF_tree_reduction.pdf)

vuillaut (commented):
why not changing the classifiers parameters as well?

moralejo:
> why not changing the classifiers parameters as well?

The tests for that (since they involve background, ideally from real data) are more complicated. 
The main goal of this is to reduce the memory footprint of the RF training & application, and the biggest RFs are those for regression. Currently it is 67% for regression, 37% for classification (according to models' .sav files). With this change the total RF size will become ~55% of what it is now, with a share of 60% for classification, 40% for regression. If we aim for a further reduction we should indeed target the classifiers.

lstchain/data/lstchain_standard_config.json
gabemery: I don't see a reason not to. Especially if there is no effect for the standard analysis before n_estimator = 5. I don't expect a simple different parametrization to significantly change the behaviour vs number of estimator.

@moralejo Can you share the code you used to remove estimators in already trained RFs? So that I can test it later?

moralejo: > I don't see a reason not to. Especially if there is no effect for the standard analysis before n_estimator = 5. I don't expect a simple different parametrization to significantly change the behaviour vs number of estimator.
> 
> @moralejo Can you share the code you used to remove estimators in already trained RFs? So that I can test it later?

Apologies, I had not read this yet. I will contact you @gabemery

rlopezcoto (approved):

morcuended:
there is something wrong with mamba setup step in the CI (is taking > 50 min)

gabemery (approved):

moralejo:
> there is something wrong with mamba setup step in the CI (is taking > 50 min)

Can we ignore it?

morcuended:
Tests will not run either until we merge #4563 or the problem with mamba is solved. Anyway, I think the changes here are not tested. I'd say it can be merged

Issue #1293, opened 2024-09-18
Telescope pointing from first event may be badly mistaken


moralejo:
https://github.com/cta-observatory/cta-lstchain/blob/e4a5b7a36dfd643acf47b9006f51fe882b6dbc5c/lstchain/high_level/hdu_table.py#L270-L272

@juanjq  has seen some instance of the telescope pointing (RA,dec) for the first event in a DL2 file being quite off from the true pointing for (most of) the run. Example: run 17821. This is the only pointing stored in the DL3 file for the whole run, as RA_PNT, DEC_PNT

We think this has consequences for the ring background model: the model for such runs is badly mistaken, probably because symmetry is assumed around a wrong sky position.

morcuended:
Are these pointing coordinates from the first event internally used in Gammapy for ring background calculation? Or is it whatever coordinates you define? What pointing information do you suggest including the HDU table?

maxnoe:
@moralejo Do you understand why this happens? Is the timestamp wrong or the interpolation or the drive log?

juanjq:
In principle it seems that gammapy assumes symmetry around pointing, that is written in the DL3 and it does not need to be input. And @maxnoe I don't know 100%, but I don't think the issue is related to the dragon timestamps, we think this happens because sometimes when wobbling, LST can oscillate in the first instants of a run, resulting in a error of few 0.1 deg approx, then stabilizes. (Or this is what it seems looking the specific set of runs I'm working with: 17820-17825, in most of them a small oscillation in pointing can be seen at the start of the run) So then taking the first pointing as general is not that accurate. So I don't know which option can be better but the mean position can be taken, or maybe excluding start and end of the run, I guess there is not a unique solution for that.

SeiyaNozaki:
I asked Armand before and it seems this is a real feature. You can also see this mispointing at the beginning of each tracking in the drive log. I had similar issues for source-dep analysis (fixed in #1211), so I agree with the suggestion by @juanjq (not take pointing info excluding the start of run)
https://www.lst1.iac.es/operations/drive/Log_cmd.2024-06-15.html

morcuended:
Indeed when checking the center of off regions in Gammapy we can see this dispersion. Agree as well with @juanjq proposal of excluding the start/end and then taking the average

maxnoe:
[...]

PR #1302, opened 2024-10-08
Computing the run pointing in sky coordinates

moralejo:
Addresses #1293 
So far the first event's pointing was used. This is problematic for runs in which the pointing is unstable at the beginning.

### DIFF
--- a/lstchain/high_level/hdu_table.py
+++ b/lstchain/high_level/hdu_table.py
@@ -264,7 +264,7 @@ def get_pointing_params(data, source_pos, time_utc, exclude_fraction=0.2):
     and average them for the run (excluding the first events, for which there
     may be some mispointing due to not-yet-stable tracking).
 
-    exclude_fraction: fraction of the run that will me excluded from the 
+    exclude_fraction: fraction of the events that will be excluded from the 
     averaging (the excluded events are the ones at the beginning of the run)
 
     Note: The angular difference from the source is just used for logging here.
--- a/lstchain/high_level/hdu_table.py
+++ b/lstchain/high_level/hdu_table.py
@@ -297,7 +297,7 @@ def get_pointing_params(data, source_pos, time_utc, exclude_fraction=0.2):
     log.info("Median of angle between pointing direction and mean pointing:")
     log.info(f"  All events: {median_angle_wrt_mean_pointing:.3f}")
     if median_angle_wrt_mean_pointing > max_median_angle:
-        log.warn(f"   ==> The telescope pointing seems to be unstable during the Run!")
+        log.warn("   ==> The telescope pointing seems to be unstable during the Run!")
     log.info(f"  Excluding the first {100*exclude_fraction:.1f}% of events: {median_angle_wrt_mean_pointing_2:.3f}")
 
     source_pointing_diff = source_pos.separation(pnt_icrs)
--- a/lstchain/high_level/hdu_table.py
+++ b/lstchain/high_level/hdu_table.py
@@ -297,7 +297,7 @@ def get_pointing_params(data, source_pos, time_utc, exclude_fraction=0.2):
     log.info("Median of angle between pointing direction and mean pointing:")
     log.info(f"  All events: {median_angle_wrt_mean_pointing:.3f}")
     if median_angle_wrt_mean_pointing > max_median_angle:
-        log.warn("   ==> The telescope pointing seems to be unstable during the Run!")
+        log.warning("   ==> The telescope pointing seems to be unstable during the Run!")
     log.info(f"  Excluding the first {100*exclude_fraction:.1f}% of events: {median_angle_wrt_mean_pointing_2:.3f}")
 
     source_pointing_diff = source_pos.separation(pnt_icrs)
--- a/lstchain/high_level/hdu_table.py
+++ b/lstchain/high_level/hdu_table.py
@@ -275,23 +275,35 @@ def get_pointing_params(data, source_pos, time_utc, exclude_fraction=0.2):
     pointing_alt = data["pointing_alt"]
     pointing_az = data["pointing_az"]
 
+    max_median_angle = 0.05 * u.deg # Is this a reasonable limit?
+    # If median of the angle between the pointing directions and their mean in the run
+    # is larger than max_median_angle a warning will be displayed about unstable pointing.
+    
     first_event = int(exclude_fraction * len(data))
     with erfa_astrom.set(ErfaAstromInterpolator(300 * u.s)):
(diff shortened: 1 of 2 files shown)


moralejo:
Possible improvements:

- check stability of pointing (and at least display a warning)
- return the "first reliable event" and use it to omit the previous ones from the DL3 file?

moralejo:
I hope it is ok to do the change in docs/conf.py in this PR
That was just a deprecated option which made docs building fail.

lstchain/high_level/hdu_table.py
```diff
@@ -257,21 +258,35 @@ def get_timing_params(data, epoch=LST_EPOCH):
...
-        alt=pointing_alt[0],
-        az=pointing_az[0],
-        frame=AltAz(obstime=time_utc[0], location=LST1_LOCATION),
-    ).transform_to(frame="icrs")
+    first_event = int(exclude_fraction * len(data))
+    with erfa_astrom.set(ErfaAstromInterpolator(300 * u.s)):
+        evtwise_pnt_icrs = SkyCoord(
+            alt=pointing_alt[first_event:],
+            az=pointing_az[first_event:],
+            frame=AltAz(obstime=time_utc[first_event:], location=LST1_LOCATION),
```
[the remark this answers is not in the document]
maxnoe: Not for Cherenkov light produced in 10 to 20 km height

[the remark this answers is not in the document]
moralejo: Here we want to know where the gamma-ray source "really is",  ¿right?,  rather than where its optical image (if any) would appear on the camera. Any bias in g-ray direction reconstruction related to refraction must be addressed via simulations (i.e. by training algorithms with MC which implements C-light refraction) .

This said, I had not even thought about that... so good that you ask (and good that the default is P=0)

moralejo:
> I hope it is ok to do the change in docs/conf.py in this PR That was just a deprecated option which made docs building fail.

Done now in #1303

morcuended:
Something is not okay with the downloading of the test dataset in the CI

moralejo:
One unrelated test failed. Trying to fix it in #1307

lstchain/high_level/hdu_table.py
```diff
@@ -257,26 +258,52 @@ def get_timing_params(data, epoch=LST_EPOCH):
     return time_pars, time_utc
 
 
-def get_pointing_params(data, source_pos, time_utc):
+def get_pointing_params(data, source_pos, time_utc, exclude_fraction=0.2):
```
morcuended, line 261: I propose to change the name of the function to something more descriptive like `get_average_pointing_coords`

moralejo: It is not clear it will always be the average. Eventually we will have a table with timestamps and pointings. We may also switch to the median at some point...

moralejo: I prefer not to change it right now for that reason

morcuended (dismissed):
Looks good to me

lstchain/high_level/hdu_table.py
```diff
@@ -257,26 +258,52 @@ def get_timing_params(data, epoch=LST_EPOCH):
...
+        ).transform_to(frame="icrs")
+
+    mean_ra = circmean(evtwise_pnt_icrs[first_event:].ra)
+    mean_dec = np.mean(evtwise_pnt_icrs[first_event:].dec)
+    pnt_icrs = SkyCoord(ra=mean_ra, dec=mean_dec, frame="icrs")
+
+    median_angle_wrt_mean_pointing = np.median(pnt_icrs.separation(evtwise_pnt_icrs))
+    median_angle_wrt_mean_pointing_2 = np.median(pnt_icrs.separation(evtwise_pnt_icrs[first_event:]))
+    
+    log.info("Median of angle between pointing direction and mean pointing:")
```
maxnoe, line 297: This compares the median over all events to the median over the events without the ones excluded by "fraction".

Is this really what we want? Normal runs are long (~20 min) and the unstable pointing we saw was for the first ~35 seconds in the bad cases. 

So the median won't be affected by the the first 35 seconds of unstable pointing and then stable pointing.

We looked at the deviation of the individual pointings from the "good pointing" in chunks of 2 seconds.

mexanick: If we want to detect unstable outlier values at the beginning, why don't we just use a sigma clipping routine from scipy or astropy?

moralejo: > This compares the median over all events to the median over the events without the ones excluded by "fraction".
> 
> Is this really what we want? Normal runs are long (~20 min) and the unstable pointing we saw was for the first ~35 seconds in the bad cases.
> 
> So the median won't be affected by the the first 35 seconds of unstable pointing and then stable pointing.
> 
> We looked at the deviation of the individual pointings from the "good pointing" in chunks of 2 seconds.

Right, this will not be sensitive to the instabilities if they are so small... Perhaps it makes more sense to simply write out in the log what fraction of the run has a pointing beyond, say, 0.05 deg of the target. But I do not know where it would make sense to set a threshold for a warning.

moralejo: > If we want to detect unstable outlier values at the beginning, why don't we just use a sigma clipping routine from scipy or astropy?

For simplicity, we know the problem occurs only at the beginning of the run. At least the problem we are addressing in this PR. And sigma clipping requires some settings, not obvious to me which ones would be adequate here.

morcuended (dismissed):
It looks good to me as it is. Just stress @maxnoe comment https://github.com/cta-observatory/cta-lstchain/issues/1293#issuecomment-2360582874:

> Maybe we should identify the time range of this unstable pointing and exclude it via the GTI.

So in the long run we want to exclude completely from the analysis those first seconds of the runs via GTI definitions (if there is unstable pointing)

lstchain/high_level/hdu_table.py
```diff
@@ -257,26 +258,52 @@ def get_timing_params(data, epoch=LST_EPOCH):
...
-    pnt_icrs = SkyCoord(
-        alt=pointing_alt[0],
-        az=pointing_az[0],
-        frame=AltAz(obstime=time_utc[0], location=LST1_LOCATION),
-    ).transform_to(frame="icrs")
+    max_off_angle = 0.05 * u.deg 
+    # Keep track of how often pointing is beyond max_off_angle w.r.t. the median
+    # run pointing
+    
+    # If median of the angle between the pointing directions and their mean in the run
```
maxnoe, line 282: This comment no longer applies

morcuended (approved):

maxnoe (approved):

PR #1305, opened 2024-10-08
Update FF charge limits in data check

moralejo:


morcuended (approved):
If you want to add a brief comment describing why this change in the expected charge, it is welcome. Otherwise, fine with me as it is.

PR #1311, opened 2024-10-10
Support Python 3.12

morcuended:


maxnoe:
Using 3.12 would require updating the dependencies on at least gammapy and matplotlib

morcuended:
This needs a bit of work and testing. I'll leave it here for the future. The current scikit-learn pinned version is not compatible with Python 3.12 either.

maxnoe:
I think we can close this in favor of #1361 ?

PR #1313, opened 2024-10-27
Notebook for quick simulation of LST1 observations

moralejo:
The telescope performance is stored in the csv files The background is obtained from good-quality real data, the gamma IRF from MC


jsitarek:
in the "pulsar mode" field it would be safer to put the overwritting of alpha and printing of it into ```if pulsar_mode:``` clause, otherwise somebody can by mistake overwrite alpha despite having the pulsar_mode set to false

jsitarek:
in the integrate function, why not passing the dFdE function as a parameter and computing dfde1 and dfde2 inside the integrate function?

jsitarek:
there is a typo: "jus trplace"

jsitarek:
I am not sure if it is a good idea to randomize the position of the true energies when you calculate the energy migration. If I understand the code correctly this means that from one execution to another you will have a different number of times you sample a given true energy bin, and this will shift the events back and forth. Maybe best to have the fine binning in Etrue uniform and only randomize (or also use constant fine binning) in the migration factor

jsitarek:
As the final plot of the notebook will likely end up in many of the proposals, I would recommend to made some changes to it to unify how they look like:
increase the marker size (the middle range points for a strong source are not visible)
add some basic information that were used during the calculation into the legend of the plot (observation time, zenith angle, maybe also if pulsar mode or extension were defined)
add somewhere in the corner of the figure the version number of the macro

notebooks/LST1_observation_simulator.ipynb
```diff
@@ -0,0 +1,965 @@
+{
```
morcuended, line 1: <p>I changed the way the background and gamma rate files are read. The initial implementation is not working unless you define in advance LSTCHAIN variable. In this way, that is unnecessary, simply using the functionality lstchain already has.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/1313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

morcuended, line 1: <p>This way of looking for the GAMMAPY_DATA folder in case it is not previously defined is not working for me. I'd say it would work only if you have downloaded the dataset somewhere even if not export the path.</p><p> </p><p>I'd say they way to go is that the user set this up in advance (put some instruction to do so). If running in the container we could simply export the path to a common GAMMAPY_DATASET path.</p><p> </p><p>If running locally I see no way out to downloading and setting GAMMAPY_DATA. Or maybe simply download the <code>ebl_dominguez11.fits.gz</code> and read it, instead of using the read_builtin method.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/1313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

moralejo: Thanks, I will do this. Since we anyway copy the ebl file, I see no need to check if GAMMAPY_DATA is set or not.

moralejo:
> in the integrate function, why not passing the dFdE function as a parameter and computing dfde1 and dfde2 inside the integrate function?

Because in this way it is obvious, at the very moment you call the function, that the integral is some sort of approximation (specifically, a power-law).  I really see no advantage in passing the function.

moralejo:
> As the final plot of the notebook will likely end up in many of the proposals, I would recommend to made some changes to it to unify how they look like: increase the marker size (the middle range points for a strong source are not visible) add some basic information that were used during the calculation into the legend of the plot (observation time, zenith angle, maybe also if pulsar mode or extension were defined) add somewhere in the corner of the figure the version number of the macro

Done:

image

notebooks/LST1_observation_simulator.ipynb
```diff
@@ -0,0 +1,1013 @@
...
```
morcuended, line 1: <pre class="ql-syntax" spellcheck="false">Line #14.    from lstchain import version
</pre><p>maybe more explicit: import lstchain, then later lstchain.__version__</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/1313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

morcuended, line 1: <pre class="ql-syntax" spellcheck="false">Line #21.    #     ebl_file = subprocess.run(['find', str(gammapy_dir), '-name', 'ebl_dominguez11.fits.gz'], 
</pre><p>I think this can be removed, and simply leave a comment saying that if you have GAMMAPY_DATA env variable set, you can also use the builtin models. But this find subprocess will probably not work for all users</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/cta-lstchain/pull/1313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

morcuended (approved):
I left some last minor comments, but overall it looks good.

notebooks/LST1_observation_simulator.ipynb
moralejo: Added some comments. I left there the commented code, since it works at least in some cases.

moralejo: Changed to
from lstchain import version as lstchain_version

morcuended (approved):

PR #1316, opened 2024-11-14
Close EventSource instances

moralejo:
Avoid a lot of zombie processes when using these functions repeatedly


maxnoe:
better directly use the context manager:

```
with EventSource(...) as mc_reader:
    ...
```

gabemery (commented):
I agree with this change, but also with the comment by maxnoe. So I will wait for its implementation before approving the PR.

moralejo:
I hope I didn't mess ups with the indents... I wonder if it was really worth. I will check again later, back to Draft

morcuended (approved):

Issue #1317, opened 2024-11-22
Make RF response more homogeneous vs. zenith


moralejo:
We know that at high zeniths (say, above ~50 deg) the output of the random forests depends strongly on zenith (because image parameters change quickly with zenith).

Since the RFs are trained on MC with a discrete distribution of pointings, when applied to the data, the distributions of reconstructed quantities (e.g. gammaness or energy) show sudden jumps when the telescope pointing crosses the middle point in zenith between two training nodes (sin_azimuth is also part of the training, but luckily its effect is negligible compared to that of zenith). In the past, for the analysis of some high-zenith observations, we have dealt with this problem just using a different training sample, with smaller steps in zenith.

A possible general solution (that would allow us to use our existing high-zenith "coarse-grid" training MC) could be the following: 

- for any given event, compute the RF output with the standard set of parameters
- compute it again with the same parameters except for zenith (taking a value just on the other side of the closest zenith boundary between training nodes)
- Each of the two values would correspond to the zeniths of the two relevant training nodes. Interpolate them linearly in cos(zenith) to the actual zenith of the event.

This would get rid of the jumps, and result in a more accurate reconstruction between the training nodes.

All we need is to know the pointings used in the training (we could save them together with the RFs). Then replace all calls to "predict" by calls to a new function which calls predict twice and does the interpolation.

Note that besides real data, this will also affect the MC _test_ nodes, which correspond to different pointings than the training ones. Hence I think this change would reduce the systematic errors of the instrument response functions (e.g. eff. area) at high zeniths.

moralejo:
I tested the cosZD interpolation for energy reconstruction, and seems to work very well (see below)
Implementation is rather simple and it increases the time spent by the RF prediction by a factor a bit more than 2 (two calls to predict instead of one), which is not too bad - plus we recently reduced the RFs size, which makes them faster.
This only has to be done in the application part of the RF. During the training the RFs are applied to the same training nodes (different events than those used in training) so there is no mismatch in pointing to worry about.

![image](https://github.com/user-attachments/assets/5d9303ad-6134-4c44-9ef5-3f881bbc8be4)

vuillaut:
Hi
Another approach could be combining linear regression with random forests.
The linear regression would first capture the global trend and smooth things, then random forests make a finer prediction but the steps would already be smoothed.

Something like this:
```
linear_reg.fit(X, y)
residuals = linear_reg.predict(X)
random_forest.fit(X, residuals)

[...]

PR #1320, opened 2024-11-27
Interpolation of RF predictions with cosZD, for homogeneous performance

moralejo:
This should solve the issue of RF "performance jumps" at high zenith angles, when the pointing goes through the middle points between training nodes. See #1317

**NOTE: this is a different implementation of the interpolation approach proposed by @gabemery, see branch https://github.com/cta-observatory/cta-lstchain/tree/dl2_RF_interpolate)**

If the interpolation option is activated we call the RF predictors twice, for each of the two closest MC training nodes, then interpolate (or extrapolate) the values to the actual telescope pointing for each event.

Currently the training sample pointings are obtained from the path to the training sample(provided via config file). A better solution would be to add to the .sav files an array with the zenith and azimuth values of the MC training nodes.

### DIFF
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -60,22 +60,25 @@ def add_zd_interpolation_info(dl2table, training_pointings):
     between the zenith pointings of the MC data in the training sample on
     which the RFs were trained.
 
-    Parameters:
-    -----------
-    dl2table: pandas dataframe. Four columns will be added: alt0, alt1, w0, w1
-    alt0 and alt1 are the alt_tel values (telescope elevation, in radians) of
-    the closest and second-closest training MC pointings (closest in elevation,
-    on the same side of culmination) for each event in the table. The values
-    w0 and w1 are the corresponding weights that, multiplied by the RF
-    predictions at those two pointings, provide the interpolated result for
-    each event's pointing
+    Parameters
+    ----------
+    dl2table : pandas.DataFrame
+        DataFrame containing DL2 information, including 'alt_tel' and 'az_tel'.
+        Four columns will be added: alt0, alt1, w0, w1.
+        alt0 and alt1 are the alt_tel values (telescope elevation, in radians) of
+        the closest and second-closest training MC pointings (closest in elevation,
+        on the same side of culmination) for each event in the table. The values
+        w0 and w1 are the corresponding weights that, multiplied by the RF
+        predictions at those two pointings, provide the interpolated result for
+        each event's pointing.
 
-    training_pointings: astropy Table containing the pointings (zd,
-    az) of the MC training nodes
+    training_pointings : astropy.table.Table 
+        Table containing the pointings (zd, az) of the MC training nodes.
 
-    Returns:
+    Returns
     -------
-    DL2 pandas dataframe with additional columns alt0, alt1, w0, w1
+    pandas.DataFrame
+        Updated DL2 pandas dataframe with additional columns alt0, alt1, w0, w1.
 
     """
 
@@ -143,25 +146,30 @@ def predict_with_zd_interpolation(rf, param_array, features):
     sceond-closest pointing. Then the values are interpolated (linearly in
     cos(zenith)) to the actual zenith pointing (90 deg - alt_tel) of the event.
 
-    rf: sklearn.ensemble.RandomForestRegressor or RandomForestClassifier,
-    the random forest we want to apply (must contain alt_tel among the
-    training parameters)
+    Parameters
+    ----------
+    rf : sklearn.ensemble.RandomForestRegressor or RandomForestClassifier,
+        The random forest we want to apply (must contain alt_tel among the
+        training parameters).
+    param_array : pandas.DataFrame
+        Dataframe containing the features needed by the RF.
+        It must also contain four additional columns: alt0, alt1, w0, w1, which
+        can be added with the function add_zd_interpolation_info. These are the
+        event-wise telescope elevations for the closest and 2nd-closest training
+        pointings (alt0 and alt1), and the event-wise weights (w0 and w1) which
+        must be applied to the RF prediction at the two pointings to obtain the
+        interpolated value at the actual telescope pointing. Since the weights
+        are the same (for a given event) for different RFs, it does not make
+        sense to compute them here - they are pre-calculated by
+        `add_zd_interpolation_info`.
+    features : list of str
+        List of the names of the image features used by the RF.
 
-    param_array: pandas dataframe containing the features needed by theRF
-    It must also contain four additional columns: alt0, alt1, w0, w1, which
-    can be added with the function add_zd_interpolation_info. These are the
-    event-wise telescope elevations for the closest and 2nd-closest training
-    pointings (alt0 and alt1), and the event-wise weights (w0 and w1) which
-    must be applied to the RF prediction at the two pointings to obtain the
-    interpolated value at the actual telescope pointing. Since the weights
-    are the same (for a given event) for different RFs, it does not make
-    sense to compute them here - they are pre-calculated by
-    add_zd_interpolation_info
-
-    features: list of the names of the image features used by the RF
-
-    Return: interpolated RF predictions. 1D array for regressors (log energy,
-    or disp_norm), 2D (events, # of classes) for classifiers
+    Return
+    ------
+    numpy.ndarray
+        Interpolated RF predictions. 1D array for regressors (log energy,
+        or disp_norm), 2D (events, # of classes) for classifiers.
 
     """
 
@@ -207,7 +215,7 @@ def train_energy(train, custom_config=None):
     Parameters
     ----------
     train: `pandas.DataFrame`
-    custom_config: dictionary
+    custom_config : dict
         Modified configuration to update the standard one
 
     Returns
@@ -780,12 +788,13 @@ def apply_models(dl1,
     effective_focal_length: `astropy.unit`
     custom_config: dictionary
         Modified configuration to update the standard one
-    interpolate_rf: dictionary. Contains three booleans, 'energy_regression',
+    interpolate_rf : dict
+        Contains three booleans, 'energy_regression',
         'particle_classification', 'disp', indicating which RF predictions
-        should be interpolated linearly in cos(zenith)
-    training_pointings: astropy Table azimuth (az), zenith (zd)
-         pointings of the MC sample used in the training. Needed for the
-         interpolation of RF predictions.
+        should be interpolated linearly in cos(zenith).
+    training_pointings : astropy.table.Table 
+        Table with azimuth (az), zenith (zd) pointings of the MC sample used
+        in the training. Needed for the interpolation of RF predictions.
 
     Returns
     -------
--- a/lstchain/reco/dl1_to_dl2.py
+++ b/lstchain/reco/dl1_to_dl2.py
@@ -199,7 +199,7 @@ def predict_with_zd_interpolation(rf, param_array, features, is_proba=False):
 
     features: list of the names of the image features used by the RF
 
-    is_proba: if True predict_proba will be called instead of predict 
+    is_proba: if True predict_proba will be called instead of predict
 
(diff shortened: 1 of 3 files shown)


vuillaut:
Hi @moralejo 
Thank you for making this implementation.
Early feedback: 
would it be possible to add the training and testing nodes of the current production as an ascii file in the lstchain data files to allow for testing without necessarily being on the cluster?
(astropy tables can be dumped to ascii easily) 
I completely agree these pointing nodes should be saved with the models during training and then loaded from there in the future.

moralejo:
> Hi @moralejo Thank you for making this implementation. Early feedback: would it be possible to add the training and testing nodes of the current production as an ascii file in the lstchain data files to allow for testing without necessarily being on the cluster? (astropy tables can be dumped to ascii easily)

Will do. What is better, just the zenith and azimuth values, or a list of the directory names?

vuillaut:
> > Hi @moralejo Thank you for making this implementation. Early feedback: would it be possible to add the training and testing nodes of the current production as an ascii file in the lstchain data files to allow for testing without necessarily being on the cluster? (astropy tables can be dumped to ascii easily)
> 
> Will do. What is better, just the zenith and azimuth values, or a list of the directory names?

I think the table of values that you can use directly at inference would be great (with appropriate metadata such as the date of export).

moralejo:
> > 
> > Will do. What is better, just the zenith and azimuth values, or a list of the directory names?
> 
> I think the table of values that you can use directly at inference would be great (with appropriate metadata such as the date of export).

Can you clarify exactly what metadata (and how? Just lines starting with # or whatever?)

moralejo:
What if we just write  in the config the two arrays of zeniths and azimuths?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -51,6 +53,159 @@
...
+    # Compute the weights that multiplied times the RF predictions at the
+    # closest (0) and 2nd-closest (1) nodes (in alt_tel) result in the
+    # interpolated value. Take care of cases in which the two closest nodes
+    # happen to have the same zenith (or very close)! (if so, both nodes are
+    # set to have equal weight in the interpolation)
+    w1 = np.where(np.isclose(closest_alt, second_closest_alt, atol=1e-4, rtol=0),
+                  0.5, (cos_tel_zd - c0) / (c1 - c0))
+    w0 = 1 - w1
+
+    # Update the dataframe:
```
vuillaut, line 124: I don't see why converting to Series is necessary?

lstchain/reco/dl1_to_dl2.py
```diff
@@ -51,6 +53,159 @@
...
+    dl2table = dl2table.assign(alt0=pd.Series(closest_alt).values,
+                               alt1=pd.Series(second_closest_alt).values,
```
vuillaut, line 126: please use more explicit names for alt0, alt1, c0 and c1 in the dl2table that will end up being written

lstchain/reco/dl1_to_dl2.py
```diff
@@ -51,6 +53,159 @@
...
+    Return: interpolated RF predictions. 1D array for regressors (log energy,
+    or disp_norm), 2D (events, # of classes) for classifiers
+
+    """
+
+    # Type of RF (classifier or regressor):
+    is_classifier = isinstance(rf, RandomForestClassifier)
+
+    # keep original alt_tel values:
+    param_array.rename(columns={"alt_tel": "original_alt_tel"}, inplace=True)
```
vuillaut, line 177: I think it would be easier to change the features rather than this game of name-swapping:
```
 features_copy =  features.copy()  # avoid modifying features outside the function
 alt_index_in_features = features_copy.index('alt_tel')
 features_copy[alt_index_in_features] = 'alt0'
 predict_0 = rf.predict(param_array[features_copy])
 features_copy[alt_index_in_features] = 'alt1'
 predict_1 = ...
```
that's it

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -74,11 +75,107 @@
...
+
+    """
+
+    training_zd_deg = []
+    training_az_deg = []
+
+    for dir in training_dirs:
+        c1 = dir.find('_theta_') + 7
+        c2 = dir.find('_az_', c1) + 4
+        c3 = dir.find('_', c2)
```
vuillaut, line 105:
```
# Regular expression to extract theta and az values
pattern = r"theta_(\d+\.\d+)_az_(\d+\.\d+)_"

# Search for matches
match = re.search(pattern, name)

if match:
    theta = float(match.group(1))  # First captured group is theta
    az = float(match.group(2))     # Second captured group is az
    print(f"Theta: {theta}, Az: {az}")
else:
    print("Pattern not found!")
```
would be more robust to find digital numbers

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -74,11 +75,107 @@
...
 
+    interpolate_energy = False
+    interpolate_gammaness = False
+    interpolate_direction = False
+    # Read in the settings for the interpolation of Random Forest predictions
+    # in cos(zd). If activated this avoids the jumps of performance produced
+    # by the discrete set of pointings in the RF training sample.
+    if 'random_forest_zd_interpolation' in config.keys():
+        zdinter = config['random_forest_zd_interpolation']
+        if 'interpolate_energy' in zdinter.keys():
```
vuillaut, line 128: slightly more readable:
```
interpolate_energy = zdinter.get('interpolate_energy', False)
```
and then you don't need to set it first, it is set to False here if `interpolate_energy` not in zdinter

vuillaut (changes requested):
Hi Abelardo,
Sorry for the delay (and I still need to finish this later...).
I have made a couple of comments for improvement in the code/readability but overall the core logic looks good to me.

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -74,11 +75,107 @@
...
+            zdinter = config['random_forest_zd_interpolation']
+            if 'dl1_training_dir' in zdinter.keys():
+                dl1_training_dir = zdinter['dl1_training_dir']
+
+        if dl1_training_dir is None:
+            # Build name of DL1 MC training files assuming standard pattern:
+            dummy = models_path.as_posix().replace('/data/models', '/data/mc/DL1')
+            dl1_training_dir = (Path(dummy[:dummy.rfind('/dec')] +
+                                     '/TrainingDataset/GammaDiffuse/' +
+                                     dummy[dummy.rfind('/dec') + 1:]))
```
morcuended, line 161: I think this should go to `lstchain.paths` module.

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -74,11 +75,107 @@
...
+    if True in interpolate_rf.values():
+        logger.warning('Cos(zenith) interpolation will be used in:')
+        if interpolate_energy:
+            logger.warning('   energy reconstruction Random Forest')
+        if interpolate_gammaness:
+            logger.warning('   g/h classification Random Forest')
+        if interpolate_direction:
+            logger.warning('   direction reconstruction Random Forest')
+
+        if 'random_forest_zd_interpolation' in config.keys():
```
morcuended, line 151:
```suggestion
        if 'random_forest_zd_interpolation' in config:
```

lstchain/scripts/lstchain_dl1_to_dl2.py
```diff
@@ -74,11 +75,107 @@
...
+
+    interpolate_rf = {'energy_regression': interpolate_energy,
+                      'particle_classification': interpolate_gammaness,
+                      'disp': interpolate_direction
+                      }
+
+    dl1_training_dir = None
+    training_pointings = None
+    if True in interpolate_rf.values():
+        logger.warning('Cos(zenith) interpolation will be used in:')
```
morcuended, line 143: log.info

morcuended (changes requested):
Left some minor comments.

morcuended:
> Can you clarify exactly what metadata (and how? Just lines starting with # or whatever?)

Regarding metadata, if you're using astropy Tables to later write ECSV files, you can define the metadata as:

```python
pointing_table = Table(...)
pointing_table.meta["export_date"] = datetime.today().strftime('%Y-%m-%d')
pointing_table.write("filepath_pointings.ecsv", format="ascii.ecsv")
```

lstchain/scripts/lstchain_dl1_to_dl2.py
moralejo: This won't be needed with the ECVS file approach.

moralejo:
> 
> 1. the training directions are saved as ascii files, produced automatically with the models. At inference this ascii file is loaded with the model to get training direction. But we could actually produce these ascii files manually for the current productions and add them to the models directories in the cluster to allow for interpolated inference already.
> 
> To save the pointing directions as ascii files, I'd recommend starting from [astropy tables and save to `ecsv`](https://docs.astropy.org/en/latest/io/ascii/ecsv.html)

I have created the ECVS files for all directories of the new production, e.g.:
/fefs/aswg/data/models/AllSky/20240918_v0.10.12_allsky_nsb_tuning_0.00/dec_4822/training_dirs.ecsv
Please check if this is ok, I will change the code to use these.

lstchain/scripts/lstchain_dl1_to_dl2.py
moralejo: No longer applies

lstchain/reco/dl1_to_dl2.py
moralejo: Much better indeed

moralejo: It seems that there is a check for the feature name, that is why I had gone for the column renaming (although I had forgotten by now)

lstchain/reco/dl1_to_dl2.py
moralejo: Can't recall now the exact error or warning, but I was forced to do this for it to work.

lstchain/reco/dl1_to_dl2.py
moralejo: I am not convinced, anything sufficiently descriptive to make looking into the code's comments unnecessary is probably too long  'alt_closest_mc_training_node'  'alt_second_closest_mc_training_node', then the same replacing but "alt" with "weight"...   The code itself becomes hard to read

lstchain/reco/dl1_to_dl2.py
moralejo: I removed the conversion and it still works, thanks!

morcuended (approved):
It looks good to me. I've just left some modifications on the docstrings following the numpydocs standard so it is later properly render in the documentation.

morcuended (approved):

moralejo:
Thanks @morcuended !
@vuillaut about the way to store and retrieve the training node pointings: is that ok? Can we merge?

morcuended (commented):
I left one last comment.

> For this we would need to store also RFs trained with several pointings. Is that the case now?

No, right now dummy models are created from the test gamma MC, with single pointing

lstchain/reco/dl1_to_dl2.py
```diff
@@ -602,6 +757,10 @@ def apply_models(dl1,
                  cls_disp_sign=None,
                  effective_focal_length=29.30565 * u.m,
                  custom_config=None,
+                 interpolate_rf={'energy_regression': False,
+                                 'particle_classification': False,
+                                 'disp': False},
```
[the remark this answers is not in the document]
moralejo: Done, thanks! I did not know about that problem, and it is indeed dangerous...

moralejo:
> I left one last comment.
> 
> > For this we would need to store also RFs trained with several pointings. Is that the case now?
> 
> No, right now dummy models are created from the test gamma MC, with single pointing

Ok, then we cannot introduce a unit test now. This is more than two months old, we need to move forward...

morcuended (approved):

vuillaut (approved):
Thank you for addressing my comments, I think it's good to go now !

Issue #1322, opened 2024-12-04
Only one offset bin when creating IRF with diffuse mc and energy dependant theta cut


mdebony:
When creating IRFs with the script lstchain_create_irf_files.py using a gamma diffuse MC file and energy dependent theta cut, the produced IRFs contained only one large offset bin without regards for the setting provided in the configuration file.

I think the issue come from this part of the code :
https://github.com/cta-observatory/cta-lstchain/blob/9aaa78b0bd911d35fc2b3fefa03b064bb21b8262/lstchain/tools/lstchain_create_irf_files.py#L467C1-L480C18

I could propose a fix if you want.

moralejo:
Thanks for raising this, and great if you can propose a fix.
 
Related to it: how are the Ereco-dependent cuts determined when diffuse gamma MC is used? Are the efficiencies computed with all gammas, regardless of the offset angle? Probably it is better to limit the offset, so that the target efficiency is achieved near the center of the FoV. 

And another question: when Ereco-dependent theta cuts are requested, are the Aeff and migration matrix, in each offset bin, computed with the **theta** cut already applied?  If the answer is yes for the mig matrix, this will only work well for pointlike sources. If it is no, it will be sub-optimal for pointlike sources (worse E-resolution than the one we really have after the theta cut). As for Aeff, is it the total one (only gammaness cut, not theta), to be later scaled, using the psf, depending on the theta cut?

maxnoe:
> And another question: when Ereco-dependent theta cuts are requested, are the Aeff and migration matrix, in each offset bin, computed with the theta cut already applied? 

They must be, that's the definition of point-like IRFs.

mdebony:
I will check all of those points to be sure everything is done as expected.

> Related to it: how are the Ereco-dependent cuts determined when diffuse gamma MC is used? Are the efficiencies computed with all gammas, regardless of the offset angle? Probably it is better to limit the offset, so that the target efficiency is achieved near the center of the FoV.
A possibility would be to limit to the maximum offset used in the configuration file for the irfs.

moralejo:
> > And another question: when Ereco-dependent theta cuts are requested, are the Aeff and migration matrix, in each offset bin, computed with the theta cut already applied?
> 
> They must be, that's the definition of point-like IRFs.

[...]

PR #1324, opened 2024-12-05
Fix issue with FoV offset binnning when using energy dependant theta cut and diffuse MC

mdebony:
This PR fix the issue #1322 . The bloc of code managing the FoV binning as been moved too.

The PR also add a warning message if the FoV binning is wider than the simulated FoV, and restrict the gammas used for the computation of energy dependant cuts (gh and theta) to the ones within the FoV range considered for the IRF.


morcuended:
Hi @mdebony, I think the downloading of the test dataset is failing because you're doing it from your forked repository. You should be included in the lst-devs group of the cta-observatory GitHub organization first, and then open commit the changes directly over the lstchain repo. @rlopezcoto, @maxnoe, can you please add @mdebony?

moralejo:
Hi @mdebony, I added you with write permissions, you should now be able to create a branch from main, and introduce the change there. Apologies for the delay! I close this, please open a new one with the change in the branch.

PR #1326, opened 2024-12-16
Pin scikit-learn to 1.2.x

morcuended:
When doing a new lstchain installation, usually the sklearn version ends up being later than 1.2.x (setup.py requires `=~1.2`, which will install the latest 1.x version), the one used to produce existing RF models. One needs to use the same sklearn version in dl1 to dl2 as the one used to produce the models.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
         'pyirf~=0.10.0',
         'scipy>=1.8,<1.12',
         'seaborn',
-        'scikit-learn~=1.2',
+        'scikit-learn~=1.2.2',
         'tables',
         'toml',
         'protozfits>=2.5,<3',


moralejo:
But with the change minor releases 1.2.x will not be used either, right? (is that intended?)

morcuended:
> But with the change minor releases 1.2.x will not be used either, right? (is that intended?)

With this change any 1.2.x version could be installed (by default the last one). One needs the same version used for the training of the models and their application. I've always come across models trained with 1.2.X likely because the installations were done using the `environment.yml` file that has sklearn 1.2 version fixed. But any new installation using pip or conda-forged version will install the latest sklearn 1.X version available.

maxnoe:
Does it really make sense to pin the version here? This will prevent people from using lstchain for models they have already trained with newer versions of scikit learn.

I'd say pinning here only makes sense if we are sure that all trained models were trained using 1.2

morcuended:
Hard to tell. Indeed, new versions of scikit learn could've been used to train models. So pinning the version here does not make sense. But letting it with more freedom you may find the problem of incompatibility, then needing to install the same version used for the training.

morcuended:
I'm no longer convinced that this requirement should be fixed. But the problem is there anyway, so I'm closing this and opening an issue so we don't forget about it.

PR #1328, opened 2024-12-19
Fix issue with FoV offset binnning when using energy dependant theta cut and diffuse MC

mdebony:
This PR fix the issue https://github.com/cta-observatory/cta-lstchain/issues/1322 . The bloc of code managing the FoV binning as been moved too.

The PR also add a warning message if the FoV binning is wider than the simulated FoV, and restrict the gammas used for the computation of energy dependant cuts (gh and theta) to the ones within the FoV range considered for the IRF.

### DIFF
--- a/lstchain/tools/lstchain_create_irf_files.py
+++ b/lstchain/tools/lstchain_create_irf_files.py
@@ -709,7 +709,8 @@ class IRFFITSWriter(Tool):
                 self.hdus.append(
                     create_rad_max_hdu(
                         self.theta_cuts["cut"][:, np.newaxis],
-                        reco_energy_bins, fov_offset_bins,
+                        reco_energy_bins, fov_offset_bins[[fov_offset_bins.argmin(),
+                                                           fov_offset_bins.argmax()]],
                         **extra_headers
                     )
                 )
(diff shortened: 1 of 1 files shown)


lstchain/tools/lstchain_create_irf_files.py
```diff
@@ -398,8 +398,21 @@ def start(self):
...
         source_offset_bins = self.data_bin.source_offset_bins()
 
+        if self.mc_particle["gamma"]["mc_type"] in ["point_like", "ring_wobble"]:
+            # The 4 is semi-arbitray. This keeps the same precision as the previous code
+            mean_fov_offset = np.round(get_mc_fov_offset(self.mc_particle["gamma"]["file"]), 4)
+            self.log.info(f"Single offset for point like gamma MC with offset {mean_fov_offset}")
+            fov_offset_bins = [mean_fov_offset - 0.1, mean_fov_offset + 0.1] * u.deg
+        else:
+            fov_offset_bins = self.data_bin.fov_offset_bins()
+            self.log.info("Multiple offset for diffuse gamma MC")
```
moralejo, line 408: Perhaps you can print in the log the binning (might be somewhere else, but better be explicit on what it is doing)

mdebony: I've added in the log line the FoV binning used.

moralejo:
> @moralejo This is the new PR. Now test are working, but the codecov check are failling. I'm not sure what I need to change.

codecov is not strictly required. I think you would have to cover a larger part of the code with CI tests

moralejo (approved):

PR #1329, opened 2024-12-19
Allow dla1b to take v0.9 dl1 files as input

moralejo:


lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -223,10 +224,16 @@ def main():
...
 
-    subarray_info = SubarrayDescription.from_hdf(args.input_file)
     tel_id = config["allowed_tels"][0] if "allowed_tels" in config else 1
-    optics = subarray_info.tel[tel_id].optics
-    camera_geom = subarray_info.tel[tel_id].camera.geometry
+    try:
+        subarray_info = SubarrayDescription.from_hdf(args.input_file)
+        optics = subarray_info.tel[tel_id].optics
+        camera_geom = subarray_info.tel[tel_id].camera.geometry
+    except OSError:
```
morcuended, line 232: Is `OSError` raised when trying to run with a v0.9 file without the try block?

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -223,10 +224,16 @@ def main():
...
+        print("Subarray description table is not readable because of version incompatibility.")
+        print("The standard LST optics and camera geometry will be used.")
```
morcuended, line 234: use log.warning() instead of print

morcuended (changes requested):

lstchain/scripts/lstchain_dl1ab.py
moralejo: Yes, I just used the same as in previous fixes of the same problem (in dl1 to dl2). This is what it yields:
OSError: Unsupported version of subarray table: 1.0

morcuended:
Looks good to me now (if it is ready for review)

morcuended (approved):
The exception type will be changed in ctapipe. Until then, it looks good as is.

moralejo:
> The exception type will be changed in ctapipe. Until then, it looks good as is.
Ok, thanks!

Issue #1340, opened 2025-01-23
No info in DL2 files on RFs used?


moralejo:
@vuillaut @rlopezcoto 
There seems to be no record inside DL2 files of what RandomForest were used to obtain the reconstructed quantities.

For the automation of the DL2 and DL3 production (with 9x10 combinations of declination line and NSB level, each with its own set of RFs) it would be convenient to keep at least the path to the used RF models (the path contains both the NSB level and the declination). In this way we can later easily find which IRFs to use for creating the DL3.

Any suggestion on the best way to do this?

vuillaut:
Hi
That should be stored as metadata of the files, along with other metadata.
ctapipe provenance and tools should allow to implement this quite easily... looking into it 👀

vuillaut:
After playing a bit with ctapipe `Provenance` and exchanging with @maxnoe , here is a summary and some thoughts:

1. ctapipe Tool automatically tracks the provenance and write it in a _ToolName.log_ **external** file 
2. there is an [on-going discussion](https://github.com/cta-observatory/ctapipe/issues/1610) in ctapipe to write this provenance directly in the output file (e.g. HDF5 file)
3. still I think we can implement something for lstchain. I believe we want something quite formatted and machine-readable to automate the DL3 creation.

Requirements:
- provenance information in the HDF5 files
- machine-readable provenance information (e.g. in JSON)
- unambiguous location to store provenance for each stage

Optional:
- keep provenance information from input file (e.g. when creating a DL2 file from a DL1 file, the DL1 provenance is copied the same way the other dataset are copied)


Solution (proposal):
- create a new group `provenance` at the root of the file and have each tool write a new dataset in this group. e.g. `dl2.root.provenance['dl1_to_dl2']`

Note: writing the provenance in the `attrs` of a dataset (e.g. `dl2.root.dl2.event.telescope.parameters.LST_LSTCam.attrs`) is not ok as several tools can write into the same dataset. 


The provenance would look like that:

```
In: prov.keys()
Out: dict_keys(['activity_name', 'activity_uuid', 'start', 'stop', 'system', 'input', 'output', 'config'])

In: prov['config']['DL1ToDL2Tool']
Out: 
{'config_file': '/Users/thomas/Work/Projets/CTA/dev/cta-observatory/cta-lstchain/lstchain/data/lstchain_standard_config.json',
 'config_files': [],
[...]

PR #1343, opened 2025-01-30
DL1 to DL2 tool

vuillaut:
In an attempt to solve #1340 , I transformed the DL1 to DL2 script into a ctapipe tool.

I also added functions to write and read the provenance into an HDF5 file.
Advantages:
- the provenance can't be lost, it's inside the file
- it add an unambiguous 'provenance' group where each tool can write, without overwriting other provenances, and keeping track of older ones

Drawbacks:
- the provenance can't be read using pytables (or by extension vitables) because it does not handle variable length strings:
```
  variable length strings are not supported yet

The leaf will become an ``UnImplemented`` node.
```
- Other kind of files are produced by lstchain (`.fits`, `.sav` for the models...) and might not be suitable to save the provenance


morcuended:
Shall we move it to the `tools/` folder?

morcuended:
For the CI/docs action, remove the script entry from `docs/lstchain_api/scripts/index.rst`

```rST
.. _lstchain_dl1_to_dl2:

lstchain_dl1_to_dl2
+++++++++++++++++++

.. automodule:: lstchain.scripts.lstchain_dl1_to_dl2

Usage
-----
.. argparse::
   :module: lstchain.scripts.lstchain_dl1_to_dl2
   :func: parser
   :prog: lstchain_dl1_to_dl2
```
and add to `docs/lstchain_api/tools/index.rst`

```rST
.. automodapi:: lstchain.tools.lstchain_dl1_to_dl2
   :no-inheritance-diagram:
```

moralejo:
> Shall we move it to the `tools/` folder?

I see no clear advantage in that, and it might make some doc (or school presentation) obsolete. The way of executing it is the same (well, it has additional options, but the basic commands are the same), so the explanations on how to run it will still be ok.

moralejo:
> Drawbacks:
> 
> * the provenance can't be read using pytables (or by extension vitables) because it does not handle variable length strings:
> 
> ```
>   variable length strings are not supported yet
> 
> The leaf will become an ``UnImplemented`` node.
> ```

How does one read the provenance, then?

moralejo:
I think tests fail because the case in which no config file is explicitly provided the default is not taken.

morcuended:
> How does one read the provenance, then?

I'd say it is:

```python
from lstchain.io.provenance import read_dl2_provenance
read_dl2_provenance(dl2_hdf5_file_path)
```
I would add it to the documentation. 

> I see no clear advantage in that, and it might make some doc (or school presentation) obsolete. The way of executing it is the same (well, it has additional options, but the basic commands are the same), so the explanations on how to run it will still be ok.

I would still do it for internal consistency with the rest of scripts/tools. The change will not have any impact on the users' side. Just some modifications in the API docs would be needed.

moralejo:
traits.Path does not take None as default argument... Any idea how to solve this?

moralejo:
> traits.Path does not take None as default argument... Any idea how to solve this?

Should be solved now

moralejo:
> > How does one read the provenance, then?
> 
> I'd say it is:
> 
> ```python
> from lstchain.io.provenance import read_dl2_provenance
> read_dl2_provenance(dl2_hdf5_file_path)
> ```
> 
> I would add it to the documentation.
> 

Ok!

> > I see no clear advantage in that, and it might make some doc (or school presentation) obsolete. The way of executing it is the same (well, it has additional options, but the basic commands are the same), so the explanations on how to run it will still be ok.
> 
> I would still do it for internal consistency with the rest of scripts/tools. The change will not have any impact on the users' side. Just some modifications in the API docs would be needed.

Ook... @vuillaut , can you do these two changes, so we can move on?

moralejo:
There is also the issue of docs failing.  @morcuended, will that be solved also with the suggestions you made?

moralejo:
> > traits.Path does not take None as default argument... Any idea how to solve this?
> 
> Should be solved now

CI tests fail now somewhere else... seems a dl2 test file is not produced or something like that.

lstchain/tools/lstchain_dl1_to_dl2.py
```diff
@@ -38,44 +38,131 @@
...
+
+    Returns:
+    --------
+    str
+        Name of the DL2 file
+    """
+    return dl1_filename.replace('dl1', 'dl2', 1)
+
+class DL1ToDL2Tool(Tool):
+    name = "DL1 to DL2 Tool"
```
maxnoe, line 65: this should be the cli name, i.e. `lstchain-dl1-to-dl2` or similar

lstchain/tools/lstchain_dl1_to_dl2.py
```diff
@@ -38,44 +38,131 @@
...
+            if len(self.input_files) == 1:
+                models_dict[models_key] = models_path
+            else:
+                models_dict[models_key] = joblib.load(models_path)
+
+        for input_dl1file in self.input_files:
+            output_filepath = apply_to_file(input_dl1file, models_dict, self.output_dir, config, 
+                                            self.path_models, self.overwrite)
+            p = Provenance()
+            p.add_input_file(input_dl1file, role='dl1 input file')
```
maxnoe, line 139: To make sure provenance is also correct when errors happen, you should add input / output files as close as possible to opening them.

moralejo (approved):

vuillaut:
Hi @moralejo, @morcuended,
Sorry I was on leave until now, thank you for taking care of last changes and merging !

PR #1345, opened 2025-01-31
Correct units in DL1ab output, when processing v0.9 files

moralejo:
v0.9 and v0.10 DL1 files differ in that the latter have units for some of the columns, which are missing in the former. With this change we ensure that the DL1 parameters table has consistent units for all DL1ab outputs, regardless of whether the input file was v0.9 or v0.10


vuillaut (approved):
looks good

morcuended (approved):

PR #1348, opened 2025-02-05
Energy dependent cuts for IRFs and DL3

gabemery:
These changes introduce the option to provide energy-dependent cuts on gammaness, theta or alpha during the IRFs creation. And provides associated metadata in the IRFs and DL3 files.

Why is it needed : Allows to provide energy dependent cuts based on criteria different from gamma efficiency. Such as best sensitivity.

Tool usage change with the new implementation :
Removed flags  --energy-dependent-gh,  --energy-dependent-theta and  --energy-dependent-alpha and replaced them by a --cut_strategy-{gh, theta, alpha} which receive the technique used as a string input.

The implementation is near complete and can be reviewed, but some points need discussion/further work.
To do before merging :

- [ ] Decide how to store the energy dependent cuts : Should the input be stored in the metadata? Before or after re-binning? Other strategy?
- [ ] Implement tests

### DIFF
--- a/docs/examples/irf_dl3_tool_config_edepcuts.json
+++ b/docs/examples/irf_dl3_tool_config_edepcuts.json
@@ -1,0 +1,56 @@
+{
+  "EventSelector": {
+    "filters": {
+      "intensity": [50, Infinity],
+      "width": [0, Infinity],
+      "length": [0, Infinity],
+      "r": [0, 1],
+      "wl": [0.01, 1],
+      "leakage_intensity_width_2": [0, 1],
+      "event_type": [32, 32]
+    }
+  },
+  "IRFFITSWriter": {
+    "cut_strategy_gh": "edep_custom",
+    "cut_strategy_theta": "edep_custom",
+    "cut_strategy_alpha": "edep_custom",
+    "interpolate_custom": "nearest"
+  },
+  "DL3Cuts": {
+    "min_event_p_en_bin": 100,
+    "energy_dependent_gh_cut": {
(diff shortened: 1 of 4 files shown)


morcuended:
Hi @gabemery, thanks for the implementation, I think it makes sense to support this. Maybe have a look at how it was recently done in ctapipe, so here we use similar naming and so on (https://github.com/cta-observatory/ctapipe/pull/2473)

gabemery:
@morcuended Thanks for bringing the ctapipe version to my attention. I will have a look and see what can be implemented without re-writing the full tool.

gabemery:
After having a look, there is nothing to directly implement from the ctapipe irf production into this PR since the pipelines are very different.
It could be interesting to fully transition to the ctapipe tools later. But I would say this is out of the scope of this feature addition.

moralejo:
@morcuended can you approve? I had a quick look and it seems ok to me, but probably you know this better.

morcuended:
Let me have a look first

morcuended (commented):
I leave some comments. It looks good overall.

moralejo:
@gabemery what is the status of this?


About this change:
> Tool usage change with the new implementation :
> Removed flags --energy-dependent-gh, --energy-dependent-theta and --energy-dependent-alpha and replaced them by a --cut_strategy-{gh, theta, alpha} which receives the technique used as a string input.

Is it really necessary? Can't we keep it such that the old behaviour is achieved with the same flags as before? 
For the new feature, you can just use e.g. --energy-dependent-gh and providing the custom cuts (instead of an efficiency value). Or am I missing something?

PR #1349, opened 2025-02-07
Set info logging level in find_tailcuts

moralejo:
Not sure if there is a better way to set it. Doing it in the script which calls the function did not work.

### DIFF
--- a/lstchain/image/cleaning.py
+++ b/lstchain/image/cleaning.py
@@ -86,7 +86,9 @@ def find_tailcuts(input_dir, run_number):
     newconfig : dict
         cleaning configuration for running the DL1ab stage
     """
-
+    
+    log.setLevel(logging.INFO)
+       
     # subrun-wise dl1 file names:
     dl1_filenames = Path(input_dir,
                          run_to_dl1_filename(1, run_number, 0).replace(


morcuended:
> @morcuended Are the loggers of the script and of the called functions independent? Should I pass the logging level explicitly as an argument of find_tailcuts? I want it always set to "info", so the original code should do.

In principle not, I think it should be controlled by the level set in the main script calling the other module (see example at the top of https://docs.python.org/3/library/logging.html). Maybe need to set explicitly the level for file handler as well

moralejo:
I will try, thanks a lot for the info!

moralejo:
Adding the console handler resulted in duplication of the (incomplete) output in stdout. 
Then adding the setting for the file handler did not help either. Will keep trying.

moralejo:
@morcuended  I tried to follow the suggestions in the link you sent and did not help. The only way I seem to be able to activate INFO logging is inside cleaning.py

morcuended (approved):
I do not know the reason why. It is still strange to me that it is not controlled by the main script calling the function.

Therefore, use the definition inside the function for the time being.

gabemery:
I think you need to specify which module you are specifying the log level for. For exemple this is how I set up the logging level for BaccMod in a notebook : 
```
from baccmod import RadialAcceptanceMapCreator, Grid3DAcceptanceMapCreator
logging.getLogger('baccmod').setLevel(logging.WARNING)
```

moralejo:
> I think you need to specify which module you are specifying the log level for. For exemple this is how I set up the logging level for BaccMod in a notebook :
> 
> ```
> from baccmod import RadialAcceptanceMapCreator, Grid3DAcceptanceMapCreator
> logging.getLogger('baccmod').setLevel(logging.WARNING)
> ```
> 
> The name used can be on a of any scope (package, module). So here you could do :
> 
> logging.getLogger('lstchain.image.cleaning').setLevel(logging.INFO) in lstchain_find_tailcuts.py

Thanks, that's it indeed. I will do it like that.

gabemery (approved):

morcuended:
Good! For me, it is still a mystery that you cannot control the level with the root logger, without having to specify explicitly the submodule, but for all lstchain modules at the same time. Maybe I'm misunderstanding something from the logging system.

morcuended (approved):
Better this way than defining it inside the function.

gabemery:
> Good! For me, it is still a mystery that you cannot control the level with the root logger, without having to specify explicitly the submodule, but for all lstchain modules at the same time. Maybe I'm misunderstanding something from the logging system.
> 
> @moralejo, are the rest of INFO messages from other parts of the code shown as well?

You can set the level for all lstchain in one go with 
logging.getLogger('lstchain').setLevel(logging.INFO)

Current implementation will show the INFO messages of the script itself (level defined in line 51) and of methods in lstchain.image.cleaning' but not from elsewhere.

morcuended:
> You can set the level for all lstchain in one go with logging.getLogger('lstchain').setLevel(logging.INFO)
> 
> Current implementation will show the INFO messages of the script itself (level defined in line 51) and of methods in lstchain.image.cleaning' but not from elsewhere.

I think I see now my wrong understanding. 

```python
log = logging.getLogger(__name__)
def main():
    args = parser.parse_args()
log.setLevel(logging.INFO)
```
`setLevel` here only acts over `log` which is defined within the file scope, but not setting it anywhere else (as I thought it was doing, like if it was the root logger acting across all submodules). 

Keep in mind that any other INFO message (default is WARNING) will not appear outside `image.cleaning` or this script as @gabemery says.

moralejo:
> 
> @moralejo, are the rest of INFO messages from other parts of the code shown as well?

I got the same output as when I defined the level inside the function

moralejo:
> 
> `setLevel` here only acts over `log`, defined within the file scope, but not setting it anywhere else (as I thought it was doing, like if it was the root logger acting across all submodules).
> 
> Keep in mind that no other INFO message (default is WARNING) will appear outside `image.cleaning` or this script as @gabemery says, unless defining it lstchain module-wise.

Ok, thanks!  Not a problem in this case.

PR #1350, opened 2025-02-13
Remove images from list of nodes to be copied

moralejo:
A solution to too bulky outputs of DL1ab, as proposed by @maxnoe

### DIFF
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -269,8 +269,11 @@ def main():
         parameters_to_update["calibration_id"] = np.int32
 
     nodes_keys = get_dataset_keys(args.input_file)
-    if args.no_image:
-        nodes_keys.remove(dl1_images_lstcam_key)
+    # Remove the images key from the list of nodes to be copied: 
+    # it will be added later if so chosen by the user.
+    # Copying it here sometimes resulted in too bulky files, due
+    # to empty chunks.
+    nodes_keys.remove(dl1_images_lstcam_key)
 
     metadata = global_metadata()
 


vuillaut (changes requested):
rather than simply removing the feature, can we simply turn "no-image" to default ?
So, change it to 
```
parser.add_argument(
    '--keep-images', action='store_true',
    help='Pass this argument to write the images in the new DL1 files.',
)
```

maxnoe:
> rather than simply removing the feature, can we simply turn "no-image" to default ? So, change it to
> 
> ```
> parser.add_argument(
>     '--keep-images', action='store_true',
>     help='Pass this argument to write the images in the new DL1 files.',
> )
> ```

I think you are misunderstanding what this does.

We do keep the images in the output. This just prevents them to be written and then over-written immediately afterwards. (See discussion in slack)

morcuended (approved):
Looks good to me

lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -269,8 +269,11 @@ def main():
         parameters_to_update["calibration_id"] = np.int32
 
     nodes_keys = get_dataset_keys(args.input_file)
-    if args.no_image:
-        nodes_keys.remove(dl1_images_lstcam_key)
+    # Remove the images key from the list of nodes to be copied: 
+    # it will be added later if so chosen by the user.
+    # Copying it here sometimes resulted in too bulky files, due
+    # to empty chunks.
+    nodes_keys.remove(dl1_images_lstcam_key)
```
maxnoe, line 276: It probably doesn't play a large role for the file size, but I think the same applies to the parameters table. Maybe compare once with no images?

moralejo: I thought the same and had already tested this morning, it did not make a difference (I tried with images, but total size was identical to within bytes). I have no idea of why the difference. This is a one-file test, I cannot tell if it would be the same for all files.

moralejo:
> rather than simply removing the feature, can we simply turn "no-image" to default ? So, change it to
> 
> ```
> parser.add_argument(
>     '--keep-images', action='store_true',
>     help='Pass this argument to write the images in the new DL1 files.',
> )
> ```

Hi @vuillaut , can you mark this as solved? It appears as "requested changes" and we can't move on.

vuillaut (approved):

vuillaut:
> > rather than simply removing the feature, can we simply turn "no-image" to default ? So, change it to
> > ```
> > parser.add_argument(
> >     '--keep-images', action='store_true',
> >     help='Pass this argument to write the images in the new DL1 files.',
> > )
> > ```
> 
> I think you are misunderstanding what this does.
> 
> We do keep the images in the output. This just prevents them to be written and then over-written immediately afterwards. (See discussion in slack)

Indeed I misunderstood this. Thanks for the explanation (I missed the slack discussion).

PR #1355, opened 2025-03-13
Keep all peak times in DL1a

moralejo:
Currently the output DL1 file keeps only the times of pixels that fulfill all cleaning conditions, except the time condition. This change makes that all pixel times are kept. This is relevant e.g. if one wants to apply lower tailcuts to re-generate DL1b. Equivalently, if one wants to re-generate DL1b with a scaling (>1) of the pixel charges and the same tailcuts (because times will be missing  for the additional pixels which survive the cleaning charge-wise)

### DIFF
--- a/lstchain/scripts/lstchain_dl1ab.py
+++ b/lstchain/scripts/lstchain_dl1ab.py
@@ -406,13 +406,9 @@ def main():
                     # if delta_time has been set, we require at least one
                     # neighbor within delta_time to accept a pixel in the image:
                     if delta_time is not None:
-                        # make sure only signal pixels are used in the time
-                        # check:
-                        cleaned_pixel_times = np.where(signal_pixels,
-                                                       peak_time, np.nan)
                         new_mask = apply_time_delta_cleaning(camera_geom,
                                                              signal_pixels,
-                                                             cleaned_pixel_times,
+                                                             peak_time,
                                                              1,
                                                              delta_time)
                         signal_pixels = new_mask


lstchain/scripts/lstchain_dl1ab.py
```diff
@@ -406,10 +406,10 @@ def main():
                     # if delta_time has been set, we require at least one
                     # neighbor within delta_time to accept a pixel in the image:
                     if delta_time is not None:
-                        cleaned_pixel_times = peak_time
-                        # makes sure only signal pixels are used in the time
+                        # make sure only signal pixels are used in the time
                         # check:
-                        cleaned_pixel_times[~signal_pixels] = np.nan
+                        cleaned_pixel_times = np.where(signal_pixels,
```
maxnoe, line 411: Why are we modifying peak times in any case? All ctapipe functions should work with the cleaning mask instead of modifying the input arrays

moralejo: Indeed, that was the bug (which I introduced 4 years ago). I thought cleaned_pixel_times would be a copy of the original peak_times, and the nans would not be written out.

maxnoe: But why is it even needed inside this function?

moralejo: The cleaning may use a timing constraint. The line in question is intended to ignore pixels which do not survive the rest of the (charge-based) cleaning conditions, when evaluating a pixel's time with respect to that of its neighbours (i.e. do not include in the constraint a neighbour which most likely contains only noise)

maxnoe: But that's already covered by passing the mask of the previous step, right?

https://github.com/cta-observatory/ctapipe/blob/ec47070cd1691deafb952f5b4f51672891dde09b/src/ctapipe/image/cleaning.py#L281-L287

moralejo: Yes, I have just noticed that. Probably it was not the case when the cleaning was  implemented within lstchain.

maxnoe (approved):

PR #1357, opened 2025-03-13
MC charge scaling factor

moralejo:
The option to scale the DL1a charges in MC images was no longer working.

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -354,8 +354,6 @@ def r0_to_dl1(
     metadata = global_metadata()
     write_metadata(metadata, output_filename)
 
-    cal_mc = load_calibrator_from_config(config, subarray)
-
     # minimum number of pe in a pixel to include it
     # in calculation of muon ring time (peak sample):
     min_pe_for_muon_t_calc = 10.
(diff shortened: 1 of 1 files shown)


lstchain/reco/r0_to_dl1.py
```diff
@@ -507,11 +515,6 @@ def r0_to_dl1(
             # write sub tables
             if is_simu:
                 write_subarray_tables(writer, event, metadata)
-                cal_mc(event)
```
moralejo, line 510: Why was this here? Isn't the calibration done by r1_dl1_calibrator below, both for data and mc?

moralejo: Small pixel charges change a bit (even when no additional random NSB is added) - not clear if this cal_mc had some role...

moralejo: No, there seems to be some random component in the signal extraction, even running twice the original version results in slightly different pixel charges

moralejo: (deleted comments about tests were wrong - I had not really switched off the additional NSB generation)

FrancaCassol (approved):

PR #1358, opened 2025-03-24
Functions to obtain recommended intensity cut

moralejo:
The cut is based on the intensity spectrum (dR/dI) obtained from a DL2 table.
The idea is to use the cut as a "software threshold" which improves the agreement between data and MC by removing events too close to trigger threshold (whose configuration is fixed in MC but variable in data). 

The proposed intensity cut is just a fixed factor (1.3) above the intensity for which 50% of the peak rate (dR/dI) of the intensity spectrum is reached. The factor is chosen such that the bulk of the good-quality data taken in dark conditions (hence processed with 8-4 p.e. cleaning) have a cut around 50 p.e. 

image

50 p.e. is close to the peak of the intensity spectrum for such data, and hence will mostly remove the low-energy events close to threshold for which we may expect trigger-induced differences between data and MC. A few example dR/dI spectra from high-rate low-zenith runs:

image

### DIFF
--- a/lstchain/reco/utils.py
+++ b/lstchain/reco/utils.py
@@ -20,6 +20,7 @@ from ctapipe.coordinates import CameraFrame
 from ctapipe.containers import EventType
 from ctapipe_io_lst import OPTICS
 from ctapipe_io_lst.constants import LST1_LOCATION
+from scipy.signal import find_peaks
 
 from . import disp
 
@@ -40,6 +41,8 @@ __all__ = [
     "get_effective_time",
     "get_event_pos_in_camera",
     "get_geomagnetic_delta",
+    "get_intensity_threshold",
+    "get_intensity_cut",
     "impute_pointing",
     "linear_imputer",
     "polar_to_cartesian",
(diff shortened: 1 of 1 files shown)


lstchain/reco/utils.py
```diff
@@ -778,6 +781,127 @@ def get_geomagnetic_delta(zen, az, geomag_dec=None, geomag_inc=None, time=None):
...
+    step = 0.02 # in log10(intensity), for histogramming & peak finding
+    
+    efftime, _ = get_effective_time(data)
+
+    nbins = 1 + int( (np.log10(max_intensity) - np.log10(min_intensity)) / step)
+    bins = np.logspace(np.log10(min_intensity), 
+                       np.log10(max_intensity), nbins)
+    binwidth = np.diff(bins)
+    nevents, _ = np.histogram(data['intensity'], bins=bins)
+    nevents = nevents.astype('float') / (binwidth*efftime.to_value(u.s))
```
rlopezcoto, line 831: why changing the value of a variable here instead of defining a new one?

lstchain/reco/utils.py
```diff
@@ -778,6 +781,127 @@ def get_geomagnetic_delta(zen, az, geomag_dec=None, geomag_inc=None, time=None):
...
+    xx = np.logspace(np.log10(min_intensity), 
+                     np.log10(xmax), 
+                     1+int((np.log10(xmax) - np.log10(min_intensity))/finestep))[::-1]
+
+    # Linear interpolation in dR/dI (differential rate) vs. log10(intensity/p.e.):    
+    yy = np.interp(np.log10(xx), np.log10(bincenters), nevents)
+
+    x50 = np.nan
+    y50 = np.nan
+    for xxx, yyy in zip(xx, yy):
```
rlopezcoto, line 857: wouldn't it be better to use vectorize this instead of a for loop?

lstchain/reco/utils.py
```diff
@@ -778,6 +781,127 @@ def get_geomagnetic_delta(zen, az, geomag_dec=None, geomag_inc=None, time=None):
...
+
+    Returns
+    -------
+    intensity_cut: (p.e.) Recommended intensity cut (both in data and MC)
+    to achieve a good match between data and MC
+    """
+
+    # Factor by which we multiply the "intensity threshold" (intensity at which
+    # 50% of peak rate dR/dI is reached) to obtain the (minimum) intensity cut. 
+    factor = 1.3
```
rlopezcoto, line 891: maybe we could add a plot to this review to justify this factor in case it needs to be revisited in the future

lstchain/reco/utils.py
moralejo: This takes a completely negligible amount of time, and is run once per run

lstchain/reco/utils.py
moralejo: Right, the name was also misleading. Now it is drdi

lstchain/reco/utils.py
```diff
@@ -778,6 +781,128 @@ def get_geomagnetic_delta(zen, az, geomag_dec=None, geomag_inc=None, time=None):
...
+    nevents, _ = np.histogram(data['intensity'], bins=bins)
+    drdi = nevents.astype('float') / (binwidth*efftime.to_value(u.s))
+    bincenters = (bins[1:]*bins[:-1])**0.5 # geometrical mean (log bin center)
+ 
+    peaks, properties = find_peaks(np.log10(drdi), 
+                                   prominence=0.04, # ~10% in log10 scale
+                                   width=int(0.1/step)) # ~25% in log10 scale
+
+    # If no peak is found, nans are returned (except for the histogram data):
+    if len(peaks) == 0:
```
ednaruiz, line 840: Hi @moralejo , just a small suggestion, to add a logging statement here to improve traceability for debugging,
something like 
logging.warning("No valid peaks found in intensity spectrum. Possible issues in input data.")

lstchain/reco/utils.py
```diff
@@ -778,6 +781,128 @@ def get_geomagnetic_delta(zen, az, geomag_dec=None, geomag_inc=None, time=None):
...
+    # Factor by which we multiply the "intensity threshold" (intensity at which
+    # 50% of peak rate dR/dI is reached) to obtain the (minimum) intensity cut. 
+    factor = 1.3 
+    # This value results in ~50 p.e. for the bulk of the LST1 data taken in dark 
+    # conditions (this cut provides good data-MC agreement in image parameters,
+    # as shown in the LST1 performance paper, ApJ 956
+
+    default_cut = 50 # p.e.
+    # We return the default of 50 p.e., if factor * intensity_threshold is below it. 
+
```
ednaruiz, line 899: maybe a logging here as well, same reason as above. During code execution, it should be transparent/traceable that we choose 50pe as default if the peak rate is not found.

rlopezcoto (approved):

PR #1359, opened 2025-03-31
select cosmics for dR/dI spectrum

moralejo:
Although interleaved pedestals and flatfield events should not be a nuisance around the peak of the intensity distribution for cosmics, it is better to make the histogram only with events tagged as cosmics.


ednaruiz (approved):
the change is reasonable and correctly implemented.

PR #1361, opened 2025-04-24
Start updating code for ctapipe 0.22

moralejo:
We should move on with this (even update to more recent ctapipe & ctapipe_io_lst), but to be honest I do not know how much needs to be updated in lstchain...

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,8 @@ dependencies:
   - numpy
   - astropy=5
   - pip
-  - ctapipe=0.19.2
+  - ctapipe=0.22.0
+  - ctapipe_io_lst=0.25
   - ctaplot~=0.6.4
   - gammapy=1.1
   - h5py
@@ -17,9 +18,8 @@ dependencies:
   - iminuit>=2
   - joblib~=1.2.0
   - toml
-  - protozfits=2.5
   - pyparsing
-  - scipy~=1.11.4
+  - scipy <1.14
   - scikit-learn=1.2
   - sphinx
   - sphinx-automodapi
@@ -32,6 +32,5 @@ dependencies:
   - pandas>=2
   - pymongo
   - seaborn
-  - ctapipe_io_lst=0.24
   - pytest
   - pyirf~=0.10.0
--- a/lstchain/calib/camera/flatfield.py
+++ b/lstchain/calib/camera/flatfield.py
@@ -135,7 +135,7 @@ class FlasherFlatFieldCalculator(FlatFieldCalculator):
         peak_pos = 0
         if self.extractor:
             broken_pixels = event.mon.tel[self.tel_id].pixel_status.hardware_failing_pixels
-            dl1 = self.extractor(waveforms, self.tel_id, no_gain_selection, broken_pixels=broken_pixels)
+            dl1 = self.extractor(waveforms, self.tel_id, selected_gain_channel=None, broken_pixels=broken_pixels)
             charge = dl1.image
             peak_pos = dl1.peak_time
 
@@ -143,7 +143,7 @@ class FlasherFlatFieldCalculator(FlatFieldCalculator):
         # (e.g. drs4 waveform time shifts for LST)
         time_shift = event.calibration.tel[self.tel_id].dl1.time_shift
         if time_shift is not None:
-                peak_pos -= time_shift
+            peak_pos -= time_shift
 
         return charge, peak_pos
 
--- a/lstchain/reco/reconstructor.py
+++ b/lstchain/reco/reconstructor.py
@@ -215,7 +215,7 @@ class TimeWaveformFitter(TelescopeComponent):
         if dl1_calib.pedestal_offset is not None:
             waveform = waveform - dl1_calib.pedestal_offset[:, np.newaxis]
 
-        n_pixels, n_samples = waveform.shape
+        n_samples = waveform.shape[2]
         times = np.arange(0, n_samples) * dt
         selected_gains = event.r1.tel[telescope_id].selected_gain_channel
         is_high_gain = (selected_gains == 0)
@@ -289,7 +289,7 @@ class TimeWaveformFitter(TelescopeComponent):
         p_y = pix_y[mask_pixel]
         pix_area = geometry.pix_area[mask_pixel].to_value(unit ** 2)
 
-        data = waveform
+        data = waveform[0]
         error = self.error
 
         filter_pixels = np.nonzero(~mask_pixel)
--- a/lstchain/tools/lstchain_create_calibration_file.py
+++ b/lstchain/tools/lstchain_create_calibration_file.py
@@ -108,7 +108,7 @@ class CalibrationHDF5Writer(Tool):
     def setup(self):
 
         self.log.info("Opening file")
-        self.eventsource = EventSource.from_config(parent=self)
+        self.eventsource = EventSource(parent=self)
 
         self.processor = CalibrationCalculator.from_name(
             self.calibration_product,
--- a/setup.py
+++ b/setup.py
@@ -41,9 +41,9 @@ setup(
     packages=find_packages(exclude="lstchain._dev_version"),
     install_requires=[
         'astropy~=5.0',
-        'bokeh~=2.0',
-        'ctapipe~=0.19.2',
-        'ctapipe_io_lst~=0.24.0',
+        'bokeh~=3.0',
+        'ctapipe~=0.22.0',
+        'ctapipe_io_lst~=0.25.0',
         'ctaplot~=0.6.4',
         'eventio>=1.9.1,<2.0.0a0',  # at least 1.1.1, but not 2
         'gammapy~=1.1',
@@ -55,7 +55,7 @@ setup(
         'numpy',
         'pandas',
         'pyirf~=0.10.0',
-        'scipy>=1.8,<1.12',
+        'scipy>=1.8,<1.14',
         'seaborn',
         'scikit-learn~=1.2',
         'tables',


maxnoe:
@moralejo You should update the ctapipe version tested in CI in `.github/workflows`

moralejo:
> @moralejo You should update the ctapipe version tested in CI in `.github/workflows`

thx, will do

maxnoe:
It seems there is also an incompatibility with gammapy

maxnoe:
I think this is between scipy and gammapy and should be fixed by updating gammapy to 1.2

maxnoe:
Add `MissingReferenceMetadata` to the ignored warnings in `pyproject.toml`

maxnoe:
>  E             File "/home/runner/work/cta-lstchain/cta-lstchain/lstchain/scripts/lstchain_data_r0_to_dl1.py", line 230, in main
> E               r0_to_dl1.r0_to_dl1(
> E             File "/home/runner/work/cta-lstchain/cta-lstchain/lstchain/reco/r0_to_dl1.py", line 718, in r0_to_dl1
> E               event.r1.tel[telescope_id].waveform *= ~bad_waveform
> E           ValueError: operands could not be broadcast together with shapes (1,1855,36) (1855,1855) (1,1855,36) 

ctapipe has changed the shape of waveform to always be 3d, also for gain selected data.

moralejo:
> > Add `MissingReferenceMetadata` to the ignored warnings in `pyproject.toml`
> 
> What will that do?

I still get the warnings, not sure what I did wrong...

maxnoe:
> I still get the warnings, not sure what I did wrong...

I don't see them anymore in the current run. They are also gone if you update to a newer version of ctapipe. We were a bit too heavy on this warning, raising it also for files we know can not have this metadata.

maxnoe:
In current ctapipe, the extractors should just return a 2d image if you give it all gains and no `selected_gain_channel`, so the worarounds here should no longer be necessary

maxnoe:
`ctapipe_io_lst` currently only allows ctapipe up to 0.24:

https://github.com/cta-observatory/ctapipe_io_lst/pull/241

moralejo:
> `ctapipe_io_lst` currently only allows ctapipe up to 0.24:
> 
> [cta-observatory/ctapipe_io_lst#241](https://github.com/cta-observatory/ctapipe_io_lst/pull/241)

Yes, that was the problem, the "<0.25.0a0" misled me, thinking 0.25.0 would be allowed

moralejo:
Solved a few issues with API changes, but now I am stuck with a problem with provenance which makes CI tests fail, likely related to https://github.com/cta-observatory/ctapipe/pull/2688

maxnoe:
> Solved a few issues with API changes, but now I am stuck with a problem with provenance which makes CI tests fail, likely related to https://github.com/cta-observatory/ctapipe/pull/2688

Fixed by using None as default and checking for None. `Undefined` is really an internal traitlets thing and shouldn't be used by users.

environment.yml
```diff
@@ -19,7 +20,7 @@ dependencies:
   - toml
   - protozfits=2.6
   - pyparsing
-  - scipy~=1.11.4
+  - scipy <1.14
```
maxnoe, line 23: this pin shouldn't be needed I think

maxnoe:
This one unfortunately needs a fix in ctapipe:

```
   0%|                                                  | 0/188 [00:00<?, ?it/s]
2025-04-25 11:50:03,885 ERROR [lstchain.CatBCalibrationHDF5Writer] (tool.run): Caught unexpected exception: Unsupported pointing frame: CoordinateFrameType.UNKNOWN
  File "/home/runner/micromamba/envs/lst-dev/lib/python3.10/site-packages/ctapipe/core/tool.py", line 445, in run
    self.start()
  File "/home/runner/work/cta-lstchain/cta-lstchain/lstchain/tools/lstchain_create_cat_B_calibration_file.py", line 175, in start
    for count, event in enumerate(tqdm(eventsource)):
  File "/home/runner/micromamba/envs/lst-dev/lib/python3.10/site-packages/tqdm/std.py", line 1181, in __iter__
    for obj in iterable:
  File "/home/runner/micromamba/envs/lst-dev/lib/python3.10/site-packages/ctapipe/io/eventsource.py", line 318, in __iter__
    for event in self._generator():
  File "/home/runner/micromamba/envs/lst-dev/lib/python3.10/site-packages/ctapipe/io/hdf5eventsource.py", line 692, in _generator
    self._fill_array_pointing(data)
  File "/home/runner/micromamba/envs/lst-dev/lib/python3.10/site-packages/ctapipe/io/hdf5eventsource.py", line 782, in _fill_array_pointing
    raise ValueError(f"Unsupported pointing frame: {frame}")
ValueError: Unsupported pointing frame: CoordinateFrameType.UNKNOWN
```

lstchain/image/tests/test_modifier.py
```diff
@@ -59,7 +59,7 @@ def test_calculate_noise_parameters(mc_gamma_testfile, observed_dl1_files):
         mc_gamma_testfile,
         observed_dl1_files["dl1_file1"]
     )
-    assert np.isclose(extra_noise_in_dim_pixels, 1.96, atol=0.01)
+    assert np.isclose(extra_noise_in_dim_pixels, 1.455, atol=0.01)
```
maxnoe, line 62: Shouldn't we understand why the noise changed significantly?

moralejo: Yes, although as I commented in the commit, the value is probably not meaningful because of the poor stats...

maxnoe:
https://github.com/cta-observatory/ctapipe/pull/2744

lstchain/image/tests/test_modifier.py
moralejo: Ok, comparing charges it seems to be a missing pulse integration correction, should be easy to find,.

maxnoe: Missing now? I.e. before it was made and now it is not?

moralejo: yes, but perhaps specific to that part of the code

moralejo: I think I know what it is (and probably is missing only for interleaved events)

lstchain/reco/r0_to_dl1.py
```diff
@@ -571,10 +571,10 @@ def r0_to_dl1(
                     r1.waveform = r1.waveform[r1.selected_gain_channel, PIXEL_INDEX][np.newaxis, :, :] # Add the "channel" axis, all ctapipe wfs are 3D now
 
                     event.calibration.tel[tel_id].dl1.time_shift = \
-                        event.calibration.tel[tel_id].dl1.time_shift[r1.selected_gain_channel, PIXEL_INDEX]
+                        event.calibration.tel[tel_id].dl1.time_shift[:, PIXEL_INDEX]
```
maxnoe, line 574: these lines can be removed completely now, right? They are no-ops

lstchain/reco/r0_to_dl1.py
```diff
@@ -571,10 +571,10 @@ def r0_to_dl1(
                     r1.waveform = r1.waveform[r1.selected_gain_channel, PIXEL_INDEX][np.newaxis, :, :] # Add the "channel" axis, all ctapipe wfs are 3D now
```
maxnoe, line 571:
```suggestion
                    # select gain but keep waveform 3d
                    r1.waveform = r1.waveform[np.newaxis, r1.selected_gain_channel, PIXEL_INDEX]
```

lstchain/reco/r0_to_dl1.py
moralejo: Right, I was assuming there was a pixel reordering but that is not the case.

moralejo:
Now tests will fail again in the (DL1ab)  NSB tuning, I put back the original reference value. Once we understand why the integration correction is not applied (at least for the interleaved pedestals) and solve it, the test should pass.

lstchain/calib/camera/flatfield.py
```diff
@@ -135,15 +135,15 @@ def _extract_charge(self, event):
         peak_pos = 0
         if self.extractor:
             broken_pixels = event.mon.tel[self.tel_id].pixel_status.hardware_failing_pixels
-            dl1 = self.extractor(waveforms, self.tel_id, no_gain_selection, broken_pixels=broken_pixels)
+            dl1 = self.extractor(waveforms, self.tel_id, selected_gain_channel=None, broken_pixels=broken_pixels)
```
maxnoe, line 138: `no_gain_selection` can be deleted now

lstchain/image/muon/muon_analysis.py
```diff
@@ -291,12 +289,13 @@ def analyze_muon_event(subarray, tel_id, event_id, image, good_ring_config, plot
 
         # We do the calculation of the ring completeness (i.e. fraction of whole circle) using the pixels
         # within the "width" fitted using MuonIntensityFitter
+
+        print('HOLA', params['ring_completeness_threshold'])
```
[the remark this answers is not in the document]
moralejo: ouch....

lstchain/image/tests/test_modifier.py
moralejo: The reason is that the default for absolute_factor in DL1CameraCalibrationContainer has changed from 1 to None:

https://github.com/cta-observatory/ctapipe/blob/3f2178084855b3fc3d5cf68ad69665c1eebb4de2/ctapipe/containers.py#L462-L466

https://github.com/cta-observatory/ctapipe/blob/5a1340adaade6167ea9e9ea73ebd175614a6ebf9/src/ctapipe/containers.py#L590-L595

When it is None, neither the absolute factor nor the relative factor (where the correction for pulse integration set in the event source config is applied) are used

moralejo:
> Now tests will fail again in the (DL1ab) NSB tuning, I put back the original reference value. Once we understand why the integration correction is not applied (at least for the interleaved pedestals) and solve it, the test should pass.

Solved now

moralejo:
I think the remaining non-passing tests will be solved with this issue pointed out by @maxnoe:
https://github.com/cta-observatory/ctapipe/pull/2744
So we need a bugfix release of ctapipe to proceed.

moralejo:
The only pending failing test was working for me "locally" (i.e. at the cluster) but I cannot test now - bridge computer still down.

maxnoe:
> Anyway, it looks like the output of the calibration is not what it used to be, and hence the error in the plotting part. 

Did you check that there is an actual change in the output? Maybe it's just the new matplotlib version complaining about it where the old one didn't

moralejo:
> > Anyway, it looks like the output of the calibration is not what it used to be, and hence the error in the plotting part.
> 
> Did you check that there is an actual change in the output? Maybe it's just the new matplotlib version complaining about it where the old one didn't.

No I couldn't check yet, it was just speculation. On the other hand, it did not fail in my local test, which is weird. 
In any case, auto-axis setting failing when nans or infs are present used to be there, right? (I've found it often)

> What version of mpl do you have locally? Is it the up-to-date 3.10 as it is in the CI?

Can't connect to the IT yet...

FrancaCassol:
Hi @maxnoe 
> 
> Did you check that there is an actual change in the output? Maybe it's just the new matplotlib version complaining about it where the old one didn't.
> 
> What version of mpl do you have locally? Is it the up-to-date 3.10 as it is in the CI?

Locally the test works for me and I have mpl 3.10.1 ...

moralejo:
> Hi @maxnoe
> 
> > Did you check that there is an actual change in the output? Maybe it's just the new matplotlib version complaining about it where the old one didn't.
> > What version of mpl do you have locally? Is it the up-to-date 3.10 as it is in the CI?
> 
> Locally the test works for me and I have mpl 3.10.1 ...

Same here. I cannot understand, the input file is the same as it used to be, so it is not about a change in the contents. And the point of failure seems to be just creating a camera display, before values are set in.

moralejo:
> Does the test work locally when you set `MPLBACKEND=Agg`? Maybe it's a difference in mpl backends and the CI does not have a proper display environment.

It still works when I set  MPLBACKEND=Agg (as env variable)

maxnoe:
I removed the explicit figure numbers, we had a similar issue in `lstcam_calib`

moralejo:
> I assume you ran this test in isolation?
> 
> I think the issue is that if all tests are run, there is a conflict in figure numbers and the wrong figure is used to plot.

For me, locally, it worked both running in isolation and running them all.
When running them all at once, I got that one test "xfailed" - couldn't find it - was that it?

maxnoe:
A remaining difference could be that the CI runs tests in parallel (-n 2)

lstchain/data/lstchain_standard_config.json
```diff
@@ -287,13 +287,14 @@
...
-      "write_waveforms": true,
+      "write_dl1_images": false,
+      "write_dl1_parameters": false,
+      "write_r0_waveforms": false,
+      "write_r1_waveforms": true,
       "transform_waveform": true,
       "waveform_dtype": "uint16",
       "waveform_offset": 400,
-      "waveform_scale": 80  
+      "waveform_scale": 60
```
maxnoe, line 297: @FrancaCassol Fine with changing the scale here?

Should we change the offset to 300 so that it is equivlent to the 60/5 we use in the EVB?

PR #1367, opened 2025-06-30
Scikit learn version

moralejo:
Pin version so that we are not forced to re-create the current standard RF models


moralejo:
Ok, it does not work, this is why we had updated the version...

moralejo:
1.3 does not work either, and >= 1.4 is already incompatible with 1.2 models. Closing.

Issue #1366, opened 2025-06-30
models from scikit-learn < v1.6 incompatible


vuillaut:
Issue brought to me by @moralejo 

With lstchain v0.11.0, joblib cannot load the model from previous trainings:

```
import joblib
model_filename="/fefs/aswg/data/models/AllSky/20240918_v0.10.12_allsky_nsb_tuning_0.00/dec_3476/cls_disp_sign.sav"
model = joblib.load(model_filename)
```

```
/fefs/aswg/software/conda/envs/lstchain-v0.11.0/lib/python3.13/site-packages/sklearn/base.py:380: InconsistentVersionWarning: Trying to unpickle estimator DecisionTreeClassifier from version 1.2.2 when using version 1.6.1. This might lead to breaking code or invalid results. Use at your own risk. For more info please refer to:
https://scikit-learn.org/stable/model_persistence.html#security-maintainability-limitations
  warnings.warn(
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 model = joblib.load(model_filename)

File /fefs/aswg/software/conda/envs/lstchain-v0.11.0/lib/python3.13/site-packages/joblib/numpy_pickle.py:658, in load(filename, mmap_mode)
    652             if isinstance(fobj, str):
    653                 # if the returned file object is a string, this means we
    654                 # try to load a pickle file generated with an version of
    655                 # Joblib so we load it with joblib compatibility function.
    656                 return load_compatibility(fobj)
--> 658             obj = _unpickle(fobj, filename, mmap_mode)
    659 return obj

File /fefs/aswg/software/conda/envs/lstchain-v0.11.0/lib/python3.13/site-packages/joblib/numpy_pickle.py:577, in _unpickle(fobj, filename, mmap_mode)
    575 obj = None
    576 try:
--> 577     obj = unpickler.load()
    578     if unpickler.compat_mode:
    579         warnings.warn("The file '%s' has been generated with a "
    580                       "joblib version less than 0.10. "
    581                       "Please regenerate this pickle file."
    582                       % filename,
    583                       DeprecationWarning, stacklevel=3)

File /fefs/aswg/software/conda/envs/lstchain-v0.11.0/lib/python3.13/pickle.py:1256, in _Unpickler.load(self)
   1254             raise EOFError
   1255         assert isinstance(key, bytes_types)
-> 1256         dispatch[key[0]](self)
   1257 except _Stop as stopinst:
[...]

PR #1369, opened 2025-07-01
fix scikit-learn v1.3

vuillaut:
This solution is only to keep track of the current solution applied on the cluster to load former models.

Fixes https://github.com/cta-observatory/cta-lstchain/issues/1366

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.11", "3.12"]
+        python-version: ["3.11"]
         ctapipe-version: ["v0.25.1"]
 
     steps:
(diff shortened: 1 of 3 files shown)


vuillaut:
> since this is only a temporary fix to process data, I would go for alpha/beta of v0.11 or something like that: https://semver.org/#spec-item-10
> 
> opinions?

If we want to have a proper release to be used by users of lstosa, yes that should be the way to go.
I'd then make the change starting from v0.11.0 rather than the current main branch and we could tag this branch as `v0.11.0-alpha.1` 

I'd avoid merging in main and force python3.11 there, it could have bad consequences down the road and is not worth for a retro-compatibility bugfix in my opinion.

rlopezcoto:
great, sounds perfect, thanks @vuillaut!

vuillaut:
So, I pushed a new branch from v0.11.0: `v0.11.0_sklearn12` with the same changes as here.
Feel free to make a release from that branch whenever you want.

rlopezcoto:
great, done, thanks again @vuillaut

vuillaut:
Great ! Closing :)

PR #1374, opened 2025-07-08
Attempt to solve issues with new version of Bokeh.

moralejo:
Basically, vertical sliders still do not work, and the workaround that was working in bokeh 2.4 is no longer working.
So I now use horizontal sliders, which required some re-ordering of panels.


moralejo:
I also took the chance to remove some redundant/unnecessary code.

PR #1376, opened 2025-07-17
Fix pixel status in R0V files

moralejo:
This sets explicitly the lowest bit of the pixel status to 0 if the pixel is not saved. Before this, the original value was kept.


maxnoe (approved):

moralejo:
Does not work now, PixelStatus.DVR_STATUS_0 has to be cast as well

PR #1377, opened 2025-07-18
Added command-line option to fix R0V files which lack pixel status info

moralejo:
Option to correct R0V files that were created correctly, except for the pixel status bits (see #1376)
This PR is to correct _already existing_ R0V files (whereas #1376 solves the original problem in the R0V production)

### DIFF
--- a/lstchain/scripts/lstchain_r0g_to_r0v.py
+++ b/lstchain/scripts/lstchain_r0g_to_r0v.py
@@ -127,11 +127,7 @@ def main():
             rows_per_tile = header["ZTILELEN"]
 
             if args.fix_pixel_status:
-                r0v_input = True
-                if "LSTDVR" in header:
-                    r0v_input = header["LSTDVR"]
-                else:
-                    r0v_input = False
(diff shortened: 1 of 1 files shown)


lstchain/scripts/lstchain_r0g_to_r0v.py
```diff
@@ -116,6 +126,15 @@ def main():
             n_tiles = header["NAXIS2"]
             rows_per_tile = header["ZTILELEN"]
 
+            if args.fix_pixel_status:
+                r0v_input = True
+                if "LSTDVR" in header:
+                    r0v_input = header["LSTDVR"]
+                else:
+                    r0v_input = False
```
maxnoe, line 134:
```suggestion
                r0v_input = header.get("LSTDVR", False)
```

moralejo: If it is not R0V the file may not even contain the LSTDVR field , this is why I did it like that. I think header.get it would yield an error which would be less clear than the current one

PR #1384, opened 2025-09-04
Using the subrun index gives an error in the cherenkov transparency script for days in which some subruns failed and were discarded

marialainez:
For example, for the day 20250823, in which several subruns of run 21523 had to be discarded, the cherenkov transparency script gives this error: 
```
  File "/var/spool/slurm//job48183790/slurm_script", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/fefs/aswg/software/conda/envs/osa-v0.11.3/lib/python3.11/site-packages/lstchain/scripts/lstchain_cherenkov_transparency.py", line 218, in main
    dummy[jj] = nn
    ~~~~~^^^^
IndexError: index 106 is out of bounds for axis 0 with size 83
/fefs/aswg/software/conda/envs/osa-v0.11.3/lib/python3.11/site-packages/tables/file.py:130: UnclosedFileWarning: Closing remaining open file: /fefs/onsite/data/lst-pipe/LSTN-01/DL1/20250823/v0.11/tailcut84/datacheck/datacheck_dl1_LST-1.Run21523.h5
  warnings.warn(UnclosedFileWarning(msg))
``` 
Using this fix I could correctly add the "cosmics_intensity_spectrum" table to the datacheck file.

### DIFF
--- a/lstchain/scripts/lstchain_cherenkov_transparency.py
+++ b/lstchain/scripts/lstchain_cherenkov_transparency.py
@@ -174,11 +174,12 @@ def main():
                 dummy = np.zeros(num_subruns)
                 dummy2 = np.zeros(num_subruns)
                 for jj, nn, nn2 in zip(
-                        np.arange(0,num_subruns),
+                        a.root.dl1datacheck.pedestals.col('subrun_index'),
                         a.root.dl1datacheck.pedestals.col('num_events'),
                         a.root.dl1datacheck.pedestals.col('num_cleaned_events')):
-                    dummy[jj] = nn
-                    dummy2[jj] = nn2
+                    new_index = np.where(a.root.dl1datacheck.cosmics.col('subrun_index') == jj)[0][0]
+                    dummy[new_index] = nn
+                    dummy2[new_index] = nn2
                 num_pedestals.append(dummy)
                 num_cleaned_pedestals.append(dummy2)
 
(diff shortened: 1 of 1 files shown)


moralejo:
This was intended for the case in which subruns had missing pedestals or flat-field tables, not for the case in which the full subrun was missing.

That just 83 of 106 subruns are processed is very worrying, is it understood why  this happened?

lstchain/scripts/lstchain_cherenkov_transparency.py
```diff
@@ -174,7 +174,7 @@ def main():
                 dummy = np.zeros(num_subruns)
                 dummy2 = np.zeros(num_subruns)
                 for jj, nn, nn2 in zip(
-                        a.root.dl1datacheck.pedestals.col('subrun_index'),
+                        np.arange(0,num_subruns),
```
moralejo, line 177: This will not work as intended. In case pedestals (or flat-field) table is missing, we want to fill the values for the existing subruns in the right subrun index. But now jj is no longer the subrun index.

moralejo:
In case some subruns are completely missing (in cosmics, pedestals and flat-field tables, as when the subrun processing went wrong), and, on top of that, some subruns have the cosmics table, but not the pedestals or flat-field tables, what we need is to put the info of the existing  pedestals/FF in the right positions (consistent with the cosmics), and fill 0's or nans for the subruns in which the pedestals/FF tables are missing.

moralejo:
> @marialainez can you please try the current version on the problematic run? (even if it worked, we have to understand why runs like 21523 have so many missing subruns)

@FrancaCassol  I think this is may be a case in which the lack of interleaved FF (because of calibox issues) actually spoils the (probably ok-ish) level-A calibration. Is that possible?

marialainez:
The lstchain_cherenkov_transparency worked correctly with this branch for 20250823.

moralejo:
Why can I merge while 3 out of 4 CI tests are still running?

PR #1395, opened 2025-12-12
Correct params light yield vs cleaning

aaguasca:
This PR applies an empirical correction on the dR/dI index and rate at 422 p.e. for runs with cleaning settings above tail cut 8 and 4 (picture and boundary threshold, respectively). 

The correction employs a single value per cleaning setting for each parameter.

Moreover, this PR adds the picture and boundary threshold values in the cosmics_intensity_spectrum table.

This PR addresses one of the tasks in #1396

### DIFF
--- a/lstchain/conftest.py
+++ b/lstchain/conftest.py
@@ -206,7 +206,6 @@ def observed_dl1_files(temp_dir_observed_files, run_summary_path):
 
 
 
-@pytest.mark.private_data
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


moralejo (changes requested):
Looks good, other than the minor comments I made.

aaguasca:
Thank you very much for the feedback. I updated the code accordingly. Please let me know if the description meets the idea you had in mind.

PR #1400, opened 2026-01-19
Modify fov_offset parameters in irf_dl3_tool_config.json

moralejo:
Updated field values for fov_offset parameters in the configuration. The previous lower edge was >0, so it resulted in a hole in the effective are vs. offset. The proposed binning is relatively coarse (0.5 deg in offset angle) in order to have sufficient statistics from the existing diffuse gamma MC.

### DIFF
--- a/docs/examples/irf_dl3_tool_config.json
+++ b/docs/examples/irf_dl3_tool_config.json
@@ -39,7 +39,7 @@
     "energy_migration_min": 0.2,
     "energy_migration_max": 5,
     "energy_migration_n_bins": 30,
-    "fov_offset_min": 0.,
+    "fov_offset_min": 0,
     "fov_offset_max": 2.5,
     "fov_offset_n_edges": 6,
     "bkg_fov_offset_min": 0,


docs/examples/irf_dl3_tool_config.json
```diff
@@ -39,9 +39,9 @@
     "energy_migration_min": 0.2,
     "energy_migration_max": 5,
     "energy_migration_n_bins": 30,
-    "fov_offset_min": 0.1,
-    "fov_offset_max": 1.1,
-    "fov_offset_n_edges": 9,
+    "fov_offset_min": 0.,
+    "fov_offset_max": 2.5,
+    "fov_offset_n_edges": 6,
```
rlopezcoto: why changing the number of edges? this will result in a file that is not compatible with older versions of the code because arrays will not have the same length

moralejo: This just sets the binning in the IRF files that go into the DL3, I think it should not be a problem.

PR #1405, opened 2026-03-12
Changes needed to process raw data calibrated by the EvB

moralejo:
These (together with changes in ctapipe_io_lst) allow to process files in which R1 waveforms are calibrated and written out by the event builder.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - astropy>=6.1,<8
   - pip
   - ctapipe >=0.25.0
-  - ctapipe_io_lst >=0.26.0,<0.28.0a0
+  - ctapipe_io_lst ~=0.28.0
   - ctaplot~=0.6.4
   - gammapy=1.3
   - h5py
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - astropy>=6.1,<8
   - pip
   - ctapipe >=0.25.0
-  - ctapipe_io_lst ~=0.29.0
+  - ctapipe_io_lst =0.29.0
   - ctaplot~=0.6.4
   - gammapy=1.3
   - h5py
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - astropy>=6.1,<8
   - pip
   - ctapipe >=0.25.0
-  - ctapipe_io_lst ~=0.28.0
+  - ctapipe_io_lst ~=0.29.0
   - ctaplot~=0.6.4
   - gammapy=1.3
   - h5py
--- a/environment.yml
+++ b/environment.yml
@@ -18,7 +18,7 @@ dependencies:
   - iminuit>=2
   - joblib~=1.4.0
   - toml
-  - protozfits=2.7
+  - protozfits>=2.11
   - pyparsing
   - scipy
   - scikit-learn>=1.2
--- a/lstchain/calib/camera/calibration_calculator.py
+++ b/lstchain/calib/camera/calibration_calculator.py
@@ -216,60 +216,49 @@ class LSTCalibrationCalculator(CalibrationCalculator):
         # define unusables on number of estimated pe
         npe_deviation =  calib_data.n_pe - npe_median[:,np.newaxis]
 
+        selected_gain = event.r1.tel[self.tel_id].selected_gain_channel
+        if selected_gain is None:
+            n_gains = 2
+        else:
+            n_gains = 1
+
         # cut on the base of the pe statistical uncertainty over the camera
         tot_std = self.expected_npe_std(npe_median, ff_data.n_events)
-
-        npe_outliers = (
-            np.logical_or(npe_deviation < self.npe_median_cut_outliers[0] * tot_std[:,np.newaxis],
-                          npe_deviation > self.npe_median_cut_outliers[1] * tot_std[:,np.newaxis]))
+        if selected_gain is None:
+            npe_outliers = (
+                np.logical_or(npe_deviation < self.npe_median_cut_outliers[0] * tot_std[:,np.newaxis],
+                              npe_deviation > self.npe_median_cut_outliers[1] * tot_std[:,np.newaxis]))
+        else:
+            npe_outliers = (
+                np.logical_or(npe_deviation < self.npe_median_cut_outliers[0] *
+                              tot_std[selected_gain],
+                              npe_deviation > self.npe_median_cut_outliers[1] *
+                              tot_std[selected_gain]))
 
         # calibration unusable pixels are an OR of all masks
         calib_data.unusable_pixels = np.logical_or(unusable_pixels, npe_outliers).filled(True)
-        
+
         # give to the unusable pixels the median camera value for the dc_to_pe and pedestal
         # (these are the starting data for the Cat-B calibration)
-        selected_gain = event.r1.tel[self.tel_id].selected_gain_channel
-        if selected_gain is None:
-            dc_to_pe_masked = np.ma.array(calib_data.dc_to_pe,
-                                          mask=calib_data.unusable_pixels)
-            pedestal_per_sample_masked = np.ma.array(calib_data.pedestal_per_sample,
-                                                     mask=calib_data.unusable_pixels)
-        else:
-            if calib_data.dc_to_pe.shape == calib_data.unusable_pixels.shape:
-                dc_to_pe = calib_data.dc_to_pe
-            else:
-                dc_to_pe = np.nan * np.ones_like(calib_data.unusable_pixels)
-                dc_to_pe[selected_gain, np.arange(constants.N_PIXELS)] = calib_data.dc_to_pe
-            dc_to_pe_masked = np.ma.array(dc_to_pe,  mask=calib_data.unusable_pixels)
-
-            if (calib_data.pedestal_per_sample.shape == calib_data.unusable_pixels.shape):
-                pedestal_per_sample = calib_data.pedestal_per_sample
-            else:
-                pedestal_per_sample = np.nan * np.ones_like(calib_data.unusable_pixels)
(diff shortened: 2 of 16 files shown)


lstchain/scripts/tests/test_lstchain_scripts.py
```diff
@@ -39,7 +39,7 @@ def find_entry_points(package_name):
     """from: https://stackoverflow.com/a/47383763/3838691"""
     entrypoints = [
         ep.name
-        for ep in pkg_resources.iter_entry_points("console_scripts")
+        for ep in entry_points(group="console_scripts")
         if ep.module_name.startswith(package_name)
```
maxnoe:
```suggestion
        if ep.value.startswith(package_name)
```

lstchain/calib/camera/flatfield.py
```diff
@@ -117,18 +117,22 @@ def _extract_charge(self, event):
...
+            n_pixels = 1855
+            selected_gain = event.r1.tel[self.tel_id].selected_gain_channel
+            if selected_gain is None:  # No gain selection
+                gain_index = np.zeros((waveforms.shape[0], waveforms.shape[1]),
+                                      dtype=np.int64)
+                gain_index[1] = 1
+                waveforms *= self.time_sampling_corrector.get_corrections(event,self.tel_id)
+                [gain_index, np.arange(n_pixels)]
+            else:
+                waveforms[0] *= self.time_sampling_corrector.get_corrections(event,self.tel_id)
```
rlopezcoto, line 139: shouldn't `gain_index` be also set here?

moralejo: I think this is not used by default (this was the correction of **charge** from the inhomogeneities of the drs4 sampling).  But it should work, in case selected_gain is not None, then I think the corrections are just (N_PIXELS, n_samples) and the array of waveforms is (1, N_PIXELS, n_samples)

moralejo: @FrancaCassol can probably tell for sure

FrancaCassol: Actually all the calibration code is moved in lstcam_calib. Are you doing the changes for the Cat-B calibration?

lstchain/calib/camera/calibration_calculator.py
```diff
@@ -226,19 +227,50 @@ def calculate_calibration_coefficients(self, event):
...
-        dc_to_pe_masked = np.ma.array(calib_data.dc_to_pe, mask=calib_data.unusable_pixels)
+        # (these are the starting data for the Cat-B calibration)
+        selected_gain = event.r1.tel[self.tel_id].selected_gain_channel
+        if selected_gain is None:
+            dc_to_pe_masked = np.ma.array(calib_data.dc_to_pe,
+                                          mask=calib_data.unusable_pixels)
+            pedestal_per_sample_masked = np.ma.array(calib_data.pedestal_per_sample,
+                                                     mask=calib_data.unusable_pixels)
+        else:
+            if calib_data.dc_to_pe.shape == calib_data.unusable_pixels.shape:
```
FrancaCassol: why you need to put this condition?
dc_to_pe will have always the same shape than the unusable pixels

lstchain/calib/camera/flatfield.py
moralejo: These are minimal changes to get the pipeline to work with single-gain data. As we have discussed in other channel, lstosa is still using the CatB calibration in lstchain

lstchain/calib/camera/calibration_calculator.py
moralejo: That was not the case in the tests I did with calibrated data from March 2025. But I do not know why.

maxnoe: THat's maybe an issue in the source, not padding some arrays that are shorter due to missing pixels.

In general, `ctapipe_io_lst` should only give out arrays that have the nominal length, not shorter arrays if modules/pixels are mssing.

moralejo: The test is on data with all pixels. I get an array dc_to_pe which is (1, 1855) whereas unusable_pixels is (2, 1855)

maxnoe: I think since the changes in ctapipe about the waveform shape, we expect all calibrations to always have both gains.

lstchain/calib/camera/calibration_calculator.py
FrancaCassol, line 186: You need to gain select also this term if it is the case

FrancaCassol, line 219: This is at the origin of the different shapes, we need to gain select it, if the case. Then you can remove all the if related to the shape I would say

maxnoe: Do we need this code still? Or can we just remove it as it is now in lstcam_calib and the standard processing uses `lstcam_calib` produced calibration files?

Maybe a good point to fully remove it now instead of trying to fix it.

moralejo: I prefer to avoid that (using this PR for clean-up), also because although we do not use it for the "core" of the pipeline (e.g. we now do the pedestal cleaning using the Cat-B file), removing it would require to remove other parts which depend on it (at least the noise-tuning at the DL1a level) which may still be useful.

moralejo: > You need to gain select also this term if it is the case

I am not sure that is needed given the other changes you suggested (please check)

FrancaCassol:
@abelardo, It is also necessary to update the visualisation/plot_calib.py module to plot only one gain, if the case.

lstchain/calib/camera/calibration_calculator.py
FrancaCassol: yes, you need it. In you case it works because you do not read the systematics file.

moralejo: With the latest commits I think that change is not needed. I have kept the 2 gains in all arrays filled by he calibration calculations.

moralejo: And this makes that calibration_calculator.py goes back to its original state (latest commit)... no change needed in it.

FrancaCassol (dismissed):

lstchain/scripts/lstchain_r0g_to_r0v.py
```diff
@@ -186,6 +187,10 @@ def main():
                         if len(wf) != num_gains * num_samples * ordered_pix_mask.sum():
                             raise RuntimeError('The pixel selection file is not consistent with the input R0V file!')
 
+                # Write pixel_time_shift only for selected pixels:
+                new_pixel_time_shift = pixel_time_shift[ordered_pix_mask]
+                event.pixel_time_shift.data = new_pixel_time_shift.tobytes()
```
maxnoe:
```suggestion
                event.pixel_time_shift = protozfits.numpy_to_any_array(new_pixel_time_shift)
```

moralejo: That does not seem to work... 
`    event.pixel_time_shift = protozfits.numpy_to_any_array(new_pixel_time_shift)
    ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: Assignment not allowed to message field "pixel_time_shift" in protocol message object.`

moralejo: Back to previous version (seems to work) - could it give any trouble?

maxnoe: should be fine... but I will check tomorrow how to properly overwrite the full object.

moralejo: > should be fine... but I will check tomorrow how to properly overwrite the full object.

Ok. I confirmed it did work: I get identical DL2 data starting from R0V or R0,  for the EvB-calibrated test data.

maxnoe: I think you could do either:

```
                event.pixel_time_shift.CopyFrom(protozfits.numpy_to_any_array(new_pixel_time_shift))
```

or


```          
                event.ClearField("pixel_time_shift")
                event.pixel_time_shift = protozfits.numpy_to_any_array(new_pixel_time_shift)
```

maxnoe: but just overriding the bytes should also be fine!

moralejo: Ok, I'll leave like this since I have tested it thoroughly.

moralejo: For the record: the test on the 20250326 20527 Crab run show performance comparable to regular Crab runs calibrated offline (in terms of ROC curves and datacheck plots). Also, the results when starting from R0 and from R0V are identical at the DL2 level, as expected.

moralejo:
@wip ready for review

environment.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   - astropy>=6.1,<8
   - pip
   - ctapipe >=0.25.0
-  - ctapipe_io_lst ~=0.29.0
+  - ctapipe_io_lst =0.29.0
```
rlopezcoto, line 10: I think it should be `==`

maxnoe: `== 0.29.0` fixes to `0.29.0` exactly, `= 0.29` for conda means `newest 0.29.x`

moralejo: I want to fix 0.29.0 because of the (unexpected and unrelated) 0.29.1 release (it should have been 0.30.0, I guess)

maxnoe: > (it should have been 0.30.0, I guess)

Why? AFAIk it didn't make any breaking changes and only added compatibility with newer ctapipe versions.

Is that not the case? Sorry then

rlopezcoto (approved):

PR #1413, opened 2026-06-05
Fix pixel time shift filling in r0_to_dl1.py

moralejo:
Make storing of pixel_time_shift in interleaved*.h5 work.


moralejo:
Closing this. Putting the pixel_time_shift in the attrs does not work, only one event (last one, I think) is written out. There seems to be no way of adding the field to the R1CameraContainer until we update ctapipe (although at some point it was working for me locally, not sure why - I have failed to reproduce it).
A better approach is to write the dl1a pixel times and use that for the correction.

Issue #1321, opened 2024-11-27
Allow for passing the target coordinates to `extract_source_position`


morcuended:
https://github.com/cta-observatory/cta-lstchain/blob/9aaa78b0bd911d35fc2b3fefa03b064bb21b8262/lstchain/reco/utils.py#L104-L132

For cases in which the name solver of astropy does not work (e.g. GRBs, GW, neutrino alerts) one needs to pass the coordinates directly to make this function work. Used in the explore DL2 notebook (theta2 distribution calculation)

Nanakjoth:
Hi @morcuended , this issue is still needed ?— can I take it on,

PR #1414, opened 2026-06-06
Add extract_source_position_from_coord function

zhipzhang:
add function to solve #1321


zhipzhang:
Looks like there have some issues with the network to download the geometry files/ test files.

PR #1415, opened 2026-06-06
Store & use DL1a images in interleaved*.h5 files

moralejo:
Use DL1a pixel times from CatA DL1 file for the calculation of Cat-B time corrections

### DIFF
--- a/lstchain/data/lstchain_standard_config.json
+++ b/lstchain/data/lstchain_standard_config.json
@@ -287,7 +287,7 @@
   "write_interleaved_events":{
     "DataWriter": {
       "overwrite":  true,
-      "write_dl1_images": false,
+      "write_dl1_images": true,
       "write_dl1_parameters": false,
       "write_r0_waveforms": false,
       "write_r1_waveforms": true,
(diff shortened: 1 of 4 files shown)


lstchain/data/lstchain_standard_config.json
```diff
@@ -287,7 +287,7 @@
   "write_interleaved_events":{
     "DataWriter": {
       "overwrite":  true,
-      "write_dl1_images": false,
+      "write_dl1_images": true,
```
rlopezcoto, line 290: why changing the default behaviour?

rlopezcoto (commented):
I think @FrancaCassol should have a look before moving on with this one

lstchain/data/lstchain_standard_config.json
moralejo: The point of the change is to use the already calculated peak_times (in the creation of the first DL1 file) to compute the Cat-B corrections to the time flat-fielding. For that we need to write them out to the interleaved*h5 files (we may consider to write only the times, and not the rest of the image info, to save some space).
The original method used re-extracted pulse times from R1 waveforms (but using the calibration integrator, not the analysis one), and hence also missed the Fourier time corrections, so the calculation was quite inaccurate,

FrancaCassol (approved):

PR #1419, opened 2026-06-23
Improve the condition to identify Cat-B calibration files which were

moralejo:
Improve the condition to identify Cat-B calibration files which were produced from gain-selected interleaved events


alexcervino (approved):
Thanks a lot!

PR #1420, opened 2026-07-01
remove hardcoded LST-1 name in file name pattern

FrancaCassol:


lstchain/scripts/lstchain_create_run_summary.py
```diff
@@ -180,7 +180,7 @@ def guess_run_type(date_path, run_number, counters, n_events=500):
     run_type: str
         Type of run (DRS4, PEDCALIB, DATA, ERROR)
     """
-    pattern = f"LST-1.1.Run{run_number:05d}.0000*.fits.fz"
+    pattern = f"LST-*.1.Run{run_number:05d}.0000*.fits.fz"
```
moralejo, line 183: Not new in this PR, but why the asterisk before ".fits"? subrun id is just 4 digits.

maxnoe: This allows using it with the test data that has suffixes like "_first_50"

FrancaCassol: Hi @maxnoe, do you understand what is the problem with the tests? (I run them several time)

moralejo: It seems it is because it now installs numpy >=2 whereas it used to use 1.26, but I do not know why that changed.

PR #1421, opened 2026-07-08
Bugfix: calibration events with single gain in R0V

guillaumegrolleron:
Bugfix in r0_to_dl1.py: dropping FLATFIELD/SKY_PEDESTAL events lacking a two-gain R0 waveform.

I've seen this bug when reprocessing DL1 from R0V for a significant part of the Crab runs (11014 e.g.)

### DIFF
--- a/lstchain/reco/r0_to_dl1.py
+++ b/lstchain/reco/r0_to_dl1.py
@@ -556,6 +556,20 @@ def r0_to_dl1(
                         rel_factor_2d[1, sg == 1] = rel_factor[sg == 1]
                         tel_dl1.relative_factor = rel_factor_2d
 
+                # Skip events tagged as interleaved (flat-field / pedestal) but
+                # stored with a single gain, i.e. without a two-gain R0 waveform
+                # (mislabeled cosmics, or interleaved events gain-selected by the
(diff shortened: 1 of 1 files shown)


moralejo:
You mean that some events have this issue, right? - not every interleaved event.

The problem is that in the current data (full online calibration) interleaved events really have a single gain, and  this fix would mean that all interleaved events would be skipped for those new files.

guillaumegrolleron:
Hi @moralejo, 
Yes indeed this bug concerns some events only, e.g. about 50 are concerned in the subrun 0000 of 11014. 

You're right, I don't know what is the best way to proceed. The issue appears only if we want to reprocess older R0V with lstchain > 0.10. I did this patch I just wanted to keep you informed.

moralejo:
A few lines below we do have this:
`is_calibrated = EVBPreprocessingFlag.PE_CALIBRATION in tdp_action`

The data in which interleaved events are _really_ gain selected (i.e. not accidentally, but on purpose) are those in which the EvB did the full calibration to p.e. So the condition to skip single-gain events can be applied to events in which is_calibrated is False  (just by initialising is_calibrated earlier). Can you give it a try?

PR #1423, opened 2026-07-22
New scripts to facilitate the DL2 to DL3 stage

moralejo:
New script to obtain the recommended minimum intensity cut  for a given run in the DL2 => DL3 step
New script to find out which RF models have been used to produce a given DL2 file

### DIFF
--- a/lstchain/scripts/lstchain_get_intensity_cut.py
+++ b/lstchain/scripts/lstchain_get_intensity_cut.py
@@ -21,7 +21,7 @@ def main():
 
     min_intensity = get_intensity_cut(read_table(file, dl2_params_lstcam_key))
     log.info('Recommended minimum intensity cut (p.e.):')
-    log.info(f'{int(min_intensity+0.5):d}')
+    log.info(f'{min_intensity:.0f}')
 
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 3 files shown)


lstchain/scripts/lstchain_get_intensity_cut.py
```diff
@@ -0,0 +1,27 @@
...
+    log.setLevel(logging.INFO)
+    handler = logging.StreamHandler()
+    logging.getLogger().addHandler(handler)
+
+    args = parser.parse_args()
+    file = args.file
+
+    min_intensity = get_intensity_cut(read_table(file, dl2_params_lstcam_key))
+    log.info('Recommended minimum intensity cut (p.e.):')
+    log.info(f'{int(min_intensity+0.5):d}')
```
rlopezcoto: why 0.5?

moralejo: It is just a pedestrian way of rounding to integer number of p.e. :)

rlopezcoto: what about `log.info(f'{min_intensity:.0f}')`?

rlopezcoto (approved):

PR #1427, opened 2026-09-16
Correct tailcut determination description

moralejo:
Updated the description to reflect the use of the 95% quantile for determining tailcuts instead of the median pixel charge.


rlopezcoto:
A test did not pass even though it is not related to this PR

moralejo:
Yes, I was looking into it. Very strange, nothing changed in the relevant function, yet the result of the psf smearing is slightly different:

> 
> =========================== short test summary info ============================
> FAILED lstchain/image/tests/test_modifier.py::test_psf_smearer[0.5] - assert np.False_
>  +  where np.False_ = <function isclose at 0x7f61871244b0>(np.float64(10000.0), np.float64(10150.0), rtol=0.01)
>  +    where <function isclose at 0x7f61871244b0> = np.isclose
> ====== 1 failed, 236 passed, 1 xfailed, 984 warnings in 308.59s (0:05:08) ======
> Error: Process completed with exit code 1.

moralejo:
> Yes, I was looking into it. Very strange, nothing changed in the relevant function, yet the result of the psf smearing is slightly different:
> 
> > =========================== short test summary info ============================
> > FAILED lstchain/image/tests/test_modifier.py::test_psf_smearer[0.5] - assert np.False_
> > 
> > * where np.False_ = <function isclose at 0x7f61871244b0>(np.float64(10000.0), np.float64(10150.0), rtol=0.01)
> > * where <function isclose at 0x7f61871244b0> = np.isclose
> >   ====== 1 failed, 236 passed, 1 xfailed, 984 warnings in 308.59s (0:05:08) ======
> >   Error: Process completed with exit code 1.

Must have to do with a new version of some dependency

moralejo:
> > Yes, I was looking into it. Very strange, nothing changed in the relevant function, yet the result of the psf smearing is slightly different:
> > > =========================== short test summary info ============================
> > > FAILED lstchain/image/tests/test_modifier.py::test_psf_smearer[0.5] - assert np.False_
> > > 
> > > * where np.False_ = <function isclose at 0x7f61871244b0>(np.float64(10000.0), np.float64(10150.0), rtol=0.01)
> > > * where <function isclose at 0x7f61871244b0> = np.isclose
> > >   ====== 1 failed, 236 passed, 1 xfailed, 984 warnings in 308.59s (0:05:08) ======
> > >   Error: Process completed with exit code 1.
> 
> Must have to do with a new version of some dependency

Strange, this just uses numpy's random generator of a multinomial, perhaps the tolerance is too tight and this is a first-time fluke? (seed is not fixed)

rlopezcoto:
we should not use random numbers to build tests

moralejo:
Must be something else, it fails again and it had never failed before...

moralejo:
> we should not use random numbers to build tests

If probability of failing the test is, say, <1e-7, I see no problem, although I agree it is not really necessary. Anyway it does not seem to be the problem here.

moralejo:
> > we should not use random numbers to build tests
> 
> If probability of failing the test is, say, <1e-7, I see no problem, although I agree it is not really necessary. Anyway it does not seem to be the problem here.

Actually the test is deterministic, the seed is set before the test. That is not the problem. Neither does it seem to be (as suggested by Gemini) that the tested image has integer dtype (fails exactly in the same way when forced to be float)

moralejo:
I think the problem is related to some change in how the CI tests are executed: the same seed now results (because of different parallelization or whatever) in a different random realization, and the 1% tolerance is indeed too tight a limit when we are spreading as much as ~half of the central pixel's photoelectrons to its neighbors (with an starting number of 1e4), and the central pixel is receiving about 1/2 * 1/6 of each neighbor's charge... So  I just set the tolerance to 3%.

rlopezcoto:
The test failed, but it seemed a connection problem, now everything works

rlopezcoto (approved):

